PR #46, opened 2018-01-24
Add CLI command to set the status of a FC

ingkebil:
This is a first part of making fastq management easier. This adds a command to set the status of a FC so we can use it in the FC removal script.

Once this is approved we can then go ahead and:
1. import fastq's to HK
2. manage the removal of fastq files through HK


ingkebil:
See also #45

PR #54, opened 2018-01-31
Added timeout in the ticket creation request

patrikgrenfeldt:
set 5 minutes timeout, currently we get a 504 after 1 minute


patrikgrenfeldt:
I think we should tested in the staging environment first to see that we don't have any formal errors and then I think we need to test this in production since it is a problem that occurs once in a while and I don't know how to reproduce it

ingkebil:
This might got to do with the actual nginx proxy and we should check how fast the proxy times out as well.

patrikgrenfeldt:
I have not heard anyone mention this as a problem for a long time, closing the issue

PR #60, opened 2018-02-05
Beacon

northwestwitch:
Added cg upload beacon and cg clean beacon functionalities


northwestwitch:
Ok, I'll try it! :)

northwestwitch:
I've uploaded the cgbeacon package into PyPI so the dependency issue should be fixed. I had a syntax error in cgbeacon that made the Travis CI fail. I've fixed and uploaded the patched cgbeacon to PyPI. Now there seems to be another issue in a piece of code that I've never touched! :-/

PR #61, opened 2018-02-05
make optional sample attributes non-mandatory

patrikgrenfeldt:
Order portal crashed when comment was missing, probably when supplying an order as an excel file 
Now is volume, concentration and comment handled as optionals


ingkebil:
👍

PR #62, opened 2018-02-06
match against the subgroup that contains the tickenumber

patrikgrenfeldt:


ingkebil:
Aha, this is then the backend validation for the same frontend fix?

PR #66, opened 2018-02-09
all required changes to integrate beacon in cg

northwestwitch:
Started from scratch from an updated branch

### DIFF
--- a/cg/apps/scoutapi.py
+++ b/cg/apps/scoutapi.py
@@ -4,6 +4,7 @@ from typing import List
 
 from pymongo import MongoClient
 from scout.adapter.mongo import MongoAdapter
+from scout.adapter.mongo.panel import PanelHandler
 from scout.export.panel import export_panels as scout_export_panels
 from scout.load import load_scout
 from scout.parse.case import parse_case_data
@@ -36,6 +37,6 @@ class ScoutAPI(MongoAdapter):
             LOG.debug("load new Scout case")
             load_scout(self, config_data)
 
-    def export_panels(self, panels: List[str]):
+    def export_panels(self, panels: List[str], versions: None):
         """Pass through to export of a list of gene panels."""
-        return scout_export_panels(self, panels)
+        return scout_export_panels(self, panels, versions)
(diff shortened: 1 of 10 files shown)


northwestwitch:
@patrikgrenfeldt and @ingkebil I've tried with rebase and still got the same error, this time I've created a new updated branch and introduced the changes in that. Still fails the Travis-ci with the same initial error ,related to some flowcell test. Gee I have no clue, I didn't even touch those files!

ingkebil:
What was "the same error"?

northwestwitch:
After having uploaded the cgbeacon package in pypi it didn't complain
anymore for the requirement. There was instead a fowcell-related test that
failed, you can see the exact error message if you click on the red cross
of the Travis-ci that didn't pass. Patrick has tested the master branch and
said that it doesn't pass the test itself, so we don't think it's my code.
We should fix the master!

ingkebil:
Sure! Sounds one a plan :)

wkr,
Kenny
--
Composed and sent on the road

northwestwitch:
Guys, I think the problem might be not in the code but in the flowcell ID. I've tried to change the ID of the flowcell and I get a different error message this time. I don't really know how to fix this test since I haven't understood the characteristics of the flowcell that has to be tested. Could we take a look at the code together?

northwestwitch:
Fixed the issue with the flowcell test (thanks @patrikgrenfeldt!). I'll merge to master and then update cg on the production environment, unless you stop me because of other issues in cg I'm not aware of?

cg/apps/scoutapi.py
```diff
@@ -36,6 +37,6 @@ def upload(self, data: dict, threshold: int=5, force: bool=False):
             LOG.debug("load new Scout case")
             load_scout(self, config_data)
 
-    def export_panels(self, panels: List[str]):
+    def export_panels(self, panels: List[str], versions: None):
```
ingkebil: Please refrain from changing function signatures like this. I presume you wanted to add a default to `versions` and not demand that `versions` is of type `None` ;)

ingkebil: Fixed in bdc72a2e4eb7d6f7 :)

PR #75, opened 2018-02-27
Added mip files vcf2cytosure and subsample_mt

henrikstranneheim:
- Added tests


ingkebil:
... the tests fail no?

henrikstranneheim:
Yes, Travis is not using the required version of Trailblazer and it look like like some unrelated flowcell test fails. However locally all tests pass.

Skickat från min iPhone

PR #93, opened 2018-03-13
Do not store undetermined fastq files in HK for pooled lanes

### LINKED ISSUE #50: cg tranfer flowcell adds undet fastq files for pooled X samples
Might be intended behavior if the linking to the analysis folder excludes the undet files, but I have not tested that yet.

ingkebil:
This would fix #50


barrystokman:
Can't say much about the is_sample_pooled method.

Change to flowcell method looks good if we want undetermined fastq files from samples that are not pooled to still be added.

Is there anything that needs to be done for pooled, undetermined samples?

ingkebil:
Nope, we can discard those files. They would be nice for statistics at some point: e.g. it would be nice to know which indexes are in the undetermined to be able to determine index bleeding (from neighbouring lanes), contaminations, leftovers from previous runs, ... But I'm precalculating those already, we so can discard the actual files :)

PR #99, opened 2018-03-19
Last analysis 2 chanjo

### LINKED ISSUE #95: cg upload doesn't replace coverage data for re-runs
Whenever a re-run analysis is completed and cg upload uploads it to scout, then the coverage subroutine is invoked from cg upload without the (replace) -r True parameter, so the coverage data is not uploaded in chanjo db and chanjo-report uses old analyses when used to create coverage reports from scout. For more info check https://github.com/Clinical-Genomics/scout/issues/699.

northwestwitch:
This should fix #95 issue, hopefully. Once fixed the issue for the current and future re-runs we should compile a list of all re-runs with outdated coverage, I'm on it!

### DIFF
--- a/cg/cli/upload.py
+++ b/cg/cli/upload.py
@@ -28,19 +28,17 @@ def upload(context, family_id):
         analysis_obj = family_obj.analyses[0]
         if analysis_obj.uploaded_at is not None:
             message = f"analysis already uploaded: {analysis_obj.uploaded_at.date()}"
-            print("Im in upload")
-            context.invoke(coverage, re_upload='-r', family_id=family_id)
             click.echo(click.style(message, fg='yellow'))
         else:
-            #context.invoke(coverage, family_id=family_id)
-            #context.invoke(validate, family_id=family_id)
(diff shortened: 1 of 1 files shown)


ingkebil:
Have you discussed with @emiliaol and @emmser if this is needed?

northwestwitch:
It was actually @emmser that noticed that the coverage data was not updated and opened the bug in scout. I've discussed about with @vwirta because I wasn't sure that this was a bug at all, but he confirmed that we want the coverage to be up to date with the analysis re-runs, so it's something we should fix.

ingkebil:
Perfect! That wasn't clear from the context.

Nice catch and good solution!

ingkebil:
Feel free to merge. We can sit together tomorrow and deploy together :)

northwestwitch:
Ok, see you later! :)

ingkebil:
@emmser @emiliaol @b4ckm4n This is now fixed and pushed into production.

Result: no need to run `cg upload coverage -r` anymore :)

PR #104, opened 2018-03-22
check if google authorized

patrikgrenfeldt:


cg/server/invoices/views.py
```diff
@@ -127,6 +133,7 @@ def invoice_template(invoice_id):
 
 @BLUEPRINT.route('/<int:invoice_id>/invoice_file/<cost_center>')
 def modified_invoice(invoice_id, cost_center):
+    return redirect(url_for('admin.index'))
```
ingkebil: This is just a quickfix?

cg/server/app.py
```diff
@@ -42,24 +40,31 @@ def configure_extensions(app: Flask):
 
 
 def register_blueprints(app: Flask):
+
+    if not app.config['CG_ENABLE_ADMIN']:
+        return
```
ingkebil: What does this do?

ingkebil: Aha! Ok, I think I get it :)

We exit here and set up auth+admin_views when we want to have admin views.

cg/server/invoices/views.py
patrikgrenfeldt: It is, I felt it took to long to figure out how to fix it. I guess we should plan implementing a solution in a coming sprint or ask Maya to take care of it since she created those views.

ingkebil: @mayabrandi ping pong. You should prolly know that we disabled the invoices on clinical-api for now. We don't seem to be able to get them safely behind admin views.

Issue #105, opened 2018-03-28
As a bioinformatician working with production analysis, I want the bug where cases missing information in status DB blocks the automatic start of the following analysis in the queue fixed to minimize manual intervention


emiliaol:
If a case that's marked for analysis (or that has no status at all) in the family table in status DB miss any information like capture kit cg analysis auto will throw a python error and exit. This aborts the whole process and it does not try to start any subsequent cases.

Actions:
* Add a warning that's sent to a suitable place so we can pick these errors up from the crontab (this will be fixed in a related user story)
* Make the process skip and not exit on the error so it moves on to the next  case

**DoD:**
See patch DoDs
Automated test not needed if manual test will suffice.

barrystokman:
Unclear what the error message is, or how to reproduce the defect. Cannot be solved for now

ingkebil:
Could you also add an `angrybird` so we can check if this now works?

emiliaol:
Found one! alivebedbug
Please don't actually start the analysis of this (or make sure to cancel it if you do). I have not set this to analyze in status db so please do so when you want to.

ingkebil:
I need to set it to analyze to test the automation. Just ot make sure: this one will fail, right? I need an analysis that fails on missing LIMS info

ingkebil:
So, `alivebedbug` will not autostart because it contains samples that are externally sequenced ;)

That being said, output from `cg get family alivebedbug` seems to be wrong in that aspect.

emiliaol:
I didn't check it that thoroughly. Just picked it from the log... Anyway. I created a new family for you to test called **prostud** that will fail on this:

```(prod170926) [hiseq.clinical@rastapopoulos ~]$ cg analysis auto
prostud: start analysis
Traceback (most recent call last):
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/bin/cg", line 11, in <module>
    load_entry_point('cg==1.4.6', 'console_scripts', 'cg')()
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
[...]

PR #107, opened 2018-04-05
Add a catch for LimsDataError so the auto start of analyses does not …

ingkebil:
…stop

See #105


ingkebil:
... waiting for input so this can be tested :)

ingkebil:
Tested and works!

PR #124, opened 2018-04-24
added startwith option to start mip6

emmser:
Added a new mip6 branch to be in use after upgrade to mip6, this is a mip6-feature. Will be tested after upgrade.

### DIFF
--- a/cg/cli/analysis.py
+++ b/cg/cli/analysis.py
@@ -12,14 +12,16 @@ from cg.store import Store
 LOG = logging.getLogger(__name__)
 PRIORITY_OPTION = click.option('-p', '--priority', type=click.Choice(['low', 'normal', 'high']))
 EMAIL_OPTION = click.option('-e', '--email', help='email to send errors to')
+START_WITH_PROGRAM = click.option('-sw', '--startwith', help='start mip-pipeline with this program and run downstream processes')
 
 
 @click.group(invoke_without_command=True)
 @PRIORITY_OPTION
 @EMAIL_OPTION
+@START_WITH_PROGRAM
 @click.option('-f', '--family', 'family_id', help='family to prepare and start an analysis for')
 @click.pass_context
-def analysis(context, priority, email, family_id):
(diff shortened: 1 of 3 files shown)


cg/cli/analysis.py
```diff
@@ -12,14 +12,16 @@
 LOG = logging.getLogger(__name__)
 PRIORITY_OPTION = click.option('-p', '--priority', type=click.Choice(['low', 'normal', 'high']))
 EMAIL_OPTION = click.option('-e', '--email', help='email to send errors to')
+START_WITH_PROGRAM = click.option('-sw', '--startwith', help='start mip-pipeline with this program and run downstream processes')
```
ingkebil: I would prefer to name this `--start-with`. All words separated by a `-`. This will translate to a `start_with` variable name in python.

cg/cli/analysis.py
```diff
@@ -12,14 +12,16 @@
...
 
 
 @click.group(invoke_without_command=True)
 @PRIORITY_OPTION
 @EMAIL_OPTION
+@START_WITH_PROGRAM
 @click.option('-f', '--family', 'family_id', help='family to prepare and start an analysis for')
 @click.pass_context
-def analysis(context, priority, email, family_id):
+def analysis(context, priority, email, family_id, startwith):
```
ingkebil: ... like here: `startwith` in python (as opposed to in the CLI), will be named `start_with`

cg/cli/analysis.py
```diff
@@ -54,7 +56,7 @@ def analysis(context, priority, email, family_id):
             context.invoke(config, family_id=family_id)
             context.invoke(link, family_id=family_id)
             context.invoke(panel, family_id=family_id)
-            context.invoke(start, family_id=family_id, priority=priority, email=email)
+            context.invoke(start, family_id=family_id, priority=priority, email=email, startwith=startwith)
```
ingkebil: ... Meaning that we need to chagne all python `startwith` to `start_with`

PR #140, opened 2018-05-23
Add capture kit to rml samples

patrikgrenfeldt:


emiliaol:
Tested on stage. Works like a charm

PR #142, opened 2018-05-29
Automate delivery report

patrikgrenfeldt:
I think this i ready for testing.

The coverage will be slightly wrong in the delivery report and that issue is dealt with in another branch


patrikgrenfeldt:
we will merge into beta instead

PR #154, opened 2018-07-05
Put genotype re-upload option into production

northwestwitch:
Introduced a flag to allow force genotype re-upload. I don't know if this would work because I couldn't test it locally (I have no databases and files etc), so test should be done on stage.

Actions
- test reupload on stage (see comment of chiara)


cg/meta/upload/gt.py
```diff
@@ -50,6 +50,6 @@ def _analysis_sex(self, hk_version: hk.models.Version) -> dict:
             data[sample_data['id']] = sample_data['predicted_sex']
         return data
 
-    def upload(self, data: dict):
+    def upload(self, data: dict, replace: bool=False):
```
ingkebil: This is a function signature change. Did you check if this function is used anywhere else in `cg` to make sure it doesn't break anything else?

ingkebil:
So, we have the problem that we cannot push the validation changes into production yet. Because of that, we need to minimize the amount of changes we make to `master`. If we really want to have PR in production now, you have to merge these changes into `beta` and validate them along.

Thanks for your understanding :)

cg/meta/upload/gt.py
northwestwitch: I did, and it should be ok, I don't see it called from anywhere alse. But to be sure we should test it probably.

northwestwitch:
I think it's not extremely urgent but it should be implemented soon because sometimes they're stuck with the genotype analysis. But I think it should be fine to wait and do it after the holidays.

northwestwitch:
@emiliaol @emmser @b4ckm4n if you need to remove the sequencing data from genotype to re-upload the same sample I think you can do it directly from genotype, with the command:
"genotype delete -a sequence sample_name" .
At least I see this option in the code!

emiliaol:
Cool, thanks Chiara. I'll test it. 
And we are almost done with the validation so we can test this next week.

ingkebil:
@northwestwitch could you point me towards how to test this? Should we sit together for this tomorrow?

northwestwitch:
Today I've stuff to do in Uppsala around midday so I'm working from home, but to answer your question you could try to re-upload the genotype of a case (ask production if they have any sample you could use). You should see a difference in the database at the level of two tables:

Connect to genotype and In the table named '**analysis**' there are two analyses for each sample: one is the genotype and one is the sequence. If you re-upload the sequence the entry of type 'sequence' should be updated.

There will be changes also in the '**genotype**' table. the 51 analyses pointing the the old analysis should disappear to be replaced by the 51 pointing to the new sequence analysis. 

You can probably see the same using the app but I don't know exactly because still don't have access to it (what about fixing it now???).

barrystokman:
Tested on stage (PASSED):

table: analysis
family: supermaggot
sample_ids: ACC4668A8, ACC4668A9, ACC4668A10
ids of type 'sequence': 9071, 9072, 9073
ids of type 'genotype': 8890, 8899, 8900

table: genotype
records with analysis_id 9071, 9072, 9073 present

```cg upload genotypes -r supermaggot```

Table ```analysis``` in ```genotype-stage``` updated:

new ids of type ' sequence': 9074, 9075, 9076
created_at timestamp of sequences

ids of type 'genotype' unchanged

Table ```genotype``` in ```genotype-stage``` updated:

old analysis_ids replaced by new analysis_ids.

Evidence:
genotype reupload after reupload family supermaggot new ids - genotype table
genotype reupload after reupload family supermaggot old ids - genotype table
genotype reupload after reupload family supermaggot - analysis table
genotype reupload before reupload family supermaggot - cli messages
genotype reupload before reupload family supermaggot - genotype table
genotype reupload before reupload family supermaggot - analysis table

barrystokman:
@emmser Could you take a look at the documentation I added for ```cg upload genotypes --re-upload```?

PR #167, opened 2018-08-22
Upload rank model version and genome build version to Scout

patrikgrenfeldt:


cg/meta/analysis.py
```diff
@@ -252,3 +259,46 @@ def convert_panels(customer: str, default_panels: List[str]) -> List[str]:
...
+
+    def _open_bundle_file(self, relative_file_path: str) -> Any:
+        """Open a bundle file and return it as an Python object."""
+
+        full_file_path = self.Path(self.deliver.get_post_analysis_files_root_dir()).joinpath(
+            relative_file_path)
+        open_file = self.ruamel.yaml.safe_load(self.Path(full_file_path).open())
+        return open_file
+
+    def get_latest_data(self, family_id: str) -> dict:
```
ingkebil: What about renaming to `get_latest_metadata`

PR #171, opened 2018-08-27
Block submission of non current orderforms

patrikgrenfeldt:
Let customers use only the current orderform
- [ ] Document need to update order portal when orderform changes
- [ ] Document each column that is hardcoded in the orderform AM-doc


patrikgrenfeldt:
Implemented by #231

PR #184, opened 2018-09-11
Warning on missing application version

patrikgrenfeldt:


patrikgrenfeldt:
Tested on stage with "Sequencing" type and an Application without any version

PR #186, opened 2018-09-12
replaced generic view

patrikgrenfeldt:


cg/server/admin.py
```diff
@@ -92,6 +92,12 @@ class SampleView(BaseView):
     form_excluded_columns = ['is_external']
 
 
+class PoolView(BaseView):
+    column_searchable_list = ['name', 'order', 'ticket_number', 'customer.internal_id']
+    column_filters = ['customer.internal_id', 'application_version.application']
+    column_editable_list = ['sequenced_at', 'ticket_number']
```
ingkebil: Why would we want to have `sequenced_at` editable?

patrikgrenfeldt: discussed and approved

PR #188, opened 2018-09-14
No invoice

mayabrandi:
Adding "no_invoice" to pools, to be able to exclude pools from the invoice view just as with samples. 

Has been tested locally against gc-stage and works fine.


ingkebil:
Don't forget to bumpversion when you merge things into master!

Issue #109, opened 2018-04-06
`cg clean mipauto` fails on deleted temp analysis files


ingkebil:
Problem: when running `cg clean mipauto 2018-03-16` the command will exit on missing mip config file.

stacktrace:
```properelf: cleaning MIP output
Traceback (most recent call last):
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/bin/cg", line 11, in <module>
    load_entry_point('cg==1.4.6', 'console_scripts', 'cg')()
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/lib/python3.6/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/lib/python3.6/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/lib/python3.6/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/lib/python3.6/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/lib/python3.6/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/lib/python3.6/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/mnt/hds/proj/bioinfo/SERVER/miniconda/envs/prod170926/lib/python3.6/site-packages/cg/cli/clean.py", line 146, in mipauto
    sampleinfo_path = context.obj['tb'].get_sampleinfo(tb_analysis)
[...]

PR #190, opened 2018-09-18
Skip deleting mip analysis when missing sample_info

ingkebil:
We automatically delete finished analysis after 21 days. CG now fails when the sample_info.yaml file is missing (e.g. when the analysis was deleted manually). Got tired of setting already deleted analysis to 'deleted' in TB. So here we go: we skip them for now.


patrikgrenfeldt:
I think this PR relates to #109

PR #193, opened 2018-09-21
exception for missing value

mayabrandi:


mayabrandi:
Missing date made the view chrash. Important to put this in production asap so that valtteri can invoice today.
/M

@ingkebil @vwirta

PR #194, opened 2018-09-21
bug fix

mayabrandi:


ingkebil:
Hi guys, as neither of you have been introduced to how to perform PRs in CG, let's go over this one together on Wednesday,
Wkr

mayabrandi:
I forced Henrik to merge because Valtteri needed the fix urgently and because I knew it was not going too break anything. No blame on Henrik. However I’m going to go over the code next week to see if there is a better way to catch the problem.

/Maya

Skickat från min iPhone

PR #195, opened 2018-09-22
Create CONTRIBUTING.md

ingkebil:


ingkebil:
I wrote it out in full here even though most of this info is available in the dev manual. We will have to remember to keep those two in sync, possibly by adding that this file exists in the dev manual.

Or what do you think?

ingkebil:
Will add to dev manual instead.

PR #196, opened 2018-09-26
lower filename when testing for type

patrikgrenfeldt:


patrikgrenfeldt:
solution as described in https://stackoverflow.com/questions/45637600/using-endswith-with-case-insensivity-in-python/45637618

PR #197, opened 2018-09-27
Other elution buffer

patrikgrenfeldt:
How to test:

0. Install on OP-stage and clinical-api-stage
1. Create microbial order file with the new UDF/Other Elution Buffer' filled for atleast one sample
2. Go to OP-stage
3. Add microbial order file
4. Open the sample(s) 

Expected outcome: 
The Other Buffer value should be seen in the Other elution buffer field


patrikgrenfeldt:
Installed on stage

PR #206, opened 2018-10-03
Flash on err

mayabrandi:
Added flash on error but had to restructure code at the same time. Hence the big PR.

### DIFF
--- a/cg/meta/invoice.py
+++ b/cg/meta/invoice.py
@@ -22,7 +22,7 @@ class InvoiceAPI():
         else:
             contact_customer = self.customer_obj
         contact_user = self.db.user(contact_customer.invoice_contact)
-        
+
         if not (contact_customer and contact_user):
             self.log.append(msg)
             return None
@@ -33,7 +33,6 @@ class InvoiceAPI():
                 'reference': contact_customer.invoice_reference,
                 'address': contact_customer.invoice_address,
             }
-        print(contact)
         if None in contact.values():
             self.log.append(msg)
             return None
@@ -141,8 +140,6 @@ class InvoiceAPI():
             priority = record.priority_human
 
         full_price = getattr(record.application_version, f"price_{priority}")
-        print(priority)
-        print(full_price)
         if not discount:
             discount=0
         discount_factor = float(100 - discount) / 100
--- a/cg/meta/invoice.py
+++ b/cg/meta/invoice.py
@@ -8,59 +8,80 @@ from cg.server.ext import lims as genologics_lims
 
 class InvoiceAPI():
 
-    def __init__(self, db: Store, lims_api: lims.LimsAPI):
+    def __init__(self, db: Store, lims_api: lims.LimsAPI, invoice_obj: models.Invoice):
         self.db = db
         self.lims_api = lims_api
+        self.log = []
+        self.invoice_obj = invoice_obj
+        self.customer_obj  = invoice_obj.customer
 
+    def prepare_contact_info(self, costcenter):
+        msg = 'Could not open/generate invoice. Contact information missing in database. See log files.'
+        if costcenter.lower() == 'kth':
+            contact_customer = self.db.customer('cust999')
+            contact_user = self.db.user('valtteri.wirta@ki.se')
+        else:
+            contact_customer = self.customer_obj
+            contact_user = self.db.user(self.customer_obj.invoice_contact)
+        if not (contact_customer and contact_user):
+            self.log.append(msg)
+            return None
+        contact = {
+                'name': contact_user.name,
+                'email': contact_user.email,
+                'customer_name': contact_customer.name,
+                'reference': contact_customer.invoice_reference,
+                'address': contact_customer.invoice_address,
+            }
+        if None in contact.values():
+            self.log.append(msg)
+            return None
+        return contact
+        
 
-
-
-
-
-    def prepare(self, costcenter: str, invoice_obj: models.Invoice) -> dict:
+    def prepare(self, costcenter: str) -> dict:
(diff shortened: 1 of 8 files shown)


ingkebil:
Please don't merge yet, I've got some remarks

cg/meta/invoice.py
```diff
@@ -108,17 +135,25 @@ def get_price(self, discount: int, record: models.Sample):
...
-                total_price += self.get_price(discount, record)
-        elif invoice_obj.samples:
-            for record in invoice_obj.samples:
-                total_price += self.get_price(discount, record)
+        if self.invoice_obj.pools:
+            for record in self.invoice_obj.pools:
+                if self.get_price(discount, record):
+                    total_price += self.get_price(discount, record)
+                else:
+                    return ''
```
ingkebil: Since python 3, we are able to give hints as to what the return type of a function is. What is the return type of this function? :)

cg/server/invoices/templates/invoices/layout.html
```diff
@@ -46,7 +46,20 @@
...
+        {% for message in messages %}
+        <div class="alert alert-danger alert-dismissible fade show" role="alert">
+            {{ message }}
+            <button type="button" class="close" data-dismiss="alert" aria-label="Close">
+              <span aria-hidden="true">&times;</span>
+            </button>
+          </div>
+        {% endfor %}
+        </ul>
+      {% endif %}
```
ingkebil: A `ul` cannot have a `div` as direct child: https://stackoverflow.com/questions/11755628/can-i-use-div-as-a-direct-child-of-ul

Can we use a `div` instead of a `ul`?

cg/server/invoices/templates/invoices/new.html
```diff
@@ -83,15 +87,19 @@ <h1>Make new invoice ({{record_type}})</h1>
...
+      }
+      
+      checkTotal()
     }
-  checkTotal()
-  }
-</script>
+  </script>
 
 <script language="JavaScript">
```
ingkebil: Even though this works perfectly, there is no need to wrap all javascript functions in separate `script` tags. You could wrap them in just one.

cg/server/invoices/templates/invoices/new.html
```diff
@@ -83,15 +87,19 @@ <h1>Make new invoice ({{record_type}})</h1>
...
 {% block scripts %}
-<script language="JavaScript">
-  function toggle() {
-    checkboxes = document.getElementsByName('records');
-    for(var checkbox in checkboxes) {
-      checkboxes[checkbox].checked = !checkboxes[checkbox].checked;
+
+<script language="JavaScript"> 
+    function check_all() {
+      checkboxes = document.getElementsByName('records');
```
ingkebil: Any reason why you're not using the jQuery selector here?

cg/server/invoices/views.py
```diff
@@ -11,68 +11,95 @@
...
+    ki_excel_file = request.files.get('KI_excel')
+    if ki_excel_file:
+        invoice_obj.excel_ki = ki_excel_file.stream.read()
+    db.commit()
+    return url_for('invoices.invoice', invoice_id=invoice_id)
 
+@BLUEPRINT.route('/', methods=['GET', 'POST'])
+def index():
+    """Show invoices."""
+    if not loged_in(): return redirect(url_for('admin.index'))
```
ingkebil: *logged

cg/store/api/find.py
```diff
@@ -274,6 +274,12 @@ def invoice(self, invoice_id: int) -> models.Invoice:
         """Fetch an invoice."""
         return self.Invoice.get(invoice_id)
 
+    def invoice_samples(self, *, invoice_id: int = None) -> List[models.Sample]:
+        pools = self.Pool.query.filter_by(invoice_id=invoice_id).all()
+        samples = self.Sample.query.filter_by(invoice_id=invoice_id).all()
+        return pools + samples
```
ingkebil: hehe, this is so clean :)

cg/store/api/trends.py
```diff
@@ -39,19 +38,49 @@ def samples_per_month(self, year):
...
+                .filter(
+                models.Sample.received_at > self.get_from_date(year),
+                models.Sample.received_at < self.get_until_date(year),
+                models.Sample.received_at,
+                models.Application.category,
+            )
+                .group_by('category', 'month_no')
+        )
+
+        for application_cathegory, results in groupby(query, key=lambda result: result.category):
```
ingkebil: this is really neat

cg/meta/invoice.py
```diff
@@ -8,59 +8,80 @@
 
 class InvoiceAPI():
 
-    def __init__(self, db: Store, lims_api: lims.LimsAPI):
+    def __init__(self, db: Store, lims_api: lims.LimsAPI, invoice_obj: models.Invoice):
```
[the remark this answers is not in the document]
mayabrandi: I see your point. Didnt think of that the class was called API. I thought of it more as an Invoice class. With functions to generate whats needed to build one invoice. Then it made more sense to define the invoice model instance in the init.  I was actually thinking of putting even mote things into the init, instead of retuning bunches of stuff from all the functions. But then I thought I shouldnt spend to much time on code finish. 

So after this explanation, you want me to take away the definition of the invoice model instance from the init? Or you want me to rename the class from InvoiceAPI to Invoice? Or i should just leave it as it is?

Im open for anything :)

cg/store/api/trends.py
patrikgrenfeldt: this code belongs to another branch?

PR #209, opened 2018-10-03
Fix errr when passing sampleinfo_path to `clean mip`

ingkebil:
Seems like you need to open the file and pass the file obj to a click method.


barrystokman:
Does the file need to be closed again? Using ```with open(sampleinfo_path, 'r') as sample_info:``` for instance?

ingkebil:
Good point! Changing.

ingkebil:
Tested on stage and it works :)

PR #218, opened 2018-10-19
made pool name non numeric

patrikgrenfeldt:
EOL states that nothing downstream should be affected by this.

How to test:
0. install clinical-api on stage on clinical-db
1. login in OP
2. upload an RML-OF that has samples with non-numeric poolnames

Expected outcome:
    Each sample has the non-numeric poolname shown in the OrderPortal


patrikgrenfeldt:
@ingkebil Houston, do we have clearance for takeoff?

PR #248, opened 2018-12-03
Add pull request template

ingkebil:
I would like to play around with pull request templates as a way to save our test results with PRs. This template would be prefilled in the description of the pull request at the time of opening. The template is based on what Patrik has been using so far.

Now that I write it like that, it seems that my suggested template only handles the 'how to test' and not 'how to document'.


emmser:
I don't understand what you want me to review? As you say this template is not enough for rolling releases instead of test specifications, but I don't think that your intention was to create that template now?

PR #257, opened 2018-12-07
Allow non saved fields for existing samples to be optional

patrikgrenfeldt:
This PR fixes the problem to submit orders with families that you have added existing samples to.

How to test:
1. run test in https://github.com/Clinical-Genomics/cgweb/pull/98

Expected outcome:
Success


cg/meta/orders/schema.py
```diff
@@ -110,22 +110,22 @@ def validate(self, value):
 
     # This information is required
     'name': validators.RegexValidator(NAME_PATTERN),
-    'container': validators.Any(CONTAINER_OPTIONS),
+    'container': validators.Optional(TypeValidator(str, allow_none=True), None),
```
ingkebil: Should prolly move to the "Order portal specific" block on line 109

cg/meta/orders/schema.py
```diff
@@ -110,22 +110,22 @@ def validate(self, value):
...
     'data_analysis': str,
     'application': str,
     'sex': validators.Any(SEX_OPTIONS),
     'family_name': validators.RegexValidator(NAME_PATTERN),
     'require_qcok': bool,
     'tumour': bool,
-    'source': str,
+    'source': validators.Optional(TypeValidator(str, allow_none=True), None),
```
ingkebil: Should prolly move to the "Order portal specific" block on line 109

cg/meta/orders/schema.py
```diff
@@ -110,22 +110,22 @@ def validate(self, value):
...
     'priority': validators.Any(PRIORITY_OPTIONS),
 
     # Required if Plate
     'container_name': validators.Optional(str, None),
     'well_position': validators.Optional(TypeValidator(str, allow_none=True), None),
 
     # This information is required for Balsamic analysis (cancer)
-    'capture_kit': validators.Any(CAPTUREKIT_CANCER_OPTIONS),
+    'capture_kit': validators.Optional(TypeValidator(str, allow_none=True), None),
```
ingkebil: hehe, it's not required when it's optional ;) Don't have a good suggestion on how to solve this tho

cg/meta/orders/schema.py
```diff
@@ -152,21 +152,21 @@ def validate(self, value):
 
     # required
     'name': validators.RegexValidator(NAME_PATTERN),
-    'capture_kit': validators.Any(CAPTUREKIT_OPTIONS),
+    'capture_kit': validators.Optional(TypeValidator(str, allow_none=True), None),
```
ingkebil: optional, not required, should prolly move to an optional block

cg/meta/orders/schema.py
```diff
@@ -152,21 +152,21 @@ def validate(self, value):
...
 
     # Required if data analysis in Scout
     'panels': ListValidator(str, min_items=0),
     # todo: find out if "Additional Gene List" is "lost in translation", implement in OP or remove from OF
     'status': validators.Any(STATUS_OPTIONS),
 
     # Required if samples are part of trio/family
-    'mother': validators.Optional(validators.RegexValidator(NAME_PATTERN), None),
-    'father': validators.Optional(validators.RegexValidator(NAME_PATTERN), None),
+    'mother': validators.Optional(TypeValidator(str, allow_none=True), None),
```
ingkebil: So, this used to be a regexvalidator with a pattern. Now it is an optional string. Does that mean we cannot have an optional regex?

cg/meta/orders/schema.py
patrikgrenfeldt: The meaning of "Order portal specific" in this context is whether the attribute exists in the Orderform or not and thus I think it is placed under the correct comment? 

What might be a bit confusing here is that some fields are required when the information comes from the Orderform but is optional when you select an already existing sample in the Orderportal.  Maybe we should implement something along the lines of OrValidator( 'internal_id', 'container' ) ?

cg/meta/orders/schema.py
patrikgrenfeldt: I think I will update the comments..

cg/meta/orders/schema.py
patrikgrenfeldt: Implemented now :)

patrikgrenfeldt:
@ingkebil can you review it again, the changes and comments that I made?

emiliaol:
Also tested with the cgweb PR with great success!

PR #301, opened 2019-03-15
List all panels for all customers

patrikgrenfeldt:
This PR lets the user select all panels in the Orderportal

How to test:
1. install clinical-api and cgweb on stage (add instructions on how to as needed) on clinical-db
1. login to OP stage
1. create an MIP order
1. add family
1. open panel list

Expected outcome:
All the panels in statusDB should be selectable

### DIFF
--- a/cg/store/api/find.py
+++ b/cg/store/api/find.py
@@ -172,6 +172,10 @@ class FindHandler:
         """Find a panel by abbreviation."""
         return self.Panel.query.filter_by(abbrev=abbrev).first()
 
+    def panels(self):
+        """Returns all panels."""
+        return self.Panel.query.order_by(models.Panel.abbrev)
+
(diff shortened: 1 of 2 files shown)


cg/store/api/find.py
```diff
@@ -172,6 +172,10 @@ def panel(self, abbrev):
         """Find a panel by abbreviation."""
         return self.Panel.query.filter_by(abbrev=abbrev).first()
 
+    def panels(self):
+        """Returns all panels."""
+        return self.Panel.query.order_by(models.Panel.abbrev)
```
ingkebil: I guess the panels are actually really stored in scout, right? Do you know when we are creating the panels in statusdb?

patrikgrenfeldt: I think we are adding them one-by-one

PR #318, opened 2019-03-29
accepts only OF rml 1604:9

patrikgrenfeldt:
This PR accepts only OF rml 1604:9

How to test:
0. install branch both on clinical-api and cgweb stage on clinical-db
1. login to OP
1. import an 1604:9 OF with some samples

Expected outcome:
The samples are imported to be further handled in the OP


emiliaol:
So I should make the suggested changes of the new OF in AM and make sure that it is ready to be published, test that OF and then we publish/update at the same time?

patrikgrenfeldt:
Yes :)

PR #334, opened 2019-04-17
Uploads delivery report separate from scout upload

patrikgrenfeldt:
This PR separates the upload of the delivery report from the regular case upload.

How to test separation from regular upload:
0. install this branch of cg on stage on hasta
1. us
1. cg upload -f [CASE-ID]

Expected outcome:
The case uploads without mention delivery report

How to test separation into its own upload:
1. us
1. cg upload delivery-report [CASE-ID]

Expected outcome:
A delivery report is created and uploaded to housekeeper and scout

### DIFF
--- a/cg/apps/scoutapi.py
+++ b/cg/apps/scoutapi.py
@@ -52,10 +52,11 @@ class ScoutAPI(MongoAdapter):
             panel genes: panel genes list
         """
         gene_panel = self.gene_panel(panel_id=panel_id, version=version)
+
         if gene_panel:
             return gene_panel.get('genes')
-        else:
-            return None
+
+        return None
 
     def get_cases(self, case_id=None, institute=None, reruns=None, finished=None,
                   causatives=None, research_requested=None, is_research=None, status=None):
@@ -76,9 +77,9 @@ class ScoutAPI(MongoAdapter):
         return models
 
     def upload_delivery_report(self,
-                             report_path: str,
-                             case_id: str,
-                             update: bool = False):
+                               report_path: str,
+                               case_id: str,
+                               update: bool = False):
         """ Load a delivery report into a case in the database
 
         If the report already exists the function will exit.
--- a/cg/apps/scoutapi.py
+++ b/cg/apps/scoutapi.py
@@ -88,7 +88,7 @@ class ScoutAPI(MongoAdapter):
 
         Args:
             report_path (string):       Path to delivery report
-            case_id     (string):       Optional case identifier
+            case_id     (string):       Case identifier
             update      (bool):         If an existing report should be replaced
 
         Returns:
--- a/cg/apps/scoutapi.py
+++ b/cg/apps/scoutapi.py
@@ -4,7 +4,7 @@ from typing import List
 
 from pymongo import MongoClient
 from scout.adapter.mongo import MongoAdapter
-from scout.adapter.mongo.panel import PanelHandler
+from scout.load.report import load_delivery_report
 from scout.export.panel import export_panels as scout_export_panels
 from scout.load import load_scout
 from scout.parse.case import parse_case_data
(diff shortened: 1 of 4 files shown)


patrikgrenfeldt:
The Travis PR build fails because of the lint warning: "ScoutAPI.get_cases: Too many arguments (9/8)", I'd like to avoid fixing that since I guess it involves changing Scout as well and I'd like to do that when developing in scout instead of the other way around.

ingkebil:
> The Travis PR build fails because of the lint warning: "ScoutAPI.get_cases: Too many arguments (9/8)", I'd like to avoid fixing that since I guess it involves changing Scout as well and I'd like to do that when developing in scout instead of the other way around.

Alright, let's add a linting exception to this errr

cg/apps/scoutapi.py
```diff
@@ -52,10 +52,14 @@ def get_genes(self, panel_id: str, version: str=None) -> list:
             panel genes: panel genes list
         """
         gene_panel = self.gene_panel(panel_id=panel_id, version=version)
-        return gene_panel.get('genes')
 
-    def get_cases(self, case_id=None, institute=None, reruns=None, finished=None, causatives=None, research_requested=None,
-          is_research=None, status=None):
+        if gene_panel:
```
ingkebil: Is there any way we can have `self.gene_panel` raise an exception when no panels are found instead of returning None? And then we catch the exception here instead of checking if the expression evaluates to False.

cg/apps/scoutapi.py
```diff
@@ -52,10 +52,14 @@ def get_genes(self, panel_id: str, version: str=None) -> list:
...
+            return gene_panel.get('genes')
+
+        return None
```
ingkebil: Same here: can we raise an exception instead of returning None?

cg/apps/scoutapi.py
```diff
@@ -66,8 +70,31 @@ def get_cases(self, case_id=None, institute=None, reruns=None, finished=None, ca
...
+                               update: bool = False):
+        """ Load a delivery report into a case in the database
+
+        If the report already exists the function will exit.
+        If the user want to load a report that is already in the database
+        'update' has to be 'True'
+
+        Args:
+            report_path (string):       Path to delivery report
+            case_id     (string):       Optional case identifier
```
ingkebil: According to the function signature, it doesn't look optional ;)

cg/cli/upload.py
```diff
@@ -147,9 +149,20 @@ def delivery_report(context, family_id, print_console):
...
+            _add_delivery_report_to_scout(context, added_file.full_path, family_id)
+            click.echo(click.style('uploaded to scout', fg='green'))
             _update_delivery_report_date(status_api, family_id)
+        else:
+            click.echo(click.style('already uploaded to housekeeper, skipping'))
+
+
+def _add_delivery_report_to_scout(context, path, case_id):
+    scout_api = scoutapi.ScoutAPI(context.obj)
+    scout_api.upload_delivery_report(path, case_id, update=True)
```
ingkebil: Is there a way we can get somesort of status of success returned? e.g. True when the upload succeeds and False when it fails?

barrystokman:
#### Prepare test data in Scout:

Remove case 18081 (dearzebra) from Scout:
image
image

Make sure there's no delivery report:
image

#### Test separation from regular upload:

```cg upload -f dearzebra```

image
image
image

Expected outcome:
The case uploads without mention delivery report :heavy_check_mark: 

#### Test separation into its own upload:

``` cg upload delivery-report dearzebra```

image
image

Expected outcome:
A delivery report is created and uploaded to housekeeper and scout :heavy_check_mark:

Both tests passed.

cg/apps/scoutapi.py
patrikgrenfeldt: removed

cg/apps/scoutapi.py
patrikgrenfeldt: fixed

cg/apps/scoutapi.py
patrikgrenfeldt: removed

cg/cli/upload.py
patrikgrenfeldt: fixed

PR #339, opened 2019-04-25
Stop storing instead of duplicating analysis

patrikgrenfeldt:
This PR makes cg throw an error if try to store an analysis that has already been created in status-db

How to test duplication check:
1. us
1. find/create case that has been stored already
1. remove bundle in `housekeeper delete bundle [casename]`
1. `cg store analysis [path_to_analysis_config]`

Expected outcome:
cg should raise an AnalysisDuplicationError exception 

How to test normal behaviour:
1. us
1. find/create case that has NOT been stored already
1. `cg store analysis [path_to_analysis_config]`

Expected outcome:
cg should store the analysis, i.e. create analysis record in status db and upload to housekeeper

### DIFF
--- a/cg/cli/store.py
+++ b/cg/cli/store.py
@@ -6,7 +6,7 @@ from pathlib import Path
 import click
 
 from cg.apps import hk, tb
-from cg.exc import AnalysisNotFinishedError
+from cg.exc import AnalysisNotFinishedError, AnalysisDuplicationError
 from cg.store import Store
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


cg/cli/store.py
```diff
@@ -77,7 +77,10 @@ def _include_files_in_housekeeper(bundle_obj, context, hk_api, version_obj):
 def _add_new_complete_analysis_record(bundle_data, family_obj, status, version_obj):
 
     pipeline = family_obj.links[0].sample.data_analysis
-    pipeline = pipeline if pipeline else 'mip' # TODO remove this default from here
+
+    if status.analysis(family=family_obj, started_at=version_obj.created_at):
```
ingkebil: haha, not assigning it was the solution to the faulty query? :D

cg/exc.py
```diff
@@ -14,6 +14,10 @@ class AnalysisNotFinishedError(CgError):
     pass
 
 
+class AnalysisDuplicationError(CgError):
+    """Error related to trying to create duplicate analysis objects"""
+
```
ingkebil:
```suggestion
    pass
```

cg/cli/store.py
patrikgrenfeldt: Well, the faulty query was due to uninitiated mock object

barrystokman:
Tested and approved!

cg/exc.py
patrikgrenfeldt: removed because of lint warning

PR #344, opened 2019-04-26
adds upload delivery-report to scout cli

patrikgrenfeldt:
this PR adds CLI command to upload delivery-report to scout.

How to test:

1. us
1. find case with report in hk and case exists in scout
1. Change the report that housekeeper points to 'housekeeper get - V CASE-ID'
1. cg upload delivery-report-to-scout CASE_ID

Expected outcome:
delivery-report CASE_ID uploaded to scout

### DIFF
--- a/cg/cli/upload.py
+++ b/cg/cli/upload.py
@@ -153,7 +153,7 @@ def delivery_report(context, family_id, print_console):
 
         if added_file:
             click.echo(click.style('uploaded to housekeeper', fg='green'))
-            _add_delivery_report_to_scout(context, added_file.full_path, family_id)
+            context.invoke(delivery_report_to_scout, case_id=family_id)
             click.echo(click.style('uploaded to scout', fg='green'))
             _update_delivery_report_date(status_api, family_id)
         else:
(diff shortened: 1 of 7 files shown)


cg/cli/upload.py
```diff
@@ -190,6 +190,29 @@ def _update_delivery_report_date(status_api, family_id):
...
+        _add_delivery_report_to_scout(context, report, case_id)
+
+def _get_delivery_report_from_hk(hk_api: hk.HousekeeperAPI, family_id):
+    delivery_report_tag_name = 'delivery-report'
+    version_obj = hk_api.last_version(family_id)
+    uploaded_delivery_report_files = hk_api.get_files(bundle=family_id,
+                                                      tags=[delivery_report_tag_name],
+                                                      version=version_obj.id)
+
+    return uploaded_delivery_report_files[0].full_path
```
ingkebil: Does this function raise an exception when the delivery report is not found or does it return `None`? If it is the latter, can we raise an exception?

ingkebil:
Did you see the tests pass but throw 16000 (!) warnings?

patrikgrenfeldt:
Yes unfortunately.. deprecation warnings, Barry has been looking at it

Den lör 27 apr. 2019 21:28Kenny Billiau <notifications@github.com> skrev:

cg/cli/upload.py
patrikgrenfeldt: We will have an exception at this point if no delivery-report exists

barrystokman:
> this PR adds CLI command to upload delivery-report to scout.
> 
> How to test:
> 0. install on stage (add instructions on how to as needed) :heavy_check_mark: 
>
> us :heavy_check_mark: 
> find case with report in hk but not in scout

ableakita  has delivery report in HK:

image

ableakita has Scout id 18069:
image

so we delete it from scout:
image
image

 

> cg upload delivery-report-to-scout CASE_ID ❌
image

> Expected outcome:
> delivery-report CASE_ID uploaded to scout

Test failed, please check if all the steps were executed correctly.

cg/cli/upload.py
ingkebil: Will the return raise an exception because the list has no items, if so, I'd prefer an exception that says no delivery report is found instead.

patrikgrenfeldt: fixed

patrikgrenfeldt:
@barrystokman, I think we have to change the file that housekeeper points to

PR #393, opened 2019-07-16
Wrong family/sample name condition in analysis link

hassanfa:
This PR adds a fix for the leaking wrong family/name.

Note: still working on the test as it is failing. I don't know much about fixtures on cg, and I get the following error: `<Result KeyError('housekeeper')>`. Any input or help is welcome :-). Of course, the def add_* are shameless copy from other tests.

**How to test**:

TBA

**Expected outcome**:
`cg analysis link -f wrong_family_name` should provide a proper error message

**Review:**
- [ ] code approved by
- [ ] tests executed by

This is patch|minor|major **version bump** because ...

### DIFF
--- a/cg/cli/analysis.py
+++ b/cg/cli/analysis.py
@@ -104,10 +104,16 @@ def link(context, family_id, sample_id):
     if family_id and (sample_id is None):
         # link all samples in a family
         family_obj = context.obj['db'].family(family_id)
+        if family_obj is None:
+            LOG.error(f"Family '{family}' does not exist.")
+            context.abort()
         link_objs = family_obj.links
     elif sample_id and (family_id is None):
         # link sample in all its families
         sample_obj = context.obj['db'].sample(sample_id)
+        if sample_obj is None:
+            LOG.error(f"Sample '{family}' does not exist.")
+            context.abort()
         link_objs = sample_obj.links
     elif sample_id and family_id:
         # link only one sample in a family
--- a/cg/cli/analysis.py
+++ b/cg/cli/analysis.py
@@ -112,7 +112,7 @@ def link(context, family_id, sample_id):
         # link sample in all its families
         sample_obj = context.obj['db'].sample(sample_id)
         if sample_obj is None:
-            LOG.error(f"Sample '{family}' does not exist.")
+            LOG.error(f"Sample '{sample}' does not exist.")
             context.abort()
         link_objs = sample_obj.links
     elif sample_id and family_id:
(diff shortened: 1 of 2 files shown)


hassanfa:
pinging @ingkebil @barrystokman

ingkebil:
Brilliant! Thanks a bunch :)

I'll code review this week as well as look into the failing tests.

hassanfa:
Ok, it is a bit tricky to make tests work. I think I'll need help on this when you have time :-) It'll be also good for me to understand how conftests are setup so I can contribute more in future. Any time you guys have time would work me as well.

cg/cli/analysis.py
```diff
@@ -112,7 +112,7 @@ def link(context, family_id, sample_id):
         # link sample in all its families
         sample_obj = context.obj['db'].sample(sample_id)
         if sample_obj is None:
-            LOG.error(f"Sample '{family}' does not exist.")
+            LOG.error(f"Sample '{sample}' does not exist.")
```
patrikgrenfeldt: 1. Should it be sample_id perhaps?
2. Please use %s to outputing when logging to avoid the lint warning
:)

cg/cli/analysis.py
```diff
@@ -117,6 +123,9 @@ def link(context, family_id, sample_id):
         context.abort()
 
     for link_obj in link_objs:
+        if link_obj is None:
+            LOG.error(f"Sample '{sample}' or family '{family}' does not exist.")
```
patrikgrenfeldt: Should it be sample_id and family_id?

cg/cli/analysis.py
ingkebil: In case the linter doesn't say why: https://docs.python.org/3/howto/logging.html#optimization

cg/cli/analysis.py
hassanfa: Correct!

cg/cli/analysis.py
hassanfa: Yes, it should be sample_id.

And about log, I took it from other examples in cg code base to use f-string :-P. I'll change it to %s 😛

ingkebil: yeah, this indeed hasn't been applied consistently throughout the code base ;)

patrikgrenfeldt:
@hassanfa, I think that the test error occurred because this cli tries to instantiate the AnalysisAPI which in turn requires housekeeper. Maybe we need to use a config instead of just the db-uri in the test..

hassanfa:
> @hassanfa, I think that the test error occurred because this cli tries to instantiate the AnalysisAPI which in turn requires housekeeper. Maybe we need to use a config instead of just the db-uri in the test..

Cool! I'll look into it later this week. 👍

patrikgrenfeldt:
@hassanfa can we close this issue?

hassanfa:
👍

PR #406, opened 2019-08-20
Order RNA in the order-portal

patrikgrenfeldt:
These PR's adds the possibility to order MIP RNA in the OP.

**How to test**:
1. install on stage of clinical-db `bash servers/resources/clinical-db.scilifelab.se/update-clinical-api-stage.sh order-rna`
1. install on stage of clinical-db `bash servers/resources/clinical-db.scilifelab.se/update-cgweb-stage.sh order-rna`
1. update stage database by running the sql script that add fields in sample for RNA
1. create an order via the 1508:19 orderform, from AM
1. upload in OP-stage 
1. submit order 

**Expected outcome**:
Order should be accepted and we should end up in the family/cases view

**Review:**
- [ ] ~~tests executed by~~

This is the part of the RNA MIP validation described here: ams#2016


cg/meta/orders/api.py
```diff
@@ -171,6 +171,10 @@ def submit_mip_balsamic(self, data: dict) -> dict:
         """Submit a batch of samples for sequencing and analysis."""
         return self.submit_family_samples(data)
 
+    def submit_mip_rna(self, data: dict) -> dict:
+        """Submit a batch of samples for sequencing and analysis."""
+        return self.submit_family_samples(data)
```
ingkebil:
```suggestion
        return self.submit_case_samples(data)
```

cg/meta/orders/api.py
```diff
@@ -171,6 +171,10 @@ def submit_mip_balsamic(self, data: dict) -> dict:
         """Submit a batch of samples for sequencing and analysis."""
         return self.submit_family_samples(data)
```
ingkebil:
```suggestion
        return self.submit_case_samples(data)
```

ingkebil: this suggestion can be applied to most of the functions above this line as well ;)

cg/store/models.py
```diff
@@ -464,37 +464,38 @@ class Pool(Model):
...
-    beaconized_at = Column(types.Text)
-
-    created_at = Column(types.DateTime, default=dt.datetime.now)
-    customer_id = Column(ForeignKey('customer.id', ondelete='CASCADE'), nullable=False)
-    application_version_id = Column(ForeignKey('application_version.id'), nullable=False)
-
-    deliveries = orm.relationship('Delivery', backref='sample')
+    sex = Column(types.Enum('male', 'female', 'unknown'), nullable=False)
+    ticket_number = Column(types.Integer)
+    time_point = Column(types.Integer)
```
ingkebil: The alter column says time_point is nullable. Just pointing it out, don't know what the best option is

cg/store/models.py
```diff
@@ -464,37 +464,38 @@ class Pool(Model):
...
+    beaconized_at = Column(types.Text)
+    capture_kit = Column(types.String(64))
+    comment = Column(types.Text)
+    created_at = Column(types.DateTime, default=dt.datetime.now)
+    customer_id = Column(ForeignKey('customer.id', ondelete='CASCADE'), nullable=False)
+    data_analysis = Column(types.String(16))
+    delivered_at = Column(types.DateTime)
+    deliveries = orm.relationship('Delivery', backref='sample')
+    downsampled_to = Column(types.BigInteger)
+    from_sample = Column(types.String(128))
```
ingkebil: The alter table statement says this column is nullable

tests/apps/lims/test_apps_lims_orderform.py
```diff
@@ -358,6 +356,58 @@ def test_parsing_mip_balsamic_orderform(mip_balsamic_orderform):
     assert mother_sample.get('mother') is None
 
 
+def test_parsing_mip_rna_orderform(mip_rna_orderform):
```
ingkebil: The problem with having 30 asserts in one test is that once the first one fails, we will not know if the others are failing as well. Does it make sense to split the asserts into their own test?

cg/meta/orders/api.py
patrikgrenfeldt: fixed

cg/store/models.py
patrikgrenfeldt: must be nullable since we use same record for MIP and balsamic which lacks timepoint

cg/meta/orders/api.py
patrikgrenfeldt: fixed

cg/store/models.py
patrikgrenfeldt: must be nullable since we use same record for MIP and balsamic which lacks from_sample

tests/apps/lims/test_apps_lims_orderform.py
patrikgrenfeldt: I guess we could but if the first fails then youll be fixing and testing until your're through anyways

cg/store/models.py
ingkebil: aha, alright, `nullable=True` is the default :)

tests/apps/lims/test_apps_lims_orderform.py
ingkebil: yes. It's up to you, I'm just pointing it out.

Source: https://testandcode.com/65

ingkebil:
I approve the code presuming fixing the tests will not add any new logic. Good work!

PR #409, opened 2019-08-23
add link to family sample in sample and family views

patrikgrenfeldt:
This PR add link to family sample in sample and family views

**How to test Family link to FamilySample**:
1. install on stage of the clinical-db machine: `bash servers/resources/clinical-db.scilifelab.se/update-clinical-api-stage.sh link-family-and-sample-to-familysample`
1. login to clinical-api-stage
1. go to family view
1. click the link in the column internal id in a family

**Expected outcome**:
The familysample view is shown for the family that was chosen

**How to test Sample link to FamilySample**:
1. go to sample view
1. click the link in the column internal id in a sample

**Expected outcome**:
The familysample view is shown for the sample that was chosen

**Review:**

This is minor **version bump** because it adds functionality in a backwards compatible manner

### DIFF
--- a/cg/server/admin.py
+++ b/cg/server/admin.py
@@ -24,7 +24,7 @@ def view_family_sample_link(unused1, unused2, model, unused3):
     return Markup(
         u"<a href='%s'>%s</a>" % (
             url_for('familysample.index_view', search=model.internal_id),
-            model
+            model.internal_id
         )
     )
 
--- a/cg/server/admin.py
+++ b/cg/server/admin.py
@@ -17,6 +17,18 @@ class BaseView(ModelView):
         return redirect(url_for('google.login', next=request.url))
 
 
+def view_family_sample_link(unused1, unused2, model, unused3):
+    """column formatter to open the family-sample view"""
+    del unused1, unused2, unused3
+
+    return Markup(
+        u"<a href='%s'>%s</a>" % (
+            url_for('familysample.index_view', search=model.internal_id),
+            model
+        )
+    )
+
+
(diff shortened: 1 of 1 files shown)


cg/server/admin.py
```diff
@@ -17,6 +17,18 @@ def inaccessible_callback(self, name, **kwargs):
         return redirect(url_for('google.login', next=request.url))
 
 
+def view_family_sample_link(unused1, unused2, model, unused3):
```
ingkebil: I guess the `unsused` ones are params that are passed by default?

ingkebil:
Code looks good!

cg/server/admin.py
patrikgrenfeldt: yes, I'm leaning towards another way of doing it, to use underscore (_, _, model, _), but then we can't delete them like we do now, what's your preference?

patrikgrenfeldt:
Test outcome: 
Family view with one searched family: 
image

Link in family view leads to FamilySample view:
image

Sample view with one searched sample:
image

Link in sample view leads to FamilySample view:
image

cg/server/admin.py
ingkebil: The latter definitely is how I've seen everyone else dealing with unused and unneeded variables. So let's go for that

ingkebil:
I'm approving pending the requested change

cg/server/admin.py
patrikgrenfeldt: It turns out that if you use the underscore more than once in the signature then it's considered to be a name collision...

PR #423, opened 2019-09-10
Microbial invoice issue - bug fix

mayabrandi:
This is a old discussion that needs to be brought up again. For some reason the microbial samples are not found when the the filter is set to "is None", but "== None" does the trick. Vallteri needs to invoice microbial samples and the fix is therefor important to get into prod soon.

**How to test and Expected outcome**:
Install on stage and go to https://clinical-api-stage.scilifelab.se/invoices/new/Microbial
The dropdown should show a list of customers to invoice. See screen shot. This compared to the [current version in production](https://clinical-api.scilifelab.se/invoices/new/Microbial), where no customers show up in the scroll down list. 

Skärmavbild 2019-09-10 kl  15 03 13

**Review:**
- [x] tests executed by @ingkebil 

This is patch **version bump** because no new functionality nor change in functionality.

### DIFF
--- a/cg/store/api/status.py
+++ b/cg/store/api/status.py
@@ -513,7 +513,7 @@ class StatusHandler(BaseHandler):
         records = (
             self.MicrobialSample.query.filter(
                 models.MicrobialSample.delivered_at is not None,
-                models.MicrobialSample.invoice_id is None
+                models.MicrobialSample.invoice_id == None
             )
         )
         customers_to_invoice = [record.microbial_order.customer for record in records.all()]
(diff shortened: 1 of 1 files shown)


ingkebil:
The linting fails on this change. Could you add a comment that disables the linting for this line so that we avoid changing your patch back at a later date?

It's something like `# pylint: disable=C0321` (with the correct number)

ingkebil:
I've aded the pylint ignore comment. The build still fails but at least it is documented in the code that we should leave that line as is.

Approving, merging, and deploying

ingkebil:
@vwirta I presume you can now invoice microbial

vwirta:
thank you both @ingkebil @mayabrandi !
Seems to work!

PR #434, opened 2019-09-22
Don't exit on failing store

ingkebil:
This PR adds a fix for deviation Prod#47 which halts storing analyses on missing files.

**How to test**:
1. install on stage of the coffee machine: `bash servers/resources/hasta.scilifelab.se/update-cg-stage.sh master`
1. activate stage: `us`
1. Copy latest production cg to stage: (on clinical-db) dbcopy-prod-to-stage.sh cg
1. Copy latest production trailblazer to stage: (on clinical-db) dbcopy-prod-to-stage.sh trailblazer
1. (On hasta) copy poeticmite from production to stage: `cp /home/proj/production/rare-disease/cases/poeticmite/ /home/proj/stage/rare-disease/cases`
1. append .bak to the mip config in `/home/proj/stage/rare-disease/cases/poeticmite`
1. run: `cg store completed`

Checkpoint!

1. install on stage: `bash servers/resources/hasta.scilifelab.se/update-cg-stage.sh production#47`. OK, this command will fail. (Never use a hashtag in the branch name.) Solution: git clone cg to ~/git/<yourname>/cg, switch to the right branch and pip install it from there.
1. run: `cg store completed`

**Expected outcome**:
1. first checkpoint: storing analyses should exit on missing file!
1. second checkpoint:
  - storing analyses should continue.
  - the exit code should be 1: `echo $?`

**Review:**
- [ ] code approved by
- [ ] tests executed by
- [ ] "Merge and deploy" approved by @ingkebil 

This is minor **version bump** as it changes existing functionality in a backwards compatible manner.


cg/cli/store.py
```diff
@@ -108,11 +109,19 @@ def _add_new_analysis_to_the_status_api(bundle_obj, status):
 def completed(context):
     """Store all completed analyses."""
     hk_api = context.obj['hk_api']
+    exit_code = 0
     for analysis_obj in context.obj['tb_api'].analyses(status='completed', deleted=False):
         existing_record = hk_api.version(analysis_obj.family, analysis_obj.started_at)
         if existing_record:
-            LOG.debug(f"analysis stored: {analysis_obj.family} - {analysis_obj.started_at}")
+            LOG.debug("analysis stored: %s - %s", analysis_obj.family, analysis_obj.started_at)
```
patrikgrenfeldt:
```suggestion
            LOG.info("analysis stored: %s - %s", analysis_obj.family, analysis_obj.started_at)
```

cg/cli/store.py
```diff
@@ -108,11 +109,19 @@ def _add_new_analysis_to_the_status_api(bundle_obj, status):
...
             continue
         click.echo(click.style(f"storing family: {analysis_obj.family}", fg='blue'))
         with Path(analysis_obj.config_path).open() as config_stream:
-            context.invoke(analysis, config_stream=config_stream)
+            try:
+                context.invoke(analysis, config_stream=config_stream)
+            except Exception: # pylint: disable=broad-except
```
patrikgrenfeldt: I guess we're trying to handle something here based on what actually went wrong in production, could we handle that exception like that and let the rest pass instead of handling all exceptions the same?

cg/cli/store.py
```diff
@@ -108,11 +109,19 @@ def _add_new_analysis_to_the_status_api(bundle_obj, status):
 def completed(context):
     """Store all completed analyses."""
     hk_api = context.obj['hk_api']
+    exit_code = 0
```
patrikgrenfeldt:
```suggestion
    exit_code = FAIL
```
suggestion something like this to avoid magic numbers

cg/cli/store.py
```diff
@@ -108,11 +109,19 @@ def _add_new_analysis_to_the_status_api(bundle_obj, status):
...
             continue
         click.echo(click.style(f"storing family: {analysis_obj.family}", fg='blue'))
         with Path(analysis_obj.config_path).open() as config_stream:
-            context.invoke(analysis, config_stream=config_stream)
+            try:
+                context.invoke(analysis, config_stream=config_stream)
+            except Exception: # pylint: disable=broad-except
+                LOG.error("uploading family failed: %s", analysis_obj.family.internal_id,
+                          exc_info=True)
+                exit_code = 1
```
patrikgrenfeldt:
```suggestion
                exit_code = SUCCESS
```
suggestion something like this to avoid magic numbers

cg/cli/store.py
ingkebil: I don't know what exception to expect here. I don't want the for-loop to stop and I want to be notified something went wrong, so the exception information is forwarded to the logger and the exit code is set to 1.

This is how we do it for uploading analyses btw.

cg/cli/store.py
```diff
@@ -108,11 +109,19 @@ def _add_new_analysis_to_the_status_api(bundle_obj, status):
 def completed(context):
     """Store all completed analyses."""
     hk_api = context.obj['hk_api']
+    exit_code = FAIL
```
ingkebil: @patrikgrenfeldt where is this constant defined?

patrikgrenfeldt: nowhere yet, that's why I wrote "something like..."

cg/cli/store.py
patrikgrenfeldt: `            try:
                context.invoke(analysis, config_stream=config_stream
            except AnalysisException: LOG.warning("Could not create config for %s", config_stream) 
            except Exception e: LOG.error("Unknown error for %s, %s", config_stream, e)

patrikgrenfeldt: I'm think the intention is handle different things differently, not to swallow everything the same, I'd rather not silence the linter but push this to master with a warning. 
I.E. we don't have to work our asses of to get the linting down to zero if we "know" the linter is right and there is a proper solution, there is just not enough time/motivation to fix it at the current point in time..

patrikgrenfeldt: What do you say?

ingkebil: I see where you're trying to go.

This is what I'm trying to achieve: We are running `cg store completed` only in the crontab. We want to catch any exceptions that manage to bubble up to this point in the code and let a human deal with them later. So we print the stacktrace in the logger and move on to the next case. Wrapping e.g. an AnalysisNotFoundException and doing exactly the same as for any other excpetion is just more type work with the same end result, nl. We want a human to look at the stacktrace.
There is only "success" and "fail", and a warning is considered a success on this level.

Hope that makes sense.

ingkebil: I'm closing this.

cg/cli/store.py
```diff
@@ -112,11 +113,21 @@ def _add_new_analysis_to_the_status_api(bundle_obj, status):
...
             continue
         click.echo(click.style(f"storing family: {analysis_obj.family}", fg='blue'))
         with Path(analysis_obj.config_path).open() as config_stream:
-            context.invoke(analysis, config_stream=config_stream)
+            try:
+                context.invoke(analysis, config_stream=config_stream)
+            except Exception: # pylint: disable=broad-except
+                LOG.error("uploading family failed: %s", analysis_obj.family.internal_id,
+                          exc_info=True)
+                exit_code = SUCCESS
```
patrikgrenfeldt:
```suggestion
                exit_code = FAIL
```

cg/cli/store.py
```diff
@@ -112,11 +113,21 @@ def _add_new_analysis_to_the_status_api(bundle_obj, status):
 def completed(context):
     """Store all completed analyses."""
     hk_api = context.obj['hk_api']
+    FAIL = 1
+    SUCCESS = 0
+    exit_code = FAIL
```
patrikgrenfeldt:
```suggestion
    exit_code = SUCCESS
```

patrikgrenfeldt:
The work on this code continues in https://github.com/Clinical-Genomics/cg/pull/591

PR #441, opened 2019-09-27
Fix balsamic tumour handling

patrikgrenfeldt:
This PR fixes the cg crash when tumour sample was missing

**How to test**:
1. install on stage 
1. activate stage: `us`
1. run following command: `cg analysis balsamic config --dry vitalmouse`

**Expected outcome**:
`cg` should be providing you with an error message instead of crashing
Take a screenshot and attach or copy/paste the output.

**Review:**

This is patch **version bump** because is a fix no new func


patrikgrenfeldt:
Test result:
image

cg/cli/balsamic.py
```diff
@@ -90,13 +93,27 @@ def config(context, dry, target_bed, case_id):
...
+    normal_path = None
+    if len(normal_paths) > 1:
+        click.echo(f"Too many normal samples found: {len(normal_paths)}", color="red")
+        context.abort()
 
-    if not tumor_path:
-        click.echo("No tumor found!", color="red")
-        raise
+    if len(normal_paths) == 1:
+        normal_path = normal_paths.pop()
```
ingkebil: Alright.

So, for tumour, we check:
- zero files
- more than one file

for normal we check
- more than one
- just one

Can you motivate that? :)

hassanfa:
I second tumor instead of tumour, it also matches balsamic where we use tumor instead of tumour. (Now that I have CG on watch, I see it all :-)

ingkebil:
(the English spelling is tumour btw ;))

Yeah, just pick one and add it to the cg ontology :)

hassanfa:
Sure thing. Here is the ontology 😅

https://bioportal.bioontology.org/ontologies/NCIT/?p=classes&conceptid=http%3A%2F%2Fncicb.nci.nih.gov%2Fxml%2Fowl%2FEVS%2FThesaurus.owl%23C18009&jump_to_nav=true

ingkebil:
Tumor it is! :grinning:

cg/cli/balsamic.py
patrikgrenfeldt: Motivation: Balsamic is a cancer pipeline that can work with tumor only or tumor/normal as a pair and thus we need to handle the faulty inputs; 0 tumor, multiple tumor samples and multiple normals. Balsamic CAN handle multiple tumor samples but we have not implemented that call here and implementing that would be something else than a mere fix and refactoring?

patrikgrenfeldt: @ingkebil, happy with motivation?

hassanfa: Note: Multiple tumor samples will only work with QC workflow. Calling somatic variants is not possible with multiple tumor samples.

ingkebil: yes! Thanks for clarifying!

ingkebil:
Approved!

cg/cli/balsamic.py
patrikgrenfeldt: Resolved?

PR #509, opened 2019-12-16
Log more information on ticket creation failures

patrikgrenfeldt:
This PR adds more information in the log when ticket creation fails

**How to test**:
- [x] open travis-CI checks

**Expected test outcome**:
- [x] tests should pass

**Review:**
- [x] "Merge and deploy" approved by @patrikgrenfeldt 

This is minor **version bump** because we only add logging capability to the OsTicket module

### DIFF
--- a/tests/apps/conftest.py
+++ b/tests/apps/conftest.py
@@ -1,0 +1,32 @@
+"""Fixtures for testing apps"""
+
+import pytest
+import requests
+
+
+@pytest.fixture
+def response():
+    """Mock a requests.response object"""
+
+    class MockResponse(requests.Response):
+        """Mock requests.response class"""
+
+        def __init__(self):
+            pass
+
+        @property
+        def ok(self):
+            """Mock ok"""
+            return False
+
+        @property
+        def text(self):
+            """Mock text"""
+            return "response text"
+
+        @property
+        def reason(self):
+            """Mock reason"""
+            return "response reason"
+
+    return MockResponse()
--- a/tests/apps/test_osticket.py
+++ b/tests/apps/test_osticket.py
@@ -3,22 +3,17 @@
 import logging
 
 import pytest
-import requests
 
 from cg.apps.osticket import OsTicket
 from cg.exc import TicketCreationError
 
 
-def test_osticket_respone_500(monkeypatch, caplog):
+def test_osticket_respone_500(monkeypatch, caplog, response):
     """Test that we log properly when we get a non successful response"""
 
     # GIVEN a ticket server always gives a failure in response
     osticket_api = OsTicket()
-    result = requests.Response
-    result.ok = False
-    result.reason = "response reason"
-    result.text = "response text"
-    monkeypatch.setattr(requests, "post", lambda *args, **kwargs: result)
+    monkeypatch.setattr("requests.post", lambda *args, **kwargs: response)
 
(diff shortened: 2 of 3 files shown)


tests/apps/conftest.py
```diff
@@ -0,0 +1,32 @@
...
+    class MockResponse(requests.Response):
+        """Mock requests.response class"""
+
+        def __init__(self):
+            pass
+
+        @property
+        def ok(self):
+            """Mock ok"""
+            return False
```
ingkebil: wait, the OK is `False`? :D

tests/apps/test_osticket.py
```diff
@@ -0,0 +1,30 @@
...
+from cg.apps.osticket import OsTicket
+from cg.exc import TicketCreationError
+
+
+def test_osticket_respone_500(monkeypatch, caplog, response):
+    """Test that we log properly when we get a non successful response"""
+
+    # GIVEN a ticket server always gives a failure in response
+    osticket_api = OsTicket()
+    monkeypatch.setattr("requests.post", lambda *args, **kwargs: response)
```
ingkebil: Can you explain what's happening here?

tests/apps/conftest.py
patrikgrenfeldt: Yes, it's failing, right ;)

tests/apps/test_osticket.py
patrikgrenfeldt: Yes, when post is called on the requests API we return our handcrafted response that will provoke the behaviour we want to test

ingkebil: :clock1: tik tok .. now I get it! The lambda is to 'catch the post'.

Disclaimer: I've only used monkeypatch with a pass-on function, not an anonymous function. As in:

```
    def mock_get(*args, **kwargs):
        return MockResponse()

    monkeypatch.setattr(requests, "get", mock_get)
```

So, I guess how you solved it here was to avoid the anonymous function, right?

tests/apps/conftest.py
ingkebil: haha! alright :)

tests/apps/test_osticket.py
patrikgrenfeldt: Maybe you could say it like that, I'm thinking of it as "connect first part (function post on requests) to a new action, our lambda expression".

So I'd say it's more like using another feature of setattr on monkeypatch, you can give the whole thing to catch as the first argument:
```    
    def mock_get(*args, **kwargs):
        return MockResponse()

    monkeypatch.setattr("requests.get", mock_get)
```

From monkeypatch.py:
        _For convenience you can specify a string as ``target`` which
        will be interpreted as a dotted import path, with the last part
        being the attribute name.  Example:
        ``monkeypatch.setattr("os.getcwd", lambda: "/")``
        would set the ``getcwd`` function of the ``os`` module._

patrikgrenfeldt:
Test result: 
image
👍 
(The linting is red due to non-called super in __init__, I'm not sure this is the best way to solve it so I'll leave it as is instead of silencing it)

tests/apps/test_osticket.py
ingkebil: Sweeet :)

patrikgrenfeldt:

PR #515, opened 2019-12-19
Fix family check

moonso:
This PR adds fixes a bug that the scout upload fails when individuals in a case are unlinked

**How to prepare for test**:
On hasta:
- [x] activate stage: `us`
- [x] try to upload `usableyak` with command `cg upload scout usableyak` where all individuals are unlinked. Check that it fails
- [x] install on stage of hasta: `bash servers/resources/hasta.scilifelab.se/update-cg-stage.sh fix-family-check`

**How to test**:
- [x] Try to upload the same case again

**Expected test outcome**:
- [x] The case should now be uploaded to scout

**Review:**
- [ ] code approved by
- [x] tests executed by @moonso and @b4ckm4n 

This is patch **version bump** because it is a small bug fix

### DIFF
--- a/cg/meta/upload/scoutapi.py
+++ b/cg/meta/upload/scoutapi.py
@@ -1,21 +1,28 @@
+"""File includes api to uploading data into Scout"""
 # -*- coding: utf-8 -*-
 import logging
 from pathlib import Path
 
-from cg.apps import hk, scoutapi, madeline
+from cg.apps import hk, madeline, scoutapi
 from cg.meta.analysis import AnalysisAPI
-from cg.store import models, Store
+from cg.store import Store, models
 from ruamel import yaml
 
 LOG = logging.getLogger(__name__)
 
 
-class UploadScoutAPI(object):
+class UploadScoutAPI:
+    """Class that handles everything that has to do with uploading to Scout"""
 
-    def __init__(self, status_api: Store, hk_api: hk.HousekeeperAPI,
-                 scout_api: scoutapi.ScoutAPI,
-                 analysis_api: AnalysisAPI, madeline_exe: str, madeline=madeline,
-                 ):
+    def __init__(
+        self,
+        status_api: Store,
+        hk_api: hk.HousekeeperAPI,
+        scout_api: scoutapi.ScoutAPI,
+        analysis_api: AnalysisAPI,
+        madeline_exe: str,
+        madeline=madeline,
+    ):
         self.status = status_api
         self.housekeeper = hk_api
         self.scout = scout_api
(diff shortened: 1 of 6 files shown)


moonso:
Skärmavbild 2019-12-20 kl  14 09 38
Before update

moonso:
Skärmavbild 2019-12-20 kl  14 23 19
After update

cg/meta/upload/scoutapi.py
```diff
@@ -1,69 +1,100 @@
...
+            vcf2cytosure_file = self.housekeeper.files(
+                version=hk_version_id, tags=vcf2cytosure_tags
+            ).first()
+            vcf2cytosure_path = (
+                vcf2cytosure_file.full_path if vcf2cytosure_file else None
+            )
+            sample = {
+                "analysis_type": link_obj.sample.application_version.application.analysis_type,
+                "sample_id": sample_id,
+                "capture_kit": None,
```
patrikgrenfeldt: Why is capture kit set to None?

moonso:
ping @patrikgrenfeldt , @barrystokman , @ingkebil this is tested and ready. Please have a look as this bug is making the life hard for production

cg/meta/upload/scoutapi.py
```diff
@@ -142,22 +184,31 @@ def _include_files(self, data, hk_version, scout_hk_map, extra_tag=None, skip_mi
                     raise FileNotFoundError(f"missing file: {scout_key}")
 
     @staticmethod
-    def _is_family_case(data):
-        return any(sample['father'] or sample['mother'] for sample in data['samples'])
+    def _is_family_case(data: dict):
+        """Check if there are any linked individuals in a case"""
+        for sample in data["samples"]:
+            if sample["mother"] and sample["mother"] != "0":
```
patrikgrenfeldt:
```suggestion
            if sample["mother"] != "0":
```
should give same result, right?

cg/meta/upload/scoutapi.py
```diff
@@ -142,22 +184,31 @@ def _include_files(self, data, hk_version, scout_hk_map, extra_tag=None, skip_mi
...
+                return True
+            if sample["father"] and sample["father"] != "0":
```
patrikgrenfeldt:
```suggestion
            if sample["father"] != "0":
```

tests/meta/upload/test_meta_upload_scoutapi.py
```diff
@@ -3,35 +3,73 @@
...
 import yaml
+
 from cg.apps.hk import HousekeeperAPI
 from cg.meta.upload.scoutapi import UploadScoutAPI
 from cg.store import Store
 
 
-def test_generate_config_adds_rank_model_version(analysis: Store.Analysis, upload_scout_api:
-                                                 UploadScoutAPI):
+# pylint: disable=protected-access
```
patrikgrenfeldt: I would recommend moving protected members into a utility class if you want to test them :)

tests/meta/upload/test_meta_upload_scoutapi.py
```diff
@@ -3,35 +3,73 @@
...
+            {"sample_id": "ADM3", "father": "0", "mother": "0"},
+        ]
+    }
+    # WHEN running the check if family is linked
+    res = upload_scout_api._is_family_case(family_data)
+    # THEN assert that the test returns False
+    assert res is False
+
+
+# pylint: disable=protected-access
```
patrikgrenfeldt: same here

cg/meta/upload/scoutapi.py
moonso: No idea, I have not touched that code. It was Blacked

cg/meta/upload/scoutapi.py
moonso: If for some reason `sample['mother'] = None` your suggestion would not work. I did this test since this is the old behaviour and am not sure if it could show up again

tests/meta/upload/test_meta_upload_scoutapi.py
moonso: I do not understand what you mean? I discussed this with @barrystokman before implementing. Perhaps you have another solution?

cg/meta/upload/scoutapi.py
moonso: Aha I moved the code into a new function but did not change anything

tests/meta/upload/test_meta_upload_scoutapi.py
moonso: I guess we want to test privat methods. The alternative would be to let the linter complain like we do in many other places, especially in the tests. I think that this was a nice way to solve the linting problem here

moonso:
As mentioned this is solving a production bug that makes it difficult to upload some cases to Scout. Perhaps we could deploy this and have these detailed discussions later. I think at least @emiliaol would be happy

cg/meta/upload/scoutapi.py
patrikgrenfeldt: If you don't need it in Scout, why bother with it at all here? I think we could remove it if there is no need for it in Scout..

cg/meta/upload/scoutapi.py
patrikgrenfeldt: That's right, lets move on ;)

tests/meta/upload/conftest.py
```diff
@@ -115,19 +151,37 @@ def import_reads(*args, **kwargs):
...
+        return self._name
+
+    @name.setter
+    def name(self, value):
+        self._name = value
+
+    @name.deleter
+    def name(self):
+        del self._name
+
```
patrikgrenfeldt: You wrote that these were added for the linting and I don't understand for which warnings?

tests/meta/upload/test_meta_upload_scoutapi.py
patrikgrenfeldt: Yes, in one place I created a small [class](https://github.com/Clinical-Genomics/cg/blob/master/cg/meta/report/sample_calculator.py) that is a public utility class and should be tested as such whereas private things are considered implementation details and it's not obvious how to test them without locking down the specific implementation..

tests/meta/upload/test_meta_upload_scoutapi.py
patrikgrenfeldt: There is a nice discussion [here](https://stackoverflow.com/questions/105007/should-i-test-private-methods-or-only-public-ones) if you want to dig into this topic

cg/meta/upload/scoutapi.py
patrikgrenfeldt: but I agree to it being done later

patrikgrenfeldt:
If you want to skip the discussion and just deploy I suggest ditching the hiding of the pylint-warnings (which are not obviously wrong in my opinion) and then do that

moonso:
You know what? I have spent a lot of time to get this don asap, involved several people in discussions about implementation and to get help to prepare for tests. Everything to get this into production before christmas. It is up to you three who can merge and deploy this to decide if you want to continue discussing implementation choices or not. That thread you refer to is stopped because the discussion is opinion based, there is no right way of doing it. So if you want to block a bug fix to production based on that it is up to you.

Merry christmas and happy new year

patrikgrenfeldt:
Testing protected members locks down the implementation and I am suggesting that we keep the warnings about doing so

tests/meta/upload/test_meta_upload_scoutapi.py
patrikgrenfeldt:
```suggestion
```

tests/meta/upload/test_meta_upload_scoutapi.py
patrikgrenfeldt:
```suggestion
```

patrikgrenfeldt:
deployed

patrikgrenfeldt:
> You know what? I have spent a lot of time to get this done asap, involved several people in discussions about implementation and to get help to prepare for tests. Everything to get this into production before christmas. It is up to you three who can merge and deploy this to decide if you want to continue discussing implementation choices or not. That thread you refer to is stopped because the discussion is opinion based, there is no right way of doing it. So if you want to block a bug fix to production based on that it is up to you.
> 
> Merry christmas and happy new year

I don't see your point really, what topics are not opinion based? Sure I didn't let your PR through because I had just reviewed it and I wanted to let you address the questions that it raised. When you told me it was urgent I even told you on which two lines I thought we should address it and what the minimal changes ought to be (remove those lines). I think it is quite unfair that you tell me I'm blocking the issue when you yourself refuse to make such a trivial change to your solution in order to get it through. I feel quite bad about your tone in the comment you made.

Merry Christmas and happy new year

moonso:
I'm sorry if I made you feel bad, I was frustrated and needed to leave work. It seems like there are many different opinions on Clinical Genomics about linting in general. We should start a discussion where we agree on some guidelines for how to handle linting in different parts of the code. Also we had a discussion about the PR-review-deploy process, we should agree on how to handle different situations like urgent bug fixes, added features or more fundamental changes. It would be nice to avoid situations like this as much as possible.

moonso:
And back to the specific discussion about testing private members. I have read the material you pointed to and some other resources. My impression is that there are many arguments before and against testing private members, there is no real consensus on what is "correct". So after reading this I'm fine with both ways. However I think it looks cumbersome to wrap the private members in a public one just for testing. It seems much easier and straight forward to just test the private member directly.

PR #540, opened 2020-02-03
Decouple chanjo

adrosenbaum:
This PR removes all usage of the chanjo API, and uses the CLI instead. 

**How to prepare for test**:
- [x] enter hasta
- [x] install this branch on stage ```bash /home/proj/stage/servers/resources/hasta.scilifelab.se/update-cg-stage.sh decouple-chanjo```
- [x] Update stage servers repo with the new cg config file ```bash /home/proj/stage/resources/hasta.scilifelab.se/update-servers-stage.sh cg-chanjo-config```
- [x] Find a case with .bed files available in housekeeper stage. 

**How to test**:
- [x] run following command: ```cg upload coverage <case-id>```
- [x] run following command: ```cg upload validation <case-id>```
- [x] run following command: ```cg upload delivery-report --print <case-id>l```

**Expected test outcome**:

The results before and after updating the chanjo API should be exactly the same

**results**

**pre update**
Save output from subcommands ```validate```and ```delivery-report``` prior to updating changes on stage:

pre-change-report

pre-change-validate

**post update**

after deleting the samples in justhusky from chanjo,  ```cg upload coverage justhusky``` is run

cg-upload-coverage

Next the case justhusky is validated to cg (and the stdout is saved to a file):

cg-upload-validate-post

The same is done creating a delivery-report

cg-upload-report-post

try to upload the case to chanjo again, without the --replace/-r flag

cg-upload-coverage-again

and with the  --replace/-r flag

cg-upload-coverage-replace

Finally, just to assure that we get the same output as before the update, the output files from after the update are compared with the files created before the update

compare-validate
compare-report

and they are identical

**Review:**
- [x] code approved by @patrikgrenfeldt 
- [x] tests executed by @adrosenbaum 
- [x] "Merge and deploy" approved by @patrikgrenfeldt 

This is major **version bump** because this change is not backwards compatible


adrosenbaum:
Added you as reviewer on this PR @moonso. I think you are right, the gene ids should be written to a tmp file and then passed to the chanjo command. I'll try to fix that!

cg/meta/upload/coverage.py
```diff
@@ -1,57 +1,65 @@
...
-                 chanjo_api: coverage.ChanjoAPI):
+    def __init__(
+        self,
+        status_api: Store,
+        hk_api: hk.HousekeeperAPI,
+        chanjo_api: coverage.ChanjoAPI,
+    ):
         self.status = status_api
-        self.hk = hk_api
+        self.hk_api = hk_api
```
moonso: Why not `self.status_api` and `self.chanjo_api` if `self.hk_api`?

tests/meta/upload/conftest.py
```diff
@@ -128,6 +129,35 @@ def convert_panels(customer_id, panels):
         return ""
 
 
+class MockCoverage:
```
patrikgrenfeldt:
```suggestion
class MockCoverage(ChanjoAPI):
```
Would this be useful?

tests/meta/upload/test_meta_upload_coverage.py
```diff
@@ -0,0 +1,66 @@
+"""Tests for coverage meta API"""
+
+import datetime
+
+from cg.meta.upload.coverage import UploadCoverageApi
+from cg.apps.coverage.api import ChanjoAPI
+
+CHANJO_CONFIG = {"chanjo": {"config_path": "config_path", "binary_path": "chanjo"}}
```
patrikgrenfeldt: Fixture?

tests/meta/upload/test_meta_upload_coverage.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class Analysis:
```
patrikgrenfeldt:
```suggestion
class MockAnalysis:
```

tests/meta/upload/test_meta_upload_coverage.py
```diff
@@ -0,0 +1,66 @@
...
+
+    # THEN this returns the data needed to upload samples to chanjo
+    assert results["family"] == family_name
+    for sample in results["samples"]:
+        assert set(sample.keys()) == set(["coverage", "sample", "sample_name"])
+
+
+def test_upload(housekeeper_api, analysis_store, mocker):
+    """test uploading with chanjo"""
+    # GIVEN a coverage api and a data dictionary
```
patrikgrenfeldt: Any data dictionary?

tests/meta/upload/test_meta_upload_coverage.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @property
+    def family(self):
+        """ mock family """
+        return self.family_obj
+
+
+def test_data(coverage_upload_api, analysis_store):
+    """test getting data for chanjo"""
+    # GIVEN a coverage api and an analysis object
```
patrikgrenfeldt: Any analysis object?

tests/meta/upload/test_meta_upload_coverage.py
```diff
@@ -0,0 +1,66 @@
...
+    )
+    family_name = "yellowhog"
+    family_obj = analysis_store.family(family_name)
+    analysis_obj = Analysis(family_obj=family_obj)
+    data = coverage_api.data(analysis_obj=analysis_obj)
+
+    # WHEN uploading samples in data dictionary
+    coverage_api.upload(data=data, replace=True)
+
+    # THEN
```
patrikgrenfeldt: THEN what? 😕

tests/apps/coverage/test_coverage.py
```diff
@@ -0,0 +1,183 @@
...
+    """Test delete method"""
+    # GIVEN a process with a mocked run_command method
+    sample_id = "sample_id"
+    mocked_run_command = mocker.patch.object(Process, "run_command")
+    api = ChanjoAPI(CHANJO_CONFIG)
+
+    # WHEN deleting a sample with the api
+    api.delete_sample(sample_id=sample_id)
+
+    # THEN run_command should be called with the correct arguments
```
patrikgrenfeldt: I guess "correct" could be considered a bit vague? I think the intention of any THEN clause is to give some information about the system under test and and I'm uncertain what to learn from "correct"? 

I suggest we specify what "correct" is so the test can be used to rectify the implementation when it deviates from the business rule, right?

tests/apps/coverage/test_coverage.py
```diff
@@ -0,0 +1,183 @@
+"""Tests for chanjo coverage api"""
+
+from pathlib import Path
+
+from cg.apps.coverage.api import ChanjoAPI
+from cg.utils.commands import Process
+
+CHANJO_CONFIG = {"chanjo": {"config_path": "config_path", "binary_path": "chanjo"}}
```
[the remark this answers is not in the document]
adrosenbaum: > Great work! Had some thoughts :)

Thank you for the comments! I'll see what I can do 😄

tests/meta/upload/test_meta_upload_coverage.py
adrosenbaum: I find this tricky, the upload method does not have any return statement... I thought of adding one, however, I didn't want to change the already existing structure of the API in this PR. I guess I can test it by mocking the Process object again, and use .assert_called_with again. would that be ok?

tests/meta/upload/test_meta_upload_coverage.py
adrosenbaum: it is created below. Do you think this should be given as a fixture instead?

tests/apps/coverage/test_coverage.py
adrosenbaum: I get your point. Would it be enough to specify in the THEN-comment what the 'correct' arguments for the call should be?

tests/meta/upload/test_meta_upload_coverage.py
patrikgrenfeldt: ok

patrikgrenfeldt:

PR #589, opened 2020-03-10
Update treshold for total price 

mayabrandi:
This PR updates the treshold for total invoice price.

**How to prepare for test**:
- [x] ssh to clinical-db depending on type of change
- [x] install on stage:
`bash servers/resources/clinical-db.sclifelab.se/update-clinical-api-stage.sh [THIS-BRANCH-NAME]`

**How to test**:
- [x] go to https://clinical-api-stage.scilifelab.se/invoices
- [x] generate a new invoice for cust002 and check the "select subset" box.

**Expected test outcome**:
- [x] check that the total price is almost 750000

**Review:**
- [x] code approved by @adrosenbaum 
- [x] tests executed by @mayabrandi 
- [x] "Merge and deploy" approved by @barrystokman 


cg/server/invoices/templates/invoices/new.html
```diff
@@ -100,7 +100,7 @@ <h1>Make new invoice ({{record_type}})</h1>
   function check_subset() {
     total_price = 0;
     $("input[name='records']").each(function() {
-      if (total_price <= 500000) {
+      if (total_price + parseInt(this.id) <= 750000) {
```
patrikgrenfeldt: If the threshold needs to be updated once in a while, should it be a config-setting?

patrikgrenfeldt:
I suggest naming the PR with a sentence describing the intention of the change (e.g. Instead of "update setup.py" write "Fix bumpversion"). What do you say?

cg/server/invoices/templates/invoices/new.html
mayabrandi: sure. You want to have a config here: cg/server/invoices/config.py?

patrikgrenfeldt: No I think I'd rather want it as either a property on customer if it varies with customer or if it is not varying with customer having it in the ordinary config yaml for cg under a new section invoicing: 

Does it make sense to you?

mayabrandi: I added it to the server config instead. Seemed to make more sense.

mayabrandi:
Tested according to description above. 
The total price was just below 75000kr. Ticking one more sample gave a price of 756779. 👍 

Skärmavbild 2020-03-19 kl  11 43 12


Ready for deploy!

patrikgrenfeldt:
@mayabrandi Great that you are so quick solving things! 

You asked me the other day on Slack if you should put the threshold value in the config.py and I said you should put it in cg's configs (.yaml or the like). 
The reason is that config.py reads configs. It is *not* a config. 

Could please see if you can fix this?

mayabrandi:
The conversation was not on slack but here on GitHub!  Se above! I didn’t ask about this config. I asked if I should create a config under in cg/server/invoices/config.py. You suggested I should add to the existing yaml config for cg. Then I found this existing config instead which I thought made more sense. But sure lest move it to the other one.

patrikgrenfeldt:
Great!

Den mån 23 mars 2020 10:52mayabrandi <notifications@github.com> skrev:

PR #599, opened 2020-03-25
Add command that 'prunes' old bundle versions

adrosenbaum:
This PR adds a command that will remove files with certain tags from bundle versions that are not latest.

**How to prepare for test**:
- [x] ssh to hasta
- [x] install on stage:
`bash /home/proj/stage/servers/resources/hasta.scilifelab.se/update-cg-stage.sh prune-hk-versions`
- [x] Find or prepare cases on stage that have various housekeeper versions, with bam, cram, crai, and bai, files in all versions

**How to test**:
- [x] login to hasta
- [x] do `us`
- [x] do `cg clean hk-past-files --dry-run -t cram -t bam -t bam-index -t cram-index --case-id <case_id>`
- [x] Repeat with`--yes` flag `cg clean hk-past-files --dry-run -t cram -t bam -t bam-index -t cram-index --yes --case-if <case_id>`
- [x] Repeat without `dry-run` flag `cg clean hk-past-files -t cram -t bam -t bam-index -t cram-index --yes --case-if <case_id>`
- [x] Repeat without `--case-id` and with `--dry-run` flag  `cg clean hk-past-files --dry-run -t cram -t bam -t bam-index -t cram-index --yes` 

`housekeeper get -V justhusky`
Screen Shot 2020-03-27 at 07 45 26
...
Screen Shot 2020-03-27 at 07 47 01

In housekeeper bundles
Screen Shot 2020-03-27 at 07 47 25

**Expected test outcome**:
- [x] When running `cg clean hk-past-files --dry-run -t cram -t bam -t bam-index -t cram-index --case-id <case_id>` This should log all files that would be removed and await a confirmation to delete them (None of the files should be deleted since `--dry-run is activated`)
- [x] Check that all of the logged files are from older hk versions
- [x] Check that none of the files have actually been deleted, and that they are still in housekeeper
- [x] When running `cg clean hk-past-files --dry-run -t cram -t bam -t bam-index -t cram-index --yes --case-if <case_id>` All files in older housekeeper versions that would be deleted are displayed, without waiting for user confirmation.
- [x] Check that all of the logged files are from older hk versions
- [x] Check that none of the files have actually been deleted, and that they are still in 
housekeeper
Screen Shot 2020-03-27 at 07 51 23
- [x] When running `cg clean hk-past-files -t cram -t bam -t bam-index -t cram-index --yes --case-if <case_id>` The files are actually deleted from housekeeper and disk

Screen Shot 2020-03-27 at 07 54 06

Screen Shot 2020-03-27 at 07 52 57

Screen Shot 2020-03-27 at 10 58 09

- [x] When running `cg clean hk-past-files --dry-run -t cram -t bam -t bam-index -t cram-index --yes` All files that can be removed are logged.

Screen Shot 2020-03-27 at 08 34 40

**Review:**
- [ ] code approved by
- [x] tests executed by @adrosenbaum 


adrosenbaum:
I think this should do the trick @patrikgrenfeldt @henrikstranneheim

cg/cli/clean.py
```diff
@@ -137,6 +137,46 @@ def scoutauto(context, days_old: int, yes: bool = False, dry_run: bool = False):
...
+        last_version = context.obj["hk"].last_version(bundle=case_id)
+        LOG.info("Searching %s bundle for outdated files", case_id)
+        hk_files = []
+        for tag in tags:
+            hk_files.extend(context.obj["hk"].get_files(bundle=case_id, tags=[tag]))
+        for hk_file in hk_files:
+            file_path = Path(hk_file.full_path)
+            if hk_file.version_id == last_version.id:
+                continue
+            if os.stat(file_path).st_nlink > 1:
```
henrikstranneheim: Why not? We will leave some files elsewhere, but is that not ok? They should be removed from our production flow, by this process or other processes?

cg/cli/clean.py
```diff
@@ -137,6 +137,46 @@ def scoutauto(context, days_old: int, yes: bool = False, dry_run: bool = False):
         context.invoke(scout, bundle=bundle, yes=yes, dry_run=dry_run)
 
 
+@clean.command("prune-hk")
```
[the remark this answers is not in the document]
adrosenbaum: Agree!

cg/cli/clean.py
adrosenbaum: You are probably right! I added this assuming that maybe we have no good overview of where the files are linked.

adrosenbaum:
@patrikgrenfeldt travis complains that the file I have touched in this PR would be formatted by black, but it was blacked before committing... Am I missing something? 😅

patrikgrenfeldt:
The command seems to do what's intended. But why go through cg to accomplish this and not directly to hk?

adrosenbaum:
> The command seems to do what's intended. But why go through cg to accomplish this and not directly to hk?

I thought of that first. It seems to be no problem getting all files in the hk-database with the hk_api.get_files() method. However, this will return a list of hk file objects, and there seems to be no field telling what bundle these come from. And we need the bundle id to fetch the last version of that bundle, to make sure the file isn't in the latest version. So in the end I found it easier to start with a case id.

adrosenbaum:
And sure, we could still do it the other way, but then I would have to add new methods to the hk-api, and if the plan in the future is to decouple cg and hk, then maybe this is a bad idea. But let me know if you have some other suggestion @patrikgrenfeldt!

cg/cli/clean.py
```diff
@@ -133,6 +132,39 @@ def scoutauto(context, days_old: int, yes: bool = False, dry_run: bool = False):
         context.invoke(scout, bundle=bundle, yes=yes, dry_run=dry_run)
 
 
+@clean.command("hk-past-files")
+@click.option("-c", "--case-id", type=str)
+@click.option("-t", "--tags", multiple=True)
+@click.option("-y", "--yes", is_flag=True, help="skip checks")
+@click.option("-d", "--dry-run", is_flag=True, help="Shows cases and files that would be cleaned")
+@click.pass_context
+def hk_past_files(context, case_id, tags, yes, dry_run):
```
patrikgrenfeldt:
```suggestion
def obsoleted_files_from_hk(context, case_id, tags, yes, dry_run):
```

cg/cli/clean.py
```diff
@@ -133,6 +132,39 @@ def scoutauto(context, days_old: int, yes: bool = False, dry_run: bool = False):
...
+    for case in cases:
+        case_id = case.internal_id
+        last_version = context.obj["hk"].last_version(bundle=case_id)
+        LOG.info("Searching %s bundle for outdated files", case_id)
+        hk_files = []
+        for tag in tags:
+            hk_files.extend(context.obj["hk"].get_files(bundle=case_id, tags=[tag]))
+        for hk_file in hk_files:
+            file_path = Path(hk_file.full_path)
+            if hk_file.version_id == last_version.id:
```
patrikgrenfeldt: I don't know if it can happen but if a file is part both of the latest bundle version AND a previous one we could end up removing it?

adrosenbaum: That's a good question. I haven't thought of that, and just assumed that is never the case. But I think you are right, two file objects in hk could point to the same path... I will try to look into that!

adrosenbaum: I have no idea if this actually will occur, but I've added a check in the last commit that would prevent removing such files. Maybe you can have a look at it @patrikgrenfeldt

patrikgrenfeldt: super!

patrikgrenfeldt:

PR #629, opened 2020-04-21
add condition fo check if target bed supplied

patrikgrenfeldt:
This PR fixes the orblem supplying target bed at command line when starting balsamic

**How to prepare for test**:
- [x] install on stage:
`bash servers/resources/hasta.scilifelab.se/update-cg-stage.sh [THIS-BRANCH-NAME]`

**How to test**:
- [x] login to hasta
- [x] cg --log-level=DEBUG workflow balsamic config-case --target-bed /home/proj/production/cancer/reference/panel/gmcksolid_4.1_hg19_design.bed usablecod -d

**Expected test outcome**:
- [x] check that the bed is the supplied one and not fetched from LIMS

**Review:**
- [x] code approved by @patrikgrenfeldt 

### DIFF
--- a/cg/cli/workflow/balsamic/base.py
+++ b/cg/cli/workflow/balsamic/base.py
@@ -13,7 +13,7 @@ from cg.apps.balsamic.fastq import FastqHandler
 from cg.cli.workflow.balsamic.store import store as store_cmd
 from cg.cli.workflow.balsamic.deliver import deliver as deliver_cmd, CASE_TAGS, SAMPLE_TAGS
 from cg.cli.workflow.get_links import get_links
-from cg.exc import LimsDataError, BalsamicStartError, CgError
+from cg.exc import LimsDataError, BalsamicStartError
 from cg.meta.deliver import DeliverAPI
 from cg.meta.workflow.base import get_target_bed_from_lims
 from cg.meta.workflow.balsamic import AnalysisAPI
--- a/cg/cli/workflow/balsamic/base.py
+++ b/cg/cli/workflow/balsamic/base.py
@@ -13,7 +13,7 @@ from cg.apps.balsamic.fastq import FastqHandler
 from cg.cli.workflow.balsamic.store import store as store_cmd
 from cg.cli.workflow.balsamic.deliver import deliver as deliver_cmd, CASE_TAGS, SAMPLE_TAGS
 from cg.cli.workflow.get_links import get_links
-from cg.exc import LimsDataError, BalsamicStartError
+from cg.exc import LimsDataError, BalsamicStartError, CgError
 from cg.meta.deliver import DeliverAPI
 from cg.meta.workflow.base import get_target_bed_from_lims
 from cg.meta.workflow.balsamic import AnalysisAPI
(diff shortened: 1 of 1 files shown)


patrikgrenfeldt:
Test result:
image

👍

hassanfa:
> Test result:
> image

I think the test is not correct. Balsamic command is wrong. There should be a fastq file after `--tumor`

patrikgrenfeldt:
@keyvanelhami  deployed on hasta:
image

patrikgrenfeldt:
> I think the test is not correct. Balsamic command is wrong. There should be a fastq file after `--tumor`

Oh no, yet another bug!

patrikgrenfeldt:
> > I think the test is not correct. Balsamic command is wrong. There should be a fastq file after `--tumor`
> 
> Oh no, yet another bug!

@hassanfa Would it be ok if I left it for you to fix it in cg?

hassanfa:
Code seems ok, I wonder why was this merged although the test result was in fact _not_ ok...

henrikstranneheim:
@hassanfa It would be great if you can try to find some time to help out with this. Obviously, you know how to run Balsamic depending on input condition so I think it is vital that you are part of this process.

hassanfa:
Maybe Patrick and I can have a look into this? I am not sure if I can even pull it off, the complexity of this code is increased beyond my skills...

Reason: This supposedly new bug (I don't know if even if it is a bug, could be visual or miscaptured screenshot) is introducd during past couple of patches to cg. We have a quite strict PR review process.

- So if this is a bug, burden is on the developer who introduced it.

- If it is not a bug, but screenshot issue, why was his merged with wrong tests performed?

patrikgrenfeldt:
> Code seems ok, I wonder why was this merged although the test result was in fact _not_ ok...

When I said that the test was OK it meant that the test result was OK compared to what was stated as **Expected test outcome** in the PR description. Since this patch was crucial for KE yesterday I wanted to merge even though you spotted another bug. It seems to me that this bug is not caused by the code in this patch.

keyvanelhami:
@henrikstranneheim @patrikgrenfeldt @hassanfa 
I have production samples that needs to be started so this issue should be prioritized.

hassanfa:
@keyvanelhami What type of error do you reiceve? I have a suspicion the error you are getting is not because of `cg`. If the error you are getting is related to the recent development project testing for hamilton, etc, and you get this: `raise BalsamicStartError("Must have exactly one tumor sample! Found %s samples." % nr_paths)` It is because the case you have includes multiple samples that are marked as tumor.

Double check the case you've created...

hassanfa:
I talked to keyvan about this, and the error he gets is because he supplied multiple tumor samples in a case. Although balsamic supports multiple tumor samples, it is not supported in cg. I tested it and received same error, it is also obvious in the code. The error I am referring to is: `raise BalsamicStartError("Must have exactly one tumor sample! Found %s samples." % nr_paths)` 

It seems the "bug" that he received might just be a user error after all...

@keyvanelhami As I explained, if you want to run a case with multiple tumor samples then just use balsamic CLI instead.

It could be that the error Keyvan received is partially because of a mixed case and/or bad check on target bed file.

There is no need to support this in cg. As I see it, not every feature in a workflow should be implemented in cg. E.g. the possibility to run multiple tumor samples is a biproduct of one of the early developments that I didn't remove from balsamic. And now it is occasionally used for some research cases. But if we see rise of use in it, we should implement it in cg.

Finally, I appreciate if/when both parties (@keyvanelhamk and Patrik) kept in the me in the loop as well when these errors arise. At the very least I can comment one why this is happening on BALSAMIC side.

PR #649, opened 2020-05-18
Run tests on github actions

moonso:
This PR changes so that the tests and coverage report is handled by GitHub Actions instead of Travis. 

## Arguments for switching
- The files for running tests on GitHub are very simple to work with. 
- There are a number of apps that can be used together with GitHub Actions allowing for smart improvements
- Tests completes much faster on Github
- A lot of the linting problems have been solved by third party solutions like code factor
- It is nice to have everything within the same ecosystem

One example of an app is the linting that is used here. It supportes multiple linters for multiple languages and it can be modified to perform the linting whenever it can do it. Like allow it to black a file.

## Arguments against switching
- Running the tests on travis works

**How to test**:
- [x] Tests are run by GitHub Actions

**Expected test outcome**:
- [x] Tests should be executed by GitHub Actions :octocat: 
- [x] Coverage report should be uploaded from :octocat: 

**Review:**
- [ ] code approved by
- [x] tests executed by :octocat: 


moonso:
Any objections @barrystokman , @patrikgrenfeldt ? We could always revert and use travis if there are any problems

.github/workflows/coveralls.yml
```diff
@@ -0,0 +1,39 @@
...
+    # Set up python
+    - name: Set up Python ${{ matrix.python-version}}
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+
+    - name: Production Dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -e .
```
patrikgrenfeldt, line 25:
```suggestion
        pip install .
```
remove -e to make it more production like

ditto in the following installs

.github/workflows/coveralls.yml
```diff
@@ -0,0 +1,39 @@
...
+
+    # Set up python
+    - name: Set up Python ${{ matrix.python-version}}
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+
+    - name: Production Dependencies
+      run: |
+        python -m pip install --upgrade pip
```
patrikgrenfeldt, line 24: I would consider using the actual pip-version that we use in production

.github/workflows/coveralls.yml
```diff
@@ -0,0 +1,39 @@
...
+        pip install -e .
+
+    - name: Development Dependencies
+      run: |
+        pip install -r requirements-dev.txt
+        pip install pytest-cov
+        pip install coveralls
```
patrikgrenfeldt, line 31: Install the deps on one line to make sure it considers all requirements at the same time?

ditto in the following installs

.github/workflows/linting.yml
```diff
@@ -0,0 +1,48 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install .
+        pip check
+
+    - name: Development Dependencies
+      run: |
+        pip install -r requirements-dev.txt
+        pip install flake8
+        pip install black
```
patrikgrenfeldt, line 34: If these two are dev-reqs then they could go into the dev-reqs file, otherwise they could be skipped here?

moonso: I do not agree. These packages are not necessary for development only for linting. I suggest that packages that are required for executing tests etc should be part of req-dev

.github/workflows/coveralls.yml
moonso: I changed so that the check is run before the extra packages are installed. I figure this makes sense since the `requirements-dev` are necessary for development. The other are only used in CI

.github/workflows/coveralls.yml
moonso: Fixed that!

PR #650, opened 2020-05-21
Vogue load bioinfo all

hassanfa:
This PR adds bioinfo load all option to `cg upload vogue`

**How to prepare for test**:
- [x] install on stage:
```
bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-servers-stage.sh vogue_update_scripts_scoutify

bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-vogue-stage.sh master

bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-cg-stage.sh vogue_load_bioinfo_all

```

**How to test**:
- [x] login to hasta
- [x] do us
- [x] run `cg upload vogue bioinfo-all`

**Expected test outcome**:
- [x] check that `vogue-stage.scilifelab.se` has all cases shown in picard QC

**Review:**
- [x] code approved by @moonso 
- [x] tests executed by @hassanfa 


hassanfa:
Turns out I don't have permission to housekeeper files and I can't upload to vogue database. Made this PR so someone else can do that instead.

barrystokman:
@hassanfa Error mesage and stack trace occurs while running `cg upload vogue bioinfo-all`, can you have a look? Code seems fine to me.

image

hassanfa:
@barrystokman  thanks a lot for testing and looking into this. I'll fix those errors asap.

hassanfa:
Found the issue.

`context.invoke` doesn't seem to be able to catch exceptions. And this happens because the workflow name is wrong in Housekeeper. It is "fastq" and not "mip".

I wrapped it in `try...except` for now, but I also found another solution: https://stackoverflow.com/questions/55626606/invoke-python-click-command-with-custom-class-from-another-command 

A feedback on which solution is preferred is much appreciated. I prefer `try...except` for being less complex.

barrystokman:
@hassanfa I've fixed the data for sharpparrot on stage, and it works now. The command now fails on `--analysis-type` `MIP` because it expects 'mip'. Both `MIP` and `mip` are present in analyses with both older and newer MIP versions.

image

barrystokman:
```
    raise DocumentTooLarge("%r command document too large" % (operation,))
pymongo.errors.DocumentTooLarge: 'update' command document too large
```


image
image

moonso:
Wow that looks fishy. Max size of a mongodb document is 16MB, that's alooooot of text

hassanfa:
Yeap... https://docs.mongodb.com/manual/reference/limits/#bson-documents We hit the limit boys... 😞

barrystokman:
Bumped
image

hassanfa:
The idea behind `vogue load bioinfo` is as follow:
1. we wanted to store all raw multiqc_data.json in MongoDB instance in a single document, so we can keep track of all the data. This collection is called: `bioinfo_raw`.

2. The data from `bioinfo_raw` gets processed. Process takes a subset of "validated" information, cleans it up, does a bit of transformation on dictionaries and whatnot, then adds it to `bioinfo_process` collection. `bioinfo_process` contains case level information.

3. At third stage, samples from `bioinfo_process` get processed once more and stored in `bioinfo_sample` which is cleaned up data ready to be plotted.

The error above fails at step 1, where it tries to store raw `multiqc_data.json`.

moonso:
But I would say that something is wrong in the model if you want to store >16mb documents in the database...
Why not store paths to the multiqc file instead?

moonso:
What is the `bioinfo_raw`collection used for? It sounds like from the description above that some of the information is triplicated in the database. Sounds to me like step 2 would do the job. Then handle step 3 in the application and skip step 1

hassanfa:
So bug is in Vogue that keeps updating the document in collection and it keeps adding and adding adding... The document grows and finally `update` in `pymongo` fails.

Solution:
Step 1. Simplify collection for raw data. Preferably remove it and only rely on input json data.
Step 2. update a document by overwriting it.

I have no quick fix for this.

moonso:
Sounds good! Tell me if you need any input

hassanfa:
Looking into this, and Vogue is loading only relevant part of MultiQC. So 2019-Hassan already thought about it 🎉 .

Working on fixing duplicate data.

moonso:
Thats great @hassanfa !! Should be a lot less work then 👯‍♂️

hassanfa:
@moonso Do you know what happened here? My PR got automatically changed with "Fix code" now I have 112 files changed....

Is this the new Github action thing?

image

hassanfa:
Install in Stage env:

image

Run `cg upload vogue bioinfo-all`:
image
more:
image

Results are in vogue-stage.scilifelab.se:
image

cg/cli/upload/vogue.py
```diff
@@ -105,6 +108,8 @@ def samples(context, days: int):
 def bioinfo(context, case_name, cleanup, target_load, dry):
     """Load bioinfo case results to the trending database"""
 
+    valid_workflows = ["mip"]
```
moonso, line 111: Suggest to set this as a constant somewhere

cg/cli/upload/vogue.py
```diff
@@ -132,6 +137,8 @@ def bioinfo(context, case_name, cleanup, target_load, dry):
         load_bioinfo_raw_inputs["analysis_workflow_name"],
         load_bioinfo_raw_inputs["analysis_workflow_version"],
     ) = _get_analysis_workflow_details(context, case_name)
+    if load_bioinfo_raw_inputs["analysis_workflow_name"] not in valid_workflows:
```
moonso, line 140: I had a really hard time understanding the above two lines. Could you do something like:

```python
workflow_name, workflow_version = _get_analysis_workflow_details(context, case_name)
if workflow_name not in valid_workflows:
    raise AnalysisUploadError(f"Case upload failed: {case_name}. Reason: Bad workflow name.")
load_bioinfo_raw_inputs["analysis_workflow_name"] = workflow_name
load_bioinfo_raw_inputs["analysis_workflow_version"] = workflow_version
```

cg/cli/upload/vogue.py
```diff
@@ -150,6 +157,41 @@ def bioinfo(context, case_name, cleanup, target_load, dry):
             context.obj["vogue_upload_api"].load_bioinfo_sample(load_bioinfo_raw_inputs)
 
 
+@vogue.command("bioinfo-all", short_help="Load all mip bioinfo results into vogue")
```
moonso, line 160: Is it only mip who produce bioinfo results :)
```suggestion
@vogue.command("bioinfo-all", short_help="Load all bioinfo results into vogue")
```

cg/cli/upload/vogue.py
```diff
@@ -150,6 +157,41 @@ def bioinfo(context, case_name, cleanup, target_load, dry):
...
+@click.option("--dry/--no-dry", default=False, help="Dry run...")
```
moonso, line 161:
```suggestion
@click.option("--dry/--no-dry", is_flag=True, help="Dry run...")
```

cg/cli/upload/vogue.py
```diff
@@ -150,6 +157,41 @@ def bioinfo(context, case_name, cleanup, target_load, dry):
...
+        if len(list(multiqc_file_obj)) == 0:
+            continue
+
+        # confirm that file exists
+        existing_multiqc_file = multiqc_file_obj[0].full_path
+        if not Path(existing_multiqc_file).exists():
+            continue
+
+        click.echo(
+            click.style(f"Found multiqc for {case_name}, {existing_multiqc_file}", fg="blue")
```
moonso, line 187:
```suggestion
        LOG.info("Found multiqc for %s, %s", case_name, existing_multiqc_file)
```

cg/cli/upload/vogue.py
```diff
@@ -150,6 +157,41 @@ def bioinfo(context, case_name, cleanup, target_load, dry):
...
+        )
+        if len(list(multiqc_file_obj)) == 0:
+            continue
+
+        # confirm that file exists
+        existing_multiqc_file = multiqc_file_obj[0].full_path
+        if not Path(existing_multiqc_file).exists():
+            continue
+
+        click.echo(
```
moonso, line 186:
```suggestion
```

cg/cli/upload/vogue.py
```diff
@@ -150,6 +157,41 @@ def bioinfo(context, case_name, cleanup, target_load, dry):
...
+            click.style(f"Found multiqc for {case_name}, {existing_multiqc_file}", fg="blue")
+        )
```
moonso, line 188:
```suggestion
```

cg/cli/upload/vogue.py
```diff
@@ -150,6 +157,41 @@ def bioinfo(context, case_name, cleanup, target_load, dry):
...
 
+@vogue.command("bioinfo-all", short_help="Load all mip bioinfo results into vogue")
+@click.option("--dry/--no-dry", default=False, help="Dry run...")
+@click.pass_context
+def bioinfo_all(context, dry):
+    """Load all cases with recent analysis and a multiqc-json to the trending database."""
+
+    hk_api = context.obj["housekeeper_api"]
+    cases = context.obj["db"].families()
+    for case in cases:
```
moonso, line 168: Really nice code here @hassanfa 🦀

cg/cli/upload/vogue.py
```diff
@@ -150,6 +157,41 @@ def bioinfo(context, case_name, cleanup, target_load, dry):
...
+        click.echo(
+            click.style(f"Found multiqc for {case_name}, {existing_multiqc_file}", fg="blue")
+        )
+        try:
+            context.invoke(bioinfo, case_name=case_name, cleanup=True, target_load="all", dry=dry)
+        except AnalysisUploadError:
+            LOG.error("Case upload failed: %s", case_name, exc_info=True)
+
+
 def _get_multiqc_latest_file(context, case_name):
```
moonso, line 195: Two things here 

- we don't need to send the whole context as we only need the housekeeper api
- Lets stick to python type hints as this is used everywhere in CG

I'll add a suggestion 
```suggestion
def _get_multiqc_latest_file(hk_api: HousekeeperAPI, case_name: str) -> str:
```

cg/cli/upload/vogue.py
```diff
@@ -199,4 +244,4 @@ def _get_analysis_workflow_details(context, case_name):
         workflow_name = family_obj.analyses[0].pipeline
         workflow_version = family_obj.analyses[0].pipeline_version
 
-    return workflow_name, workflow_version
+    return workflow_name.lower(), workflow_version
```
moonso, line 247: Same here, no need for sending the whole context. Just pick out the store before.

cg/cli/upload/vogue.py
```diff
@@ -150,6 +157,41 @@ def bioinfo(context, case_name, cleanup, target_load, dry):
...
 
 
+@vogue.command("bioinfo-all", short_help="Load all mip bioinfo results into vogue")
+@click.option("--dry/--no-dry", default=False, help="Dry run...")
+@click.pass_context
+def bioinfo_all(context, dry):
+    """Load all cases with recent analysis and a multiqc-json to the trending database."""
+
+    hk_api = context.obj["housekeeper_api"]
+    cases = context.obj["db"].families()
```
moonso, line 167:
```suggestion
    store = context.obj["db"]
    cases = store.families()
```

cg/cli/upload/vogue.py
```diff
@@ -105,6 +108,8 @@ def samples(context, days: int):
 def bioinfo(context, case_name, cleanup, target_load, dry):
     """Load bioinfo case results to the trending database"""
 
+    valid_workflows = ["mip"]
+
```
moonso, line 112: Related to what I suggested below. To make code more transparent I suggest to pick out what is needed from the context and send only that to the helper functions.

```python
hk_api = context.obj['db']
load_bioinfo_raw_inputs["samples"] = _get_samples(hk_api, case_name)
```
etc..

cg/cli/upload/vogue.py
hassanfa: 👍 good idea. Should I keep it in `cg/cli/upload/vogue.py` as well or add it to `cg/constants.py`? I think constants.py is a better place imho.

cg/cli/upload/vogue.py
hassanfa: I had it like this before, but changed it for sake of fewer lines. I'll do it this way. 👍

cg/cli/upload/vogue.py
hassanfa: For now yes 😉 . Microsalt is next!

cg/cli/upload/vogue.py
hassanfa: 👍

cg/cli/upload/vogue.py
hassanfa: 👍

cg/cli/upload/vogue.py
moonso: Your call. If it has no value for the rest of (cg) it could be placed at the top of the file

cg/cli/upload/vogue.py
moonso: I prefer readability over compactness!

cg/cli/upload/vogue.py
moonso: What about balsamic?

hassanfa: balsamic doesn't have any data stored in HK yet. I suppose it'll be loaded when it is stored.

cg/cli/upload/vogue.py
hassanfa: Added it top of the file.

cg/cli/upload/vogue.py
hassanfa: review addressed. Microsalt will be another PR if you agree.

cg/cli/upload/vogue.py
hassanfa: done

cg/cli/upload/vogue.py
hassanfa: done

cg/cli/upload/vogue.py
hassanfa: 👹 🦀

cg/cli/upload/vogue.py
hassanfa: I made some changes according to you comment. There is one left, which is: `get_links` and it seems it needs a click.Context: https://github.com/Clinical-Genomics/cg/blob/6577794299c6ae61a81057540ed83f1d2dcc246d/cg/cli/workflow/get_links.py#L10 I kept it as it is for this one.

cg/cli/upload/vogue.py
hassanfa: I made changes according to comment.

cg/cli/upload/vogue.py
hassanfa: addressed

cg/cli/upload/vogue.py
hassanfa: I updated the code according to your suggestion. Let me know if this is A-ok.

cg/cli/upload/vogue.py
moonso: Ok, that one could be fixed by raising exceptions instead of `context.abort`. Otherwise it only uses `store` from the context. Feel free to adress that if you like :)

moonso:
Nicely done @hassanfa ! 🚀 I'm approving as is but it would be nice if you could add at least one test that run the cli command and perhaps remove the context from the link function. I know it could mean some extra work and I'm happy to assist you if needed. If we cannot find any time we can merge.

henrikstranneheim:
Great work guys!!

moonso:
I couldn't keep my fingers from the get_links function, I fixed that. Please have a look

hassanfa:
oh wow! Great job Måns. 👏 Lets talk today about how we can proceed with this.

moonso:
Skärmavbild 2020-06-03 kl  09 40 45

Merged and deployed 🎉

PR #651, opened 2020-05-27
better gh workflows

moonso:
This PR updates the GitHub Actions workflows. 
It was confusing when mixing when different workflows should be executed (`push`, `pull_request`). Now everything will run for every push and opened PR.

Black is now allowed to edit files, if this PR is merged then all files will automatically be blacked by the bot in every PR. You might see that in the "files changed" :)


moonso:
Could someone have a quick look at this @barrystokman or @patrikgrenfeldt ?

barrystokman:
> Could someone have a quick look at this @barrystokman or @patrikgrenfeldt ?

I'll have a look

moonso:
I wonder if it blacked everything again since we use line length of 100 in the action...

.github/workflows/lint_and_install.yml
```diff
@@ -0,0 +1,123 @@
...
+      # [0|0|436] [mans.magnusson@hasta:~] [P_main] 3s $ date
+      # Wed May 27 16:44:09 CEST 2020
+      # [0|0|436] [mans.magnusson@hasta:~] [P_main] $ python --version
+      # Python 3.7.1
+      # (prod170926) [hiseq.clinical@hippocampus ~]$ date
+      # Wed May 27 16:44:09 CEST 2020
+      # (prod170926) [hiseq.clinical@hippocampus ~]$ python --version
+      # Python 3.6.2 :: Continuum Analytics, Inc.
+      # (prod170926) [hiseq.clinical@hippocampus ~]$ pip --version
+      # pip 9.0.1
```
barrystokman, line 64: Can we store this information somewhere instead of in comments here?

moonso: I was also thinking about it but could not get any idea where... any suggestion? I think it is good to show why we are choosing these specific versions and where to update information when environments are updated

barrystokman: Maybe a markdown file in `cg/.github/workflows`?

moonso: Good idea! I added that

moonso:
I'll merge and deploy then 👍

moonso:
Skärmavbild 2020-05-28 kl  13 08 24

PR #674, opened 2020-06-17
feat(hk_tags): Added files and hk_tags for mip dna to cg

henrikstranneheim:
This PR adds/fixes:
- Added new files and hk_tags for files already in the deliverables file, which was not currently part of cg constant dict
- Ordered dict alphabetically

**How to prepare for test**:
- [x] install on stage:
`bash servers/resources/hasta.scilifelab.se/update-cg-stage.sh mip_rd_dna_hk_tags`

**How to test**:
- [x] login to hasta
- [x] `us`
- [x] `housekeeper delete bundle onemite`
-[x] `cg workflow mip-dna store analysis /home/proj/stage/rare-disease/cases/onemite/analysis/onemite_config.yaml` 

**Expected test outcome**:
- [x] check that the:
1. chanjo-sexcheck tag and file are present in HK as : "chanjo", "sex-check" ✅ 
1. coverage and sambamba-depth tags and file are present in HK as: "coverage", "sambamba-depth"  ✅ 
1. smn-calling and smncopynumbercaller in HK as: "smncopynumbercaller", "smn-calling"  ✅ 

**Review:**
- [ ] code approved by
- [x] tests executed by @henrikstranneheim 


henrikstranneheim:
**How to test**:
- [x] login to hasta
- [x] `us`
- [x] `housekeeper delete bundle onemite`
-[x] `cg workflow mip-dna store analysis /home/proj/stage/rare-disease/cases/onemite/analysis/onemite_config.yaml` 

**Expected test outcome**:
- [x] check that the:
1. chanjo-sexcheck tag and file are present in HK ✅ 
1. coverage and sambamba-depth tags and file are present in HK ✅ 
1. smn-calling and smncopynumbercaller ✅ 

Screenshot 2020-06-17 at 15 55 49
Screenshot 2020-06-17 at 15 56 08

Screenshot 2020-06-17 at 15 56 48

cg/constants.py
```diff
@@ -67,64 +67,70 @@
 # presence of mandatory files. Keys = tags found in deliverables, values = MIP standard tags and
 # mandatory flag
 MIP_DNA_TAGS = {
+    tuple(["config_analysis"]): {"tags": ["mip-config"], "is_mandatory": True},
```
moonso, line 70: Great, one small remark. Why do we create tuples with `tuple` all the time?
In python:
```python
tuple(['a','b']) == ('a','b')
```
BUT remember that
```python
('a','b') != ('b','a')
```
which also means that
```python
tuple(['a','b']) != tuple(['b','a'])
```

moonso:
```python
Python 3.7.7 (default, Apr 21 2020, 11:51:15)
Type 'copyright', 'credits' or 'license' for more information
IPython 7.13.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: assert tuple(['a','b']) == ('a','b')

In [2]: assert ('a','b') != ('b','a')

In [3]: assert tuple(['a','b']) != tuple(['b','a'])

In [4]: assert tuple(['a','b']) == tuple(['b','a'])
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-4-5ace5e5729c5> in <module>
----> 1 assert tuple(['a','b']) == tuple(['b','a'])

AssertionError:
```

barrystokman: Dictionary keys must be hashable, so I went with tuples.

image

I also check if tags are *in* the keys, not if the tags *are* the keys.

Do see your point though. I realise it works fine for my code, but it's a bit tricky for other developers that don't know the reasoning behind it.

henrikstranneheim: If you want I can remove the tuple constructor and just let it be a tuple by just keeping the `()`.

barrystokman: Defining singleton tuples like that will make them of the type of the object in them, you would need a trailing comma to force the type tuple:

image

So it would need to be 
`("config_analysis",): {"tags": ["mip-config"], "is_mandatory": True},`
or since we can check for more than one distinct value in the key
`(["config_analysis"],): {"tags": ["mip-config"], "is_mandatory": True},` and 
`(["multiqc_ar", "html"],): {"tags": ["multiqc-html"], "is_mandatory": True},` and so on.

I didn't like that so I choose to use `tuple(...)`, because it looked more clear to me.

I'm not opposed to your suggestion, I'm just explaining the reasoning behind my choice :)

henrikstranneheim: Thx @barrystokman ! That made it more clear. I removed the tuple and added a trailing comma. I think this will be great - it is a little less clutter and still very clear that this is a tuple. I hope that is fine!

henrikstranneheim:
@barrystokman @moonso I redid the tests for onemite and also for sharpparrot (RNA) and it works. 
Could you do another code review and approve if there is nothing to remark upon?
Screenshot 2020-06-18 at 12 06 17

tests/meta/store/snapshots/snap_test_cg_meta_store_base.py
```diff
@@ -27,7 +28,7 @@
     {
         "archive": False,
         "path": "/path/to/rare-disease/cases/case_id/analysis/sample_id/arriba_ar/sample_id_lanes_1234_trim_arriba.pdf",
-        "tags": ["arriba-ar", "arriba-report", "meta", "mip-rna", "sample_id"],
+        "tags": ["arriba-ar", "arriba_report", "meta", "mip-rna", "sample_id"],
```
barrystokman, line 31: Snapshottest picked up that the tag is `arriba_report` while it should be `arriba-report`. Should the snapshot  be updated?

barrystokman: sharpparrot bundle created a while ago:
image

sharpparrot bundle cerated with this branch just now:
image

Something seems to have changed.

henrikstranneheim: strange it looks like this in this branch:
```
  ("arriba_ar", "arriba_report"): {"tags": ["arriba-ar", "arriba-report"], "is_mandatory": True},
```

henrikstranneheim: I will checkout it out again on stage and try to recreate the bundle again...

cg/constants.py
moonso: Ok I see your point @barrystokman , I'm in favour of

```python
("config_analysis",): {"tags": ["mip-config"], "is_mandatory": True}
```
But I won't object if you declare by using the word `tuple`

moonso: If you check if the tag is IN the tuple I would choose to not use tuples at all and just have unique strings as keys and duplicate values instead, example:

now
```python
{tuple(["config_analysis", "config"]): {"tags": ["mip-config"], "is_mandatory": True}}
```
could be
```python
{
    "config_analysis": {"tags": ["mip-config"], "is_mandatory": True},
    "config": {"tags": ["mip-config"], "is_mandatory": True}
}
```
This will make the dictionary search more effective, at least if I understand what you mean. I thought we used tuples as keys because we search the dictionary with tuples...

tests/meta/store/snapshots/snap_test_cg_meta_store_mip.py
```diff
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
```
henrikstranneheim, line 1:
```suggestion
```

henrikstranneheim:
I am having a bit of trouble with these snapshots, but hopefully this is ok now. I will retry this on stage

cg/constants.py
barrystokman:
```
("multiqc_ar", "html"): {"tags": ["multiqc-html"], "is_mandatory": True},
("multiqc_ar", "json"): {"tags": ["multiqc-json"], "is_mandatory": True},
```

Here's a case where the outcome is dependent on two tags in the deliverables file, these can't be mapped one-on-one. I basically check per file in the deliverables.yaml if all the tags in the key (for instance `multiqc_ar` and `html`) are in the tags of the file (the number of files can be anything, so like `some_tag`, `another_tag`, `html`, `tag4: the taggening`, `multiqc_ar`. If that's the case I know the Housekeeper tag should be `multiqc-html` and not the other one.

I'm pretty sure more experienced developers can come up with a much more elegant solution, but it's the best I could come up with :frowning_face:

henrikstranneheim:
@barrystokman Redid the sharpparrot test after changing the order of tuples in the dict. Now it looks like before, but I do not like that switching the order around modifies the behavior. What do you think - feels error prone?
Screenshot 2020-06-18 at 17 41 58

cg/constants.py
moonso: Yes that sounds a bit tricky... so I guess you have to loop over all tags in the dictionary every time you want to check what tags to fetch? Where is the code for doing that?

barrystokman:
The problem is that before you changed the order the code tried, first of all, to determine if all tags in ('arriba_ar', 'arriba_ar')* are in ['ACC5963A2', 'arriba_ar', 'arriba_report', 'meta', 'mip-rna'], which they are. Then `arriba_report` doesn't get converted to `arriba-report`.

Simply looping over all the tags didn't work because sometimes a combination of tags determines the actual HK tag, for instance `("multiqc_ar", "html")` should convert to `["multiqc-html"]`, while `("multiqc_ar", "json")` should be `["multiqc-json"]`.

I'll think about a more robust solution.

*could be just ('arriba_ar',), different topic...

barrystokman:
All guthub checks were green a  week ago, but now:
image

cg/meta/store/base.py
```diff
@@ -40,87 +54,94 @@ def add_new_analysis(bundle_data, case_obj, status, version_obj):
...
+    tags = _convert_tags(tags, tag_map, tag_map_key, is_index)
 
+    return tags
 
-def build_bundle(config_data: dict, analysisinfo_data: dict, deliverables: dict) -> dict:
-    """Create a new bundle to store in Housekeeper"""
 
-    analysis_type = config_data["samples"][0]["type"]
+def _convert_tags(tags: dict, tag_map: dict, tag_map_key: tuple, is_index: bool = False) -> list:
+    """ Convert tags from external deliverables tags to standard internal housekeeper tags """
```
henrikstranneheim, line 117:
```suggestion
    """ Filter and convert tags from external deliverables tags to standard internal housekeeper tags """
```

cg/meta/store/base.py
```diff
@@ -40,87 +54,94 @@ def add_new_analysis(bundle_data, case_obj, status, version_obj):
...
     """
-
-    all_deliverable_tags = [file["tags"] for file in files]
-    all_mandatory_tags = [tag for tag in pipeline_tags if pipeline_tags[tag]["is_mandatory"]]
-    deliverable_tags = _flatten(all_deliverable_tags)
-    mandatory_tags = _flatten(all_mandatory_tags)
+    deliverable_tags = [file_["tag_map_key"] for file_ in files]
+    mandatory_tags = [tag for tag in pipeline_tags if pipeline_tags[tag]["is_mandatory"]]
 
     tags_are_missing, missing_tags = _determine_missing_files(mandatory_tags, deliverable_tags)
```
henrikstranneheim, line 146:
```suggestion
    tags_are_missing, missing_tags = _determine_missing_tags(mandatory_tags, deliverable_tags)
```

cg/meta/store/base.py
```diff
@@ -130,20 +151,10 @@ def _check_mandatory_tags(files: list, pipeline_tags: dict):
         )
 
 
-def _determine_missing_files(mandatory_tags: set, found_tags: set) -> tuple:
-    """Determines if mandatory are missing, and which ones"""
+def _determine_missing_files(mandatory_tags: list, found_tags: list) -> tuple:
+    """Determines if mandatory are files missing in the deliverables, and which ones"""
```
henrikstranneheim, line 155:
```suggestion
    """Determines if mandatory tags are missing and hence if files are missing in the deliverables, and returns any missing tags"""
```

cg/meta/store/base.py
```diff
@@ -40,87 +54,94 @@ def add_new_analysis(bundle_data, case_obj, status, version_obj):
...
+                "archive": False,
+                "tag_map_key": tag_map_key,
+            }
+            parsed_files.append(parsed_index_file)
+
+    return parsed_files
+
+
+def get_tags(
+    file: dict, pipeline_tags: list, tag_map: dict, tag_map_key: tuple, is_index: bool = False
```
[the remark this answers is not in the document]
barrystokman: Good point, will change it.

barrystokman:
Retest after implementing suggested changes:
1) store
image
image
2) store, bundle exists
image
3) store, analysis already exists
image

barrystokman:
Test with DNA family (onemite):
image

`housekeeper get -V onemite`:
image

Specific tags:
image

barrystokman:
Bumped:
image
Deployed:
image

PR #709, opened 2020-08-18
Upload old balsamic data

Mropat:
This PR adds a script for one-time upload of the old BALSAMIC analyses to Housekeeper and StatusDB

### How to prepare for test
- [x] ssh to hasta 
- [x] install on stage:
- [x] install on stage:
`bash /home/proj/production/servers/resources/clinical-db.scilifelab.se/update-clinical-api-stage.sh [THIS-BRANCH-NAME]`

### How to test
- [x] login to hasta
- [x] run script upload-balsamic-housekeeper.py > outlog.txt
Provide cg-stage config parameters for status and housekeeper as options
The script will find the latest analysis for case id, add bundle to Housekeeper and update StatusDB with new analysis

### Expected test outcome
- [x] check the log to see which bundles were stored and errors raised

## Review
- [x] code approved by @moonso 
- [x] tests executed by @Mropat 


scripts/upload-balsamic-housekeeper.py
```diff
@@ -0,0 +1,115 @@
...
+                    fam_names[case_id] = {
+                        "bv": config_dict.get("analysis").get("BALSAMIC_version"),
+                        "config_name": f"{case_dir}/{obj.name}/{obj2.name}",
+                        "date": creation_date,
+                        "path": obj.name,
+                        "hk": Path(report_dir, obj.name + ".hk"),
+                    }
+    for case_id in fam_names:
+        items = fam_names.get(case_id)
+        if items:
```
moonso, line 54:
```suggestion
        if not items:
            continue
```
And then you can move the following try-except block one indentation to the left making it more flat

moonso (approved):
Hi, nice. If this was not just a one off script I would suggest that you moved the two large loops into functions but I guess that does not matter so much in this case 😃

Mropat:
Successfully stored case on stage:
image

Script successfully uploaded ~200 cases.

Mropat:
For 228 cases on STAGE, this error was raised:

image

For those cases, no bundle was created:

image

No bundle exists in Housekeeper for either of those cases

Mropat:
Viewing the housekeeper paths for one of the cases:
image

It appears that the files were copied to the directory, even thought there is no housekeeper bundle in the database. This could happen if copying Housekeeper production to stage does not wipe the stored data, only updates database. This could be understandable because it's the second time i'm running this script.
(These ghost files are not found in housekeeper production folder, so this shouldn't be a problem i think)
@patrikgrenfeldt, I think you may find this interesting, and I'd appreciate a comment on this

On the other hand, It appears that Housekeeper still tries to go through with committing the bundle (as not the bundle nor bundle version can be found in database), even though files already exist. This is a clearly unintended behavior by Housekeeper, because it will start linking files before checking if linking all of the bundle contents will be possible, and the error will be thrown after some files have already been linked. 
This will be a problem when adding an analysis version with the same date, but not time, as it will attempt to link to the same folder. 
This will be a problem for cases with "mip+balsamic" analysis, which are started on the same date - because both BALSAMIC and MIP produce files with identical names: multiqc_data.json, multiqc_report.html. 
This will be a problem if we try to add bundle initially - and fail to link all of the files due to permission error (happens in some cases depending on who starts the analysis).
@moonso I think this could be interesting

Otherwise, feel free to merge&deploy this!

moonso:
Ok @Mropat , great job! I think we need to have a meeting about this together with @henrikstranneheim , @patrikgrenfeldt and @hassanfa . This is definitely a part of the project [revised data management plan](https://github.com/Clinical-Genomics/project-planning/issues/1) that have highest priority atm. We need to clarify current behaviour and intended behaviour. I'll set up a meeting about this for next week. 

As a part of this I suggest we document somewhere how our stage setup is created and if that is what we want, there has been some unexpected events touching this.

moonso:
Hi @Mropat should I merge and deploy this? It looks like tests are not done since the boxes in top are not ticked, but I guess you did them all according to the comments..

Mropat:
@moonso please wait for a couple hours with this... I want to check everything one last time!

moonso:
Just shout when you are ready

Mropat:
New version of script successfully stores goodshimp, which was failing previously due to permissions:
image

Mropat:
I have modified the script a lot, so it will need a new review.

Changes in functionality:
If files exist already, the directory is cleaned before re-trying to add bundle (Should be relevant only when testing on stage)
If directory cannot be cleaned due to permissions, inform about it and skip case 
If bundle/version already exist, remove and store again (will be used to "update" some of the old bundles)

Script moved to cg.utils and added as balsamic cli command. I think it should be removed when one-time 
store is done. 
The script can be used multiple times without problems, but needs to be deleted before we merge new
 hk store commands

Attaching output:
[hkstore.txt](https://github.com/Clinical-Genomics/cg/files/5123732/hkstore.txt)

moonso:
> Script moved to cg.utils and added as balsamic cli command. I think it should be removed when one-time
store is done.

😆 this is a real hack! Why was it not possible to keep outside codebase?

Mropat:
Reverted to keeping this in scripts. Tested running on stage and it still works as intended!
If rest of the code is ok, merge and deploy whenever!

Mropat:
Added new tags this morning, proof upload works!
image

Mropat:
@moonso Ready to merge!

Mropat:
This is not a notable change and not in code base. DO we need to chengelog this? As Patch?

moonso:
> This is not a notable change and not in code base. DO we need to chengelog this? As Patch?

No I guess we can skip that. Will merge this tomorrow though, no more time now!

moonso:
Skärmavbild 2020-09-03 kl  07 58 51

Merged and deployed on hasta

PR #727, opened 2020-09-01
delete empty fastqs

moonso:
This PR adds a functionality so that empty fastq files will be removed from housekeeper and from disk. Atm there are several cases where there are fastq files added to HK even though they have no content. While looping over the cases to check which cases that are ready for fastq compression the compress API will remove the empty fastq files from housekeeper and from disk.

### How to test
Tests can only be run in automated test suite

## Review
- [ ] code approved by


cg/models/file_data.py
```diff
@@ -0,0 +1,34 @@
...
+        handle = open(path, "rb")
+        if handle.read(2).hex() == GZIP_MAGIC_NUMBER:
+            return True
+        return False
+
+    @staticmethod
+    def is_empty(path: Path) -> bool:
+        """Check if file is empty"""
+        handle = open(path, "rb")
+        if handle.read(1):
```
henrikstranneheim, line 25: What does "1" stand for? Could you give it a named constant to make it easier to understand?

tests/conftest.py
```diff
@@ -235,16 +241,27 @@ def fixture_run_name() -> str:
     return "fastq_run"
 
 
+@pytest.fixture(scope="function", name="original_fastqs")
+def fixture_original_fastq(fastq_dir: Path, run_name) -> CompressionData:
```
henrikstranneheim, line 245:
```suggestion
def fixture_original_fastq_path(fastq_dir: Path, run_name) -> CompressionData:
```

tests/models/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+        " ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ull"
+        "amco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehende"
+        "rit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaec"
+        "at cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum."
+    )
+    return _content
+
+
+@pytest.fixture(name="file_path")
+def fixture_file_path(project_dir: Path) -> Path:
```
henrikstranneheim, line 23: I think this name should better reflect the docstring

tests/models/conftest.py
```diff
@@ -0,0 +1,48 @@
...
+    """Return the path to a non existing file"""
+    return project_dir / "a_file.txt"
+
+
+@pytest.fixture(name="gzipped_file_path")
+def fixture_gzipped_file_path(file_path: Path) -> Path:
```
henrikstranneheim, line 29: Same here: I would expect to get a path to an existing file

tests/models/test_file_data.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # WHEN checking if the file is gzipped
+    res = FileData.is_gzipped(filled_file)
+
+    # THEN assert the result is False
+    assert res is False
+
+
+def test_is_gzipped_non_existing(gzipped_file_path):
+    """Test if a file that is does not exist is gzipped"""
```
henrikstranneheim, line 36:
```suggestion
    """Test if a file that does not exist is gzipped"""
```

tests/models/test_file_data.py
```diff
@@ -0,0 +1,85 @@
...
+    assert not gzipped_file_path.exists()
+
+    # WHEN checking if the file is gzipped
+    with pytest.raises(FileNotFoundError):
+        # THEN assert that a FileNotFoundError is raised
+        FileData.is_gzipped(gzipped_file_path)
+
+
+def test_is_gzipped_empty_file(gzipped_file_path):
+    """Test if a file that is does exists but is empty"""
```
henrikstranneheim, line 47:
```suggestion
    """Test if a file that does exists but is empty"""
```

henrikstranneheim (commented):
It looks good! Just to be certain - this is deleting fastq (presumed empty). The repercussion if there is something wrong the empty fastq file check could be quite grave. Do we dare to this automatically? Would it be possible to add a couple of cases to test this on alongside the unit tests

cg/models/file_data.py
moonso: It is just reading the first byte of the file. I can make a comment in the code

tests/conftest.py
moonso: It is not returning a path but a CompressionData object so the path suffix is not in place here

moonso: Changed to `original_fastq_data`

tests/models/conftest.py
moonso: changed to `non_existing_gzipped_file_path`

tests/models/conftest.py
moonso: Fixed

moonso:
Skärmavbild 2020-09-01 kl  17 39 50

Merged and deployed on hasta

PR #743, opened 2020-09-04
create pending flag outside batch

moonso:
This PR fixes a problem that when two processes of fastq compression was launched at the same time the processes would fail. This was due to the fact that the crunchy pending flag was created in the batch script. This PR solves this by creating the pending flag in the crunchy API instead of the batch script.

### Expected test outcome
- [ ] Two fastq compressions started at the same time should not be possible

## Review
- [x] code approved by PG

### DIFF
--- a/cg/apps/crunchy/crunchy.py
+++ b/cg/apps/crunchy/crunchy.py
@@ -399,6 +399,7 @@ class CrunchyAPI:
             fastq_second=compression_obj.fastq_second,
             spring_path=compression_obj.spring_path,
             flag_path=compression_obj.spring_metadata_path,
+            pending_path=compression_obj.pending_path,
             tmp_dir=tmp_dir_path,
         )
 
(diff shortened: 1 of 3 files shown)


cg/apps/crunchy/crunchy.py
```diff
@@ -327,12 +335,20 @@ def get_sbatch_path(log_dir: Path, compression: str, run_name: str = None) -> Pa
...
         return log_dir / "_".join([run_name, "decompress_spring.sh"])
 
-    def _submit_sbatch(self, sbatch_content: str, sbatch_path: Path):
+    def _submit_sbatch(self, sbatch_content: str, sbatch_path: Path, pending_path: Path):
         """Submit SLURM job"""
         LOG.info("Submit sbatch")
         if self.dry_run:
             LOG.info("Would submit sbatch %s to slurm", sbatch_path)
             return
+        LOG.info("Creating pending flag")
```
patrikgrenfeldt, line 344: do we need all of the logging?

patrikgrenfeldt (approved):
👍

cg/apps/crunchy/crunchy.py
moonso: I love it!

moonso: I find it pretty helpful when parsing the logs, there has been so many obstacles with this so I prefer to log a lot

moonso:
Skärmavbild 2020-09-04 kl  14 52 52

Merged and deployed on hasta

PR #756, opened 2020-09-15
Removes unused scout_api from cli/store

moonso:
This PR fixes a bug in cg/cli/store where a scout_api was used where it should not be

### Expected test outcome
- [ ] check that it is possible to run `cg store fastq`

## Review
- [x] code approved by MR
- [ ] tests executed by

### DIFF
--- a/cg/cli/store.py
+++ b/cg/cli/store.py
@@ -14,8 +14,15 @@ LOG = logging.getLogger(__name__)
 
 
 @click.group()
-def store():
+@click.pass_context
+def store(context):
     """Command for storing files"""
+    hk_api = hk.HousekeeperAPI(context.obj)
+    crunchy_api = crunchy.CrunchyAPI(context.obj)
+
+    compress_api = CompressAPI(hk_api=hk_api, crunchy_api=crunchy_api)
+    context.obj["compress"] = compress_api
+    context.obj["db"] = Store(context.obj.get("database"))
 
 
 @store.command("fastq")
@@ -25,12 +32,10 @@ def store():
 def fastq_cmd(context, case_id, dry_run):
     """Store fastq files for a case in Housekeeper"""
     LOG.info("Running store fastq")
-    cg_store = Store(context.obj.get("database"))
 
-    hk_api = hk.HousekeeperAPI(context.obj)
-    crunchy_api = crunchy.CrunchyAPI(context.obj)
+    compress_api = context.obj["compress"]
+    cg_store = context.obj["db"]
 
-    compress_api = CompressAPI(hk_api=hk_api, crunchy_api=crunchy_api)
     update_compress_api(compress_api, dry_run=dry_run)
 
     samples = get_fastq_individuals(cg_store, case_id)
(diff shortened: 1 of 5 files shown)


moonso:
Will add a test to this

cg/cli/store.py
```diff
@@ -14,8 +14,15 @@
...
-def store():
+@click.pass_context
+def store(context):
     """Command for storing files"""
+    hk_api = hk.HousekeeperAPI(context.obj)
+    crunchy_api = crunchy.CrunchyAPI(context.obj)
+
+    compress_api = CompressAPI(hk_api=hk_api, crunchy_api=crunchy_api)
+    context.obj["compress"] = compress_api
+    context.obj["db"] = Store(context.obj.get("database"))
```
Mropat, line 25: :1st_place_medal:

cg/cli/store.py
```diff
@@ -25,26 +32,23 @@ def store():
...
     try:
         for sample_id in samples:
             was_added = compress_api.add_decompressed_fastq(sample_id)
             if was_added is False:
                 LOG.info("skipping individual %s", sample_id)
                 continue
-            stored_inds += 1
+            stored_individuals += 1
     except CaseNotFoundError:
         return
```
Mropat, line 52: Do we know for sure if this gives rise to some logging? Maybe CaseNotFoundError could be thrown with a message, which would be logged here, instead of somewhere in compress_api? I might be wrong though, just something to consider

tests/cli/compress/conftest.py
```diff
@@ -158,6 +158,13 @@ def fixture_base_compress_context(compress_api, store):
...
+@pytest.fixture(name="store_fastq_context")
+def fixture_store_fastq_context(compress_api, store):
+    """Return a compress context"""
+    ctx = {"compress": compress_api, "db": store}
+    return ctx
+
+
 @pytest.fixture(name="populated_multiple_compress_context")
 def fixture_populated_multiple_compress_context(compress_api, populated_compress_multiple_store):
     """Return a compress context populated with a completed analysis"""
```
Mropat, line 170: :1st_place_medal:

cg/cli/store.py
moonso: It is logged in the function that is run. This is also tested if you check the test

Mropat (approved):
Awesome changes, I really like the improved test coverage!

cg/cli/store.py
```diff
@@ -14,8 +14,15 @@
 
 
 @click.group()
-def store():
+@click.pass_context
+def store(context):
```
Mropat, line 18: Personal opinion, but if no command is invoked, it can print help and return, which could add to user friendliness!
It is also possible to cover this block with tests (providing context with fake config dict, and checking exit code to see if all api initiated correctly given the config variables). No need to do it though, if it doesn't seem helpful

Mropat (approved):

cg/cli/store.py
moonso: Next time! I need to get this into production now

moonso:
Skärmavbild 2020-09-15 kl  15 53 06

Merged and deployed

PR #845, opened 2020-11-18
Load Scout from CLI

moonso:
This PR changes the behaviour so that CG uses scout cli when uploading cases instead of importing code directly from Scout

### How to prepare for test
- [ ] ssh to hasta (depending on type of change)
- [ ] install on stage:
`bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-cg-stage.sh use-cli-scout-load`

### How to test
- [ ] use stage us
- [ ] upload a case `cg load case vitalmouse`

### Expected test outcome
- [ ] check that the case was uploaded to scout

## Review
- [ ] code approved by
- [ ] tests executed by

### DIFF
--- a/cg/meta/upload/scoutapi.py
+++ b/cg/meta/upload/scoutapi.py
@@ -76,7 +76,7 @@ class UploadScoutAPI:
             }
             yield sample
 
-    def generate_config(self, analysis_obj: models.Analysis) -> dict:
+    def generate_config(self, analysis_obj: models.Analysis, rank_score_threshold: int = 5) -> dict:
         """Fetch data about an analysis to load Scout."""
         analysis_date = analysis_obj.started_at or analysis_obj.completed_at
         hk_version = self.housekeeper.version(analysis_obj.family.internal_id, analysis_date)
@@ -86,6 +86,7 @@ class UploadScoutAPI:
             "default_gene_panels": analysis_obj.family.panels,
             "family": analysis_obj.family.internal_id,
             "family_name": analysis_obj.family.name,
+            "rank_score_threshold": rank_score_threshold,
             "gene_panels": self.analysis.convert_panels(
                 analysis_obj.family.customer.internal_id, analysis_obj.family.panels
             ),
--- a/cg/apps/scoutapi.py
+++ b/cg/apps/scoutapi.py
@@ -4,6 +4,7 @@ import datetime as dt
 import logging
 from typing import List
 from pathlib import Path
+import yaml
 
 from pymongo import MongoClient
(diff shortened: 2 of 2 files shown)


Mropat (approved):
:fire:

cg/meta/upload/scoutapi.py
```diff
@@ -112,7 +127,11 @@ def generate_config(self, analysis_obj: models.Analysis, rank_score_threshold: i
         else:
             LOG.info("family of 1 sample - skip pedigree graph")
 
-        return data
+        # Validate that the config is correct
+
+        scout_case = ScoutCase(**data)
```
Mropat, line 132: could also be done with ScoutCase.parse_obj(data)
Not saying its better but i've always done that instead of unpacking

moonso: Ok! sounds like it would be similar. I like that it will strip the `None` values when converting back to dict

cg/meta/upload/scoutapi.py
```diff
@@ -74,13 +79,23 @@ def build_samples(self, analysis_obj: models.Analysis, hk_version_id: int = None
...
+        LOG.debug("Found analysis date %s", analysis_date)
+        hk_version: hk_models.Version = self.housekeeper.version(
+            analysis_obj.family.internal_id, analysis_date
+        )
+        LOG.debug("Found housekeeper version %s", hk_version.id)
+        analysis_data: dict = self.analysis.get_latest_metadata(analysis_obj.family.internal_id)
+        raw_genome_build = analysis_data.get("genome_build", "37")
+        genome_build = "37"
+        if "38" in raw_genome_build:
+            genome_build = "38"
```
Mropat, line 98:
```suggestion
        raw_genome_build = analysis_data.get("genome_build", "37")
        genome_build = "37"
        if "38" in raw_genome_build:
            genome_build = "38"
```

raw_genome_build = analysis_data.get("genome_build")
genome_build = "38" if "38" in raw_genome_build else "37"

cg/meta/upload/scoutapi.py
Mropat: That can be stripped anyway and can be configured when dumping model:
https://pydantic-docs.helpmanual.io/usage/exporting_models/#modeldict

cg/meta/upload/scoutapi.py
Mropat: or even shorter

moonso: I think that will fail if "genome_build" does not exist. But I get the point

PR #895, opened 2020-12-16
Refactor microSALT to use case_id

Mropat:
**This PR adds/fixes ...**
Some backend logic for starting / running microsalt via cg is rewritten
Now Housekeeper and StatusDB both refer to same analysis by its case_id

**New CLI commands:**

Starting workflow is possible by providing either case_id, sample_id, or ticket_id.
By default the workflow line expects case_id as the only argument. 
If the user would like to start the workflow for only a single sample in a ticket, one can provide sample_id followed by flag "-s" to indicate that only one sample is to be analyzed. 

Same applies for starting with ticket number followed by flag "-t". Since ticket and case_id under current database structure could both be used as a unique identifier for a family, starting workflow with ticket will result in the same exact output as starting with case_id, and the user will be informed which case_id corresponds to the given ticket upon starting the workflow.

`cg workflow microsalt link <case_id | sample_id -s | ticket -t>`
Link fastq files to case working directory namely cases_dir/fastq/{case_id}/{sample_id}

`cg workflow microsalt config-case <case_id | sample_id -s | ticket -t>`
Creates config file for either a single sample or a case

`cg workflow microsalt run <case_id | sample_id -s | ticket -t>`
Calls microSALT cli with correct arguments for given case/sample

`cg workflow microsalt start <case_id | sample_id -s | ticket -t>`
Calls link, config-case and run in succession for a given id

`cg workflow microsalt start-available`
Calls link, config-case and run in succession for all cases without analysis. Currently ALL samples fall into this category, and so all old samples should be put on hold prior to deploying this PR

`cg workflow microsalt store <case_id>`
Store results of analysis for given case_id. Here only case_id is a valid argument. Analyses where only one sample was analyzed cannot be stored in Housekeeper following current convention (and possibly we don't want to do so anyway?) 

`cg workflow microsalt store-available`
Checks for deliverables file at expected path and stores deliverables for all cases with status "running"

### How to prepare for test
- [x] install on stage:
`bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-cg-stage.sh fix/microsalt-start`

**Test case names with decompressed files and LIMS data:**

_luckyhog (large case)
bigdrum (small case)_

### How to test
- [x] login to hasta & us
- [x] do `cg workflow microsalt link luckyhog`
- [x] do `cg workflow microsalt config-case luckyhog`
- [x] do `cg workflow microsalt run luckyhog`

- [x] do `cg workflow microsalt store luckyhog`
image
image

Bundle saved in HK and can be retrieved with case_id
image

### Expected test outcome
- [x] check that analysis is successfully prepared and executed, and deliverables are stored in Housekeeper

## Review
- [x] code approved by MM
- [x] tests executed by MR

## Implementation Plan
- [ ] Document in Atlas
- [ ] Set all the old microbial cases on "hold" to avoid starting automatically
- [ ] Deploy this branch
- [ ] Inform production


Mropat:
cg workflow micrisalt start bigdrum works as intended
image

Mropat:
Successfully starts whole workflow for one sample only 
image

Mropat:
Starting bigdrum with ticket id works as intended
image

barrystokman:
`store-available` was not working correctly due to the field `action` to being set correctly. Should we 1) test that here too and 2) fix the existing microsalt 'families' in the database?

Mropat:
@barrystokman store-available was not working because production was not starting analyses through cg. Instead, they would use cg to make config, and then ran microsalt manually with the config as input argument. This means that the code which would set family action to running was never invoked.
Now I have tested starting the whole workflow through cg, and it workd (which will set the action properly). So the store-available can now catch those cases without changing families query

barrystokman (approved):
Looks good to me!

Mropat:
Storing in HK works again with Hermes in place.
Kicks out non existing files and stores bundle

image

tests/cli/workflow/microsalt/conftest.py
```diff
@@ -29,22 +29,19 @@ def base_context(microsalt_store, lims_api, tmpdir, queries_path, housekeeper_ap
             db=microsalt_store,
             hk_api=housekeeper_api,
             lims_api=lims_api,
-            fastq_handler=fastq_handler,
+            hermes_api=HermesApi(
```
moonso, line 32: Could this be made as a common fixture?

tests/cli/workflow/microsalt/test_microsalt_case_config.py
```diff
@@ -33,47 +31,46 @@ def test_no_sample_found(cli_runner, base_context, caplog):
...
-        assert f"No sample found for that ticket/sample_id" in caplog.text
+        assert f"No sample found with id: {microbial_sample_id}" in caplog.text
 
 
 def test_no_order_found(cli_runner, base_context, caplog):
     """Test missing order command """
 
     # GIVEN a not existing ticket
-    ticket = -1
+    ticket = "123000"
```
moonso, line 46: Convert this to fixture?

moonso (changes requested):
Wow great job! Some minor comments

cg/apps/hermes/models.py
```diff
@@ -18,3 +23,17 @@ class CGDeliverables(BaseModel):
     pipeline: str
     bundle_id: str
     files: List[CGTag]
+
+    @validator("files", each_item=True)
```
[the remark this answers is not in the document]
Mropat: What a coincidence, I was inspired by Hermes validators to do this :laughing:

moonso (approved):

Mropat:
Deployed on hasta:
image

PR #953, opened 2021-02-08
Feat/containerize cg app

Mropat:
This PR adds/fixes ...

Containerfile and its configurations for running clinical-api

### How to test
- [x] Make sure container can be built

### Expected test outcome
- [x] Make sure container can be run with podman and not fail on startup

## Review
- [x] code approved by ES
- [x] tests executed by MR
- [x] "Merge and deploy" approved by MR

- [x] **PATCH** 

## Implementation Plan
- [ ] Merge this branch


Mropat:
Preliminary ready.
Feedback appreciated

@eriksjolund I would appreciate if you had any comments regarding whether I need to declare user/uid in this file or if they should be set at runtime

Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM python:3.7-slim
```
eriksjolund, line 1: I think it should be a fully qualified name


```
FROM python:3.7-slim
```
=>

```
FROM docker.io/library/python:3.7-slim
```
(If I remember correctly)

eriksjolund (approved):
Regarding the topic of how to deal with secrets inside a Dockerfile, I don't have any good answer right now. I need to learn more. Maria we discussed this briefly before.

Dockerfile
Mropat, line 1:
```suggestion
FROM docker.io/library/python:3.7-slim
```

Containerfile
```diff
@@ -0,0 +1,40 @@
+FROM python:3.7-slim
```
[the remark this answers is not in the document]
Mropat: Done

Mropat:
Will deploy this as is, and modify if needed

PR #1020, opened 2021-03-11
Removes Pipfile and Pipfile.lock

moonso:
## Description
This PR removes the Pipfile and Pipfile.lock. We have tried to use them for almost a year but it seems like there is no interest to picl this up so right now they are more of a technical debth as @patrikgrenfeldt was afraid of from the beginning.
Let's remove these files until there are more developers who think this is important.

## Review
- [x] code approved by BS


barrystokman (approved):
👍🏻

PR #1025, opened 2021-03-11
slurm module

moonso:
## Description
This PR adds a slurm module to CG. The idea is to use this module whenever we submit batch jobs from CG to slurm.
As a proof of concept the CrunchyAPI now use the new slurm api.

### How to test
- [x] Tests are automated

### Expected test outcome
- [x] check that the tests pass and that the new module is covered by tests

Screen Shot 2021-03-12 at 08 58 35

## Review
- [x] code approved by MR
- [x] tests executed by GitHub


cg/models/slurm/sbatch.py
```diff
@@ -0,0 +1,18 @@
...
+class Sbatch(BaseModel):
+    job_name: str
+    account: str
+    number_tasks: int
+    memory: int
+    log_dir: str
+    email: str
+    hours: int
+    minutes: str = "00"
+    priority: Literal["high", "low"] = "low"
```
Mropat, line 16:
```suggestion
    priority: Literal["high", "normal", "low"] = "low"
```

cg/utils/date.py
```diff
@@ -16,20 +16,10 @@
...
-    """Check if a string is a valid date
-
-    Args:
-        date(str)
-
-    Returns:
-        bool
-    """
+def match_date(date: str) -> bool:
+    """Check if a string is a valid date"""
```
Mropat, line 20: This can do the job and parse most date string formats!
https://dateutil.readthedocs.io/en/stable/parser.html

cg/apps/slurm/slurm_api.py
```diff
@@ -0,0 +1,66 @@
...
+        return SBATCH_HEADER_TEMPLATE.format(**sbatch_parameters.dict())
+
+    @staticmethod
+    def generate_sbatch_body(commands: str, error_function: Optional[str] = None) -> str:
+        if not error_function:
+            error_function = "log 'Something went wrong, aborting'"
+
+        return SBATCH_BODY_TEMPLATE.format(**{"error_body": error_function, "commands": commands})
+
+    def submit_sbatch(self, sbatch_content: str, sbatch_path: Path) -> int:
```
Mropat, line 42: I suggest to split this method into things doing one thing at a time, One method to write sbatch content to sbatch path, and one to send it to slurm

cg/apps/slurm/slurm_api.py
```diff
@@ -0,0 +1,66 @@
...
+    def __init__(self):
+        self.process: Process = Process("sbatch")
+        self.dry_run: bool = False
+
+    def set_dry_run(self, dry_run: bool) -> None:
+        LOG.info("Set dry run to %s", dry_run)
+        self.dry_run = dry_run
+
+    @staticmethod
+    def generate_sbatch(sbatch_parameters: Sbatch) -> str:
```
Mropat, line 23: generate_sbatch_script?

cg/apps/slurm/slurm_api.py
```diff
@@ -0,0 +1,66 @@
...
+        LOG.info("Submit sbatch")
+        if self.dry_run:
+            LOG.info("Would submit sbatch %s to slurm", sbatch_path)
+            return 123456
+
+        with open(sbatch_path, mode="w+t") as sbatch_file:
+            sbatch_file.write(sbatch_content)
+
+        sbatch_parameters: List[str] = [str(sbatch_path.resolve())]
+        self.process.run_command(parameters=sbatch_parameters)
```
Mropat, line 56: Process.run_command also supports dry_run flag!

cg/apps/slurm/slurm_api.py
```diff
@@ -0,0 +1,66 @@
...
+        if self.process.stderr:
+            LOG.info(self.process.stderr)
+        LOG.info(self.process.stdout)
+
+        try:
+            job_number: int = int(self.process.stdout.strip().split()[-1])
+        except ValueError:
+            LOG.warning("Could not get slurm job number")
+            job_number = 0
```
Mropat, line 65: Just wondering, not a criticism! Does it need to return a number? Could it return nothing?

cg/apps/crunchy/files.py
```diff
@@ -11,6 +12,30 @@
 LOG = logging.getLogger(__name__)
 
 
+# Methods to get file information
+def get_log_dir(file_path: Path) -> Path:
+    """Return the path to where logs should be stored"""
+    return file_path.parent
+
+
+def get_sbatch_path(log_dir: Path, compression: str, run_name: str = None) -> Path:
```
Mropat, line 21: This can be simplified I think! But making 2 methods - one returns decompress name and one compress name... then there wont be need to pass compression="fastq"... But its like, just a picky opninion

Mropat (approved):
Sweet refactor! Left some comments, feel free to address whatever you agree with! :+1:

cg/apps/crunchy/crunchy.py
```diff
@@ -135,11 +141,15 @@ def is_fastq_compression_done(compression_obj: CompressionData) -> bool:
...
         LOG.info("SPRING metadata file found")
 
-        crunchy_metadata: CrunchyMetadata = get_crunchy_metadata(
-            compression_obj.spring_metadata_path
-        )
+        try:
+            crunchy_metadata: CrunchyMetadata = files.get_crunchy_metadata(
+                compression_obj.spring_metadata_path
+            )
+        except ValidationError:
```
Mropat, line 148: I think it would be nice to not except this here! Knowing where validation fails can be helpful. It came in handy when hunting a bug last week

cg/utils/date.py
moonso: We had this as dependency before and I wrote the module to get rid of that dependency:) It is not well maintained and fragile to have in the dependency tree

cg/apps/slurm/slurm_api.py
moonso: I'm trying to get away from having `Optional[str]` or `Optional[int]` as return type for almost all functions. I don't like it. Perhaps it should raise an exception if there is no job id returned

cg/apps/crunchy/crunchy.py
moonso: Ok I think you are right

cg/apps/slurm/slurm_api.py
moonso: Sure

cg/apps/slurm/slurm_api.py
moonso: If you like, no problem

cg/apps/slurm/slurm_api.py
moonso: Yes

cg/apps/slurm/slurm_api.py
moonso: done

cg/apps/slurm/slurm_api.py
moonso: changed name to `generate_sbatch_content`

cg/apps/slurm/slurm_api.py
moonso: changed name to `generate_sbatch_content`

cg/apps/slurm/slurm_api.py
moonso: using that as suggestion implies

cg/apps/slurm/slurm_api.py
moonso: I'll keep it to 0 and we can see if it should raise in the future

cg/apps/crunchy/files.py
moonso: Splitting into two!

cg/apps/crunchy/crunchy.py
moonso: Changed to fail hard

moonso:
Screen Shot 2021-03-12 at 09 16 34

PR #1131, opened 2021-04-23
Eat cgstats

moonso:
## Description

This PR eats more of cgstats so that we can remove cgstats as a dependency from `requirements.txt`.
There where only two things that was needed:

- Move the declaration of the database models into CG
- Add find sample functionality into CG

The full functionality of cgstats is not yet in CG but now we can go on and step by step add the stuff that is still used in production.

### How to prepare for test

Tests are automated

## Review
- [x] code approved by MR
- [x] tests executed by GitHub


Mropat:
:hungary:

Mropat (approved):
I think you should eat the models

tests/meta/upload/gisaid/conftest.py
```diff
@@ -81,16 +80,26 @@ def gisaid_case_id():
     return "gisaid_case"
 
 
+@pytest.fixture(name="gisaid_file_name")
```
Mropat, line 83: can this be saved to Path(tmpdir, "test.fasta")? then tmpdir is cleaned after pytest is done
even though this is rmoved once not used, still its kind of weird to save and delete file from repo base every time

Mropat (approved):

tests/meta/upload/gisaid/conftest.py
moonso: Haha, yes that's exactly what I did first

moonso: But there is a test that checks for the filename in a file, so it is hardcoded there. And the tests will fail

moonso:
Screen Shot 2021-04-27 at 09 22 50

deployed on hasta

PR #1333, opened 2021-11-26
Fix(ordering): accept optional empty/missing numerical values in json

patrikgrenfeldt:
### Fixed
- Optional numerical values can be empty

### How to prepare for test
- [ ] Use stage: `us`

### How to test
- [ ] submit an json mip-dna order without quantity specified

### Expected test outcome
- [ ] check that the order is accepted without pydantic errora

## Review
- [ ] code approved by
- [ ] tests executed by

## Implementation Plan
- [ ] Deploy this branch
- [ ] Inform to MG

### DIFF
--- a/cg/models/orders/samples.py
+++ b/cg/models/orders/samples.py
@@ -105,9 +105,9 @@ class Of1508Sample(OrderInSample):
     ]
     synopsis: Optional[str]
 
-    @validator("tumour_purity", "formalin_fixation_time", "post_formalin_fixation_time", pre=True)
+    @validator("tumour_purity", "formalin_fixation_time", "post_formalin_fixation_time", "quantity", pre=True)
     def str_to_int(cls, v: str) -> Optional[int]:
-        if not v:
+        if v in (None, ""):
             return None
         return int(v)
 
--- a/cg/models/orders/samples.py
+++ b/cg/models/orders/samples.py
@@ -121,7 +121,13 @@ class Of1508Sample(OrderInSample):
     ]
     synopsis: Optional[str]
 
-    @validator("tumour_purity", "formalin_fixation_time", "post_formalin_fixation_time", "quantity", pre=True)
+    @validator(
+        "tumour_purity",
+        "formalin_fixation_time",
+        "post_formalin_fixation_time",
+        "quantity",
+        pre=True,
+    )
     def str_to_int(cls, v: str) -> Optional[int]:
         return OptionalIntValidator.str_to_int(v=v)
 
--- a/cg/models/orders/samples.py
+++ b/cg/models/orders/samples.py
@@ -16,6 +16,22 @@ from cg.models.orders.sample_base import (
 from cg.store import models
 
 
+class OptionalIntValidator:
+    @classmethod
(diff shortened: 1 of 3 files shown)


Mropat (commented):
Why not just write validator code in the pydantic model?

cg/models/orders/samples.py
```diff
@@ -16,6 +16,22 @@
 from cg.store import models
 
 
+class OptionalIntValidator:
+    @classmethod
+    def str_to_int(cls, v: str) -> Optional[int]:
+        if v in (None, ""):
+            return None
+        return int(v)
```
Mropat, line 24:
```suggestion
        if v:
            return int(v)
        
```

cg/models/orders/samples.py
```diff
@@ -105,11 +121,15 @@ def required_for_new_samples(cls, value, values, **kwargs):
...
+        "formalin_fixation_time",
+        "post_formalin_fixation_time",
+        "quantity",
+        pre=True,
+    )
     def str_to_int(cls, v: str) -> Optional[int]:
-        if not v:
-            return None
-        return int(v)
+        return OptionalIntValidator.str_to_int(v=v)
```
Mropat, line 132: Maybe add str to float here to handle it for all use-cases? Even if you validate different fields in different orderforms, it still validates None to None if the value is not filled

cg/models/orders/samples.py
patrikgrenfeldt: Python treats `0` as `False` and thus we can't do `if v:` in this case, otherwise a good suggestion to clean up the code

cg/models/orders/samples.py
patrikgrenfeldt: It was my intuition to put these in the base class but I haven't figured out how to decorate them in a prettier way than this the derived classes.

patrikgrenfeldt:
> Why not just write validator code in the pydantic model?

Since it was the same in several models it's nice not to duplicate that code

cg/models/orders/samples.py
Mropat: str("0") is True. If this is a string validator, I dont see a problem with this

patrikgrenfeldt: Yes, that's right, as long as it is a string ant not an int

patrikgrenfeldt: I guess we could make sure it is a string from where we parse the json and excel but it might make more sense to have it as an int?

patrikgrenfeldt: fixed

cg/models/orders/samples.py
patrikgrenfeldt: I think I misunderstood what you meant. Now I think you meant that we should use the str_to_float(v=v) to convert both ints and floats to float!?

patrikgrenfeldt:
Test result:
image
🆗 quantity handled correctly as string, int and missing

cg/models/orders/samples.py
patrikgrenfeldt: I forgot to push the commit, I'll move it to the next PR instead

PR #1370, opened 2022-01-19
Support for nanopore data transfer via "cg add external"

henningonsbring:
## Description
When downloading external data from caesar that have been sequenced with nanopore we want that data transferred into a dedicated folder on hasta. This PR enable sending data to `/home/proj/production/external-data/<CUST_ID>/nanopore/` on hasta when using the `--nanopore` flag with the `cg add external` command.

### Added

- --nanopore flag for cg add external

### Changed

-

### Fixed

-

### How to prepare for test
- [ ] Use stage: `us`
`bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-cg-stage.sh add-external-nanopore`

### How to test
- [ ] do ...

### Expected test outcome
- [ ] check that ...

## Review
- [ ] code approved by
- [ ] tests executed by

## Implementation Plan
- [ ] Document in ...
- [ ] Deploy this branch
- [ ] Inform to ...


cg/meta/transfer/external_data.py
```diff
@@ -71,6 +73,9 @@ def transfer_sample_files_from_source(self, dry_run: bool, ticket_id: int) -> No
             source_path=self.get_source_path(customer=cust, ticket_id=ticket_id),
             destination_path=self.get_destination_path(customer=cust),
         )
+        if nanopore:
```
Vince-janv, line 76: I think this will result in an error if the nanopore folder is not created.
```suggestion
        destination_folder: Path = Path(self.destination_path % cust)
        if nanopore:
            destination_folder = destination_folder / "nanopore"
        Path(destination_folder).mkdir(exist_ok=True, parents=True)
```

Vince-janv (commented):
Still a bit unsure why we need a trailing slash for the destination path. Maybe some quick tests on hasta to see that this works as intended?

henningonsbring:
The command works as intended, see tests below:

When running `cg add external --nanopore --dry-run -t 690536` on hasta, then the sbatch script printed on the screen indicate that the following rsync command will be used:
`rsync -rvL /home/proj/stage/external-data/stage_caesar/home/cust124/outbox/690536 /home/proj/stage/external-data/cust124/nanopore`

When running that rsync command we get a folder structure that looks like this:
```
rsync -rvL /home/proj/stage/external-data/stage_caesar/home/cust124/outbox/690536 /home/proj/stage/external-data/cust124/nanopore
sending incremental file list
690536/
690536/fil1.txt
690536/fil2.txt

sent 195 bytes  received 58 bytes  506.00 bytes/sec
total size is 4  speedup is 0.02
[0|0|2046] [henning.onsbring@hasta:/home/proj/stage/external-data/cust124/nanopore] [S_main] $ ls /home/proj/stage/external-data/cust124/nanopore/
690536
[0|0|2046] [henning.onsbring@hasta:/home/proj/stage/external-data/cust124/nanopore] [S_main] $ ls /home/proj/stage/external-data/cust124/nanopore/690536/
fil1.txt  fil2.txt
```

Vince-janv:
I think a small change need to be made on line 179 in cg/meta/transfer/external_data.py. Since the we throw an error if any non-sample folder is present in the customer's external-data directory. The change below should be sufficient

`elif not sample and not self.status_db.sample(sample_folder.name) and not sample_folder.name == "nanopore"`

cg/meta/transfer/external_data.py
```diff
@@ -165,6 +176,8 @@ def curate_sample_folder(self, cust_name: str, force: bool, sample_folder: Path)
             sample and force
         ):
             sample_folder.rename(customer_folder.joinpath(sample.internal_id))
+        elif sample_folder.name == "nanopore":
```
mayabrandi, line 179: i think this check should happen outside of this function to keep it more clean

mayabrandi: Do the check just before curate_sample_folder is being called

cg/meta/transfer/external_data.py
```diff
@@ -60,17 +60,28 @@ def get_destination_path(self, customer: str, lims_sample_id: Optional[str] = ""
...
         error_function: str = ERROR_RSYNC_FUNCTION.format()
-        Path(self.destination_path % cust).mkdir(exist_ok=True)
 
-        command: str = RSYNC_CONTENTS_COMMAND.format(
-            source_path=self.get_source_path(customer=cust, ticket_id=ticket_id),
-            destination_path=self.get_destination_path(customer=cust),
-        )
+        destination_folder: Path = Path(self.destination_path % cust)
+        if nanopore:
+            destination_folder = destination_folder / "nanopore"
```
mayabrandi, line 73: I would probably make a function rsync_nanaopore_command() with a doc string that would explain something about this. I react on this hardcoded sting path without explanation. Thats why.

cg/meta/transfer/external_data.py
```diff
@@ -60,17 +60,28 @@ def get_destination_path(self, customer: str, lims_sample_id: Optional[str] = ""
...
+            command: str = RSYNC_COMMAND.format(
+                source_path=self.get_source_path(customer=cust, ticket_id=ticket_id),
+                destination_path=destination_folder,
+            )
+        else:
+            command: str = RSYNC_CONTENTS_COMMAND.format(
```
mayabrandi, line 79: And then another function rsync_command() to be consistent

cg/cli/add.py
```diff
@@ -279,11 +279,14 @@ def relationship(
     required=True,
 )
 @click.option("--dry-run", is_flag=True)
+@click.option("--nanopore", is_flag=True)
```
mayabrandi, line 282:
```suggestion
@click.option("--nanopore", is_flag=True, help="...???")
```

mayabrandi (commented):
Had a few comments!

cg/meta/transfer/external_data.py
henningonsbring: fixed

cg/meta/transfer/external_data.py
henningonsbring: look at how I solved it and let me know what you think, I am not sure I understand how you want it.

cg/cli/add.py
henningonsbring: fixat

patrikgrenfeldt:
> @patrikgrenfeldt can you have a look if these changes are correct, or if I should put back the "u" (see picture). I am sure I did not remove those "u" in this PR, so I dont understand why the diff say I have removed "u". Right now when I check master branch for cg that is in production it seems like there are no "u", so that should be correct? 

According to the Python 2 reference the string literals default to ascii and that would cause a unicode escaped _runname_ to fail on this line.

According to the Python 3 reference the string literals defaults to UTF-8 and thus the u' becomes redundant. 

Conclusion: as long as the code is being run in an Python 3 environment we should be fine with the string without u'. 

Sources:
https://docs.python.org/2/reference/lexical_analysis.html#string-literals
https://docs.python.org/3/reference/lexical_analysis.html#string-literals
https://stackoverflow.com/questions/3235386/python-using-format-on-a-unicode-escaped-string

cg/meta/transfer/external_data.py
mayabrandi: I think i meant keeping the if statement in the     def transfer_sample_files_from_source() and have one function for each command type. But this looks fine to me!

mayabrandi (approved):
👍

henningonsbring:
I close this PR since the customer has agreed to change folder names so we can use the same `cg add external` command for both nanopore and illumina.

PR #1407, opened 2022-02-15
Sarscov2 start maya

mayabrandi:
You can do something like this. I didnt test. Youll have to do that. + we have to make sure we can use the contol udf. Thats not 100 verified yet.

### DIFF
--- a/cg/meta/workflow/mutant.py
+++ b/cg/meta/workflow/mutant.py
@@ -5,6 +5,9 @@ import shutil
 from pathlib import Path
 from typing import List, Optional
 
+from genologics.entities import Sample, Artifact
+
+from cg.apps.lims import LimsAPI
 from cg.constants import Pipeline
 from cg.meta.workflow.analysis import AnalysisAPI
 from cg.meta.workflow.fastq import MicrosaltFastqHandler, MutantFastqHandler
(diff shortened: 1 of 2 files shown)


cg/meta/workflow/mutant.py
```diff
@@ -108,12 +113,15 @@ def get_sample_parameters(self, sample_obj: models.Sample) -> MutantSampleConfig
...
-        ticket_id: int = samples[0].ticket_number
-        # 1. ticket_id will lead to samples that all come from the same pool
-        # 2. In the pool from "1." there will be samples from another ticket than ticket_id given as input
-        # 3. Return a list of sample objects from "2."
+        sample: models.Sample = samples[0]
+        pools = lims.get_artifacts(samplelimsid=sample.internal_id, type="Analyte",
+                                   process_type=MASTER_STEPS_UDFS['cov_pooling_step'])
+        pool: Artifact = LimsAPI.get_latest_artifact(lims_artifacts=pools)
+        negative_controls: List[Sample] = LimsAPI.get_controls(lims_artifact=pool)
+        return [self.status_db.sample(lims_sample.id) for lims_sample in negative_controls]
```
henningonsbring, line 124: We should be clear that this is about internal negative controls, not any negative control. There will be negative controls in that pool that are from the customer that will contain reads since it is processed differently in the lab.

PR #1421, opened 2022-02-23
Fix/fohm pandas indices

Vince-janv:
## Description
The sarscov2 case luckykite failed fohm upload with the following pandas error: `pandas.core.indexing.IndexingError: Unalignable boolean Series provided as indexer (index of the boolean Series and of the indexed object do not match).`

### Fixed

- The indices for the pandas dataframes are now the customer sample names, removing the need to ignore and/or reset indices when comparing dataframes.

### How to prepare for test
- [x] ssh to relevant server (depending on type of change)
- [x] Use stage: `us`
- [x] paxa the environment: `paxa`
- [x] install on stage (example for hasta):
`bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-cg-stage.sh fix/fohm_pandas_indices`

### How to test
- [x] do `cg upload fohm preprocess-all --dry-run -c luckykite`

### Expected test outcome
- [x] check that the function reaches `PermissionError: [Errno 13] Permission denied: '/home/hiseq.clinical/.ssh/external/cust124_id_ed25519'` (Apparently that's how a dry-run ends)

## Review
- [ ] code approved by
- [ ] tests executed by

## Implementation Plan
- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/upload/fohm/fohm.py
+++ b/cg/meta/upload/fohm/fohm.py
@@ -75,7 +75,7 @@ class FOHMUploadAPI:
             self._reports_dataframe.drop_duplicates(inplace=True, ignore_index=True)
             self._reports_dataframe = self._reports_dataframe[
                 self._reports_dataframe["provnummer"].str.contains(SARS_COV_REGEX)
-            ]
+            ].reset_index()
         return self._reports_dataframe
 
     @property
@@ -88,7 +88,7 @@ class FOHMUploadAPI:
             self._pangolin_dataframe.drop_duplicates(inplace=True, ignore_index=True)
             self._pangolin_dataframe = self.pangolin_dataframe[
                 self._pangolin_dataframe["taxon"].str.contains(SARS_COV_REGEX)
-            ]
+            ].reset_index()
         return self._pangolin_dataframe
 
     @property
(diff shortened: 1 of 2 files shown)


Vince-janv:
Results of running `cg upload fohm preprocess-all --dry-run -c luckykite`
Screenshot 2022-02-23 at 11 58 55

Vince-janv:
@Mropat I also tested on the case enoughhermit and it worked as well during a dry-run. Do you think any more tests need to be performed?

Mropat:
Awesome work finding the bug that caused this!
However I think for safety's sake it would bet better to keep the current index solution and reset index like in your first commit with bug fix. This is because Mutant post-proccess logic can append results to results file when a batch is restarted. This will cause duplicate lines with identical sample names, which cannot be used as an index and cause error.

Vince-janv:
> Awesome work finding the bug that caused this! However I think for safety's sake it would bet better to keep the current index solution and reset index like in your first commit with bug fix. This is because Mutant post-proccess logic can append results to results file when a batch is restarted. This will cause duplicate lines with identical sample names, which cannot be used as an index and cause error.

I see, I'll revert to the first commit then 👍

Mropat (approved):

Vince-janv:
Stage:
Screenshot 2022-02-25 at 15 00 14

Vince-janv:
Production:

PR #1506, opened 2022-06-08
Improves logging for compression and improved compression logics

henningonsbring:
Almost nothing is getting compressed via our crontab that runs `cg compress fastq -n 40` via [cg-compress-fastq.sh](https://github.com/Clinical-Genomics/servers/blob/master/resources/hasta.scilifelab.se/crontabs/cg-compress-fastq.sh). The `-n 40` states that 40 cases should be compressed. However in our logs when running `tac /home/proj/production/logs/cg-compress-fastq.log | grep "Individuals in" | grep "2022" | less` we can see:
```
2022-06-08 18:08:06 hasta.scilifelab.se cg.cli.compress.fastq[121393] INFO 0 Individuals in 40 (completed) cases where compressed
2022-06-08 10:07:24 hasta.scilifelab.se cg.cli.compress.fastq[161798] INFO 0 Individuals in 40 (completed) cases where compressed
```

The log indicates that 0 individuals (i.e. samples) were compressed for 40 cases. This indicates that the counter that cg uses to keep track of how many cases it compressed can be increased, while number of individuals (i.e. samples) are not increased.

This PR deals with the situation where cg find a case without samples. With current code in production cg would think it successfully compress such case and increase the counter. However, since it is not sure it this is the root to the problem I also make sure to log what cases cg consider to be compressed so further troubleshooting can continue if needed.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test
- [ ] Use stage: `us`

### How to test
- [ ] do ...

### Expected test outcome
- [ ] check that ...

## Review
- [ ] code approved by
- [ ] tests executed by

## Implementation Plan
- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/cli/compress/fastq.py
+++ b/cg/cli/compress/fastq.py
@@ -72,6 +72,8 @@ def fastq_cmd(
             continue
 
         LOG.info("Searching for FASTQ files in case %s", internal_id)
+        if not case.links:
+            continue
         for link_obj in case.links:
             sample_id: str = link_obj.sample.internal_id
             case_converted: bool = compress_api.compress_fastq(sample_id)
@@ -81,6 +83,7 @@ def fastq_cmd(
             ind_conversion_count += 1
         if case_converted:
             case_conversion_count += 1
+            LOG.info("Considering case %s converted", internal_id)
 
     LOG.info(
         "%s Individuals in %s (completed) cases where compressed",


henningonsbring:
Perhaps we should just change so cg count to 40 samples compressed, instead of 40 cases compressed (i.e. use `ind_conversion_count` instead of `case_conversion_count` in the comparison done in https://github.com/Clinical-Genomics/cg/blob/master/cg/cli/compress/fastq.py:
```
        if case_conversion_count >= number_of_conversions:
            break
```

PR #1535, opened 2022-08-04
add aggregate action and updateb pip

Mropat:
### Fixed

- Updated coveralls action

### Expected test outcome
- [x] Check if coveralls check reports correctly

## Review
- [ ] code approved by
- [x] tests executed by

### DIFF
--- a/.github/workflows/tests_and_coverage.yml
+++ b/.github/workflows/tests_and_coverage.yml
@@ -1,6 +1,10 @@
 name: CG CI
 
-on: ["pull_request"]
+on:
+  pull_request:
+  push:
+    branches:
+      - master
 
 jobs:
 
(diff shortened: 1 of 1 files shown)


.github/workflows/tests_and_coverage.yml
```diff
@@ -21,7 +25,7 @@ jobs:
 
     - name: CG Dependencies
       run: |
-        python -m pip install --upgrade pip==21.0.1
+        python -m pip install --upgrade pip
```
karlnyr, line 28: will this adapt to the locked python version? Or perhaps pip version is never dependent on the python version?

karlnyr (approved):
🥇

Vince-janv (approved):
🔥

.github/workflows/tests_and_coverage.yml
Mropat: It never did hehe. And now its even breaking new coveralls if its too outdated
We locked it this way also because pip version wasn't allowed to be updated on clinical-db

karlnyr: nice 👍

PR #1575, opened 2022-09-27
fix(CgError) Show error message in log

Vince-janv:
## Description
When logging error messages with CgError as seen below, the actual error message is not shown.

#### Log

> Upload of case amusedhawk to GISAID unseccessful , case amusedhawk will be removed from delivery batch

### Fixed

- Fixes logging of error messages

### How to prepare for test
- [ ] Use stage: `us`
`bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-cg-stage.sh fix-cg-errors`

### How to test
- [ ] do ...

### Expected test outcome
- [ ] check that ...

## Review
- [ ] code approved by
- [ ] tests executed by

## Implementation Plan
- [ ] Document in ...
- [ ] Inform to ...


Vince-janv:
### With this branch:

> Upload of case amusedhawk to GISAID unseccessful No fasta file found for sample ACC10032A77, case amusedhawk will be removed from delivery batch

cg/exc.py
```diff
@@ -127,8 +126,8 @@ class HousekeeperFileMissingError(CgError):
     Exception raised when a file is missing in Housekeeper.
     """
 
-    def __init__(self, message, errors=None):
-        self.message = message
+    def __init__(self, message: str = "", errors=None):
```
henrikstranneheim, line 129:
```suggestion
    def __init__(self, message: str = EMPTY_STRING, errors=None):
```
Matter of style - what do you think of using a named constant instead. It does read better.

We have some in cg/constants/symbols.py

henrikstranneheim (approved):
Good catch! 👍

cg/exc.py
Vince-janv: I agree for those that are currently in `cg/constants/symbols.py` but for an empty string i quite like `""`. I feel like I would be more confused by `EMPTY_STRING`. I don't have strong opinion though, so feel free to overrule me 😉

henrikstranneheim: Ok, let's keep it as it is. 👍

PR #1585, opened 2022-10-03
Change default gene panel for MAF cases from None to OMIM-AUTO

peterpru:
## Description
With recent changes, when the attribute 'gene panel' is empty for a case in MIP-DNA, with delivery type ANALYSIS_FILES such as for MAF controls, it causes the delivery step of that case to fail.

There are two options to fix this problem. As the MAF cases are run through the mip-dna pipeline to generate a VCF file, the most straightforward fix is to add the OMIM gene panel so that the delivery report is completed. The alternative would be to create a new pipeline just for MAF cases consisting of a mapper and a variant caller. The current suggestion here is to complete the delivery, as we have run the full mip pipeline for which delivery is the natural conclusion.

### Added

-

### Changed

- Change default gene panel for MAF cases from None to OMIM-AUTO

### Fixed

-

### How to prepare for test
- [ ] In github: Action > Deploy branch to staging environment > empty-gene-panel-field.

### How to test
- [ ] In https://clinical-stage.scilifelab.se/ make an order for a fastq delivery.

### Expected test outcome
- [ ] check that the automatically created MAF case has as gene panel "OMIM-AUTO"
image

## Review
- [ ] code approved by
- [ ] tests executed by

## Implementation Plan
- [ ] Document in ...
- [ ] Inform to ...


cg/meta/orders/fastq_submitter.py
```diff
@@ -61,7 +61,7 @@ def create_maf_case(self, sample_obj: models.Sample):
             data_analysis=Pipeline(Pipeline.MIP_DNA),
             data_delivery=DataDelivery(DataDelivery.ANALYSIS_FILES),
             name="_".join([sample_obj.name, "MAF"]),
-            panels=None,
+            panels=["OMIM-AUTO"],
```
henrikstranneheim, line 64: I think this is called `OMIM-AUTO`. Should we add a constant class in: cg/constants/gene_panel.py that defines the panel ids? And also make the `MASTER_LIST` use that class

peterpru: I have now changed MasterList to a constant class, and refactored it everywhere I could see it be used.

cg/constants/gene_panel.py
```diff
@@ -1,30 +1,32 @@
...
+    MTDNA: str = "mtDNA"
+    NBS_M: str = "NBS-M"
+    NEURODEG: str = "NEURODEG"
+    NMD: str = "NMD"
+    PEDHEP: str = "PEDHEP"
+    PID: str = "PID"
+    PIDCAD: str = "PIDCAD"
+    OMIM_AUTO: str = "OMIM-AUTO"
+    SKD: str = "SKD"
+
```
henrikstranneheim, line 32: How about adding a classmethod here:

```
@classmethod
    def get_panel_names(cls, MasterList: MasterList) -> List[str]:
        """Return pnale names from Master list."""
        return [panel_name.value for panel_name in MasterList]
```

cg/meta/workflow/mip.py
```diff
@@ -159,8 +159,8 @@ def write_panel(self, case_id: str, content: List[str]):
     def convert_panels(customer: str, default_panels: List[str]) -> List[str]:
         """Convert between default panels and all panels included in gene list."""
         # check if all default panels are part of master list
-        if customer in COLLABORATORS and all(panel in MASTER_LIST for panel in default_panels):
-            return MASTER_LIST
+        if customer in COLLABORATORS and all(panel in dir(MasterList) for panel in default_panels):
```
henrikstranneheim, line 162: And using the classmethod here

henrikstranneheim (commented):
👍

henrikstranneheim:
@peterpru Should we also take the opportunity to rename MasterList to something more intuitive? The name should include "gene_panel" or similar at least to reflect what it is a master list of.

cg/meta/workflow/mip.py
peterpru: Added.

cg/constants/gene_panel.py
peterpru: Great, I have now added the classmethod.

peterpru:
@henrikstranneheim changing the name sounds good, I have used GenePanelMasterList, which feels a bit long, but does reflect its contents.

cg/constants/gene_panel.py
```diff
@@ -1,30 +1,40 @@
...
+    @classmethod
+    def get_panel_names(cls, panels=None) -> List[str]:
+        """Return requested panel names from Master list, or all panels if none are specified."""
+        if not panels:
+            return list(map(lambda p: p.value, cls))
```
henrikstranneheim, line 39:
```suggestion
            return list(map(lambda panel: panel.value, cls))
```
Avoid single character variables if possible they are hard to understand and not mnemonic

henrikstranneheim (approved):
Well done!

cg/constants/gene_panel.py
peterpru: That sounds good.

peterpru:
Deploying in production:
image

Issue #1614, opened 2022-11-01
microSALT Typing report - missing lab prep and sequencing method 


karlnyr:
## Description

The full name of lab prep and sequencing method are missing from the typing report:
![image](https://user-images.githubusercontent.com/45558267/199195959-93965856-f2ab-4bfb-966e-4d173c3f4513.png)

## Suggested solution

Remove the string split done in the method that fetches the information in:

https://github.com/Clinical-Genomics/cg/blob/ebeae86872cb4fbf4000f30fbb14ff3ca610f71c/cg/meta/workflow/microsalt.py#L203

and:

https://github.com/Clinical-Genomics/cg/blob/ebeae86872cb4fbf4000f30fbb14ff3ca610f71c/cg/meta/workflow/microsalt.py#L206

## This can be closed when

The full prep and sequencing method are printed in the typing report.

## Blocked by

*NA*

karlnyr:
@Karl-Svard @mathiasbio

PR #1634, opened 2022-11-13
fix: missing full atlas-document names for microsalt typing report

mathiasbio:
## Description

Earlier (before roughly summer 2022?) all library prep and sequencing method documents were in the AM system, they had a docID and a name (for example: 1830:8 - NovaSeq 6000 Sequencing method). For the typing report it was deemed that the docID was sufficient and so the full name was split on spaces and only the first element in the list, the docID was kept. With the transition to Atlas, and the different name structure there (for example: Manual cost-efficient library preparation for microbial WGS (6.11.24), this is obviously problematic. 

In the issue https://github.com/Clinical-Genomics/cg/issues/1614 the suggested solution was to remove the split, which is what has been done here. The result being that all samples regardless of having been handled in an AM or Atlas document should have the full doc name in the microsalt typing report.

The tiny issue with testing this is that it seems that most samples that have been prepped with old AM-documented methods are not updated in LIMS. However, at least sample ACC10152A1 had an Atlas documented library prep and, running:

<pre><code>cg workflow microsalt config-case -s ACC10152A1 
</code></pre>

On the changes lead to "Manual" --> "Manual cost-efficient library preparation for microbial WGS (6.11.24)" in the config-file input to microsalt which transferred into the typing report:

image

The only concern perhaps is that the reports will be a bit stretched with these long method doc names.

### Added

-

### Changed

- Removed split on spaces for the library prep and sequencing method document string for the microsalt typing report

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b fix_microsalt_missing_method_in_typingreport -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


mathiasbio:
Might need to update LIMS on stage to test entirely as it seems the samples sequenced and prepped with methods documented in Atlas don't exist in LIMS-stage due to being too new. But here's the effect on samples with docs in AM, such as ACC9068A1 in the configfile where the typing report derives its info: 
Before:
"method_libprep": "1764:16", "method_sequencing": "1830:7",
After: 
"method_libprep": "1764:16 - None", "method_sequencing": "1830:7,"
And here's an example from the typing report as shown in the description for a sample with a mix of both AM and Atlas docs:
image

Not sure if the "None" part will look too pretty in the typing-report, if you think it's necessary I can try to see what can be done about it.

henrikstranneheim (approved):
👍

mathiasbio:
Deployed in production:
image

PR #1640, opened 2022-11-22
feat(external data) Re-writing of the external-data module

Vince-janv:
## Description
The external data module was poorly written, clunky and did not facilitate things as intended. This PR greatly simplifies the process and will allow for automatic starting of cases (given that the customers have performed the upload and order correctly)
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/external_data.py
```diff
@@ -0,0 +1,53 @@
...
+
+class ExternalDataHandler:
+    """Class for handling the integration of external data into the common workflow."""
+
+    def __init__(self, config: CGConfig):
+        self.hk_api: HousekeeperAPI = config.housekeeper_api
+        self.status_db: Store = config.status_db
+        self.external_data_dir = Path(config.external.external_data_dir)
+
+    def curate_external_folder(self):
```
henrikstranneheim, line 22: What is the curate part? Looks like it just iters and adds?

cg/meta/external_data.py
```diff
@@ -0,0 +1,53 @@
...
+                ).first()
+                self.add_sample_to_hk(sample_folder=sample_folder, sample_object=sample_object)
+
+    def add_sample_to_hk(self, sample_folder: Path, sample_object: models.Sample):
+        """Creates a housekeeper bundle for a sample if non-existant and adds any file that is
+        not already added."""
+        hk_version: hk_models.Version = self.hk_api.get_create_version(
+            bundle=sample_object.internal_id
+        )
+        for fastq_file in sample_folder.iterdir():
```
henrikstranneheim, line 39: Can we assume that there will be only FASTQ files in the sample_folder? Should we add a guard to also check a file ending or be very serious and actually open the file and check proper fastq format? Experience is be very strict in the data ingestion be cause it will save you a lot error checking and issues later.

cg/meta/external_data.py
```diff
@@ -0,0 +1,53 @@
...
+            if self.hk_api.files(version=hk_version, path=fastq_file.as_posix()):
+                LOG.debug(f"File {fastq_file} already added to bundle {hk_version.bundle.name}")
+                continue
+            LOG.info(f"File {fastq_file} added to bundle {hk_version.bundle.name}")
+            self.hk_api.add_file(path=fastq_file, tags=HK_FASTQ_TAGS, version_obj=hk_version)
```
henrikstranneheim, line 44:
```suggestion
            self.hk_api.add_file(path=fastq_file, tags=HK_FASTQ_TAGS, version_obj=hk_version)
            LOG.info(f"File {fastq_file} added to bundle {hk_version.bundle.name}")
```
Otherwise we might think it was added, but the `add_file` could error

cg/meta/external_data.py
```diff
@@ -0,0 +1,53 @@
...
+        self.hk_api.commit()
+        self.start_associated_case(sample_object=sample_object)
+
+    def start_associated_case(self, sample_object: models.Sample):
```
henrikstranneheim, line 48:
```suggestion
    def set_to_analyze_for_sample_associated_cases(self, sample_object: models.Sample):
```
Can be multiple right? Maybe there is a shorter more to the point name, but this process does not really start the case analysis.

tests/meta/transfer/test_external_data.py
```diff
@@ -3,266 +3,150 @@
...
-
-    # Given a file containing a md5sum
-    md5sum_file = Path(f"{fastq_file.as_posix()}.md5")
-
-    # Then the function should extract it
-    assert extract_md5sum(md5sum_file=md5sum_file) == "a95cbb265540a2261fce941059784fd1"
+    # GIVEN a sample linked to a case set to hold
+    for case_links in sample_obj.links:
+        case_links.family.action = "hold"
+    # WHEN the function is called
```
henrikstranneheim, line 22: Let's extend `CaseActions` with "hold" and use it here
```suggestion
        case_links.family.action = CaseActions.HOLD
        
    # WHEN the function is called
```

tests/meta/transfer/test_external_data.py
```diff
@@ -3,266 +3,150 @@
...
+    external_data_handler.start_associated_case(sample_object=sample_obj)
+    # THEN the action should have changed
```
henrikstranneheim, line 24:
```suggestion
    external_data_handler.start_associated_case(sample_object=sample_obj)

    # THEN the action should have changed
```

tests/meta/transfer/test_external_data.py
```diff
@@ -3,266 +3,150 @@
...
+    assert all(case_links.family.action == "analyze" for case_links in sample_obj.links)
```
henrikstranneheim, line 25: Use CaseActions

henrikstranneheim (commented):
Nice start. 👍

cg/meta/external_data.py
Vince-janv: It's because it's still a draft 😉

cg/meta/external_data.py
Vince-janv: From my experience with the external data we rarely encounter faulty fastq files, and in the few cases we have it is almost always due to truncation (which would require reading trough the whole file). This is why I'm removing md5sums since they always matched and were very slow to calculate. I think we can trust the customers for now (they are given strict instructions) and see if problems arise.

cg/meta/external_data.py
Vince-janv: Good catch 👍

cg/meta/external_data.py
Vince-janv: True, I will figure out a more snazzy name

Vince-janv: Turns out we just delete it

Vince-janv:
Will not be worked-on anytime soon

PR #1744, opened 2023-02-14
Copy specific files instead of entire build context

seallard:
## Description
SonarCloud flagged the recursive copy in our Dockerfile. It does not really matter since the build context is our public repository, but this might change over time or it might happen that it is done locally for some reason. It is better to explicitly specify which directories/files that should be copied to the container.

### Fixed

- Replace recursive copy in Dockerfile

### How to prepare for test

- [x] Paxa the environment: `paxa` cg-stage vm1

### How to test

- [x] Deploy to the stage environment.

### Expected test outcome

- [x] Check that the deploy is successful, that the image is built and the container runs as expected.

## Review

- [x] Tests executed by SA
- [x] "Merge and deploy" approved by HS

## Implementation Plan
- [ ] Deploy this branch on cg prod

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,9 @@ ENV GOOGLE_OAUTH_CLIENT_SECRET="1"
 
 
 WORKDIR /home/src/app
-COPY cg MANIFEST.in pyproject.toml requirements* setup.py templates ./
+COPY cg ./cg
+COPY templates ./templates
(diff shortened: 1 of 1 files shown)


henrikstranneheim (approved):
👍

seallard:
Deployed successfully to stage and prod.

```
Log deploy... done.
cg, version 26.0.7
```

PR #1787, opened 2023-03-10
Move filters into separate directory

seallard:
## Description
I think we should group the filters in a separate directory.

### Fixed
- Move filters into separate directory

## Review
- [x] "Merge and deploy" approved by CO, HS, SD


ChrOertlin (approved):
I like this idea the directory was becoming quite full. In addition to this I would also move the tests to tests/store/filters/

henrikstranneheim (approved):
Awesome!

diitaz93 (approved):
👍

seallard:
I'm unable to deploy since I don't have access to Hasta yet.

PR #1793, opened 2023-03-10
Refactor set case command

seallard:
## Description
Made the set case command implementation more readable by extracting smaller functions.

### Changed
- Refactor set case command

## Review


cg/cli/set/family.py
```diff
@@ -47,40 +48,87 @@ def family(
     customer_id: Optional[str],
 ):
     """Update information about a case."""
+
+    check_nothing_to_change(action, panels, priority, customer_id, data_analysis, data_delivery)
```
henrikstranneheim, line 52: Use named args

cg/cli/set/family.py
```diff
@@ -47,40 +48,87 @@ def family(
...
-        raise click.Abort
-    if not any([action, panels, priority, customer_id, data_analysis, data_delivery]):
-        LOG.error("Nothing to change")
-        raise click.Abort
+    case: Family = get_case(family_id, status_db)
+
     if action:
-        LOG.info("Update action: %s -> %s", case_obj.action or "NA", action)
-        case_obj.action = action
+        update_action(case, action)
```
henrikstranneheim, line 58: Use named args

cg/cli/set/family.py
```diff
@@ -47,40 +48,87 @@ def family(
...
-            LOG.error("Unknown customer: %s", customer_id)
-            raise click.Abort
-        LOG.info(f"Update customer: {case_obj.customer.internal_id} -> {customer_id}")
-        case_obj.customer = customer_obj
+        update_customer(case, customer_id, status_db)
+
     if data_analysis:
-        LOG.info(f"Update data_analysis: {case_obj.data_analysis or 'NA'} -> {data_analysis}")
-        case_obj.data_analysis = data_analysis
+        update_data_analysis(case, data_analysis)
```
henrikstranneheim, line 64: Same here

cg/cli/set/family.py
```diff
@@ -47,40 +48,87 @@ def family(
...
+
     if data_delivery:
-        LOG.info(f"Update data_delivery: {case_obj.data_delivery or 'NA'} -> {data_delivery}")
-        case_obj.data_delivery = data_delivery
+        update_data_delivery(case, data_delivery)
```
henrikstranneheim, line 67: And here

cg/cli/set/family.py
```diff
@@ -47,40 +48,87 @@ def family(
...
+
     if priority:
-        LOG.info(f"Update priority: {case_obj.priority.name} -> {priority.name}")
-        case_obj.priority = priority
+        update_priority(case, priority)
 
     status_db.commit()
+
+
+def check_nothing_to_change(action, panels, priority, customer_id, data_analysis, data_delivery):
```
henrikstranneheim, line 78: Add type hint and return hint.
Should we input a list instead and move `[action, panels, priority, customer_id, data_analysis, data_delivery]` to the caller?

cg/cli/set/family.py
```diff
@@ -47,40 +48,87 @@ def family(
...
+    case: Family = status_db.family(case_id)
+
+    if case is None:
+        LOG.error(f"Can't find case {case_id}")
+        raise click.Abort
+
+    return case
+
+
+def update_action(case: Family, action: str) -> None:
```
henrikstranneheim, line 94: There are these in cg/meta/orders/case_submitter.py:

```
               self._append_ticket(ticket_id=ticket_id, case=status_db_case)
                self._update_action(action=CaseActions.ANALYZE, case=status_db_case)
                self._update_case_panel(panels=case["panels"], case=status_db_case)
```
Would be good to define the functions once and in a update.py handler to keep them collected and easy to find

henrikstranneheim (commented):
Nice to clean this 👍

cg/cli/set/family.py
seallard: Yes, I think it is a good idea to pass all of the options as a list.

cg/cli/set/family.py
seallard: Yes, I think this would be good. Maybe we could have the following directory structure for our handlers:

```
handlers/
    ├── base.py
    ├── base_find_handler.py
    ├── base_update_handler.py
    ├── base_delete_handler.py
    ├── family/
    │   ├── __init__.py
    │   ├── family_find_handler.py
    │   ├── family_update_handler.py
    │   └── family_delete_handler.py
```

seallard:
Deployed to stage.

PR #1821, opened 2023-03-20
Remove unused imports

seallard:
Removed all unused imports using autoflake:

```bash
autoflake --recursive --exclude=__init__.py --remove-all-unused-imports -i cg
autoflake --recursive --exclude=conftest.py --remove-all-unused-imports -i tests
```

Upside: less clutter and time spent importing unnecessary libraries.

### Fixed
- Remove all unused imports

## Review

- [ ] Tests executed by

### DIFF
--- a/cg/cli/add.py
+++ b/cg/cli/add.py
@@ -27,6 +27,7 @@ LOG = logging.getLogger(__name__)
 @click.group()
 def add():
     """Add new things to the database."""
+    pass
 
 
 @add.command()
--- a/cg/cli/add.py
+++ b/cg/cli/add.py
@@ -27,7 +27,6 @@ LOG = logging.getLogger(__name__)
 @click.group()
 def add():
     """Add new things to the database."""
-    pass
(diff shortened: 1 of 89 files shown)


cg/cli/add.py
```diff
@@ -27,7 +27,6 @@
 @click.group()
 def add():
     """Add new things to the database."""
-    pass
```
diitaz93, line 30: I don't think we should remove the `pass`es

diitaz93 (changes requested):
I don't understand the motivation behind removing the pass statements in empty functions

cg/cli/add.py
seallard: Good catch, seems like the tool automatically removed these since they are not needed. Maybe we want to be explicit anyway. I'm restoring them.

diitaz93 (approved):
👍

seallard:
Deployed to stage.

PR #1823, opened 2023-03-21
Lock housekeeper to latest minor release

seallard:
## Description
We don't want to use a new major version of housekeeper by mistake, so we should lock it to the most recent minor release and update the major manually together with the necessary patches.

### Fixed
- Lock housekeeper version to the latest minor version.


seallard:
Deployed to stage and prod.

PR #1825, opened 2023-03-21
Replace undefined function call

seallard:
## Description
The ```cases_to_mip_analyze``` function has been removed but is still used in the /families endpoint.
Not sure where the endpoint is used

### Fixed
- Replaced undefined function


cg/server/api.py
```diff
@@ -154,7 +154,7 @@ def parse_cases():
 def parse_families():
     """Return families."""
     if request.args.get("status") == "analysis":
-        records = db.cases_to_mip_analyze()
+        records = db.cases_to_analyze(pipeline=Pipeline.MIP_DNA)
```
diitaz93, line 157: add type hint

henrikstranneheim (approved):
👍. It might be in the status view where you can follow the sample through the system as a customer. However, I do not think it is used much.

PR #1829, opened 2023-03-22
Fix typo causing removal of failed flow cells to fail

seallard:
## Description
Fix typo causing commands related to removing failed flow cells to error out.

The `cg clean invalid-flow-cell-dirs -f` command fails.

### Fixed
- Typo in cleanup for failed flow cells


henrikstranneheim (approved):
Good catch! 👍

PR #1847, opened 2023-03-24
Remove dead case filter functions

seallard:
## Description
`family_has_correct_number_tumor_normal_samples` is only called by `get_valid_cases_to_analyze`, which in turn is not called anywhere.

### Fixed
- Remove dead functions `family_has_correct_number_tumor_normal_samples` and `get_valid_cases_to_analyze`.

### DIFF
--- a/cg/meta/workflow/balsamic.py
+++ b/cg/meta/workflow/balsamic.py
@@ -604,21 +604,24 @@ class BalsamicAnalysisAPI(AnalysisAPI):
         )
         return config_data["analysis"]["BALSAMIC_version"]
 
-    def family_has_correct_number_tumor_normal_samples(self, case_id: str) -> bool:
+    def case_has_correct_number_tumour_normal_samples(self, case_internal_id: str) -> bool:
         """Evaluates if a case has exactly one tumor and up to one normal sample in ClinicalDB.
         This check is only applied to filter jobs which start automatically"""
-        query = (
-            self.status_db.query(Sample)
-            .join(Family.links, FamilySample.sample)
-            .filter(Family.internal_id == case_id)
-            .filter(Family.data_analysis == self.pipeline)
+
+        samples: List[Sample] = self.status_db.get_samples_by_case_and_pipeline(
+            case_internal_id=case_internal_id, pipeline=self.pipeline
         )
-        return all(
-            [
-                len(query.filter(Sample.is_tumour == False).all()) <= 1,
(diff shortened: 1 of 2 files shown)


cg/meta/workflow/balsamic.py
```diff
@@ -604,30 +604,35 @@ def get_pipeline_version(self, case_id: str) -> str:
...
+        samples: List[Sample] = self.status_db.get_samples_by_case_and_pipeline(
+            case_internal_id=case_internal_id, pipeline=self.pipeline
         )
 
+        normal_samples_count: int = self.count_normal_samples(samples=samples)
+        tumour_samples_count: int = self.count_tumour_samples(samples=samples)
+
+        return normal_samples_count <= 1 and tumour_samples_count == 1
+
+    def count_normal_samples(self, samples: List[Sample]) -> int:
```
seallard, line 620: Maybe this logic should be in the find handler

seallard (commented):
Maybe the count_normal_samples and count_tumour_samples should live in the Find handler. Not sure.

cg/store/api/find_business_data.py
```diff
@@ -185,6 +185,22 @@ def get_cases_from_sample(self, sample_entry_id: str) -> Query:
             case_samples=self._get_join_case_sample_query(),
         )
 
+    def get_samples_by_case_and_pipeline(
```
seallard, line 188: Is it really necessary to filter by pipeline? Why is it not enough to just filter by case_id? Can the case have samples being run in other pipelines?

cg/meta/workflow/balsamic.py
```diff
@@ -604,30 +604,35 @@ def get_pipeline_version(self, case_id: str) -> str:
...
-            [
-                len(query.filter(Sample.is_tumour == False).all()) <= 1,
-                len(query.filter(Sample.is_tumour == True).all()) == 1,
-            ]
+
+        samples: List[Sample] = self.status_db.get_samples_by_case_and_pipeline(
+            case_internal_id=case_internal_id, pipeline=self.pipeline
         )
 
+        normal_samples_count: int = self.count_normal_samples(samples=samples)
```
seallard, line 615: I think we should skip type hints when the type is super clear by context, for instance here.

henrikstranneheim: I think it is still good to do to tell non-humans (like editors and mypy (we should add it)) what to expect

PR #1858, opened 2023-03-27
Refactor find family

seallard:
### Fixed
- Extract filters for Family by customer entry id and case name

## Review
- [ ] Tests executed by

### DIFF
--- a/cg/meta/orders/fastq_submitter.py
+++ b/cg/meta/orders/fastq_submitter.py
@@ -81,7 +81,7 @@ class FastqSubmitter(Submitter):
         if not customer:
             raise OrderError(f"Unknown customer: {customer_id}")
         new_samples = []
-        case: Family = self.status.find_family(customer=customer, name=ticket_id)
+        case: Family = self.status.get_case_by_name_and_customer(customer=customer, name=ticket_id)
         submitted_case: dict = items[0]
         with self.status.session.no_autoflush:
             for sample in items:
--- a/cg/meta/orders/fastq_submitter.py
+++ b/cg/meta/orders/fastq_submitter.py
@@ -81,7 +81,9 @@ class FastqSubmitter(Submitter):
         if not customer:
             raise OrderError(f"Unknown customer: {customer_id}")
         new_samples = []
-        case: Family = self.status.get_case_by_name_and_customer(customer=customer, name=ticket_id)
+        case: Family = self.status.get_case_by_name_and_customer(
+            customer=customer, case_name=ticket_id
+        )
         submitted_case: dict = items[0]
         with self.status.session.no_autoflush:
             for sample in items:
--- a/cg/meta/orders/metagenome_submitter.py
+++ b/cg/meta/orders/metagenome_submitter.py
@@ -89,9 +89,9 @@ class MetagenomeSubmitter(Submitter):
         case: Family = self.status.get_case_by_name_and_customer(
             customer=customer, case_name=str(ticket_id)
         )
-        case: dict = items[0]
+        case_dict: dict = items[0]
         with self.status.session.no_autoflush:
-            for sample in case["samples"]:
+            for sample in case_dict["samples"]:
                 new_sample = self.status.add_sample(
                     name=sample["name"],
                     sex="unknown",
@@ -113,11 +113,11 @@ class MetagenomeSubmitter(Submitter):
 
                 if not case:
                     case = self.status.add_case(
-                        data_analysis=Pipeline(case["data_analysis"]),
-                        data_delivery=DataDelivery(case["data_delivery"]),
+                        data_analysis=Pipeline(case_dict["data_analysis"]),
+                        data_delivery=DataDelivery(case_dict["data_delivery"]),
                         name=str(ticket_id),
                         panels=None,
-                        priority=case["priority"],
+                        priority=case_dict["priority"],
                         ticket=ticket_id,
                     )
                     case.customer = customer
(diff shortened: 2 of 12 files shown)


cg/meta/workflow/microsalt.py
```diff
@@ -307,7 +307,7 @@ def resolve_case_sample_id(
     def get_case_id_from_ticket(self, unique_id: str) -> Tuple[str, None]:
         """If ticked is provided as argument, finds the corresponding case_id and returns it.
         Since sample_id is not specified, nothing is returned as sample_id"""
-        case_obj: Family = self.status_db.find_family_by_name(unique_id)
+        case_obj: Family = self.status_db.get_case_by_name(unique_id)
```
henrikstranneheim, line 310:
```suggestion
        case: Family = self.status_db.get_case_by_name(unique_id)
```
Use named args

cg/store/filters/status_case_filters.py
```diff
@@ -111,18 +111,36 @@ def filter_case_by_internal_id(cases: Query, internal_id: str, **kwargs) -> Quer
     return cases.filter(Family.internal_id == internal_id)
 
 
+def filter_cases_by_customer_entry_id(cases: Query, customer_entry_id: int, **kwargs) -> Query:
+    """Return cases with matching customer id."""
+    return cases.filter(Customer.id == customer_entry_id)
```
henrikstranneheim, line 116: Should we start refactoring entry id to database id?

cg/store/api/find_business_data.py
```diff
@@ -248,13 +248,22 @@ def get_sequenced_samples(self, family_id: str) -> List[Sample]:
         samples: List[Sample] = self.get_samples_by_case_id(family_id)
         return [sample for sample in samples if sample.sequencing_qc]
 
-    def find_family(self, customer: Customer, name: str) -> Family:
-        """Find a family by family name within a customer."""
-        return self.Family.query.filter_by(customer=customer, name=name).first()
+    def get_case_by_name_and_customer(self, customer: Customer, name: str) -> Family:
```
henrikstranneheim, line 251:
```suggestion
    def get_case_by_name_and_customer(self, customer: Customer, case_name: str) -> Family:
```
To distinguish from customer name and make it very explicit what we expect as input

cg/meta/orders/fastq_submitter.py
```diff
@@ -81,7 +81,7 @@ def store_items_in_status(
         if not customer:
             raise OrderError(f"Unknown customer: {customer_id}")
         new_samples = []
-        case: Family = self.status.find_family(customer=customer, name=ticket_id)
+        case: Family = self.status.get_case_by_name_and_customer(customer=customer, name=ticket_id)
```
henrikstranneheim, line 84:
```suggestion
        case: Family = self.status.get_case_by_name_and_customer(customer=customer, case_name=ticket_id)
```
It is not clear that (at least to me) the ticket id is the same as the case name

cg/meta/orders/metagenome_submitter.py
```diff
@@ -86,7 +86,7 @@ def store_items_in_status(
         if customer is None:
             raise OrderError(f"unknown customer: {customer_id}")
         new_samples = []
-        case_obj = self.status.find_family(customer=customer, name=str(ticket_id))
+        case_obj = self.status.get_case_by_name_and_customer(customer=customer, name=str(ticket_id))
```
henrikstranneheim, line 89:
```suggestion
        case: Family = self.status.get_case_by_name_and_customer(customer=customer, case_name=str(ticket_id))
```

tests/cli/workflow/conftest.py
```diff
@@ -108,21 +108,21 @@ def fixture_fastq_case(case_id, family_name, sample_id, cust_sample_id, ticket:
 def dna_case(analysis_store, helpers) -> Family:
     """Case with DNA application"""
     cust = helpers.ensure_customer(analysis_store)
-    return analysis_store.find_family(cust, "dna_case")
+    return analysis_store.get_case_by_name_and_customer(cust, "dna_case")
```
henrikstranneheim, line 111: Use named args in the changed calls in this module

henrikstranneheim (approved):
Looks great 👍

cg/meta/workflow/microsalt.py
seallard: I think we should create separate PR:s where we rename all instances of model_obj, model_object, model_objects, model_objs to just model. But lets do it at a later point when the codebase is more stable.

cg/store/filters/status_case_filters.py
seallard: Not sure, maybe we can do that later in separate PR:s. I think I prefer customer_entry_id.

henrikstranneheim: A quick googling seems to indicate that entry id is a good name for a unique database id. Let's keep it and use it consistently. 👍

cg/meta/orders/fastq_submitter.py
seallard: Me neither, but that is how the function was implemented. Looking into it

seallard: In some places, the name of the case is set as the ticket id or some variation on the ticket id (for instance by adding a suffix). In other places, the name of the case does not seem to be explicitly related to the ticket id. I created a separate issue for it https://github.com/Clinical-Genomics/cg/issues/1862

cg/store/filters/status_case_filters.py
seallard: Ok!

PR #1872, opened 2023-03-29
Remove unused threshold parameter

seallard:
## Description
Remove unused threshold parameter in ScoutAPI upload.

### Fixed
- Remove unused parameter.


ivadym (approved):
👍 

Seems that it has been used for the rank score threshold before and moved to `build_load_config` 👀  Good catch!

PR #1916, opened 2023-04-13
Cache Google Oauth2 certificate

seallard:
## Description
Instead of fetching the certificate on each received request, we can cache it. This should reduce the response time for requests against our API. Google only rotates the certificate once per day.

The current implementation does not use that two certificates are returned from Google.

### Added
- Caching of Google OAuth certificate

## Review

- [ ] Tests executed by

### DIFF
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -35,7 +35,8 @@ BLUEPRINT = Blueprint("api", __name__, url_prefix="/api/v1")
 cache = TTLCache(maxsize=1, ttl=3600)
 
 
-def get_certificate_ttl(response_data):
+def get_certificate_ttl(response_data) -> int:
+    """Extract time to live in seconds for certificate from response headers."""
     expires_header = response_data.headers.get("Expires")
     expires = datetime.strptime(expires_header, "%a, %d %b %Y %H:%M:%S %Z")
     ttl = int((expires - datetime.utcnow().replace(tzinfo=timezone.utc)).total_seconds())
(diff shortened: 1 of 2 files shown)


cg/server/api.py
```diff
@@ -25,12 +27,49 @@
...
 
+cache = TTLCache(maxsize=1, ttl=3600)
+
+
+def get_certificate_ttl(response_data):
+    expires_header = response_data.headers.get("Expires")
+    expires = datetime.strptime(expires_header, "%a, %d %b %Y %H:%M:%S %Z")
+    ttl = int((expires - datetime.utcnow().replace(tzinfo=timezone.utc)).total_seconds())
+
+    return ttl
```
Vince-janv, line 43: Docstring and type hints
```suggestion
def get_certificate_ttl(response_data) -> int:
    """Returns the time-to-live of given response."""
    expires_header: str = response_data.headers.get("Expires")
    expires: datetime = datetime.strptime(expires_header, "%a, %d %b %Y %H:%M:%S %Z")
    ttl: int = int((expires - datetime.utcnow().replace(tzinfo=timezone.utc)).total_seconds())
    
    return ttl
```

Vince-janv (approved):
Very neat solution! So `TTLCache` removes the cert when the ttl expires? Just add docstrings and type hints and I'm happy 😊

seallard:
> Very neat solution! So `TTLCache` removes the cert when the ttl expires? Just add docstrings and type hints and I'm happy 😊

Yes, but it apparently only uses the ttl specified when creating the cache so I need to make some changes 🙁

seallard:
Tested in stage

PR #1921, opened 2023-04-13
Refactor case filtering

seallard:
### Changed
- Use case filter functions

## Review
- [ ] Tests executed by

### DIFF
--- a/cg/store/api/status.py
+++ b/cg/store/api/status.py
@@ -233,7 +233,7 @@ class StatusHandler(BaseHandler):
         """Return all cases that are ready to be compressed by SPRING."""
         case_filter_functions: List[CaseFilter] = [
             CaseFilter.GET_HAS_INACTIVE_ANALYSIS,
-            CaseFilter.GET_NEW,
+            CaseFilter.GET_OLD,
         ]
         return apply_case_filter(
             filter_functions=case_filter_functions,
--- a/cg/store/api/status.py
+++ b/cg/store/api/status.py
@@ -1,6 +1,6 @@
 from datetime import datetime, timedelta
 from types import SimpleNamespace
-from typing import List, Optional
+from typing import Callable, List, Optional
 
 
 from sqlalchemy.orm import Query
@@ -538,9 +538,9 @@ class StatusHandler(BaseHandler):
         priority: str,
         sample_id: str,
     ) -> Query:
-        case_q = self._get_query(table=Family)
-        filter_functions = []
-        # family filters
+        cases_query: Query = self._get_query(table=Family)
+        filter_functions: List[Callable] = []
+
         filter_case_order_date = None
         if days != 0:
             filter_case_order_date = datetime.now() - timedelta(days=days)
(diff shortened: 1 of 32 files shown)


cg/store/filters/status_case_filters.py
```diff
@@ -173,6 +173,16 @@ def filter_cases_by_name_search(cases: Query, name_search: str, **kwargs) -> Que
     return cases.filter(Family.name.like(f"%{name_search}%"))
 
 
+def filter_cases_by_pipeline_search(cases: Query, pipeline_search: str, **kwargs) -> Query:
+    """Return cases with pipeline matching the search pattern."""
+    return cases.filter(Family.data_analysis.ilike(f"%{pipeline_search}%"))
```
Vince-janv, line 178: I don't think this should be a strict search. We have several pipelines with similar names and I don't see any merit to returning cases from different pipelines here.
```suggestion
def filter_cases_by_pipeline(cases: Query, pipeline: str, **kwargs) -> Query:
    """Return cases with the given pipeline."""
    return cases.filter(Family.data_analysis==pipeline))
```

cg/store/api/status.py
```diff
@@ -538,21 +538,35 @@ def _get_filtered_case_query(
         priority: str,
         sample_id: str,
     ) -> Query:
-        case_q = self.Family.query
+        case_q = self._get_query(table=Family)
+        filter_functions = []
```
Vince-janv, line 542:
```suggestion
        cases: Query = self._get_query(table=Family)
        filter_functions: List[Callable] = []
```

cg/store/filters/status_case_filters.py
```diff
@@ -173,6 +173,16 @@ def filter_cases_by_name_search(cases: Query, name_search: str, **kwargs) -> Que
...
+
+
+def filter_cases_by_priority(cases: Query, priority: int, **kwargs) -> Query:
```
Vince-janv, line 181:
```suggestion
def filter_cases_by_priority(cases: Query, priority: str, **kwargs) -> Query:
```

cg/store/filters/status_case_filters.py
```diff
@@ -193,6 +203,8 @@ def apply_case_filter(
     internal_id_search: Optional[str] = None,
     name_search: Optional[str] = None,
     case_search: Optional[str] = None,
+    pipeline_search: Optional[str] = None,
+    priority: Optional[int] = None,
```
Vince-janv, line 207:
```suggestion
    priority: Optional[str] = None,
```

Vince-janv (approved):
Well done!

cg/store/filters/status_case_filters.py
seallard: The original filtering in the cases function uses pattern matching:
```python
case_q = case_q.filter(Family.data_analysis.ilike(f"%{data_analysis}%"))
```

I'm guessing it because the pipeline is provided from the user through the CLI. But I agree, lets use strict matching instead (but this changes the behaviour of the function).

seallard: There are some tests which explicitly check that it is possible to filter cases by partial pipeline names:
```python
    # WHEN getting active cases by partial data_analysis
    cases = base_store.cases(data_analysis=str(data_analysis)[:-1])
```
These break when changing the filtering in the cases function to only include exact matches. I'll stick to the original behaviour for now.

PR #1924, opened 2023-04-14
Extract function for filtering not analysed cases associated with sample

seallard:
## Description
Replace function call to gigantic and super complex
```python
self.status_db.cases(sample_id=sample.internal_id, exclude_analysed=True)
```

With a more specialised function
```python
self.status_db.get_cases_not_analysed_by_sample_internal_id(
    sample_internal_id=sample.internal_id
)
```

I implemented et_cases_not_analysed_by_sample_internal_id based on the logic in the cases function when called like
```python
cases(sample_id=sample.internal_id, exclude_analysed=True)
```

### Added
- ```python
  CaseFilter.GET_NOT_ANALYSED
  ```
- ```python
  filter_cases_not_analysed(cases: Query, **kwargs) -> Query
    ```
- ```python
    get_not_analysed_cases_by_sample_internal_id(
        sample_internal_id: str,
    ) -> List[Family]
  ```
- ```python
  _get_join_case_and_sample_query() -> Query
  ```

### Changed
- Replaced call to cases() with get_not_analysed_cases_by_sample_internal_id()

### DIFF
--- a/cg/meta/transfer/external_data.py
+++ b/cg/meta/transfer/external_data.py
@@ -183,7 +183,9 @@ class ExternalDataAPI(MetaAPI):
         cases_to_start: List[dict] = []
         for sample in available_samples:
             cases_to_start.extend(
-                self.status_db.get_cases_not_analysed_by_sample_internal_id(sample.internal_id)
+                self.status_db.get_cases_not_analysed_by_sample_internal_id(
+                    sample_internal_id=sample.internal_id
+                )
             )
             last_version: Version = self.housekeeper_api.get_create_version(
                 bundle_name=sample.internal_id
--- a/cg/meta/transfer/external_data.py
+++ b/cg/meta/transfer/external_data.py
@@ -183,7 +183,7 @@ class ExternalDataAPI(MetaAPI):
         cases_to_start: List[dict] = []
         for sample in available_samples:
             cases_to_start.extend(
-                self.status_db.get_cases_not_analysed_by_sample_internal_id(
+                self.status_db.get_not_analysed_cases_by_sample_internal_id(
                     sample_internal_id=sample.internal_id
                 )
             )
--- a/cg/meta/transfer/external_data.py
+++ b/cg/meta/transfer/external_data.py
@@ -183,7 +183,7 @@ class ExternalDataAPI(MetaAPI):
         cases_to_start: List[dict] = []
         for sample in available_samples:
             cases_to_start.extend(
-                self.status_db.cases(sample_id=sample.internal_id, exclude_analysed=True)
+                self.status_db.get_cases_not_analysed_by_sample_internal_id(sample.internal_id)
             )
             last_version: Version = self.housekeeper_api.get_create_version(
                 bundle_name=sample.internal_id
--- a/cg/store/api/find_business_data.py
+++ b/cg/store/api/find_business_data.py
@@ -723,11 +723,11 @@ class FindBusinessDataHandler(BaseHandler):
             pipeline=pipeline,
         ).all()
 
-    def get_cases_not_analysed_by_sample_internal_id(
+    def get_not_analysed_cases_by_sample_internal_id(
         self,
         sample_internal_id: str,
     ) -> List[Family]:
-        """Get cases to analyse by sample internal id."""
+        """Get not analysed cases by sample internal id."""
 
         query: Query = self._get_join_case_and_sample_query()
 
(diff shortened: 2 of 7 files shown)


seallard:
The original function used (cases) filters based on the sample internal id like this:
```python
            case_q = case_q.filter(Sample.internal_id.ilike(f"%{sample_id}%"))
```

Which does not really make sense to me. Why would we use matching when we have the actual sample internal id retrieved from the sample? 

```python
self.status_db.cases(sample_id=sample.internal_id, exclude_analysed=True)
```

So currently the filtering is done by exact matching on the sample internal id instead. Does it make sense to you? @diitaz93 @Vince-janv

cg/store/api/find_business_data.py
```diff
@@ -713,3 +713,24 @@ def get_running_cases_in_pipeline(self, pipeline: Pipeline) -> List[Family]:
...
+            filter_functions=[
+                CaseFilter.GET_NOT_ANALYSED,
+            ],
+        )
+
+        return apply_sample_filter(
+            samples=not_analysed_cases,
+            filter_functions=[SampleFilter.FILTER_BY_INTERNAL_ID],
+            internal_id=sample_internal_id,
+        ).all()
```
diitaz93, line 736: Is this filtering by the internal id of the sample or the case? I think the CaseFilter function should be used instead of the Sample to avoid confusion. If the intended behaviour is to filter by sample I think this is not doing it. I think is taking the internal id of the cases. Better double check :)

tests/store/api/find/test_find_business_data.py
```diff
@@ -731,3 +733,45 @@ def test_get_case_by_name_and_customer_case_found(store_with_multiple_cases_and_
...
     assert filtered_case.customer_id == customer.id
     assert filtered_case.name == case.name
+
+
+def test_get_cases_not_analysed_by_sample_internal_id_empty_query(
+    store_with_multiple_cases_and_samples: Store,
+):
+    """Test that an empty query is returned if no cases match the sample internal id."""
+    # GIVEN a store with no cases matching the sample internal id
+    sample_internal_id = "nonexistent_sample_internal_id"
```
diitaz93, line 743: there is a fixture in `test/conftest.py` called `fixture_invalid_sample_id` that maybe you can use here

diitaz93 (commented):
Great work on the tests! I would like you to double check the filtering of the case query by a sample filtering function 🙂

cg/store/filters/status_case_filters.py
```diff
@@ -96,6 +96,14 @@ def get_cases_for_analysis(cases: Query, **kwargs) -> Query:
     )
 
 
+def filter_cases_not_analysed(cases: Query, **kwargs) -> Query:
+    """Filter cases that have not been analysed and are not currently being analysed."""
+    not_analyzed_condition = not_(Family.analyses.any(Analysis.completed_at.isnot(None)))
+    not_in_progress_condition = Family.action != CaseActions.ANALYZE
```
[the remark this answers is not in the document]
seallard: The types are sqlalchemy.sql.elements.BinaryExpression but I don't think that makes the code any more readable, so I won't add them

cg/store/api/find_business_data.py
seallard: The filtering should be done with the sample internal id and the function is called with sample.internal_id, so it should be ok!

cg/meta/transfer/external_data.py
```diff
@@ -183,7 +183,9 @@ def add_transfer_to_housekeeper(
         cases_to_start: List[dict] = []
         for sample in available_samples:
             cases_to_start.extend(
-                self.status_db.cases(sample_id=sample.internal_id, exclude_analysed=True)
+                self.status_db.get_cases_not_analysed_by_sample_internal_id(
```
henrikstranneheim, line 186:
```suggestion
                self.status_db.get_not_analysed cases_by_sample_internal_id(
```

cg/store/api/find_business_data.py
```diff
@@ -722,3 +722,24 @@ def get_running_cases_in_pipeline(self, pipeline: Pipeline) -> List[Family]:
             cases=self._get_query(table=Family),
             pipeline=pipeline,
         ).all()
+
+    def get_cases_not_analysed_by_sample_internal_id(
+        self,
+        sample_internal_id: str,
+    ) -> List[Family]:
+        """Get cases to analyse by sample internal id."""
```
henrikstranneheim, line 730:
```suggestion
        """Get not analysed cases by sample internal id."""
```

henrikstranneheim (approved):
Great work 👍

Issue #1855, opened 2023-03-27
Add Balsamic validation cases  


fevac:
## Description

Add validation cases for Balsamic WES:
- livedeer (TO, WES)
- Add (TN, WES)

## Suggested solution

- Add cases to: `cg/constants/compression.py`

## This can be closed when

*Describe what needs to be done for this issue to be closed*

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

henrikstranneheim:
@ivadym @fevac Is this done or planned?

fevac:
No, I can do it this week

PR #1925, opened 2023-04-14
Add validation cases to skip compression

fevac:
## Description

Fixes https://github.com/Clinical-Genomics/cg/issues/1855
Fixes https://github.com/Clinical-Genomics/cg/issues/1971

### Added

- Balsamic WES validation cases
- RNAfusion validation cases

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


fevac:
@ivadym @mathiasbio any other case we need to add?

ivadym:
In the issue you mention a TN-WES case, should I check if we have one available or you did it already?

fevac:
> In the issue you mention a TN-WES case, should I check if we have one available or you did it already?

I didn't find any

ivadym:
I found a few, but we need to know which samples of which customers  we can use:

https://docs.google.com/spreadsheets/d/1XeepO390huwMTPFnzfxZ0YUVwOqLUfh_5vrR0HQ0vmY/edit#gid=0

vwirta:
@ivadym I think the cust052 samples in your list are not exome samples but TGA using the GMS Myeloid panel.

TN cases:
Regarding samples from cust009, cust082, cust096 - I do not think we should include these samples in the validation data set as they are from different research projects.  The cust127 sample is however interesting, because it is a ped cancer pat and has been sequenced with high depth (like we would do for exome for a cancer case).

TO cases:
Also here, we should not include samples from cust082, cust091, cust096 (various research projects).
Also probably good to avoid are various internal customers: cust000, cust900
Good candidate customer numbers would be: cust002, cust087 (surprised they have exomes) and cust127

fevac:
Just took two cases considering all mentioned before. Adding  `properpigeon` as TN and `rightthrush` as TO

ivadym (approved):
⭐

mathiasbio (approved):
Nice! :) One step closer to being able to do smaller and faster releases! 🥳

Issue #1907, opened 2023-04-06
RNAfusion: store tower id in file


fevac:
## Description

Parses the tw launch output and saves the tower id in a file.

## Suggested solution

*Write here if there is a suggested way to solve the issue*

## This can be closed when

*Describe what needs to be done for this issue to be closed*

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

PR #1933, opened 2023-04-17
feat(rnafusion): Store tower ID when launching tower workflow

fevac:
## Description

Fixes https://github.com/Clinical-Genomics/cg/issues/1907

### Added

- Store Tower ID in a trailblazer config file

### Changed

-

### Fixed

-

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b store_tower_id -a
    ```

### How to test

- [x] `cg workflow rnafusion run --compute-env <COMPUTE_ENV> cuddlyhen`

### Expected test outcome

- [x] Check that a file called tower_ids.yaml is created.
image

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/workflow/rnafusion.py
```diff
@@ -149,10 +151,24 @@ def write_params_file(
             ),
         )
 
+    def get_trailblazer_config_path(self, case_id: str) -> Path:
+        """Return the path to a trailblazer config file containing Tower IDs."""
+        return Path(self.root_dir, case_id, "tower_ids.yaml")
```
Vince-janv, line 156: I think this should be in the trailblazer_api and called something like `get_nf_tower_config_path()`

cg/meta/workflow/tower_common.py
```diff
@@ -28,8 +31,16 @@ def get_tower_launch_parameters(cls, tower_pipeline: str, command_args: dict) ->
...
+        return ["launch"] + tower_options + [tower_pipeline]
+
+    @staticmethod
+    def get_tower_id(stdout_lines: Iterable) -> str:
+        """Parse the stdout and return a workflow id."""
+        for line in stdout_lines:
+            if line.strip().startswith("Workflow"):
+                return line.strip().split()[1]
+            else:
+                continue
```
Vince-janv, line 46: Is there not a simpler way to do this? Could you paste an example output?

Vince-janv (approved):
Nice ⭐

cg/meta/workflow/rnafusion.py
fevac: I like keeping it the same name for all analyses since it is then used in `add_pending_trailblazer_analysis` and it is still correct. This is the trailblazer config that is stored in the same field in the TB database regardless of it contains slurm or tower ids. As it is right now in cg it different workflows define this function in their APIs so seem reasonable to me to keep the same logic.  I would favour keeping it as it is.

cg/meta/workflow/tower_common.py
fevac: actually the else is not required! I added an example output in the docstring

cg/meta/workflow/rnafusion.py
fevac: resolving and leaving as it is

fevac:
- Deployed in stage

image

- Deployed in prod

image

PR #1943, opened 2023-04-17
Fix revisions

seallard:
## Description
Apparently, the Backup and Backuptape models were silently dropped from the database before, or were never created.
So it is not necessary to create revisions.

### Fixed
- Remove unnecessary revisions

## Review
- [ ] Tests executed by


henrikstranneheim (approved):
👍

PR #1948, opened 2023-04-18
Remove queries from tests

seallard:
## Description
Remove all instances of direct interactions with queries in the tests (exception of stats api queries which still need to be refactored).

### Fixed
- Remove queries from tests


henrikstranneheim (approved):
Great clean-up 👍

PR #2006, opened 2023-05-05
fix(RNA-upload - remove update argument)

karlnyr:
## Description

Extra argument was passed

### Fixed

- rna upload to scout command

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/cli/upload/scout.py
+++ b/cg/cli/upload/scout.py
@@ -171,7 +171,7 @@ def upload_case_to_scout(context: CGConfig, re_upload: bool, dry_run: bool, case
 @click.argument("case_id")
 @click.pass_context
 def upload_rna_to_scout(
-    context, case_id: str, dry_run: bool, update_fusion_report: bool, research: bool
+    context, case_id: str, dry_run: bool, research: bool
 ) -> None:
     """Upload an RNA case's gene fusion report and junction splice files for all samples connect via subject_id."""
 
@@ -182,8 +182,7 @@ def upload_rna_to_scout(
         upload_rna_fusion_report_to_scout,
         case_id=case_id,
         dry_run=dry_run,
-        research=research,
-        update=update_fusion_report,
+        research=research
     )
     context.invoke(upload_rna_junctions_to_scout, case_id=case_id, dry_run=dry_run)
 


peterpru:
Hi, tested and it failed, I am not sure why. It finds all the files, uploads to the two cases, but then at the very end it gives an error:
image

karlnyr:
Running the test again after the fix:

```Command
cg upload -f smoothtuna -r
....
Upload RNA coverage bigwig file finished!
smoothtuna - uploaded at set to 2023-05-09 10:27:12
smoothtuna analysis has been successfully uploaded
```

cg/cli/upload/scout.py
```diff
@@ -252,7 +239,7 @@ def upload_multiqc_to_scout(context: CGConfig, case_id: str, dry_run: bool) -> N
 def get_upload_api(case: Family, cg_config: CGConfig) -> UploadAPI:
     """Return the upload API based on the data analysis type"""
 
-    analysis_apis: Dict[Pipeline, UploadAPI] = {
```
Vince-janv, line 255: Do you not like the verbose hint?

cg/cli/upload/scout.py
```diff
@@ -162,28 +161,16 @@ def upload_case_to_scout(context: CGConfig, re_upload: bool, dry_run: bool, case
 @click.command(name="rna-to-scout")
 @click.option("--dry-run", is_flag=True)
 @click.option("-r", "--research", is_flag=True, help="Upload research report instead of clinical")
-@click.option(
-    "-u",
-    "--update-fusion-report",
-    is_flag=True,
-    help="re-upload existing fusion report",
-)
```
Vince-janv, line 170: Why is removing this the solution?

cg/meta/upload/mip/mip_rna.py
```diff
@@ -34,14 +35,11 @@ def upload(self, ctx: click.Context, case: Family, restart: bool) -> None:
...
-                LOG.info(
-                    f"Upload of case {case.internal_id} was successful. Setting uploaded at to {dt.datetime.now()}"
-                )
+            try:
+                ctx.invoke(upload_rna_to_scout, case_id=case.internal_id)
                 self.update_uploaded_at(analysis)
-            else:
-                raise RuntimeError(f"Upload to Scout failed for sample {case.internal_id}")
+            except CalledProcessError as error:
+                LOG.error(error)
```
Vince-janv, line 42: Much better 👌

Vince-janv (commented):
I don't really understand what error we were getting 🤔

cg/cli/upload/scout.py
karlnyr: The verbose hint was incorrect - it was in matter of fact not returning the metaclass but a long union of the different upload apis which I found more complex

cg/cli/upload/scout.py
karlnyr: With the new `scout load report` command, this update flag is no longer available

diitaz93 (approved):
Looks good! 👍  I guess the flag deletion in the cli command would make this a Major update?

karlnyr:
> Looks good! 👍 I guess the flag deletion in the cli command would make this a Major update?

I don't know - I feel like the API behaves the same we are just not passing the argument any longer

cg/cli/upload/scout.py
```diff
@@ -252,7 +239,7 @@ def upload_multiqc_to_scout(context: CGConfig, case_id: str, dry_run: bool) -> N
 def get_upload_api(case: Family, cg_config: CGConfig) -> UploadAPI:
     """Return the upload API based on the data analysis type"""
 
-    analysis_apis: Dict[Pipeline, UploadAPI] = {
+    analysis_apis: dict = {
```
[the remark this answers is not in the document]
karlnyr: See the comment I made for @Vince-janv But since you both agree do you still think this is a good idea? I agree that it was better before to the eye but it assumes that you also know the metaclass of the different upload-apis

diitaz93:
> > Looks good! 👍 I guess the flag deletion in the cli command would make this a Major update?
> 
> I don't know - I feel like the API behaves the same we are just not passing the argument any longer

I'm thinking that adding the flag will raise an error now, maybe that is uncompatible with the previous version

cg/cli/upload/scout.py
Vince-janv: So scout has a new function, which removed an option for us to use?

karlnyr: Yes! The new function replaces by default:
https://github.com/Clinical-Genomics/scout/blob/main/scout/load/report.py

karlnyr: the cli call defined in:
https://github.com/Clinical-Genomics/scout/blob/main/scout/commands/load/report.py#L85

karlnyr:
> I don't really understand what error we were getting 🤔

I tried to update the description of the PR - but the sum is that when we were running `cg upload -f <case>` we would also pass the update flag which was removed from the function in the earlier PR I made. Then came the error that I had also changed the return of the cli function to be none which crashed the upload function as it expected a 0.

PR #2023, opened 2023-05-11
Fix 401 errors due to invalid certificate

seallard:
## Description
This PR addresses the 401 errors due to an outdated google certificate. The error was associated with the JWT token decoding process in the before_request function. 

This decoding most likely fails due to a global cache for the requests library introduced here: https://github.com/Clinical-Genomics/servers/issues/30. The global cache likely causes the refresh of the google certificate to fail and can be removed since it was introduced due to an issue in the pybiomart library which is no longer used.

The caching mechanism might have cached the HTTP request used to fetch Google's OAuth2 certificates, leading to the use of an outdated certificate when decoding the JWT token. By removing requests_cache, the application will now always fetch the latest certificates.

Let's keep an eye on these crontabs in case the original issue resurfaces:
- transfer-lims-received
- transfer-lims-prepared
- transfer-lims-delivered

### Fixed
- Remove global cache `requests_cache` for the `requests` library
- Improve logging for errors when decoding tokens


Vince-janv (approved):
I hope this works! 🥳

seallard:
Deployed and tested in stage.

PR #544, opened 2020-02-10
Fix/werkzeug dependency

patrikgrenfeldt:
This PR adds a fix for the werkzeug 1.0.0 problem

**How to test**:
- [x] look at travis checks

**Expected test outcome**:
- [x] all should be green indicating werkzeug was used successfully
- [x] Take a screenshot and attach or copy/paste the output.

**Review:**
- [x] code approved by @mayabrandi 
- [x] tests executed by @patrikgrenfeldt 
- [x] "Merge and deploy" approved by @patrikgrenfeldt 
Thanks for filling in who performed the code review and the test!

This is patch **version bump** because it's a fix

PR #2026, opened 2023-05-11
Remove version lock for werkzeug

seallard:
## Description
This PR removes the version lock on the werkzeug library, letting pip resolve the compatible version instead (`1.0.1` instead of `0.16.1`). The version was locked in https://github.com/Clinical-Genomics/cg/pull/544, it is not really clear what the issue was and it seems fine when deployed and tested (by uploading an order form and placing the order) in the web stage environment.

### Fixed
- Remove werkzeug lock.


henrikstranneheim (approved):
👍. How do we best test it?

seallard:
I deployed it to the stage environment and tested placing an order with and order form (the only flow which explicitly uses a function from the library) and it worked. I guess we could run the end-to-end tests that I think the front end team have implemented

henrikstranneheim:
Nice!

PR #2031, opened 2023-05-12
Add module for calculating sequencing metrics

seallard:
## Description
This PR adds a module containing logic for calculating the metrics we need to populate the SequencingStatistics model.
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


ChrOertlin (commented):
We should keep this separate per bcl converter by e.g. using a sub folder structure with bcl2fastq and bclconvert

seallard:
I was hoping we could reuse these functions for both modules since they only accept raw values and the metrics should be calculated in the same fashion no matter which Demultiplexing software was used

ChrOertlin:
Agreed, I just hope the software output structure will allow us to do so. Might be too much of a hassle to streamline both to each other. For example the bcl convert already gives pre-calculated metrics

seallard:
Oh, that is nice! Then we can use those values. 

If there are any other values which need to be calculated, I think it would be good to place that logic in a shared module

PR #2055, opened 2023-05-24
Remove unused imports

seallard:
## Description
This PR removes all unused imports in the repository using
```
pip install ruff
ruff --fix . --select F401 --exclude __init__.py
```

Where F401 is the rule for `Module imported but unused`. Some manual fixes necessary due to hard coded imports of fixtures, see https://github.com/Clinical-Genomics/cg/issues/2056

### Fixed
- Remove all unused imports


seallard:
Deployed to stage and prod.

Issue #2063, opened 2023-05-24
Fixtures that return None and perform side effects


seallard:
## Description
Some fixtures are provided as parameters to tests, but they do not return anything and only perform side effects. I guess this is ok, but their names are very non descriptive and the side effects are not clear at all. For instance `mock_config`.

Here are the ones I've found which might be refactored or given better names:
```
mock_deliverable
mock_check_processing
mock_get_archived_encryption_key_path
mock_get_archived_flow_cell_path
mock_create_rta_complete
mock_get_archived_key
mock_analysis_finish
```

seallard:
A lot of work for nothing

PR #2064, opened 2023-05-24
Remove unused fixture parameters

seallard:
## Description
This PR removes most unused fixture parameters from the tests, found with `vulture`. There were plenty more, but those are fixtures which perform side effects and are required for the tests to pass https://github.com/Clinical-Genomics/cg/issues/2063.

### Fixed
- Remove unused fixture parameters


tests/cli/workflow/balsamic/test_store_housekeeper.py
```diff
@@ -66,7 +66,7 @@ def test_without_config(cli_runner: CliRunner, balsamic_context: CGConfig, caplo
 
 
 def test_case_without_deliverables_file(
-    cli_runner: CliRunner, balsamic_context: CGConfig, mock_config: dict, caplog
+    cli_runner: CliRunner, balsamic_context: CGConfig, mock_config, caplog
```
henrikstranneheim, line 69: Was the type hint wrong or could we also remove the parameter?

henrikstranneheim (approved):
Nice to get rid of these fixture parameters 👍

tests/cli/workflow/balsamic/test_store_housekeeper.py
seallard: Yes, the type hint was wrong. The fixture does not return anything and does some setup, the test fails without it. I created an issue for it: https://github.com/Clinical-Genomics/cg/issues/2063

diitaz93 (approved):
Another great initiative 👍

PR #2075, opened 2023-05-26
Remove unused functions and classes

seallard:
## Description
This PR removes most functions and classes flagged by Vulture as unused. I avoided removing recently created functions since they might be used in some branches in development (but I might have caught some by accident... The resulting merge conflict should be easy to resolve). Some functions were only used in tests and were replaced when easily done.

### Fixed
- Remove most unused functions and classes


henrikstranneheim (approved):
Awesome 🤩! Feels a bit scary slicing across this much code, but I think we should do this and then reinstate any code that might have been needed (if any). Great initiative 👍

seallard:
Deployed to stage

PR #2083, opened 2023-05-30
Fix parsing of bcl2fastq metrics per tile

seallard:
## Description
This PR fixes the logic for populating the `SampleLaneSequencingMetrics` model, aggregating the tile statistics to per sample and lane instead.

### Fixed
- Aggregate `bcl2fastq` demultiplex tile metrics to be per sample and lane


seallard:
Deployed to stage

seallard:
Deployed to prod.

PR #2107, opened 2023-06-07
Populate new sequencing statistics table

seallard:
## Description
This PR adds function calls to start populating the sequencing metrics table in status db in the demultiplex post processing module. 

Since the `DemuxPostProcessingApi` contains methods like `finish_all_flow_cells`, I decided to remove the flow cell name as an attribute from the class. It does not make sense to keep it since the class contains logic for finishing all flow cells and not just a specific one. The methods need to accept the flow cell name as an argument since we iterate over all flow cells.

### Added
- Population of new sequencing metrics table to the existing `cg demultiplex finish` commands

### Fixed
- Remove flow cell name and directory as attributes from the `DemuxPostProcessingApi`


henrikstranneheim (approved):
Nice 👍

seallard:
Closed this PR since data can be ported with the existing script once the new command is completed. The only reason to do it now is to get some information on whether there are any issues with the parsers we have implemented.

PR #2108, opened 2023-06-07
Remove flow cell name and directory as attributes

seallard:
## Description
Since the DemuxPostProcessingApi contains methods like `finish_all_flow_cells`, it does not really make sense to store the flow cell name and directory as attributes. The class does not only process a single flow cell directory, it also contains logic for finishing all flow cells.

### Fixed
- Remove flow cell name and directory as attributes from the DemuxPostProcessingApi 

### DIFF
--- a/cg/meta/demultiplex/demux_post_processing.py
+++ b/cg/meta/demultiplex/demux_post_processing.py
@@ -34,7 +34,7 @@ LOG = logging.getLogger(__name__)
 class DemuxPostProcessingAPI:
     """Post demultiplexing API class."""
 
-    def __init__(self, config: CGConfig, flow_cell_name: Optional[str] = None) -> None:
+    def __init__(self, config: CGConfig) -> None:
         self.stats_api: StatsAPI = config.cg_stats_api
         self.demux_api: DemultiplexingAPI = config.demultiplex_api
         self.status_db: Store = config.status_db
@@ -43,10 +43,6 @@ class DemuxPostProcessingAPI:
             db=self.status_db, stats_api=self.stats_api, hk_api=self.hk_api
         )
         self.dry_run = False
-        self.flow_cell_name: Optional[str] = flow_cell_name
-        self.flow_cell_dir: Optional[Path] = (
-            Path(self.demux_api.out_dir, self.flow_cell_name) if self.flow_cell_name else None
-        )
 
     def set_dry_run(self, dry_run: bool) -> None:
         """Set dry run."""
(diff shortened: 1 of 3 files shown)


islean (approved):
Nice!

tests/meta/demultiplex/test_demux_post_processing.py
```diff
@@ -509,7 +511,9 @@ def test_is_not_bcl2fastq_folder_structure(
     demux_post_processing_api.flow_cell_dir = not_bcl2fastq_folder_structure
 
     # WHEN checking if it is a bcl2fastq folder structure
-    is_bcl2fastq_folder_structure = demux_post_processing_api.is_bcl2fastq_demux_folder_structure()
+    is_bcl2fastq_folder_structure = demux_post_processing_api.is_bcl2fastq_demux_folder_structure(
+        flow_cell_name=""
+    )
```
islean, line 516: Not too fond of sending flow_cell_name="" but I don't know if the code actually improves by setting it as an optional in the input of is_bcl2fastq_demux_folder_structure.

seallard:
Deployed to stage and tested with `cg demultiplex finish temporary /home/proj/production/demultiplexed-runs/170407_ST-E00198_0209_BHHKVCALXX`

tests/meta/demultiplex/test_demux_post_processing.py
seallard: Me neither, but the argument is not optional - the directory fixtures were created with the flow cell name set as an empty string

seallard: Letting the tests be for now.

seallard:
Deployed to prod.

seallard:
Deployed to prod.

ChrOertlin (commented):
The  class is generated per flow cell so I am not entirely sure why this would be an issue. We could also just move the flow cell finish all method out of here?

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -70,53 +66,42 @@ def transfer_flow_cell(
...
-        """Validate flow cell."""
-        LOG.info(f"Check demultiplexed flow cell {self.flow_cell_name}")
-        try:
-            flow_cell: FlowCell = FlowCell(
-                flow_cell_path=self.flow_cell_dir, bcl_converter=self.get_bcl_converter()
-            )
-            return flow_cell
-        except FlowCellError:
-            return None
-
```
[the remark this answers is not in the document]
seallard: Lets add it when it is needed then!

PR #2111, opened 2023-06-08
Outline implementation of new demultiplex finish functionality

seallard:
## Description
This PR ports part of the required functionality to the temporary `cg demultiplex finish` command, which will replace the existing finish flow cell functionality, and outlines the remaining steps to be taken.

If you feel like you can give me any feedback on whether steps are missing below for the post processing of a flow cell, it is super appreciated!

```
        1. Validate that the flow cell directory exists.
        2. Validate that the demultiplexing data transfer is complete.
        3. Create flow cell.
        4. Store flow cell in status db.
        5. Store flow cell data in housekeeper.
        6. Create sequencing metrics.
```

I'm most uncertain about:

- Is it necessary to check whether delivery already has been started and does this differ for the different sequencing platforms? (Why is there only a specific `is_hiseq_x_delivery_started` function?)
- Where can the sequencer type and name be found? ~Currently that data is retrieved from `cgstats` which will be removed.~

### Added
- Outline for porting functionality to the new `cg demultixplex finish` command 

### How to test
The PR only adds incomplete stubs to the temporary command (`cg demultiplex finish temporary`) which will fail. The intention is to get some feedback on the post processing steps after demultiplexing and outline an implementation plan.


cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -82,7 +82,43 @@ def add_sample_lane_sequencing_metrics_for_flow_cell(self, flow_cell_name: str):
...
+        try:
+            flow_cell: FlowCell = FlowCell(
+                flow_cell_path=flow_cell_dir, bcl_converter=bcl_converter
+            )
+        except FlowCellError:
+            return None
+
+        # 4. Store flow cell in status db.
+        self.status_db.session.add(flow_cell)
+        self.status_db.session.commit()
```
ChrOertlin, line 117: I think these are all the steps which are needed indeed. Would be nice to separate them into smaller functions to keep them more testable and simple.

ChrOertlin:
_Where can the sequencer type and name be found? Currently that data is retrieved from cgstats which will be removed._
This should be in the runparameter files, right?

cg/meta/demultiplex/demux_post_processing.py
seallard: Yes, that sounds like a good idea :)

ChrOertlin:
_Is it necessary to check whether delivery already has been started and does this differ for the different sequencing platforms? (Why is there only a specific is_hiseq_x_delivery_started function?)_

I think we discussed this in the meeting and it was not necessary.

diitaz93:
> _Where can the sequencer type and name be found? Currently that data is retrieved from cgstats which will be removed._ This should be in the runparameter files, right?

It is actually an attribute of the FlowCell object and is retrieved by the name of the flow cell. it is called `sequencer_type`

cg/meta/demultiplex/demux_post_processing.py
seallard: I'll keep this outline for now and refactor each step in separate PR:s.

diitaz93 (approved):
looks good 👍

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -82,7 +82,43 @@ def add_sample_lane_sequencing_metrics_for_flow_cell(self, flow_cell_name: str):
...
+        # 2. Validate that the demultiplexing data transfer is complete.
+        if not Path(flow_cell_dir, DemultiplexingDirsAndFiles.DEMUX_COMPLETE).exists():
+            return
+
+        # 3. Create flow cell. TODO: Populate all fields here.
+        try:
+            flow_cell: FlowCell = FlowCell(
+                flow_cell_path=flow_cell_dir, bcl_converter=bcl_converter
+            )
+        except FlowCellError:
```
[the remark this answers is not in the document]
seallard: Added some log messages 👍

PR #2119, opened 2023-06-12
Transfer flow cell data to housekeeper

seallard:
## Description
This PR adds logic for transferring flow cell data to housekeeper in the `cg demultiplex finish` command. Specifically, it does the following:
- Create a bundle and version for the flow cell
- Add tag for flow cell
- Add sample sheet to bundle and tag with flow cell name
- Add fastq files and tag them with sample id

Much of the added functionality already exists in `cg/meta/transfer/flowcell.py`, but it is tangled up with cg stats so I extracted the required functionality to this module. When cgstats is removed and the new `finish`command replaces the old one, the functions can replace the existing ones in the separate `cg/meta/transfer/flowcell.py` module if we want that.

### Added
- Transfer of flow cell data to housekeeper in the new `demultiplex finish flowcell` command

### How to prepare for test
   
```
bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b transfer-flow-cell-data-to-housekeeper -a
```

### How to test
```
cg demultiplex finish temporary 221019_A00689_0663_BH7FYHDRX2
```

- [x] **MINOR**


seallard:
Deployed to stage and tested.

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -126,10 +124,116 @@ def finish_flow_cell_temp(self, flow_cell_name: str) -> None:
...
+        self.add_file_if_not_exists(
+            flow_cell_directory / DemultiplexingDirsAndFiles.SAMPLE_SHEET_FILE_NAME,
+            flow_cell_name,
+            [SequencingFileTag.SAMPLE_SHEET, flow_cell_name],
+        )
+
+    def is_valid_sample_fastq_filename(self, filename: str) -> bool:
+        """Validate the file name and discard any undetermined fastq files."""
+        parts = filename.split("_")
+        return len(parts) > 1 and "Undetermined" not in filename
```
seallard, line 177: This validation is hard to understand, can you make it more explicit and easier to follow? Does it really cover all criteria for a valid sample fastq file name?

islean, line 177: Is it always the case that we have an underscore in valid file names?

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -126,10 +124,116 @@ def finish_flow_cell_temp(self, flow_cell_name: str) -> None:
...
+        ]
+        return valid_sample_fastq_file_paths
+
+    def get_sample_id_from_sample_fastq_file_path(self, fastq_file_path: Path) -> Optional[str]:
+        """Extract sample id from fastq file name."""
+        sample_directory: str = fastq_file_path.parent.name
+        directory_parts: str = sample_directory.split("_")
+
+        if len(directory_parts) > 1:
+            sample_internal_id = directory_parts[1]
```
seallard, line 194: Is the sample id always the second element in the fastq files parent directory name?

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -126,10 +124,116 @@ def finish_flow_cell_temp(self, flow_cell_name: str) -> None:
...
+        else:
+            sample_internal_id = directory_parts[0]
+
```
islean, line 197: What is the structure of the directory here? Does len(directory_parts)>1 ensure that we have exactly two directory parts and the latter is the sample_internal_id?

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -126,10 +124,116 @@ def finish_flow_cell_temp(self, flow_cell_name: str) -> None:
...
 
+    def add_flow_cell_data_to_housekeeper(
+        self, flow_cell: Flowcell, flow_cell_directory: Path
+    ) -> None:
+        """Add flow cell data to housekeeper."""
+        LOG.info(f"Add flow cell data to housekeeper for {flow_cell.name}")
+
+        self.add_bundle_and_version_if_not_exists(flow_cell_name=flow_cell.name)
+
+        tags: List[str] = [SequencingFileTag.FASTQ, SequencingFileTag.SAMPLE_SHEET, flow_cell.name]
```
seallard, line 142: It seems unnecessary to always try to add these constant tags (FASTQ and SAMPLE_SHEET).

henrikstranneheim: I guess the only time this would fail for us is if we start with a completely empty database - if they were to be removed. Having them should make it fool-proof and I guess this check does not add much cognitive load nor time running the process. I would leave it as it is.

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -126,10 +124,116 @@ def finish_flow_cell_temp(self, flow_cell_name: str) -> None:
...
+        fastq_file_paths: List[Path] = self.get_sample_fastq_file_paths(
+            flow_cell_directory=flow_cell_directory
+        )
+
+        for fastq_file_path in fastq_file_paths:
+            sample_id: str = self.get_sample_id_from_sample_fastq_file_path(
+                fastq_file_path=fastq_file_path
+            )
+
+            if sample_id:
```
seallard, line 159: Consider adding log warnings for these unexpected cases

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -126,10 +124,116 @@ def finish_flow_cell_temp(self, flow_cell_name: str) -> None:
...
+    def get_sample_fastq_file_paths(self, flow_cell_directory: Path) -> List[Path]:
+        """Get fastq file paths for flow cell."""
+        valid_sample_fastq_file_paths = [
+            file_path
+            for file_path in flow_cell_directory.glob("**/*.fastq.gz")
+            if self.is_valid_sample_fastq_filename(file_path.name)
+        ]
+        return valid_sample_fastq_file_paths
+
+    def get_sample_id_from_sample_fastq_file_path(self, fastq_file_path: Path) -> Optional[str]:
```
seallard, line 188: The return type should just be `str`

henrikstranneheim (approved):
Looks good 👍. Would be great with unit tests

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -126,10 +124,116 @@ def finish_flow_cell_temp(self, flow_cell_name: str) -> None:
...
+    def is_valid_sample_fastq_filename(self, filename: str) -> bool:
+        """Validate the file name and discard any undetermined fastq files."""
+        parts = filename.split("_")
+        return len(parts) > 1 and "Undetermined" not in filename
+
+    def get_sample_fastq_file_paths(self, flow_cell_directory: Path) -> List[Path]:
+        """Get fastq file paths for flow cell."""
+        valid_sample_fastq_file_paths = [
+            file_path
+            for file_path in flow_cell_directory.glob("**/*.fastq.gz")
```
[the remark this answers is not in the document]
seallard: Yes 👍

cg/meta/demultiplex/demux_post_processing.py
seallard: I don't see why the check is necessary though, I'll add it again if turns out to be necessary and make it more apparent why. I think we want all `fastq`files except for the `Undetermined` for now

cg/meta/demultiplex/demux_post_processing.py
seallard: Yes.

cg/meta/demultiplex/demux_post_processing.py
seallard: I think there needs to be some additional validation of the sample id that is extracted. I wonder if it always should be of a specific length etc.

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -17,9 +15,11 @@
 from cg.apps.demultiplex.demux_report import create_demux_report
 from cg.apps.housekeeper.hk import HousekeeperAPI
 from cg.constants.cgstats import STATS_HEADER
-from cg.constants.constants import FlowCellStatus
+from cg.constants.constants import FileExtensions
 from cg.constants.demultiplexing import BclConverter, DemultiplexingDirsAndFiles
+from cg.constants.housekeeper_tags import SequencingFileTag
 from cg.exc import FlowCellError
+from housekeeper.store.models import Version
```
diitaz93, line 22: An unnecessary change, I think is a good practice to keep imports in alphabetical order 😃

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -126,10 +126,119 @@ def finish_flow_cell_temp(self, flow_cell_name: str) -> None:
...
+            sample_internal_id = directory_parts[0]
+
+        return sample_internal_id
+
+    def add_bundle_and_version_if_not_exists(self, flow_cell_name: str) -> None:
+        """Add bundle if it does not exist."""
+        if not self.hk_api.bundle(name=flow_cell_name):
+            self.hk_api.create_new_bundle_and_version(name=flow_cell_name)
+
+    def ensure_tags_exist(self, tag_names: List[str]) -> None:
```
diitaz93, line 210: I would call this function `add_tags_if_non_existent`, as `ensure` sounds like it would fail if they don't exist

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -126,10 +126,119 @@ def finish_flow_cell_temp(self, flow_cell_name: str) -> None:
...
+        directory_parts: str = sample_directory.split("_")
+
+        if len(directory_parts) > 1:
+            sample_internal_id = directory_parts[1]
+        else:
+            sample_internal_id = directory_parts[0]
+
+        return sample_internal_id
+
+    def add_bundle_and_version_if_not_exists(self, flow_cell_name: str) -> None:
```
diitaz93, line 205: sounds better 😃 
```suggestion
    def add_bundle_and_version_if_non_existent(self, flow_cell_name: str) -> None:
```

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -126,10 +126,119 @@ def finish_flow_cell_temp(self, flow_cell_name: str) -> None:
...
+        if not self.hk_api.bundle(name=flow_cell_name):
+            self.hk_api.create_new_bundle_and_version(name=flow_cell_name)
+
+    def ensure_tags_exist(self, tag_names: List[str]) -> None:
+        """Ensure that tags exist in Housekeeper."""
+        for tag_name in tag_names:
+            if self.hk_api.get_tag(name=tag_name) is None:
+                self.hk_api.add_tag(name=tag_name)
+
+    def add_file_if_not_exists(
```
diitaz93, line 216:
```suggestion
    def add_file_if_non_existent(
```

diitaz93 (approved):
Great work 💯 I added some personal opinions about the names of the functions

cg/meta/demultiplex/demux_post_processing.py
diitaz93: I think it is good as it is too

seallard:
Deployed to prod.

PR #2123, opened 2023-06-13
Change family parameter to case in upload cmd

islean:
## Description

A parameter in the cg upload command is currently named "family". This PR changes the parameter to be named "case" instead.

### Changed

- Family parameter is now called case.
- Updated documentation to reflect this change.
- Removed test which was skipped every run.
- Changed parameter name case_obj -> case in UploadAPI/upload to match the instantiations.

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b family-flag-to-case-flag -a
    ```

### How to test

- [x] Upload sample case with case_id set as family.
- [x] Upload sample case with case_id set as case.

### Expected test outcome

- [x] Trying to set family fails with an unrecognized parameter error.
- [x] Trying to set case works and results in a successful upload.

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/cli/upload/base.py
```diff
@@ -39,23 +38,22 @@
...
+@click.option("-c", "--case", "case_id", help="Upload to all apps")
 @click.option(
     "-r",
     "--restart",
     is_flag=True,
     help="Force upload of an analysis that has already been uploaded or marked as started",
 )
 @click.pass_context
-def upload(context: click.Context, family_id: Optional[str], restart: bool):
+def upload(context: click.Context, case_id: Optional[str], restart: bool):
```
ChrOertlin, line 56: ```case_id``` is in this case ```case_internal_id``` which refers to the case name e.g. "updog" 

there was a discussion some time ago how we wanted to refer to such ids and the outcome was to be more explicit at least in the backend. I am not sure what we want in the cli with which the users interact. 

@henrikstranneheim, @Vince-janv, @seallard and @diitaz93 what are your thoughts?

documentation/README.md
```diff
@@ -165,7 +165,7 @@ The command will update the _total_ read counts of each sample and check against
 Much like the group of analysis subcommands you can perform an upload of analysis results (stored in Housekeeper) all at once by running:
 
 ```bash
-cg upload --family raredragon
+cg upload --case raredragon
```
ChrOertlin, line 168:
```suggestion
cg upload --case raredragon
```
ah indeed so it would be case in the cli command for the user.

ChrOertlin (approved):
It looks good so far, just a comment in the case_id naming. 

Also I think the upload must be tied into some cron jobs that run uploads to scout regularly. You need to check if this upload command is in there and likely adjust those as well.

ChrOertlin:
> @ChrOertlin Good catch with the cron jobs. From what I can gather, the three scripts with "upload" in the name all use "cg upload auto --pipeline " wherein you don't specify a specific case. Would be great if anyone could confirm that only
> 
> ```
> # upload results for families that are analyzed [every 3rd hour]
> 11 */3 * * * $CRONIC ${PRODUCTION_HOME}/servers/resources/hasta.scilifelab.se/crontabs/upload-auto-mip-dna.sh >> ${LOG_BASE}/cg.upload-mip-dna.log 2> >(tee -a ${LOG_BASE}/cg.upload-mip-dna.log >&2)
> 33 */3 * * * $CRONIC ${PRODUCTION_HOME}/servers/resources/hasta.scilifelab.se/crontabs/upload-auto-balsamic.sh >> ${LOG_BASE}/cg.upload-balsamic.log 2> >(tee -a ${LOG_BASE}/cg.upload-balsamic.log >&2)
> 53 */3 * * * $CRONIC ${PRODUCTION_HOME}/servers/resources/hasta.scilifelab.se/crontabs/upload-auto-balsamic-umi.sh >> ${LOG_BASE}/cg.upload-balsamic-umi.log 2> >(tee -a ${LOG_BASE}/cg.upload-balsamic-umi.log >&2)
> ```
> 
> are the relevant cron jobs to check.

@henrikstranneheim  I think you know more about this than I do.

cg/cli/upload/base.py
seallard: Since the field on the model is called `internal_id` I think we should refer to the variable as `case_internal_id`. I don't really think `internal_id` is a good name, but that is a separate issue.

However, I don't think the user of the CLI should know about these hair cleaving exercises and for them it would make much more sense to just refer to it as the case id.

documentation/README.md
islean: ChatGPT suggests that only "-c" and "--case" will actually set this parameter, while "case_id" would become the variable name. Hence maybe "-c, --case, case_internal_id" would be the most attractive option to go for?

henrikstranneheim:
@ChrOertlin @islean I am not aware of any automation using the case option for uploading. 👍

cg/cli/upload/base.py
diitaz93: I think we should make as few changes as possible to the CLI. I would say keep `case_id` and leave the `internal_id`s inside infrastructure

tests/store/api/test_base.py
```diff
@@ -1,20 +1,7 @@
...
-from sqlalchemy.orm import Query
 from cg.constants.subject import PhenotypeStatus
-from cg.store.api.base import BaseHandler
-from cg.store.models import Model
-
-
-@pytest.mark.parametrize("table", astuple(BaseHandler(MagicMock())))
-def test__get_query(base_store, table: Type[Model]):
-    """Tests the _get_query function for all attributes of BaseHandler ie tables in the database."""
-    assert isinstance(base_store._get_query(table=table), Query)
```
diitaz93, line 17: so happy for this 🙌

diitaz93 (approved):
Looks good 👍

cg/cli/upload/base.py
islean: @diitaz93 The CLI used family_id before though, but maybe you meant that starting with case_internal_id would imply changes to the CLI in the future to match all other usages of case_id?

ChrOertlin:
```suggestion
def upload(context: click.Context, case_internal_id: Optional[str], restart: bool)
```
so we settle for this then? since the parameter has to be changed from ```familiy_id``` to something with ```case```anyway we can be explicit and make it ```case_internal_id```?

diitaz93: yes! Sorry for not being clear. The change of family to case has to be done. But changing "id" to "internal_id" would bring unnecessary confusion. I would prefer just `_id` just in this case as it is far away from the SQLAlchemy models, but I would understand why @ChrOertlin wants to keep the ìnternal`, it is "more correct" but less practical I think.

islean: To me, it would be more confusing to have case_id in most places and case_internal_id in just one, although I agree that case_internal_id is probably better in the long-run. So my suggestion is that we use case_id for now and then open a separate pr where we do a larger overhaul of the naming. @ChrOertlin @diitaz93.

Vince-janv:
Don't forget [this line](https://github.com/Clinical-Genomics/cg/blob/c8490ec543b7aeda4700ec6291e33bd66eb439bc/cg/cli/upload/base.py#L114)!

Vince-janv:
Also, this needs to be communicated to prod-bioinfo and some atlas documents might need revising

islean:
Test was successful:

θ67° [isak.ohlsson@hasta:~] [S_base] 3m31s 128 $ cg upload --help
Usage: cg upload [OPTIONS] COMMAND [ARGS]...

  Upload results from analyses

Options:
  -c, --case TEXT  Upload to all apps
  -r, --restart    Force upload of an analysis that has already been uploaded
                   or marked as started
  --help           Show this message and exit.

Commands:
  all-fastq                   Starts upload of all not previously uploaded...
  auto                        Upload all completed analyses
  available-observations      Uploads the available observations to...
  clinical-delivery           Links the appropriate files for a case, based...
  coverage                    Upload coverage from an analysis to Chanjo.
  create-scout-load-config    Create a load config for a case in scout and...
  delivery-report-to-scout    Fetches a delivery report from housekeeper...
  fohm
  genotypes                   Upload genotypes from an analysis to...
  gens                        Upload data from an analysis to Gens.
  gisaid                      Upload mutant analysis data to GISAID.
  multiqc-to-scout            Upload multiqc report to Scout.
  nipt                        Upload NIPT result files
  observations                Upload observations from an analysis to...
  process-solved              Process cases with mutacc that has been...
  processed-solved            Upload solved cases that has been processed...
  rna-fusion-report-to-scout  Upload fusion report file for a case to...
  rna-junctions-to-scout      Upload RNA junctions splice files to Scout.
  rna-to-scout                Upload an RNA case's gene fusion report and...
  scout                       Upload variants from analysis to Scout.
  upload-case-to-scout        Upload variants and case from analysis to...
  validate                    Validate a family of samples.
  
θ68° [isak.ohlsson@hasta:~] [S_base] 22s $ cg upload -f nativepug
Error: no such option: -f

θ64° [isak.ohlsson@hasta:~] [S_base] 23s 2 $ cg upload -c nativepug
----------------- UPLOAD -----------------
Case nativepug exists in Status DB
There is no analysis for case: nativepug

PR #2133, opened 2023-06-14
fix(rnafusion): mapped reads percentage calculation (patch)

fevac:
### Fixed

- Mapped reads percentage was incorrect in the delivery report as it was dividing aligned reads (in read pairs) by total reads (in number of reads). Solved by multiplying aligned reads x 2.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b fix_mapped_reads_perc -a
    ```

### How to test

- [x] cg generate delivery-report <case_id> --dry-run
- [x] cg upload delivery-report-to-scout  <case_id>. Make sure that `Mappade sekvenser` has the right value.

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


fevac:
ready for review

cg/meta/report/rnafusion.py
```diff
@@ -55,7 +47,9 @@ def get_sample_metadata(
             input_amount=self.lims_api.get_latest_rna_input_amount(sample_id=sample.internal_id),
             insert_size=None,
             insert_size_peak=None,
-            mapped_reads=sample_metrics.reads_aligned / sample_metrics.before_filtering_total_reads,
+            mapped_reads=sample_metrics.reads_aligned
```
henrikstranneheim, line 50: Should reads_aligned be renamed to reflect that it is read pairs and not single reads - to avoid future confusion.

henrikstranneheim (approved):
Nice catch 👍

cg/meta/report/rnafusion.py
fevac: these are standard names given by multiqc. So I don't think its a great idea to change them. It is nice that the metrics collected by multiqc always get assigned the same names regardless of pipeline, etc. I do agree that several of these names are confusing. But maybe something to think about in the future? I don't think I would do it now

henrikstranneheim: Ok, let's leave it for now 👍. If you want to you could make an issue on multiqc to clarify the variable name.

fevac:
tests passing

fevac:
- Deployed in stage:
image

- Deployed in prod:
image

PR #2135, opened 2023-06-14
Add index to sample id on sequencing metrics

seallard:
## Description
This PR adds an index to the `internal_sample_id` column on the sequencing metrics table. The reason is that queries will be done on the table to extract the total number of reads for a sample, which will be slow without an index.

### Added
- Index on `sample_internal_id` column on the `SampleLaneSequencingMetrics` table.


henrikstranneheim (approved):
👍

alembic/versions/879e6bc89750_add_index_to_sequencing_metrics_sample_.py
```diff
@@ -0,0 +1,29 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_index(
+        "idx_sample_internal_id",
+        "sample_lane_sequencing_metrics",
+        ["sample_internal_id"],
+        unique=False,
```
Vince-janv, line 24: Do you know what `unique` means here?

Vince-janv (approved):
Great idea ⭐  I think there might be some other columns in cg that should be indexed but we can figure that out later.

alembic/versions/879e6bc89750_add_index_to_sequencing_metrics_sample_.py
seallard: I think it means that there can be multiple entries with the same `sample_internal_id` in the index. But the default is already false, so it should not be necessary to specify.

Vince-janv: But do we not want `unique` to be true? 🤔

seallard: I don't think so since the `SampleLaneSequencingMetrics` are calculated per sample, flow cell and lane. So there should be many entries with the same `sample_internal_ind`! Maybe I'm confused? 😖

Vince-janv: Sorry, misread the table we were adding this on! You are of course correct 👍

seallard:
Deployed to stage

PR #2137, opened 2023-06-16
Remove revision for already existing index on sample_internal_id

seallard:
This reverts commit 9cb0ddc3e41fb5859d922b6bf8dc4d889c328ec8.

## Description
This PR removes the commit which introduced an index on the `sample_internal_id`. I missed that this column is marked as a foreign key which means it already has an index. So the revision is useless and the downgrade method does not work.

So this PR removes the revision and I will have to manually edit the id of the latest applied revision in the database since the down method is broken.

### Fixed
- Revert introduction of index on already indexed column


henrikstranneheim (approved):
👍

diitaz93 (approved):
👍

seallard:
Deployed to stage and revisions applied successfully.

seallard:
Deployed to prod and revisions applied successfully.

PR #2148, opened 2023-06-19
Fix invalid flow cell name when transferring sequencing data to housekeeper

seallard:
## Description
This PR passes the correct flow cell name to the function which transfers the flow cell data to housekeeper.
The current value passed is the entire flow cell run directory name instead of just the flow cell name.

### Fixed
- Pass correct flow cell name when transferring sequencing data to housekeeper


diitaz93 (approved):
👍

PR #2152, opened 2023-06-20
fix(rnafusion): change accreditation

fevac:
### Fixed

- Rnafusion accreditation is under review. Delivery report should be non-accredited. 

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b rnafusion_remove_accreditation -a
    ```

### How to test

- [x]  cg generate delivery-report <case_id> 
- [x] cg upload delivery-report-to-scout <case_id>. Make sure that there is no logo in the delivery report. 

### Expected test outcome

- [ ] Check that ...

## Review

- [x] Tests executed by EFC

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/report/rnafusion.py
+++ b/cg/meta/report/rnafusion.py
@@ -71,7 +71,7 @@ class RnafusionReportAPI(ReportAPI):
     def get_report_accreditation(
         self, samples: List[SampleModel], analysis_metadata: AnalysisModel
     ) -> bool:
-        """Checks if the report is accredited or not. Rnafusion is an accredited workflow."""
+        """Checks if the report is accredited or not. Rnafusion is not an accredited workflow."""
         return False
 
     def get_scout_uploaded_file_from_hk(self, case_id: str, scout_tag: str) -> Optional[str]:
--- a/cg/meta/report/rnafusion.py
+++ b/cg/meta/report/rnafusion.py
@@ -72,7 +72,7 @@ class RnafusionReportAPI(ReportAPI):
         self, samples: List[SampleModel], analysis_metadata: AnalysisModel
     ) -> bool:
         """Checks if the report is accredited or not. Rnafusion is an accredited workflow."""
-        return True
+        return False
 
     def get_scout_uploaded_file_from_hk(self, case_id: str, scout_tag: str) -> Optional[str]:
         """Return the file path of the uploaded to Scout file given its tag."""


fevac:
tests are passing

cg/meta/report/rnafusion.py
```diff
@@ -71,8 +71,8 @@ def get_genome_build(self, analysis_metadata: AnalysisModel) -> str:
     def get_report_accreditation(
```
ChrOertlin, line 71:
```suggestion
    def is_accredited(
```
The current function name indicates an object is returned, although it is just a bool that is returned.

cg/meta/report/rnafusion.py
```diff
@@ -71,8 +71,8 @@ def get_genome_build(self, analysis_metadata: AnalysisModel) -> str:
...
         self, samples: List[SampleModel], analysis_metadata: AnalysisModel
     ) -> bool:
-        """Checks if the report is accredited or not. Rnafusion is an accredited workflow."""
-        return True
+        """Checks if the report is accredited or not. Rnafusion is not an accredited workflow."""
+        return False
```
ChrOertlin, line 75: This function seems to just return false no matter what. It does not seem like a check to me. 

Would it be possible to have an ```is_accredited``` value somewhere in the I suppose RNAfusion ```AnalysisModel ```?

ChrOertlin (approved):
Looks good, some things to consider!

cg/meta/report/rnafusion.py
fevac: So, for other workflows accreditation depends on other factors and not only in the workflow. For example, in balsamic some panels and analysis type are accredited and others not. I don't feel confident changing the logic only in this pipeline. However, it could be good to revisit this part of the code and refactor to make it clearer but for that I would wait until @ivadym is back as he understand the delivery report logic much better.

cg/meta/report/rnafusion.py
ChrOertlin: Sounds good!

fevac:
- Deployed in stage
image


- Deployed in prod
image

cg/meta/report/rnafusion.py
ivadym: `is_accredited` is a better name for that method, I agree. I'll refactor it in the future. Thanks!

PR #2163, opened 2023-06-22
migration(migrate data from CGstats to CG)

ChrOertlin:
## Description

This alembic script populates the SampleLaneSequencingMetrics in the CG database with information from the Unaligned table in the CGstats database.

Only information of samples that have a matching internal_id in CG and CGstats will be migrated. 

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Check that the count of distiinct flow cell names and sample internal ids of the sample_lane_sequencing metrics matches the outcome of the following queries:
```SQL
SELECT COUNT(DISTINCT(cg.flowcell.name)) AS flow_cell_overlap
FROM
    cg.flowcell
JOIN cgstats.flowcell ON cgstats.flowcell.flowcellname = cg.flowcell.name

SELECT COUNT(DISTINCT(cg.sample.internal_id)) AS sample_overlap
FROM cg.sample
JOIN cgstats.sample ON cgstats.sample.limsid= cg.sample.internal_id
```

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


diitaz93 (commented):
Not my field of expertise but looks good!

alembic/versions/3ae765d6776d_migrate_cgstats_data.py
```diff
@@ -0,0 +1,53 @@
...
+            sample_base_fraction_passing_q30,
+            sample_base_mean_quality_score,
+            created_at
+        )
+        SELECT
+            cgstats.sample.limsid AS sample_internal_id,
+            cgstats.flowcell.flowcellname AS flow_cell_name,
+            cgstats.unaligned.lane AS flow_cell_lane_number,
+            cgstats.unaligned.readcounts AS sample_total_reads_in_lane,
+            cgstats.unaligned.q30_bases_pct AS sample_base_fraction_passing_q30,
```
seallard, line 37: This field is reported as a fraction and not a percentage in our new table - do we want to transform the data we are importing to a fraction as well?

seallard: Ok, I think so to! I think it is nice to keep everything related to the migration in a single revision, but we will likely need to revise it anyway

seallard (approved):
Nice 👍  Lets do the formatting of the data in a separate revision then

alembic/versions/3ae765d6776d_migrate_cgstats_data.py
```diff
@@ -0,0 +1,53 @@
...
+            cgstats.unaligned.mean_quality_score AS sample_base_mean_quality_score,
```
seallard, line 38: We should double check the formatting of this value as well, maybe the number of decimals differ

ChrOertlin: yes, we need to fix this currently it is 4 decimals

islean (commented):
Looks nice although I am a bit suspicious of having data loads and so on in an alembic script. Once this is in production we are essentially ruling out any option to rollback to prior to this pr. Ideally we should not need to rollback ever but I think it can be quite risky to have data manipulation which is not related to the "structure" of the tables in alembic scripts.

seallard (commented):
I think I agree with @islean, we should consider keeping this in a separate script. Tying the data in the database to a specific version of the code might not be what we want to do.

ChrOertlin:
Closing this PR - migration will be done outside of alembic. Otherwise issues might right arise with up and or downgrade of alembic revisions.

PR #2183, opened 2023-06-27
Cascade delete metrics associated with flow cell

seallard:
## Description
This PR updates the relationship between the sequencing metrics and flow cell tables to cascade deletes on flow cells to all associated sequencing metrics.

### Fixed
- Cascade delete sequencing metrics when flow cell is deleted


seallard:
> Looks, good. Did you test on stage that it does not touch any other table's entries?

Only looked at the flow cell and metrics tables, seems to behave as expected

PR #2185, opened 2023-06-27
add(dynamic read count column)

ChrOertlin:
## Description

Adds a column to the statusdb sample table that automatically updates when an insert, delete or update event in the SampleLaneSequencing metrics is triggered.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 37.0.11
+current_version = 37.0.13
 commit = True
 tag = True
 tag_name = v{new_version}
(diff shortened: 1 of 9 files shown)


seallard:
Nice idea! I think the implementation is messier than just having a regular field that is manually updated though and I don't think the added complexity justifies the benefits. We could also use something like this if we want to go the route of calculating the read count directly on the model:

```python
    @property
    def read_count(self):
        return sum(metric.sample_total_reads_in_lane for metric in self.sequencing_metrics)
```

PR #2190, opened 2023-06-28
Fix housekeeper transfer of data

seallard:
## Description
This PR fixes some issues in the new demultiplex finish command pointed out by @karlnyr:
- A bundle is now created for each sample in the flow cell
- Each sample fastq file is now added to the sample bundle
- Each sample fastq file is now tagged with the flow cell name

### Fixed
- Transfer of sequencing data to housekeeper


ChrOertlin (approved):
Changes bundle creation to also take sample ids. Looks good

PR #2191, opened 2023-06-28
Fix adding sample sheet to housekeeper

seallard:
## Description
This PR fixes the issue where the path to the sample sheet in the flow cell run directory was assumed to be in the root when its actual location seems to vary. Instead of hard coding the path, a function has been added which recursively searches the flow cell run directory for it.

### Fixed
- Find sample sheet path in flow cell run directory when transferring housekeeper files in new demultiplex finish

### DIFF
--- a/cg/meta/demultiplex/demux_post_processing.py
+++ b/cg/meta/demultiplex/demux_post_processing.py
@@ -254,12 +254,6 @@ class DemuxPostProcessingAPI:
         """
         Recursively searches for the given sample sheet file in the provided flow cell directory.
 
-        Args:
-            flow_cell_directory (Path): The path to the flow cell directory.
-
-        Returns:
-            Path: The path to the found sample sheet file.
-
         Raises:
             FileNotFoundError: If the sample sheet file is not found in the flow cell directory.
         """
--- a/cg/meta/demultiplex/demux_post_processing.py
+++ b/cg/meta/demultiplex/demux_post_processing.py
@@ -1,5 +1,6 @@
 """Post-processing Demultiiplex API."""
 import logging
+import os
 import re
 import shutil
 from contextlib import redirect_stdout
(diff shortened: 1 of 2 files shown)


cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -230,18 +231,44 @@ def add_sample_fastq_files(self, flow_cell_directory: Path, flow_cell_name: str)
...
+
+        Args:
+            flow_cell_directory (Path): The path to the flow cell directory.
+
+        Returns:
+            Path: The path to the found sample sheet file.
+
+        Raises:
+            FileNotFoundError: If the sample sheet file is not found in the flow cell directory.
+        """
```
ChrOertlin, line 263:
```suggestion
        """
        Recursively searches for the given sample sheet file in the provided flow cell directory.
        """
```

ChrOertlin (approved):
Added a sample sheet file path finder to the demux post processing api and changed function parameter variables.

Looks good, just one docstring that does not follow current convention.

cg/meta/demultiplex/demux_post_processing.py
seallard: I don't think we have a guidelines for our doc strings yet, I created a suggestion for it here https://github.com/Clinical-Genomics/atlas/pull/1772

ChrOertlin: Nice! Curretnly I think we use single sentence Docstrings, at least that is how it was communicated before by @henrikstranneheim if I am not mistaken. Maybe it is better to stick to that until something is decided.

seallard: Ok! I'll keep the exception part though since I think it is valuable

seallard:
Tested in stage

PR #2192, opened 2023-06-29
add(temporary column sample table)

ChrOertlin:
## Description

This PR adds a column "calculated_read_count" to the Sample table in statusDB.

The reason for this is the parallel validation of read count caluclation using CGstats and the new SampleLaneSequencingMetrics.

Once validated, this column will be dropped and the calucaltion logic will be implemented into the reads column.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/demultiplex/demux_post_processing.py
+++ b/cg/meta/demultiplex/demux_post_processing.py
@@ -204,7 +204,7 @@ class DemuxPostProcessingAPI:
     ) -> None:
         LOG.info(f"Add flow cell data to Housekeeper for {flow_cell_name}")
 
-        self.add_bundle_and_version_if_non_existent(flow_cell_name=flow_cell_name)
+        self.add_bundle_and_version_if_non_existent(bundle_name=flow_cell_name)
 
         tags: List[str] = [SequencingFileTag.FASTQ, SequencingFileTag.SAMPLE_SHEET, flow_cell_name]
         self.add_tags_if_non_existent(tag_names=tags)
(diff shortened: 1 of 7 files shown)


cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -197,7 +197,7 @@ def update_single_sample_read_count(self, sample_id: str) -> None:
                 sample_internal_id=sample_id
             )
             LOG.info(f"Updating sample {sample_id} with read count {sample_read_count}")
-            sample.reads = sample_read_count
+            sample.calculated_read_count = sample_read_count
```
islean, line 200: So what happens to the reads column in this case?

islean (approved):
Looks nice! Just one question - is this intended to be deployed to production? Will it cause any differences in functionality that the temporary column is filled instead of reads?

cg/meta/demultiplex/demux_post_processing.py
ChrOertlin: Nothing, it is updated with the CGstats logic that we aim to deprecate. After some reviewing of the code there are some more connections that we need to sever one by one to have this all go away. So a parallel approach in this case might be healthier.

seallard (approved):
Nice 👍  This will make it possible to run the new finish command in parallel with the old one until we have removed all dependencies on cgstats and can can replace the old command completely.

ChrOertlin:
Tested in stage:
image

PR #2195, opened 2023-06-29
Revert "Revert "add(temporary column sample table)""

ChrOertlin:
# Description

Revert the revert, since the issue with alembic was not caused by the original merge.

Reverts Clinical-Genomics/cg#2194


diitaz93 (approved):
Love the revert of a revert 😆

PR #2199, opened 2023-06-29
Filter on q30 threshold for update of read count

seallard:
## Description
This PR adds a filter on the q30 value of the sequencing metrics when updating the sample read count in the post processing step of demultiplexing. So any read counts from a lane with a q30 score below the threshold for the sequencer type are discarded.

### Fixed
- Use q30 threshold when updating the sample read count

### DIFF
--- a/cg/meta/demultiplex/demux_post_processing.py
+++ b/cg/meta/demultiplex/demux_post_processing.py
@@ -200,9 +200,11 @@ class DemuxPostProcessingAPI:
         sample = self.status_db.get_sample_by_internal_id(internal_id=sample_id)
 
         if sample:
-            sample_read_count: int = self.status_db.get_number_of_reads_for_sample_passing_q30_threshold(
-                sample_internal_id=sample_id,
-                q30_threshold=q30_threshold,
+            sample_read_count: int = (
+                self.status_db.get_number_of_reads_for_sample_passing_q30_threshold(
+                    sample_internal_id=sample_id,
+                    q30_threshold=q30_threshold,
+                )
             )
             LOG.debug(f"Updating sample {sample_id} with read count {sample_read_count}")
             sample.calculated_read_count = sample_read_count
--- a/cg/meta/demultiplex/demux_post_processing.py
+++ b/cg/meta/demultiplex/demux_post_processing.py
@@ -22,6 +22,7 @@ from cg.constants.cgstats import STATS_HEADER
 from cg.constants.constants import FileExtensions
 from cg.constants.demultiplexing import BclConverter, DemultiplexingDirsAndFiles
 from cg.constants.housekeeper_tags import SequencingFileTag
+from cg.constants.sequencing import FLOWCELL_Q30_THRESHOLD
 from cg.exc import FlowCellError
 from cg.meta.demultiplex import files
 from cg.meta.transfer import TransferFlowCell
@@ -115,11 +116,7 @@ class DemuxPostProcessingAPI:
             flow_cell_directory=flow_cell_directory_path, bcl_converter=bcl_converter
         )
 
-        flow_cell_sample_ids: List[str] = self.get_sample_ids_from_sample_sheet(
-            parsed_flow_cell=parsed_flow_cell
-        )
-
-        self.update_sample_read_counts(sample_internal_ids=flow_cell_sample_ids)
+        self.update_sample_read_counts(flow_cell_data=parsed_flow_cell)
 
         self.create_delivery_file_in_flow_cell_directory(
             flow_cell_directory=flow_cell_directory_path
@@ -161,10 +158,8 @@ class DemuxPostProcessingAPI:
                 )
                 self.status_db.session.add(sample_lane_sequencing_metric)
 
-    def get_sample_ids_from_sample_sheet(
-        self, parsed_flow_cell: FlowCellDirectoryData
-    ) -> List[str]:
-        samples: List[FlowCellSample] = parsed_flow_cell.get_sample_sheet().samples
+    def get_sample_ids_from_sample_sheet(self, flow_cell_data: FlowCellDirectoryData) -> List[str]:
+        samples: List[FlowCellSample] = flow_cell_data.get_sample_sheet().samples
         sample_ids_with_indexes: List[str] = [sample.sample_id for sample in samples]
         return [sample_id_index.split("_")[0] for sample_id_index in sample_ids_with_indexes]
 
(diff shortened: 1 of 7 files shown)


cg/store/filters/status_metrics_filters.py
```diff
@@ -13,6 +13,12 @@ def filter_total_read_count_for_sample(metrics: Query, sample_internal_id: str,
     return total_reads_query
 
 
+def filter_above_q30_threshold(metrics: Query, q30_threshold: int, **kwargs) -> Query:
```
seallard, line 16: Add test for this filter

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -184,18 +179,26 @@ def is_flow_cell_directory_valid(self, flow_cell_directory: Path) -> bool:
...
+            flow_cell_data=flow_cell_data
+        )
+
         for sample_id in sample_internal_ids:
-            self.update_single_sample_read_count(sample_id)
+            self.update_single_sample_read_count(sample_id=sample_id, q30_threshold=q30_threshold)
         self.status_db.session.commit()
 
-    def update_single_sample_read_count(self, sample_id: str) -> None:
+    def update_single_sample_read_count(self, sample_id: str, q30_threshold: int) -> None:
```
henrikstranneheim, line 195: Add doc str

cg/store/filters/status_metrics_filters.py
henrikstranneheim, line 16: Would be nice with a doc str mentioning the conversion to fraction of the q30 threshold before comparing

henrikstranneheim (approved):
Nice 👍!

cg/store/api/find_business_data.py
```diff
@@ -314,12 +314,18 @@ def get_sample_by_customer_and_name(
             name=sample_name,
         ).first()
 
-    def get_number_of_reads_for_sample_from_metrics(self, sample_internal_id: str) -> int:
+    def get_number_of_reads_for_sample_from_metrics(
```
ChrOertlin, line 317:
```suggestion
    def get_number_of_reads_for_sample_passing_q30_threshold_from_metrics(
```

cg/store/api/find_business_data.py
```diff
@@ -314,12 +314,18 @@ def get_sample_by_customer_and_name(
...
+        self, sample_internal_id: str, q30_threshold: int
+    ) -> int:
         """Get number of reads for sample from sample lane sequencing metrics."""
```
ChrOertlin, line 320:
```suggestion
        """Get number of reads that pass bases percent above q30 threshold for sample from sample lane sequencing metrics."""
```

tests/store/api/find/test_find_business_data_sample.py
```diff
@@ -328,7 +328,7 @@ def test_get_number_of_reads_for_sample_from_metrics(
 
     # WHEN getting number of reads for a specific sample
     actual_total_reads = store_with_sequencing_metrics.get_number_of_reads_for_sample_from_metrics(
-        sample_internal_id=sample_id
+        sample_internal_id=sample_id, q30_threshold=0
```
ChrOertlin, line 331:
```suggestion
        sample_internal_id=sample_id, q30_threshold=0
```
Maybe have another threshold that does some filtering also instead of 0?

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -184,18 +179,26 @@ def is_flow_cell_directory_valid(self, flow_cell_directory: Path) -> bool:
     def is_demultiplexing_complete(self, flow_cell_directory: Path) -> bool:
         return Path(flow_cell_directory, DemultiplexingDirsAndFiles.DEMUX_COMPLETE).exists()
 
-    def update_sample_read_counts(self, sample_internal_ids: List[str]) -> None:
+    def update_sample_read_counts(self, flow_cell_data: FlowCellDirectoryData) -> None:
         """Update samples in status db with the sum of all read counts for the sample in the sequencing metrics table."""
+
+        q30_threshold: int = FLOWCELL_Q30_THRESHOLD[flow_cell_data.sequencer_type]
```
ChrOertlin, line 185:
```suggestion
```

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -184,18 +179,26 @@ def is_flow_cell_directory_valid(self, flow_cell_directory: Path) -> bool:
...
+
+        sample_internal_ids: List[str] = self.get_sample_ids_from_sample_sheet(
+            flow_cell_data=flow_cell_data
+        )
+
         for sample_id in sample_internal_ids:
-            self.update_single_sample_read_count(sample_id)
+            self.update_single_sample_read_count(sample_id=sample_id, q30_threshold=q30_threshold)
```
ChrOertlin, line 192:
```suggestion
            self.update_single_sample_read_count(sample_id=sample_id, q30_threshold=FLOWCELL_Q30_THRESHOLD[flow_cell_data.sequencer_type])
```

cg/store/api/find_business_data.py
seallard: I think this doc string is a DRY violation (the name contains the exact same info). Should we remove doc strings like these instead?

henrikstranneheim: I think the info that there is a threshold and exactly what table the metrics are taken from does make it useful and doesn't violate DrY

cg/meta/demultiplex/demux_post_processing.py
ChrOertlin: Indeed, I think it makes sense to mention it is fetching data from sample lane sequencing metrics and updating the calculated read counts column in the sample table

cg/meta/demultiplex/demux_post_processing.py
seallard: I think this is less readable and prefer keeping it in a variable as above, alternatively extracting it to a function like
```python
def get_q30_threshold(sequencer_type: str):
    return FLOWCELL_Q30_THRESHOLD[sequencer_type]
```

cg/store/filters/status_metrics_filters.py
seallard: I think that is an implementation detail that the caller does not need to know about, maybe it is confusing to have in the doc string?

tests/store/api/find/test_find_business_data_sample.py
seallard: Adding some tests for the filter 👍

cg/store/api/find_business_data.py
seallard: I guess, there is also the risk of leaking implementation details. Does the caller really need to know which table is used? The caller is only interested in the result and might not care about the actual implementation

seallard: Updated the doc string

cg/store/filters/status_metrics_filters.py
seallard: Leaving it as is for now

seallard: Done

seallard:
Deployed and tested in stage.

PR #2201, opened 2023-06-30
Only store sample fastqs passing q30 in Housekeeper

seallard:
## Description
This PR adds a check of the q30 value for a sample fastq file before storing its path in housekeeper, as part of ticket https://github.com/Clinical-Genomics/project-planning/issues/482.

The check is done by looking up the sequencing metric for the fastq file using the `flow_cell_name`, `sample_internal_id` and `lane` since these uniquely identify a metric which corresponds to the forward and reverse fastq-file. The `sample_internal_id` is extracted from the parent directory of the sample fastq file and the lane is extracted from the sample fastq name. The following assumptions are made:
- The sample fastq parent directory always contains the sample internal id formatted as `Sample_<sample_id>`
- The sample fastq file name always contains the lane formatted as `_L<lane_number>`

Are these assumptions reasonable @karlnyr @Vince-janv?
It seems like these assumptions will not hold for old demultiplexed flow cells. However, if a descriptive exception is thrown and logged and it does not block production, it is ok since it will work for the vast majority of flow cells.

Some additional refactoring was done in order to improve the readability of the `DemuxPostProcessingAPI`:
- Extract utils methods from `DemuxPostProcessingAPI` to a separate utils module
- Extract validation methods from `DemuxPostProcessingAPI` to a separate validation module
- Renaming of methods to accurately reflect what is done in the method body

As well as unit tests for some new and previously untested functions.

### Added
- q30 check before storing sample fastq path in housekeeper
- Unit tests for the `validation`, `utils` and `demux_post_processing` modules


cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -73,129 +80,99 @@ def transfer_flow_cell(
...
+        create_delivery_file_in_flow_cell_directory(flow_cell_directory_path)
 
-    def create_delivery_file_in_flow_cell_directory(self, flow_cell_directory: Path) -> None:
-        Path(flow_cell_directory, DemultiplexingDirsAndFiles.DELIVERY).touch()
+    def store_flow_cell_data(self, parsed_flow_cell: FlowCellDirectoryData) -> None:
+        """Store data from the flow cell directory in status db and housekeeper."""
+        self.store_flow_cell_data_in_status_db(parsed_flow_cell)
+        self.store_sequencing_metrics_for_flow_cell(parsed_flow_cell)
+        self.update_sample_read_counts(parsed_flow_cell)
+        self.store_flow_cell_data_in_housekeeper(parsed_flow_cell)
```
seallard, line 123: The logic contained in these methods should to a large extent be extracted to separate modules, such as 
```
FlowCellTransferApi
FlowCellProcessingApi
```
etc. but that will have to be done in separate PR:s.

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -209,92 +186,77 @@ def update_single_sample_read_count(self, sample_id: str, q30_threshold: int) ->
...
-        """Extract sample id from fastq file path."""
-        sample_directory: str = fastq_file_path.parent.name
-        directory_parts: str = sample_directory.split("_")
+    def add_sample_sheet(self, flow_cell_directory: Path, flow_cell_name: str) -> None:
+        """Add sample sheet path to Housekeeper."""
 
-        if len(directory_parts) > 1:
-            # The sample id is always the second part of the directory names
-            return directory_parts[1]
+        sample_sheet_file_path: Path = get_sample_sheet_path(flow_cell_directory)
```
ChrOertlin, line 253: named args

ChrOertlin (approved):

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -209,92 +186,77 @@ def update_single_sample_read_count(self, sample_id: str, q30_threshold: int) ->
...
-        for directory_path, _, files in os.walk(flow_cell_directory):
-            if DemultiplexingDirsAndFiles.SAMPLE_SHEET_FILE_NAME in files:
-                LOG.info(f"Found sample sheet in {directory_path}")
-                return Path(directory_path, DemultiplexingDirsAndFiles.SAMPLE_SHEET_FILE_NAME)
-
-        raise FileNotFoundError(
-            f"Sample sheet not found in given flow cell directory: {flow_cell_directory}"
+        sample_id = get_sample_id_from_sample_fastq(sample_fastq_path)
+        lane = get_lane_from_sample_fastq(sample_fastq_path)
+        q30_threshold: int = get_q30_threshold(sequencer_type)
```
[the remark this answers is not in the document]
seallard: Since we have a suggestion for a guideline which makes it optional for single argument functions and it seems to be uniformly supported I'm going to keep the calls as is
https://github.com/Clinical-Genomics/atlas/pull/1782

cg/meta/demultiplex/utils.py
```diff
@@ -0,0 +1,120 @@
...
+    samples: List[FlowCellSample] = flow_cell_data.get_sample_sheet().samples
+    sample_ids_with_indexes: List[str] = [sample.sample_id for sample in samples]
+    return [sample_id_index.split("_")[0] for sample_id_index in sample_ids_with_indexes]
+
+
+def get_q30_threshold(sequencer_type: Sequencers) -> int:
+    return FLOWCELL_Q30_THRESHOLD[sequencer_type]
+
+
+def get_valid_sample_fastq_paths(flow_cell_directory: Path):
```
Vince-janv, line 80: Return hint
```suggestion
def get_valid_sample_fastq_paths(flow_cell_directory: Path)-> List[Path]:
```

cg/meta/demultiplex/utils.py
```diff
@@ -0,0 +1,120 @@
...
+    )
+
+
+def parse_flow_cell_directory_data(
+    flow_cell_directory: Path, bcl_converter: str
+) -> FlowCellDirectoryData:
+    """Parse flow cell data from the flow cell directory."""
+
+    if not is_flow_cell_directory_valid(flow_cell_directory):
+        raise FlowCellError(f"Flow cell directory was not valid: {flow_cell_directory}")
```
Vince-janv, line 118: Maybe specify what invalid means here? ie not a dir or not finished

cg/meta/demultiplex/validation.py
```diff
@@ -0,0 +1,63 @@
+from pathlib import Path
+import re
+from cg.constants.constants import FileExtensions
+
+from cg.constants.demultiplexing import DemultiplexingDirsAndFiles
+
+
+def validate_sample_fastq_file(sample_fastq: Path) -> None:
```
Vince-janv, line 8: Great function 🌟

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -73,129 +80,99 @@ def transfer_flow_cell(
...
-            LOG.warning(f"Demultiplexing is not complete for flow cell {flow_cell_directory.name}")
-            return False
+        for metric in sample_lane_sequencing_metrics:
+            if not self.metric_exists_in_status_db(metric):
+                self.add_metric_to_status_db(metric)
+        self.status_db.session.commit()
 
-        return True
+    def metric_exists_in_status_db(self, metric: SampleLaneSequencingMetrics) -> bool:
+        existing_metrics_entry = (
```
Vince-janv, line 145: type hint

Vince-janv (approved):
Great work @seallard! All additions look crisp 🔥

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -73,129 +80,99 @@ def transfer_flow_cell(
...
+            self.update_sample_read_count(sample_id=sample_id, q30_threshold=q30_threshold)
 
-    def get_q30_threshold(self, sequencer_type: Sequencers) -> int:
-        return FLOWCELL_Q30_THRESHOLD[sequencer_type]
+        self.status_db.session.commit()
 
-    def update_single_sample_read_count(self, sample_id: str, q30_threshold: int) -> None:
+    def update_sample_read_count(self, sample_id: str, q30_threshold: int) -> None:
         """Update the read count for a sample in status db with all reads exceeding the q30 threshold from the sequencing metrics table."""
         sample = self.status_db.get_sample_by_internal_id(internal_id=sample_id)
```
karlnyr, line 178: Do we need to do the `if sample:` in the line below here? In what use case would we not have a sample to add the read counts to? and if we don't - would we not want to be somewhat informed? I think logging with a warning `LOG.warning(f"Cannot find {sample_id} in status_db when adding read counts")` - this or remove the if and let it fail 🤷

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -209,92 +186,77 @@ def update_single_sample_read_count(self, sample_id: str, q30_threshold: int) ->
...
         )
-        return list(flow_cell_directory.glob(fastq_sample_pattern))
+        LOG.warning(f"Flow cell name: {flow_cell_name}, sample id: {sample_id}, lane: {lane} ")
+        return False
 
-    def get_sample_id_from_sample_fastq_file_path(self, fastq_file_path: Path) -> str:
-        """Extract sample id from fastq file path."""
-        sample_directory: str = fastq_file_path.parent.name
-        directory_parts: str = sample_directory.split("_")
+    def add_sample_sheet(self, flow_cell_directory: Path, flow_cell_name: str) -> None:
```
karlnyr, line 250:
```suggestion
    def add_sample_sheet_to_housekeeper(self, flow_cell_directory: Path, flow_cell_name: str) -> None:
```

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -209,92 +186,77 @@ def update_single_sample_read_count(self, sample_id: str, q30_threshold: int) ->
...
-            flow_cell_directory=flow_cell_directory, flow_cell_name=flow_cell_name
-        )
-        self.add_sample_fastq_files(
-            flow_cell_directory=flow_cell_directory, flow_cell_name=flow_cell_name
-        )
+        self.add_sample_sheet(flow_cell_directory=flow_cell.path, flow_cell_name=flow_cell.id)
+        self.add_sample_fastq_files(flow_cell)
 
-    def add_sample_fastq_files(self, flow_cell_directory: Path, flow_cell_name: str) -> None:
+    def add_sample_fastq_files(self, flow_cell: FlowCellDirectoryData) -> None:
```
karlnyr, line 200:
```suggestion
    def add_sample_fastq_files_to_housekeeper(self, flow_cell: FlowCellDirectoryData) -> None:
```

cg/meta/demultiplex/utils.py
```diff
@@ -0,0 +1,120 @@
...
+    """Get all valid sample fastq file paths from flow cell directory."""
+    fastq_file_paths: List[Path] = get_sample_fastq_paths_from_flow_cell(flow_cell_directory)
+    valid_sample_fastq_paths: List[Path] = []
+
+    for fastq_path in fastq_file_paths:
+        try:
+            validate_sample_fastq_file(fastq_path)
+            valid_sample_fastq_paths.append(fastq_path)
+        except ValueError as e:
+            LOG.warning(f"Skipping invalid sample fastq file {fastq_path.name}: {e}")
```
karlnyr, line 90: I really like this! I am just wondering - where are these logs kept? are they printed to the demultiplexing directory? Should we perhaps add these log files for the demultiplexing onto the flow cell bundle to trace back what happened to the post demux in the future - Since we do remove, or at least plan on removing demultiplexing directories, they would be lost otherwise.

cg/meta/demultiplex/utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def get_sample_sheet_path(flow_cell_directory: Path) -> Optional[Path]:
+    """
+    Recursively searches for the given sample sheet file in the provided flow cell directory.
+
+    Raises:
+        FileNotFoundError: If the sample sheet file is not found in the flow cell directory.
+    """
+    for directory_path, _, files in os.walk(flow_cell_directory):
```
karlnyr, line 102: Nice 🚶

cg/meta/demultiplex/utils.py
karlnyr: I agree with Vincent - perhaps we can return a string from the validator with the specific metric that fails?

karlnyr (commented):
This looks really nice! I really appreciate that we bring out these validators and utils outside of the API - makes it easier to get a grab of what their intended purpose is as well!

I have some minor comments - but I think it looks good, and also nice to be able to track that the Q30 was poor and then proceed with not adding it into housekeeper 👍

cg/meta/demultiplex/utils.py
seallard: Yes, good catch 👍

cg/meta/demultiplex/demux_post_processing.py
seallard: Hm, my best guess is that it could occur due to some issue with the porting of samples from cgstats to cg. But I think @ChrOertlin is better informed - do you think this is something we should handle?

diitaz93 (approved):

diitaz93 (commented):
Will finish tomorrow

cg/meta/demultiplex/demux_post_processing.py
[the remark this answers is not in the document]
seallard: I will leave this as is since I think `get_q30_threshold_for_sequencer_type` is too verbose and it should be clear from the argument name that it is from the sequencer type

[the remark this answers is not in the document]
seallard: Not sure though, just a preference

tests/meta/demultiplex/test_utils.py
```diff
@@ -0,0 +1,92 @@
...
+    lane = 4
+    sample_fastq_path = Path(
+        f"ACC12164A17_S367_L00{lane}_R1_001{FileExtensions.FASTQ}{FileExtensions.GZIP}",
+    )
+
+    # WHEN we get lane from the sample fastq file path
+    result = get_lane_from_sample_fastq(sample_fastq_path)
+
+    # THEN we should get the correct lane
+    assert result == lane
```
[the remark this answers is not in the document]
seallard: I want to modify different parts of it in the tests, I think it is easier to keep like this

cg/meta/demultiplex/demux_post_processing.py
seallard: I guess we have to since the return type of the `get_sample_by_internal_id` function is an optional Sample

seallard: Added the log warning

cg/meta/demultiplex/validation.py
```diff
@@ -0,0 +1,63 @@
...
+        raise ValueError(f"Sample fastq must end with {FileExtensions.FASTQ}{FileExtensions.GZIP}.")
+
+    if not is_lane_in_fastq_file_name(sample_fastq):
+        raise ValueError("Sample fastq must contain lane number formatted as '_L<lane_number>'.")
+
+    if not is_sample_id_in_directory_name(sample_fastq.parent):
+        raise ValueError("Parent directory name of sample fastq must contain 'Sample_<sample_id>'.")
+
+
+def is_file_path_compressed_fastq(file_path: Path) -> bool:
```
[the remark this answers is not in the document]
seallard: I think it violates DRY and that it leaks implementation details, I'm going to leave it out for now. Maybe we can make the guidelines clearer

seallard:
Creating a separate PR with the missing integration tests, this PR is already doing too many different things and is getting out of scope.

cg/meta/demultiplex/utils.py
seallard: @Vince-janv  helped me out. The cron-tab jobs end up in `/home/proj/production/logs`. Which specific file can be determined from https://github.com/Clinical-Genomics/servers/blob/master/crontab/hasta.scilifelab.se.sh
For system-d stuff we can find the logs with `journalctl --user-unit <service-name>`.

seallard: Not sure about the configuration for how long these logs are kept, something we definitely should review.

seallard: Created an issue for your suggestion: https://github.com/Clinical-Genomics/cg/issues/2217

cg/meta/demultiplex/demux_post_processing.py
karlnyr: nice!

Issue #2216, opened 2023-07-05
Remove index checks from sample sheets


seallard:
## Description
Currently, we generate index checks in our sample sheets. These are useful to quickly get reads for a sample when something has gone wrong in lims. However, they do require us to run the demultiplexing process with stricter settings. Removing the index checks would allow us to use less strict settings for the demultiplexing process, saving us and the customers time and money.

## Suggested solution
Remove the index checks from the sample sheet generation?

## This can be closed when
The index checks have been removed from the sample sheets and the demultiplexing process can be run with less strict settings.

PR #2220, opened 2023-07-05
Remove dummy samples from sample sheets

seallard:
## Description
This PR removes dummy samples (index checks) from the sample sheets, see https://github.com/Clinical-Genomics/cg/issues/2216.

Dummy samples (a.k.a index checks) help us get reads for a sample when something has gone wrong in lims and a sample has been assigned the wrong index. However, it forces us to run the demultiplexing software (`bcl2fastq` and `bclconvert`) with stricter settings which costs us and our customers time and money. @karlnyr thinks this drawback outweighs the benefit of having dummy samples.

### Fixed
- Remove dummy samples from sample sheets

### Blocked
This is blocked until we have deprecated cgstats. Removing the index checks will require re-demultiplexing which involves deleting things in cgstats which is a mess currently.


diitaz93:
Interesting! This would make the sample sheet generation much simpler. It will be interesting to know the opinion of @Karl-Svard, @idalindegaard and @amandaraud

karlnyr:
Also, @AnnaLeinfelt I thought about the risk of this - with the speed of dragen this would equate to about 2 hours of extra time before delivery - There is actually 1 other solution.... and that is to add indexes sets which are at least 3 hamming distance away from the other indexes in the sheet. But I don't know about the number of index checks we would be able to include then.

Secondly - In today's routine we would check the top unknown barcodes - but many reads may be assigned to one of these indexes - it would, of course, be better if everyone checked all files - but I think collecting the information that some indexes that we have on-site was "unknown" in the report is better.

seallard:
> Also, @AnnaLeinfelt I thought about the risk of this - with the speed of dragen this would equate to about 2 hours of extra time before delivery - There is actually 1 other solution.... and that is to add indexes sets which are at least 3 hamming distance away from the other indexes in the sheet. But I don't know about the number of index checks we would be able to include then.
> 
> Secondly - In today's routine we would check the top unknown barcodes - but many reads may be assigned to one of these indexes checks instead, thus wont appear in the top unknown report - it would, of course, be better if everyone checked all files - but I think collecting the information that some indexes that we have on-site was "unknown" in the report is better.

Would we save 2 hours per demultiplexing run by removing the index checks? Also, how often is a sample assigned the wrong index in lims and how much work would it be to fix without the index checks?

karlnyr:
> > Also, @AnnaLeinfelt I thought about the risk of this - with the speed of dragen this would equate to about 2 hours of extra time before delivery - There is actually 1 other solution.... and that is to add indexes sets which are at least 3 hamming distance away from the other indexes in the sheet. But I don't know about the number of index checks we would be able to include then.
> > Secondly - In today's routine we would check the top unknown barcodes - but many reads may be assigned to one of these indexes checks instead, thus wont appear in the top unknown report - it would, of course, be better if everyone checked all files - but I think collecting the information that some indexes that we have on-site was "unknown" in the report is better.
> 
> Would we save 2 hours per demultiplexing run by removing the index checks? Also, how often is a sample assigned the wrong index in lims and how much work would it be to fix without the index checks?

Noooooo - By having index checks in the sample sheets we could save 2 hours of delivery time - since we could simply take the fastq files from the index that captured the reads instead of re-demultiplexing the flow cell again. 

Wrong index set in lims does happen - it is a human error so the rate is very variable. It would need to be corrected either way because we assume that LIMS is the truth so the lab staff would have to re-lims the samples again to set the correct index. On the bioinformatic side we can simply edit the sample sheet manually and set the correct index for the sample and then restart demultiplexing.

karlnyr:
> Hi! I need some more information on how we operate today, or rather, will operate with the X Plus.
> 
> For X Plus:
> 
> * Will we allow any mismatches in the read index reads? Or are we only using perfect index reads?
> * It would then be interesting to look into the **unassigned** reads with the same criteria as above.
> * @karlnyr Can you elaborate on the benefit of including indices with hamming distance of 3, what would we gain?

With the X we plan to start of by:

- Allow for mismatches if the index setup on the flow cell allows for it. This will be decided on samples within the lane, and index position.
- I don't think I understand this point
- The idea of including index checks into the sample sheet is to enable fast addition of fastq where they actually belong - this however would only be fast if the index clash is on one sample and not multiple samples. Furthermore, by limiting the index checks to a hamming distance of 3 (to allow for mismatches in indexes) we could also just miss the samples anyways - completely removing their purpose but by chance enabling the scenario I described first in this point.

seallard:
Conclusions from meeting: merge this once cgstats is deprecated.

seallard:
Closing this in favour of https://github.com/Clinical-Genomics/cg/pull/2477 to avoid merge conflicts.

PR #2229, opened 2023-07-07
Remove parsing of RunInfo from new bcl convert parserr

seallard:
## Description
This PR removes the parsing of the `RunInfo.xml` file from the new sequencing metrics parser for the `bclconvert` demultiplexing software. I could not find any usage of the parsed information from the `RunInfo.xml` file, so I think it is safe to simplify the parser and remove it.

### Fixed
- Remove parsing of unused information from `RunInfo.xml` in sequencing metrics parser

### How to test

- [ ] Do `cg demultiplex finish temporary some_demultiplexed_flow_cell`


diitaz93 (approved):
Great 👍  The least dependence on RunInfo the better

ChrOertlin:
Can we keep this for a bit longer... until we implemented the other things?

islean:
My intuition would be that this info is needed if we need to rerun with the same parameters manually at some point? I am probably the wrong person to ask but are we sure that the info is not used even if it is not mentioned in cg?

diitaz93:
> 

I think most of the important info is in the `RunParameters.xlm`. I have never found a use for `RunInfo.xlm` but if @ChrOertlin  does not agree I think we should hold it until we are all on the same page on this

seallard:
> My intuition would be that this info is needed if we need to rerun with the same parameters manually at some point? I am probably the wrong person to ask but are we sure that the info is not used even if it is not mentioned in cg?

That might be true, but a really good development rule is to never implement something unless it is needed now. Otherwise it is really easy to end up with a lot of code which does nothing and that no one understands. If we do end up needing it in the future we still have it in version control. So I think we should remove it

islean:
> > My intuition would be that this info is needed if we need to rerun with the same parameters manually at some point? I am probably the wrong person to ask but are we sure that the info is not used even if it is not mentioned in cg?
> 
> That might be true, but a really good development rule is to never implement something unless it is needed now. Otherwise it is easy to end up with a lot of code which does nothing and that no one understands. If we do end up needing it in the future we still have it in version control. So I think we should remove it

I don't think this really addressed my concern. The version control would allow us to get the code back, but my concern was that the xml files might be needed. But that is instead addressed by my noticing that this code does not produce said file :)

ChrOertlin:
Stage:
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/hiseq.clinical
done.
Log deploy... done.

Prod:
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/hiseq.clinical
done.
Log deploy... done.
cg, version 38.1.4

Issue #2056, opened 2023-05-24
Direct imports of fixtures in tests and conftests


seallard:
## Description
We have some imports of fixtures like:
```python
from tests.cli.workflow.balsamic.conftest import (
    fixture_balsamic_context,
    balsamic_lims,
    balsamic_housekeeper,
    balsamic_mock_fastq_files,
    balsamic_housekeeper_dir,
    fastq_file_l_1_r_1,
    fastq_file_l_1_r_2,
    fastq_file_l_2_r_1,
    fastq_file_l_2_r_2,
    fastq_file_l_3_r_1,
    fastq_file_l_3_r_2,
    fastq_file_l_4_r_1,
    fastq_file_l_4_r_2,
)
```

These imports makes using and changing the fixtures difficult. We should let pytest handle importing the fixtures. I guess the reason for the imports is due to the fixtures being present in conftest files not available to the test suite. And instead of fixing the issue by moving the fixtures to the correct conftest file we directly import them.

## Suggested solution
- Move fixtures as to enable pytest to automatically resolve the fixtures

## This can be closed when
- No direct imports of fixtures remain

ivadym:
It's hard to address this when we have a test file structure that doesn't allow reusing fixtures. Most of them are common for CLI and meta tests, for example. However, if we want to currently define common fixtures, we would need to add this pipeline or tool-specific definitions to the main conftest, which will become increasingly complex.

I stumbled across this at: https://github.com/Clinical-Genomics/cg/pull/2079

Has this been refined and do we have a strategy to avoid adding more complexity to this issue?

seallard:
Good point! I guess we need to settle on an approach if we wish to limit the number of lines in the contest.py in the root of the tests directory. 

I think it might be best to mirror the overall project structure in our tests directory and I'm not that concerned about the number of lines in the contest.py.

seallard:
I think the issue of a large number of fixtures in the root conftest file is more related to how we write our tests.

seallard:
The direct imports are also confusing because it is necessary to also import the fixtures used by the fixture directly. So we end up with a long list of imported fixtures which are not used directly in the test, but they need to be imported in order to use other fixtures.

seallard:
Another idea is to keep the fixtures in the test file in which it is used and only move it to a conftest if it is needed elsewhere

PR #2230, opened 2023-07-07
#2056: Move fixtures to root conftest

seallard:
## Description
This PR moves some fixtures related to testing of the demultiplex post processing into the root conftest file and removes direct imports of them in our tests.

When modifying an existing fixture, suddenly a lot of tests broke due to not having access to a new fixture used by another fixture etc. All due to the direct imports of fixtures. Instead of updating each location directly importing the relevant fixtures with direct imports of new fixtures I moved them into the root conftest.

### Fixed
- Remove direct imports of demux fixtures by moving fixtures into root conftest


diitaz93 (commented):
Great initiative, but could you place the demultiplex fixtures under the demultiplex section delimited by the comment in line 729 of `test/conftest.py`? It would make them so much easier to find

seallard:
> Great initiative, but could you place the demultiplex fixtures under the demultiplex section delimited by the comment in line 729 of `test/conftest.py`? It would make them so much easier to find

Good catch :)

PR #2235, opened 2023-07-07
Add integration test for post processing of flow cell demultiplexed with bcl2fastq

seallard:
## Description
This PR adds an integration test for the post processing of a flow cell demultiplexed with bcl2fastq (part of project https://github.com/Clinical-Genomics/project-planning/issues/482).

### Fixed
- Add integration test for post processing of flow cell


tests/conftest.py
Vince-janv, line 1: Add name and scope to the fixtures. Also have them start with `fixture_`

tests/meta/demultiplex/test_demux_post_processing.py
```diff
@@ -769,8 +770,77 @@ def test_post_processing_of_flow_cell_demultiplexed_with_bclconvert(
...
         demux_post_processing_api.demux_api.out_dir,
-        flow_cell_name_demultiplexed_with_bcl_convert,
+        flow_cell_directory_name_demultiplexed_with_bcl_convert,
+        DemultiplexingDirsAndFiles.DELIVERY,
+    )
+
+    assert delivery_path.exists()
+
+
+def test_post_processing_of_flow_cell_demultiplexed_with_bcl2fastq(
```
Vince-janv, line 780: It's hard to make such a large test this clear, well done ⭐

Vince-janv (approved):
Great test! 🎖️

tests/conftest.py
seallard: I know this is how we currently write our fixtures but I don't really see why? I dislike prepending `fixture_` since it is always clear by context that it is a fixture. And having the same name in the pytest decorator does not seem very dry.

```
@pytest.fixture(name="my_fixture_name", scope="session")
def fixture_my_fixture_name():
    return ["SVE2528A1"]
```

The fact that a function is a pytest fixture is usually clear from its use in tests (because it's passed as an argument to test functions) and from its definition (because it's decorated with @pytest.fixture).

seallard: Maybe it is because of the anti-pattern occuring in cg where we directly import fixtures? That it is confusing to try to import fixtures directly when they don't start with `fixture_`

PR #2267, opened 2023-07-19
Include time stamp in sequencing date for flow cells

seallard:
## Description
This PR fixes the missing time stamp in the flow cell model and is part of the https://github.com/Clinical-Genomics/project-planning/issues/482 project.

The timestamp was missing from the `sequenced_at` field for the flow cell in the new post processing flow. 

As far as I understand from the existing flow, the date and timestamp used for the status db flow cell model was just the current time set in cgstats when the flow cell model was created there. Not sure if that is reasonable. 

Some alternatives:
1. Use the current time stamp and keep `sequenced_at` (this PR)
2. Rename `sequenced_at` to `demultiplexed_at` and use the current time stamp
3.  Only use the date from the directory name and keep `seqenced_at`
4. Get the exact time stamp somehow from the demux files.

### Fixed
- Include time stamp in the `sequenced_at` field in the flow cell model 


ChrOertlin:
I agree with that the current implementation of the sequenced at is an approximation rather than the actual time a sequencing run has started or is finished. Maybe having a time format like yy-mm-dd would be "accurate" enough?

seallard:
> I agree with that the current implementation of the sequenced at is an approximation rather than the actual time a sequencing run has started or is finished. Maybe having a time format like yy-mm-dd would be "accurate" enough?

Hm, now I remember that @karlnyr wanted the timestamp to be included when we did the first validation round. I made this note `The sequenced_at date and timestamp differs (important for TAT and automation)`. So maybe it is better to merge it like this and use the current timestamp (as done before)?

ChrOertlin (approved):
Looks good. If there are any refinements regarding the time stamp later on we can still change. this will follow the current implementation and I agree we can adhere to that.

Vince-janv:
I think it's relevant to know when the actual sequencing was performed instead of the demultiplexing. That leaves alternatives 3 and 4. I also think a date is enough, so 3 is probably easiest

PR #2268, opened 2023-07-19
Add endpoint for retrieving sequencing metrics

seallard:
## Description
This PR adds an endpoint to the web api for retrieving sequencing metrics for a flow cell. Part of project https://github.com/Clinical-Genomics/project-planning/issues/482. 

This endpoint will be used in other repos such as `cg_lims`/`clinical_EPPs`, enabling us to remove their dependency on cgstats.
 

### Added
- Endpoint for retrieving sequencing metrics for a flow cell

### How to test
- [x] Send a request with Postman against `https://cg-statusdb-stage.scilifelab.se/api/v1/flowcells/<flow_cell_id>/sequencing_metrics`
### Expected test outcome
- [x] Check that a 200 response is given and the response contains the metrics

### DIFF
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -373,17 +373,17 @@ def get_sequencing_metrics(flow_cell_name: str):
     if not flow_cell_name:
         return jsonify({"error": "Invalid or missing flow cell id"}), http.HTTPStatus.BAD_REQUEST
 
-    metrics: List[
+    sequencing_metrics: List[
         SampleLaneSequencingMetrics
     ] = db.get_sample_lane_sequencing_metrics_by_flow_cell_name(flow_cell_name)
 
-    if not metrics:
+    if not sequencing_metrics:
         return (
             jsonify({"error": f"Sequencing metrics not found for flow cell {flow_cell_name}."}),
             http.HTTPStatus.NOT_FOUND,
         )
 
-    return jsonify([metric.to_dict() for metric in metrics])
+    return jsonify([metric.to_dict() for metric in sequencing_metrics])
 
 
 @BLUEPRINT.route("/analyses")
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -368,7 +368,7 @@ def parse_flow_cell(flowcell_id):
 
 @BLUEPRINT.route("/flowcells/<flow_cell_id>/sequencing_metrics", methods=["GET"])
 def get_sequencing_metrics(flow_cell_id: str):
-    """Return sequencing metrics for a flow cell."""
+    """Return sample lane sequencing metrics for a flow cell."""
 
     if not flow_cell_id:
         return jsonify({"error": "Invalid or missing flow cell id"}), http.HTTPStatus.BAD_REQUEST
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -19,7 +19,17 @@ from cg.meta.orders import OrdersAPI
 from cg.models.orders.order import OrderIn, OrderType
 from cg.models.orders.orderform_schema import Orderform
 from cg.server.ext import db, lims, osticket
-from cg.store.models import Analysis, Application, Customer, Family, Flowcell, Pool, Sample, User
+from cg.store.models import (
+    Analysis,
+    Application,
+    Customer,
+    Family,
+    Flowcell,
+    Pool,
+    Sample,
+    SampleLaneSequencingMetrics,
+    User,
+)
 from flask import Blueprint, abort, current_app, g, jsonify, make_response, request
 from google.auth import jwt
 from pydantic.v1 import ValidationError
@@ -354,6 +364,15 @@ def parse_flow_cell(flowcell_id):
     if flow_cell is None:
         return abort(http.HTTPStatus.NOT_FOUND)
     return jsonify(**flow_cell.to_dict(samples=True))
+
+
+@BLUEPRINT.route("/flowcells/<flow_cell_id>/sequencing_metrics", methods=["GET"])
+def get_sequencing_metrics(flow_cell_id: str):
+    """Return sequencing metrics for a flow cell."""
+    metrics: List[
+        SampleLaneSequencingMetrics
+    ] = db.get_sample_lane_sequencing_metrics_by_flow_cell_name(flow_cell_id)
+    return jsonify(metrics)
 
 
 @BLUEPRINT.route("/analyses")
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -372,7 +372,7 @@ def get_sequencing_metrics(flow_cell_id: str):
     metrics: List[
         SampleLaneSequencingMetrics
     ] = db.get_sample_lane_sequencing_metrics_by_flow_cell_name(flow_cell_id)
-    return jsonify(metrics)
+    return jsonify([metric.to_dict() for metric in metrics])
 
 
 @BLUEPRINT.route("/analyses")
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -368,7 +368,7 @@ def parse_flow_cell(flowcell_id):
 
 @BLUEPRINT.route("/flowcells/<flow_cell_id>/sequencing_metrics", methods=["GET"])
 def get_sequencing_metrics(flow_cell_id: str):
-    """Return sequencing metrics for a flow cell."""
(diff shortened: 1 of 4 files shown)


seallard:
Tested in web-stage in Postman sending a request against `https://cg-statusdb-stage.scilifelab.se/api/v1/flowcells/H27CLDRX3/sequencing_metrics`, which gives the expected response:

```
[
  {
    "created_at": "Wed, 05 Jul 2023 13:27:29 GMT", 
    "flow_cell_lane_number": 1, 
    "flow_cell_name": "H27CLDRX3", 
    "id": 9, 
    "sample_base_fraction_passing_q30": 0.93, 
    "sample_base_mean_quality_score": 35.75, 
    "sample_internal_id": "ACC11915A1", 
    "sample_total_reads_in_lane": 17731348
  }, 
  {
    "created_at": "Wed, 05 Jul 2023 13:27:29 GMT", 
    "flow_cell_lane_number": 2, 
    "flow_cell_name": "H27CLDRX3", 
    "id": 10, 
    "sample_base_fraction_passing_q30": 0.93, 
    "sample_base_mean_quality_score": 35.74, 
    "sample_internal_id": "ACC11915A1", 
    "sample_total_reads_in_lane": 17604800
  }, 
  {
    "created_at": "Wed, 05 Jul 2023 13:27:29 GMT", 
    "flow_cell_lane_number": 1, 
    "flow_cell_name": "H27CLDRX3", 
    "id": 65, 
    "sample_base_fraction_passing_q30": 0.93, 
    "sample_base_mean_quality_score": 35.77, 
    "sample_internal_id": "ACC11915A10", 
    "sample_total_reads_in_lane": 19626510
  }, 
  {
    "created_at": "Wed, 05 Jul 2023 13:27:29 GMT", 
    "flow_cell_lane_number": 2, 
    "flow_cell_name": "H27CLDRX3", 
    "id": 66, 
    "sample_base_fraction_passing_q30": 0.93, 
    "sample_base_mean_quality_score": 35.75, 
    "sample_internal_id": "ACC11915A10", 
    "sample_total_reads_in_lane": 19488650
  }, 
... (etc)
]

```

cg/server/api.py
```diff
@@ -356,6 +366,23 @@ def parse_flow_cell(flowcell_id):
     return jsonify(**flow_cell.to_dict(samples=True))
 
 
+@BLUEPRINT.route("/flowcells/<flow_cell_id>/sequencing_metrics", methods=["GET"])
+def get_sequencing_metrics(flow_cell_id: str):
+    """Return sequencing metrics for a flow cell."""
```
ChrOertlin, line 371:
```suggestion
def get_sequencing_metrics(flow_cell_id: str):
    """Return sample lane sequencing metrics for a flow cell."""
```
We can also think of making the function name more explicit to what it fetches, i.e. ```get_sample_lane_sequencing_metrics``` 
Although I think ```sequencing_metrics``` can be considered explicit enough with a clarification in the doc string.

cg/server/api.py
```diff
@@ -356,6 +366,23 @@ def parse_flow_cell(flowcell_id):
     return jsonify(**flow_cell.to_dict(samples=True))
 
 
+@BLUEPRINT.route("/flowcells/<flow_cell_id>/sequencing_metrics", methods=["GET"])
+def get_sequencing_metrics(flow_cell_id: str):
```
ChrOertlin, line 370:
```suggestion
def get_sequencing_metrics(flow_cell_name: str):
```
Worth considering that we retrieve using the flow cell name and not the database entry id. Can also be read from the function usage below I suppose, more a comment on consistency. Although the end-point is coded to retrieve for an id. So I get the choice here as well.  I suppose here the endpoint might clash with the back-end.

cg/server/api.py
```diff
@@ -356,6 +366,23 @@ def parse_flow_cell(flowcell_id):
...
+
+    if not flow_cell_id:
+        return jsonify({"error": "Invalid or missing flow cell id"}), http.HTTPStatus.BAD_REQUEST
+
+    metrics: List[
+        SampleLaneSequencingMetrics
+    ] = db.get_sample_lane_sequencing_metrics_by_flow_cell_name(flow_cell_id)
+
+    if not metrics:
+        return jsonify({"error": "Sequencing metrics not found"}), http.HTTPStatus.NOT_FOUND
```
ChrOertlin, line 381:
```suggestion
    if not metrics:
        return jsonify({"error": f"Sample Lane Sequencing metrics for flow cell {flow_cell_id} not found."}), http.HTTPStatus.NOT_FOUND
```

Might be helpful with a more specific error message.

ChrOertlin (approved):
Looks good, just some minor things.

cg/server/ext.py
```diff
@@ -32,6 +36,13 @@ def init_app(self, app):
         super(FlaskStore, self).__init__(uri)
 
 
+class CustomJSONEncoder(JSONEncoder):
```
Vince-janv, line 39: What happens if this is not handled?

cg/server/api.py
Vince-janv: I think we should go with flow_cell_name, since that is the column title. However `flow_cell_id` is used a lot in cg 😬

cg/server/api.py
Vince-janv: I think sample_lane_sequencing metrics is a mouthful. For most circumstances i think `sequencing_metrics` is enough

cg/server/api.py
```diff
@@ -356,6 +366,23 @@ def parse_flow_cell(flowcell_id):
...
 
 
+@BLUEPRINT.route("/flowcells/<flow_cell_id>/sequencing_metrics", methods=["GET"])
+def get_sequencing_metrics(flow_cell_id: str):
+    """Return sequencing metrics for a flow cell."""
+
+    if not flow_cell_id:
+        return jsonify({"error": "Invalid or missing flow cell id"}), http.HTTPStatus.BAD_REQUEST
+
+    metrics: List[
```
Vince-janv, line 376: I think metrics is a bit too general for variable names
```suggestion
    sequencing_metrics: List[
```

Vince-janv (approved):
Great that you're starting on this 🔥

cg/server/api.py
seallard: Updated the doc string to be more descriptive 👍

cg/server/api.py
seallard: Yeah, I'll change it to `flow_cell_name`, it bothers me a little since name does not imply uniqueness and resources are usually fetched with an id. But it is more consistent and something we can change in a separate PR

seallard: https://github.com/Clinical-Genomics/cg/pull/2268/commits/afa6d3f0318a5d5dc830321f6b0ddf112c6c9dfe

cg/server/api.py
seallard: https://github.com/Clinical-Genomics/cg/pull/2268/commits/914ca2c412545bd8b9e2c28ae3c74483f025bd35

cg/server/ext.py
seallard: Most fields are serialisable by default by the `JSONEncoder`, but the `Decimal` (`BigInteger`) field in the sequencing metrics table was an exception. Adding the same handling to all places where we might have a `Decimal` seemed annoying, so I thought it would be better to handle the fields which are not serialisable by default here

seallard: But yeah, if a field cannot be serialised we get an internal server error when the endpoint tries to serialise the database object to json. There are some other solutions to this problem, such as using a json serialisation library like `Marshmallow` in Python. That would require us to add separate schema models for all our database objects that we return and serialise in the API

seallard: Which seemed out of scope for this PR 😅 Not sure how we want to do this, but this seems like an OK way for now

cg/server/api.py
seallard: https://github.com/Clinical-Genomics/cg/pull/2268/commits/a7d6e1c8702948b58780099269fd52b62c258c22

cg/server/ext.py
Vince-janv: Thanks for the explanation. I think this implementation looks good 👍

seallard:
Tested again in stage

seallard:
Deployed to prod

PR #2274, opened 2023-07-21
Return if flow cell is not valid

seallard:
## Description
This PR avoids raising an exception when a flow cell could not be post processed due to not being valid and just skips it after logging an error message. This fix avoids causing the command to exit with a non zero status.

### Fixed
- Skip flow cell instead of raising exception when it is invalid


ChrOertlin (approved):
Looks reasonable to me!

PR #2275, opened 2023-07-21
refactor(demux post processing api)

ChrOertlin:
## Description

Moves store and housekeeper related functions into separate modules to make the core flow of the post processing API more understandable.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


tests/meta/demultiplex/test_demux_post_processing.py
```diff
@@ -485,46 +499,6 @@ def test_finish_all_flowcells(
     assert f"Check demultiplexed flow cell {bcl2fastq_flow_cell.full_name}" in caplog.text
 
 
-def test_add_flow_cell_data_to_housekeeper(demultiplex_context: CGConfig):
```
ChrOertlin, line 488: Removed this test. This function just calls other functions that are invoked, these are tested.

ChrOertlin:
cg demultiplex finish temporary 230622_A00621_0864_BH5F5YDSX7 finished successfully with all steps

cg/meta/demultiplex/hk_functions.py
seallard, line 1: Maybe the module can be given a more descriptive name? For example `flowcell_data_storage_housekeeper`, `demultiplexing_housekeeper_operations`, `housekeeper_flow_cell_data_storage` or so.

cg/meta/demultiplex/hk_functions.py
```diff
@@ -0,0 +1,217 @@
...
+) -> None:
+    sample_fastq_should_be_stored: bool = check_if_fastq_path_should_be_stored_in_housekeeper(
+        sample_id=sample_internal_id,
+        sample_fastq_path=sample_fastq_path,
+        sequencer_type=flow_cell.sequencer_type,
+        flow_cell_name=flow_cell.id,
+        store=store,
+    )
+
+    if sample_fastq_should_be_stored:
```
seallard, line 117: There has been a bug fix applied to this function which is missing here, see https://github.com/Clinical-Genomics/cg/commit/1ae9c379e29fc8054382bd258952370bb195e5f5#diff-afa9660521ae4a7f3be5393c339ace5b0741915c3b2a2c7f2bd04cf788d3ea26R311

cg/meta/demultiplex/store_functions.py
seallard, line 1: Maybe there is a more descriptive name for this module as well

seallard (approved):
Nice! It is good that we separate logic from the main flow into these modules with clearer responsibilities 👍  I found one function where a patch was missing, might be worth double checking the commits in the master branch as to not reintroduce any bugs. Always tricky when moving functions like this

cg/meta/demultiplex/hk_functions.py
```diff
@@ -0,0 +1,217 @@
...
+)
+from cg.utils.files import get_files_matching_pattern
+
+LOG = logging.getLogger(__name__)
+
+
+def store_flow_cell_data_in_housekeeper(
+    flow_cell: FlowCellDirectoryData,
+    hk_api: HousekeeperAPI,
+    demux_api: DemultiplexingAPI,
```
seallard, line 29: Maybe we can consider just sending in the flow cell run directory here? The demux api is not used, only the run dir attribute. I think it is cleaner to just send in that. Might be out of scope for this PR if it was like this already

seallard: Easier to test if we just send in the run dir as well

cg/meta/demultiplex/hk_functions.py
ChrOertlin: added the bug fix here.

cg/meta/demultiplex/store_functions.py
ChrOertlin: same as above.

cg/meta/demultiplex/housekeeper_storage_functions.py
```diff
@@ -0,0 +1,219 @@
...
+) -> None:
+    """Add sample fastq files from flow cell to Housekeeper."""
+
+    sample_internal_ids: List[str] = get_sample_internal_ids_from_sample_sheet(
+        sample_sheet_path=flow_cell.sample_sheet_path,
+        flow_cell_sample_type=flow_cell.sample_type,
+    )
+
+    for sample_internal_id in sample_internal_ids:
+        add_bundle_and_version_if_non_existent(bundle_name=sample_internal_id, hk_api=hk_api)
```
seallard, line 81: This needs to be removed as to not create empty sample bundles

cg/meta/demultiplex/store_storage_functions.py
Vince-janv, line 1: I think this file should be called status_db_storage_functions
Even though I'm not a big fan of the name status-db  I still think it's better than just store.

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -154,266 +145,16 @@ def finish_all_flow_cells_temp(self) -> bool:
...
-            )
-
-    def file_exists_in_latest_version_for_bundle(self, file_path: Path, bundle_name: str) -> bool:
-        """Check if file exists in latest version for bundle."""
-        latest_version: Version = self.hk_api.get_latest_bundle_version(bundle_name=bundle_name)
-        return any(
-            file_path.name == Path(bundle_file.path).name for bundle_file in latest_version.files
+        store_flow_cell_data_in_status_db(parsed_flow_cell, store=self.status_db)
+        store_sequencing_metrics_in_status_db(parsed_flow_cell, store=self.status_db)
+        update_sample_read_counts_in_status_db(parsed_flow_cell, store=self.status_db)
```
Vince-janv, line 150: We are using a mix of positional and key word arguments here. I'm fine with it since the combination of variable and function names makes it clear what's happening, but it does go against the discussion we had in the coding guidelines doc. What do you think @ChrOertlin and @seallard?

Vince-janv (commented):
Nice 👍  I'm guessing none of the functions were changed, only moved, and had arguments added? Also if we split the functions, does it not also make sense to split the tests into separate files?

cg/meta/demultiplex/housekeeper_storage_functions.py
ChrOertlin: removed it now

cg/meta/demultiplex/demux_post_processing.py
ChrOertlin: You are absolutely right, I fixed this!

ChrOertlin:
> Nice 👍 I'm guessing none of the functions were changed, only moved, and had arguments added? Also if we split the functions, does it not also make sense to split the tests into separate files?

Yes, I think they should should I will address this in a separate PR

ChrOertlin:
stage:
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... 
done.
Log deploy... done.
cg, version 40.2.2

prod:
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... 
done.
Log deploy... done.
cg, version 40.2.2

PR #2276, opened 2023-07-21
Check if sample exists before creating bundle

seallard:
## Description
This PR fixes the issue where empty bundles are created for samples in housekeeper in the new post processing flow.

### Fixed
- Creation of empty samples bundles in housekeeper


Vince-janv (approved):
Nice, this was cleaner than i imagined 🌟

PR #2280, opened 2023-07-24
Fix fastq file pattern bug

seallard:
## Description
This PR fixes a bug where we match too many sample fastq files and store them erroneously in the same sample bundle in housekeeper in the new post processing flow for flow cells.

### Fixed
- Bug adding too many fastq files to a sample bundle in housekeeper


Vince-janv (approved):
🚒

seallard:
Deployed to prod

PR #2281, opened 2023-07-24
Fix sample fastq name validation

seallard:
## Description
This PR fixes the validation checks for whether a sample fastq file really belongs to a specific sample id. 

This validation step is probably redundant given that we fixed the regex pattern used to retrieve the sample fastq files. So let me know if you think it should be discarded.

### Fixed
- Fix validation of sample fastq files ensuring they belong to a specific sample


Vince-janv (approved):
Nice to also have it adressed here 👍

seallard:
Deployed to stage and prod.

PR #2290, opened 2023-07-31
Rename trailblazer api to client

seallard:
## Description
This PR changes the name of the `TrailblazerAPI` to `TrailblazerClient`. Calling it an API is misleading since it is a client consuming an api.

Changes all instances of `TrailblazerAPI` to `TrailblazerClient`.

### Fixed
- Naming of Trailblazer client

### DIFF
--- a/cg/cli/clean.py
+++ b/cg/cli/clean.py
@@ -9,7 +9,7 @@ from cg.apps.demultiplex.demultiplex_api import DemultiplexingAPI
 from cg.apps.housekeeper.hk import HousekeeperAPI
 from cg.apps.scout.scout_export import ScoutExportCase
 from cg.apps.scout.scoutapi import ScoutAPI
-from cg.apps.tb import TrailblazerAPI
+from cg.apps.tb import TrailblazerClient
 from cg.cli.workflow.commands import (
     balsamic_past_run_dirs,
     balsamic_pon_past_run_dirs,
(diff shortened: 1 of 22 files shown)


cg/cli/clean.py
```diff
@@ -267,7 +267,7 @@ def remove_invalid_flow_cell_directories(context: CGConfig, failed_only: bool, d
     status_db: Store = context.status_db
     demux_api: DemultiplexingAPI = context.demultiplex_api
     housekeeper_api: HousekeeperAPI = context.housekeeper_api
-    trailblazer_api: TrailblazerAPI = context.trailblazer_api
+    trailblazer_api: TrailblazerClient = context.trailblazer_api
```
ChrOertlin, line 270:
```suggestion
    trailblazer_client: TrailblazerClient = context.trailblazer_client
```

ChrOertlin (commented):
It seems the changes only  address the Class name from `TrailblazerAPI` to `TrailblazerClient` but not the variables in the code such as `tb_api, `trailblazer_api` etc.  If we make such a change we should be consistent with this throughout.

cg/cli/clean.py
seallard: Yes 👍  Lets do it in a separate PR, this one is only intended to rename the class

seallard: https://github.com/Clinical-Genomics/cg/pull/2291

seallard:
> It seems the changes only address the Class name from `TrailblazerAPI` to `TrailblazerClient` but not the variables in the code such as `tb_api, `trailblazer_api` etc. If we make such a change we should be consistent with this throughout.

👍 Created a separate PR renaming all variables as well: https://github.com/Clinical-Genomics/cg/pull/2291

ChrOertlin (approved):
Looks good together with the other PR!

PR #2294, opened 2023-08-01
fix(uncouple statusDB Sample reads from CGstats)

ChrOertlin:
## Description

This PR re-routes the data flow from cgstats - cg Sample.reads to cg SampleLaneSequencingMetrics to cg Sample.reads

in addition the temporary column Sample.caluclated_read_counts will be dropped.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


ChrOertlin:
> Nice 👍 At what point will we remove the old flow? Also I'm guessing this PR is not getting merged right away?

The calculated counts are validated with the counts we get from cgstats. If we compare side by side, there are no differences between the values. So we can decouple this now.

cg/meta/transfer/flowcell.py
```diff
@@ -110,8 +110,8 @@ def _parse_flow_cell_samples(
                     sequencing_files=cgstats_sample.fastqs,
                     tag_name=SequencingFileTag.FASTQ,
                 )
-
-            status_db_sample.reads = cgstats_sample.reads
+            # commenting out for now - just for traceability. This code will be removed later.
```
seallard, line 113: Oh no zombie code! 🧟

seallard: I guess this entire module is supposed to be nuked so I'm not sure if we should touch it

seallard (approved):
👍 Maybe we don't need to touch the transfer api at all

cg/meta/transfer/flowcell.py
ChrOertlin: yes you are right we have the PRs that keep track of code, I should remove it.

ChrOertlin:
> 👍 I guess we should consider whether there are any cases where the new flow is not run in association with the old flow since that would mean the reads would not be updated for the sample. I can't think of any now.
> 
> We have the force flag implemented and no crashes are occurring for the new flow when run after the old. What about the current manual practice when a sample is assigned the wrong index? Could there be some issue with logic that is only run if the flow cell does not already exist in the new flow?
> 
> Sorry for the vague questions, trying to catch any potential issues with this
> 
> EDIT: After going through the flow again I don't think there should be any issues

We update the counts outside of the flow cell logic, so it should happen any time this command is run. 
Ideally we do not require manual intervention when cgstats is deprecated if I understood the reason correctly why we currently have manual intervention going on.

ChrOertlin:
# Calculcated_read_counts column is gone
image

# Sample gets assigned reads in the reads column when running cg demultiplex finish temporary cmd
image

# cg demultiplex finish temporary runs without issues

ChrOertlin:
stage:
Logging deploy ...
Getting deployer... done.
done.
Log deploy... done.
cg, version 40.8.2

prod:

Issue #2293, opened 2023-08-01
missing(new demux post processing misses sample flow cell links)


ChrOertlin:
## Description

Currently the new demux post processing workflow does not establish the link between flow cells and samples as is done in the old workflow:

```Python 
 def _parse_flow_cell_samples(
        self,
        cgstats_flow_cell: StatsFlowcell,
        flow_cell: Flowcell,
        flow_cell_id: str,
        store: bool,
    ) -> None:
        """Adds fastq to Housekeeper, set sequenced at for sample, add samples to flow cell."""
        for cgstats_sample in cgstats_flow_cell.samples:
            LOG.debug(f"Adding reads/FASTQs to sample: {cgstats_sample.name}")

            status_db_sample: Sample = self.db.get_sample_by_internal_id(
                internal_id=cgstats_sample.name
            )

            if not status_db_sample:
                LOG.warning(f"Unable to find sample: {cgstats_sample.name}")
                continue

            if store:
                self._store_sequencing_files(
                    flow_cell_id=flow_cell_id,
                    sample_id=status_db_sample.internal_id,
                    sequencing_files=cgstats_sample.fastqs,
                    tag_name=SequencingFileTag.FASTQ,
                )

            status_db_sample.reads = cgstats_sample.reads

            _set_status_db_sample_sequenced_at(
                status_db_sample=status_db_sample, flow_cell_sequenced_at=flow_cell.sequenced_at
            )

            if isinstance(status_db_sample, Sample):
                flow_cell.samples.append(status_db_sample)

            log_enough_reads(
                status_db_sample_reads=status_db_sample.reads,
                application_expected_reads=status_db_sample.application_version.application.expected_reads,
                cgstats_sample_name=cgstats_sample.name,
            )
```

See:

```Python
            if isinstance(status_db_sample, Sample):
                flow_cell.samples.append(status_db_sample)
```

## Suggested solution

Create flow cell sample links in statusDB like in the old work flow

## This can be closed when

When the flow cell sample links are established in the new work flow.

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

Issue #2292, opened 2023-08-01
Flowcell - Sample connection was not created when the sample recieved 0 reads


Karl-Svard:
## Description

It seems like the flowcell - sample association in cg wasn't created between `HGN7YDSX7` and `ACC12418A13`. I assume this might be because it had 0 reads across all lanes as a result of an incorrect index pair being registered in LIMS (see deviation https://github.com/Clinical-Genomics/Deviations/issues/548). 

I ran the re-demux outside the normal demultiplexing folder to avoid it being picked up by the automation as we still don't want to delete data from cgstats. After this I linked the fastq manually to housekeeper but the analysis still could not be started as it didn't find any connected flow cells:
```
[hiseq.clinical@hasta:/home/proj/production/rare-disease/cases] [P_base] 2 $ cg workflow mip-dna start poeticpuma
Case poeticpuma exists in Status DB
Starting full MIP analysis workflow for case poeticpuma
Case poeticpuma exists in Status DB
No flow cells found
Analysis cannot be started: all flow cells need to be on disk to run the analysis
```

The issue was fixed by manually updating the database:
```
INSERT INTO flowcell_sample (flowcell_id, sample_id)
SELECT flowcell.id, sample.id
FROM flowcell, sample
WHERE flowcell.name IN ('HGN7YDSX7') AND sample.internal_id IN ('ACC12418A13');
```

## Suggested solution

Investigate what the cause of this could be. Haven't seen it behaving like this in the past when fixing index issues.

## This can be closed when

*Describe what needs to be done for this issue to be closed*

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

ChrOertlin:
Can you describe what you mean by "re-demux outside the normal demultiplexing folder"?

henrikstranneheim:
@ChrOertlin Done?

ChrOertlin:
Yes, this was fixed in the attached PR, closing

PR #2295, opened 2023-08-01
add(samples to flow cell in new demux pp workflow)

ChrOertlin:
## Description

This PR adds functionality to add samples to flow cells in the new demux post processing work flow. 

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -33,13 +33,34 @@ def store_flow_cell_data_in_status_db(
...
+        sample_sheet_path=parsed_flow_cell.sample_sheet_path,
+        flow_cell_sample_type=parsed_flow_cell.sample_type,
+    )
+    samples: List[Sample] = [
+        store.get_sample_by_internal_id(sample_internal_id)
+        for sample_internal_id in sample_internal_ids
+    ]
+    for sample in samples:
+        if sample:
+            flow_cell.samples.append(sample)
```
seallard, line 60: What if the sample is already present on the flow cell?

seallard (approved):
Nice 👍 Is this something we missed from the old flow?

cg/meta/demultiplex/status_db_storage_functions.py
seallard: I think we need to add a check whether the sample already is present on the flow cell to avoid duplicates

ChrOertlin:
> Nice 👍 Is this something we missed from the old flow?

https://github.com/Clinical-Genomics/cg/issues/2293

Vince-janv:
I was under the impression the link table between sample and flow cell would be removed when removing the old flow. We can achieve the same thing with a property on sample and flow cell which looks at the sample_lane_sequencing_metrics and returns flow cells and samples respectively. I do not like the idea of having two different links between samples and flow cells 😬

ChrOertlin:
I agree. But there is currently logic in mip-dna that appears to rely on this relationship between flow cells and samples. See issue: https://github.com/Clinical-Genomics/cg/issues/2292 

We should look at this behaviour and perhaps refine in. For now, I think it makes sense to mimic the old flow as much as possible. 

What are your thoughts?

cg/meta/demultiplex/status_db_storage_functions.py
seallard: If you feel up for it, the integration tests could be updated to assert that the flow cell has been associated with the samples

ChrOertlin:
> > I agree. But there is currently logic in mip-dna that appears to rely on this relationship between flow cells and samples. See issue: #2292
> > We should look at this behaviour and perhaps refine it. Possibly also get rid of the flowcell_sample table if it is not needed anymore. For now, I think it makes sense to mimic the old flow as much as possible.
> > What are your thoughts?
> 
> I think we can even get away with only adding this and not changing anything else:
> 
> ```python
> class Sample:
> 
>     @property
>     def flowcells(self):
>         return list({self.sequencing_metrics.flow_cell})
> 
> class Flowcell:
>     @property
>     def samples(self):
>         return list({self.sequencing_metrics.sample})
> ```

Oh, this would be elegant. I will try!

cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -33,13 +33,38 @@ def store_flow_cell_data_in_status_db(
             sequencer_name=parsed_flow_cell.machine_name,
             sequenced_at=parsed_flow_cell.run_date,
         )
+        flow_cell = add_samples_to_flow_cell_in_status_db(
```
seallard, line 36: I think it is more readable to not nest function calls:
```python
           sample_internal_ids = get_sample_internal_ids_from_sample_sheet(
                sample_sheet_path=parsed_flow_cell.sample_sheet_path,
                flow_cell_sample_type=parsed_flow_cell.sample_type,
            )
        flow_cell = add_samples_to_flow_cell_in_status_db(
            flow_cell_name=flow_cell.name,
            sample_internal_ids=sample_internal_ids,
            store=store,
        )
```

seallard: With the current call, it is necessary to understand several at once:

```python
        flow_cell = add_samples_to_flow_cell_in_status_db(
            flow_cell_name=flow_cell.name,
            sample_internal_ids=get_sample_internal_ids_from_sample_sheet(
                sample_sheet_path=parsed_flow_cell.sample_sheet_path,
                flow_cell_sample_type=parsed_flow_cell.sample_type,
            ),
            store=store,
        )
```

seallard (approved):
Made a second review because I feel like I missed some suff in the first round, looks good 👍 

We can also adress the comments in separate PR:s where we clean up the flow a bit

cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -33,13 +33,38 @@ def store_flow_cell_data_in_status_db(
...
+            flow_cell_name=flow_cell.name,
+            sample_internal_ids=get_sample_internal_ids_from_sample_sheet(
+                sample_sheet_path=parsed_flow_cell.sample_sheet_path,
+                flow_cell_sample_type=parsed_flow_cell.sample_type,
+            ),
+            store=store,
```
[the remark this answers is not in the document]
ChrOertlin: Checking the store for the flow cell did not make sense there since it is the process of adding it. I changed it so it passes the flow cell now. 

We need the store the retrieve the samples to append to the flow cell.

cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -33,13 +33,38 @@ def store_flow_cell_data_in_status_db(
...
         LOG.info(f"Flow cell added to status db: {parsed_flow_cell.id}.")
     else:
         LOG.info(f"Flow cell already exists in status db: {parsed_flow_cell.id}. Skipping.")
 
 
+def add_samples_to_flow_cell_in_status_db(
+    flow_cell_name: str, sample_internal_ids: List[str], store: Store
+) -> Flowcell:
+    """Adds samples to a flow cell in status db."""
+    flow_cell: Flowcell = store.get_flow_cell_by_name(flow_cell_name=flow_cell_name)
```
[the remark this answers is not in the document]
ChrOertlin: I don't think so I changed the logic

ChrOertlin:
# cg demultiplex finish temporary runs successfully

New entries are added to flowcell_sample table
image

ChrOertlin:
Stage:
Logging deploy ...
Getting deployer... done.
done.
Log deploy... done.
cg, version 40.6.2

Prod:
Logging deploy ...
Getting deployer... done.
done.
Log deploy... done.
cg, version 40.6.2

PR #2298, opened 2023-08-03
Create parent models for archiving

Vince-janv:
## Description
To allow for easy integration of other archiving locations and software, this PR creates parent models for the handlers to inherit from so as to have common methods and signatures. It also adds the method `from_models` which instantiates the model with different input than the normal one for BaseModels.

### Added

- Method to initiate a file model with Sample and File as input
- Interface models to inherit from

### Changed

- Naming of old models

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/archive/ddn_dataflow.py
+++ b/cg/meta/archive/ddn_dataflow.py
@@ -129,7 +129,7 @@ class AuthResponse(BaseModel):
     refresh: Optional[str]
 
 
-class DDNDataFlowApi(ArchiveHandler):
+class DDNDataFlowClient(ArchiveHandler):
     """Class for archiving and retrieving folders via DDN Dataflow."""
 
     def __init__(self, config: DDNDataFlowConfig):
--- a/cg/meta/archive/ddn_dataflow.py
+++ b/cg/meta/archive/ddn_dataflow.py
@@ -10,7 +10,7 @@ from datetime import datetime
 from cg.constants.constants import APIMethods, FileFormat
 from cg.exc import DdnDataflowAuthenticationError
 from cg.io.controller import APIRequest, ReadStream
-from cg.meta.archive.models import ArchiveHandler, FileTransferData
+from cg.meta.archive.models import ArchiveInterface, ArchiveFile
 from cg.models.cg_config import DDNDataFlowConfig
 from pydantic.v1 import BaseModel
 from requests.models import Response
@@ -42,7 +42,7 @@ class ResponseFields(str, Enum):
     RETRIEVE_FILES = "files/retrieve"
 
 
-class DataFlowFileTransferData(FileTransferData):
+class MiriaFile(ArchiveFile):
     """Model for representing a singular object transfer."""
 
     _metadata = None
@@ -50,7 +50,7 @@ class DataFlowFileTransferData(FileTransferData):
     source: str
 
     @classmethod
-    def from_models(cls, file: File, sample: Sample):
+    def from_file_and_sample(cls, file: File, sample: Sample) -> "MiriaFile":
         """Instantiates the class from a File and Sample object."""
         return cls(destination=sample.internal_id, source=file.path)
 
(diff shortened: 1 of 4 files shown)


cg/meta/archive/ddn_dataflow.py
```diff
@@ -38,13 +42,18 @@ class ResponseFields(str, Enum):
     RETRIEVE_FILES = "files/retrieve"
 
 
-class TransferData(BaseModel):
+class DataFlowFileTransferData(FileTransferData):
```
Vince-janv, line 45: Please help with the naming... This is a model representing a single file and containing the information needed for the archiving, ie where the file is, any metadata and where to put it

cg/meta/archive/models.py
```diff
@@ -0,0 +1,30 @@
+"""Contains base models to be inherited from in other archive software files."""
+from typing import List
+
+from housekeeper.store.models import File
+from pydantic.v1 import BaseModel
+
+from cg.store.models import Sample
+
+
+class FileTransferData(BaseModel):
```
seallard, line 10: Is it misleading to just call this `ArchiveData`?

cg/meta/archive/ddn_dataflow.py
seallard: Some options:
- `DDNTransferData`
- `DDNArchiveData`

seallard: Maybe the base model can be called `TransferData`, `TransferTask`. Naming is hard 😅

islean: I think in DataFlow is a bit superfluous here given the naming of the class. Maybe KarolinskaFileData, KarolinskaFile or FileData would be sufficient?

cg/meta/archive/ddn_dataflow.py
```diff
@@ -38,13 +42,18 @@ class ResponseFields(str, Enum):
...
     """Model for representing a singular object transfer."""
 
     _metadata = None
     destination: str
     source: str
 
+    @classmethod
+    def from_models(cls, file: File, sample: Sample):
```
islean, line 53: from_file_and_sample maybe is clearer given that there are many models within cg?

islean (approved):
Nice work 🔥 Just a few naming suggestions but apart from that everything looks great.

cg/meta/archive/models.py
```diff
@@ -0,0 +1,30 @@
...
+    """Base class for classes representing files to be archived."""
+
+    @classmethod
+    def from_models(cls, file: File, sample: Sample):
```
seallard, line 14: Missing return tyoe

seallard (approved):
Nice 👍  Naming is hard, made some suggestions. Also, I think the base classes could be marked as abstract to convey the design more clearly (if the intention is that the base class never will be instantiated).

Out of curiosity, how many concrete implementations of ArchiveHandler will we have?

cg/meta/archive/models.py
```diff
@@ -0,0 +1,30 @@
...
+        """Instantiates the class from a File and Sample object."""
+        raise NotImplementedError
+
+
+class ArchiveHandler:
+    """Base class for classes handling different archiving programs."""
```
seallard, line 20: Could potentially be called `ArchiveHandlerInterface` if we want to make it super clear that is an abstract base class (the closest thing Python has to interfaces)

cg/meta/archive/ddn_dataflow.py
ChrOertlin: ```FileTransferInfo``` could also be applicable since it contains specifics as the where and what and not just data, but ```TransferData``` is sufficient also.

cg/meta/archive/models.py
Vince-janv: Definitely better with return type 👍

cg/meta/archive/models.py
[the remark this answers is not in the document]
Vince-janv: Great! It's exactly what I wanted to suggest with the classes, but now I can enforce it ⭐

cg/meta/archive/models.py
Vince-janv: Maybe just `ArchiveInterface`?

cg/meta/archive/ddn_dataflow.py
Vince-janv: After consulting with @islean we have decided to move away from using `DDN` or just `DataFlow`, and instead use Miria (actual software used). Here I then suggest MiriaFile and then refactor the rest in a separate PR

Vince-janv:
> Out of curiosity, how many concrete implementations of ArchiveHandler will we have?
@seallard I think two within a year or so, but If an other hospital would want their samples sequenced we would have to archive the files there in som way which might require a new interface.

Vince-janv:
### Deployed in production:
```
Log deploy... done.
cg, version 40.4.0
[vincent.janvid@hasta:~] [P_base] 2m52s $
```

PR #2299, opened 2023-08-03
fix(Archiving) Changed input and output types for archiving and retrieval functions

Vince-janv:
## Description
This PR cleans previous input and returns types of archive_folders and retrieve_folders. It also adds som exception handling if the requests fails

### Added

- Exception handling when calling the archive and retrieve endpoints
- Function `create_transfer_request` performing the path curation needed before sending the request
- Class `TransferJob` to represent the return of archive and retrieve posts

### Changed

- input to archive_folders & retrieve_folders changed from a dict to a list of already instantiated TransferData objects
- Return changed from just the request return, to the actual task_id

### DIFF
--- a/cg/apps/housekeeper/hk.py
+++ b/cg/apps/housekeeper/hk.py
@@ -55,7 +55,7 @@ class HousekeeperAPI:
         return new_bundle
 
     def set_to_archive(self, file: File, value: bool) -> None:
-        """Sets the 'to_archive' field of a file."""
+        """Sets the 'is_archiving_request' field of a file."""
         file.to_archive: bool = value
         self.commit()
 
--- a/cg/apps/housekeeper/hk.py
+++ b/cg/apps/housekeeper/hk.py
@@ -55,7 +55,7 @@ class HousekeeperAPI:
         return new_bundle
 
     def set_to_archive(self, file: File, value: bool) -> None:
-        """Sets the 'is_archiving_request' field of a file."""
+        """Sets the 'to_archive' field of a file."""
         file.to_archive: bool = value
         self.commit()
 
--- a/cg/meta/archive/ddn_dataflow.py
+++ b/cg/meta/archive/ddn_dataflow.py
@@ -4,14 +4,18 @@ from pathlib import Path
 from typing import Dict, List, Optional
 from urllib.parse import urljoin
 
-from pydantic import BaseModel
-from requests.models import Response
+from housekeeper.store.models import File
 
 from datetime import datetime
 from cg.constants.constants import APIMethods
 from cg.exc import DdnDataflowAuthenticationError
+from cg.meta.archive.models import ArchiveInterface, ArchiveFile
 from cg.io.controller import APIRequest
 from cg.models.cg_config import DDNDataFlowConfig
+from pydantic.v1 import BaseModel
+from requests.models import Response
+
+from cg.store.models import Sample
 
 OSTYPE: str = "Unix/MacOS"
 ROOT_TO_TRIM: str = "/home"
@@ -29,12 +33,17 @@ class DataflowEndpoints(str, Enum):
     RETRIEVE_FILES = "files/retrieve"
 
 
-class TransferData(BaseModel):
+class MiriaFile(ArchiveFile):
     """Model for representing a singular object transfer."""
 
     _metadata = None
     destination: str
     source: str
+
+    @classmethod
+    def from_file_and_sample(cls, file: File, sample: Sample) -> "MiriaFile":
+        """Instantiates the class from a File and Sample object."""
+        return cls(destination=sample.internal_id, source=file.path)
 
     def trim_path(self, attribute_to_trim: str):
         """Trims the given attribute (source or destination) from its root directory."""
@@ -53,29 +62,27 @@ class TransferData(BaseModel):
 class TransferPayload(BaseModel):
     """Model for representing a Dataflow transfer task."""
 
-    files_to_transfer: List[TransferData]
+    files_to_transfer: List[MiriaFile]
     osType: str = OSTYPE
     createFolder: bool = False
 
     def trim_paths(self, attribute_to_trim: str):
         """Trims the source path from its root directory for all objects in the transfer."""
-        for transfer_data in self.files_to_transfer:
-            transfer_data.trim_path(attribute_to_trim=attribute_to_trim)
+        for miria_file in self.files_to_transfer:
+            miria_file.trim_path(attribute_to_trim=attribute_to_trim)
 
     def add_repositories(self, source_prefix: str, destination_prefix: str):
         """Prepends the given repositories to the source and destination paths all objects in the
         transfer."""
-        for transfer_data in self.files_to_transfer:
-            transfer_data.add_repositories(
+        for miria_file in self.files_to_transfer:
+            miria_file.add_repositories(
                 source_prefix=source_prefix, destination_prefix=destination_prefix
             )
 
     def model_dump(self, **kwargs) -> dict:
         """Creates a correctly structured dict to be used as the request payload."""
         payload: dict = super().model_dump(exclude={"files_to_transfer"})
-        payload["pathInfo"] = [
-            transfer_data.model_dump() for transfer_data in self.files_to_transfer
-        ]
+        payload["pathInfo"] = [miria_file.model_dump() for miria_file in self.files_to_transfer]
         payload["metadataList"] = []
         return payload
 
@@ -125,7 +132,7 @@ class TransferJob(BaseModel):
     job_id: int
 
 
-class DDNDataFlowApi:
+class DDNDataFlowClient(ArchiveInterface):
     """Class for archiving and retrieving folders via DDN Dataflow."""
 
     def __init__(self, config: DDNDataFlowConfig):
@@ -183,20 +190,19 @@ class DDNDataFlowApi:
             self._refresh_auth_token()
         return {"Authorization": f"Bearer {self.auth_token}"}
 
-    def archive_folders(self, sources_and_destinations: List[TransferData]) -> int:
+    def archive_folders(self, sources_and_destinations: List[MiriaFile]) -> int:
         """Archives all folders provided, to their corresponding destination, as given by sources
         and destination in TransferData. Returns the job ID of the archiving task."""
         transfer_request: TransferPayload = self.create_transfer_request(
             sources_and_destinations, is_archiving_request=True
         )
-
         job_id: int = transfer_request.post_request(
             headers=dict(self.headers, **self.auth_header),
             url=urljoin(base=self.url, url=DataflowEndpoints.ARCHIVE_FILES),
         )
         return job_id
 
-    def retrieve_folders(self, sources_and_destinations: List[TransferData]) -> int:
+    def retrieve_folders(self, sources_and_destinations: List[MiriaFile]) -> bool:
         """Retrieves all folders provided, to their corresponding destination, as given by sources
         and destination in TransferData. Returns the job ID of the retrieval task."""
         transfer_request: TransferPayload = self.create_transfer_request(
@@ -209,7 +215,7 @@ class DDNDataFlowApi:
         return job_id
 
     def create_transfer_request(
-        self, sources_and_destinations: List[TransferData], is_archiving_request: bool
+        self, sources_and_destinations: List[MiriaFile], is_archiving_request: bool
     ) -> TransferPayload:
         """Performs the necessary curation of paths for the request to be valid, depending on if
         it is an archiving or a retrieve request.
--- a/cg/meta/archive/ddn_dataflow.py
+++ b/cg/meta/archive/ddn_dataflow.py
@@ -86,7 +86,7 @@ class TransferPayload(BaseModel):
         payload["metadataList"] = []
         return payload
 
-    def post_request(self, url: str, headers: dict) -> bool:
+    def post_request(self, url: str, headers: dict) -> Response:
         """Sends a request to the given url with, the given headers, and its own content as
         payload."""
         return APIRequest.api_request_from_content(
@@ -94,7 +94,7 @@ class TransferPayload(BaseModel):
             url=url,
             headers=headers,
             json=self.dict(),
-        ).ok
+        )
 
 
 class AuthPayload(BaseModel):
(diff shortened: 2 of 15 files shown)


cg/meta/archive/ddn_dataflow.py
```diff
@@ -186,34 +186,39 @@ def auth_header(self) -> Dict[str, str]:
...
         )
-        return transfer_request.post_request(
+        response: Response = transfer_request.post_request(
             headers=dict(self.headers, **self.auth_header),
             url=urljoin(base=self.url, url=DataflowEndpoints.ARCHIVE_FILES),
         )
+        if response.ok:
+            return response.json()["job_id"]
+        else:
+            raise IOError(response.text)
```
ChrOertlin, line 206: duplicated code below consider suggestion to adhere to DRY:
```Python

return get_response_job_id(response)

def get_response_job_id(reponse):
if response.ok:
            return response.json()["job_id"]
        else:
            raise IOError(response.text)

File: cg/meta/archive/ddn_dataflow.py [hunk 2]
```diff
@@ -186,34 +186,39 @@ def auth_header(self) -> Dict[str, str]:
...
         )
-        return transfer_request.post_request(
+        response: Response = transfer_request.post_request(
             headers=dict(self.headers, **self.auth_header),
             url=urljoin(base=self.url, url=DataflowEndpoints.RETRIEVE_FILES),
         )
+        if response.ok:
+            return response.json()["job_id"]
+        else:
+            raise IOError(response.text)
```
ChrOertlin, line 224:
```suggestion
 return get_response_job_id(response)
```

cg/meta/archive/ddn_dataflow.py
```diff
@@ -186,34 +186,39 @@ def auth_header(self) -> Dict[str, str]:
...
         return {"Authorization": f"Bearer {self.auth_token}"}
 
-    def archive_folders(self, sources_and_destinations: Dict[Path, Path]) -> bool:
-        """Archives all folders provided, to their corresponding destination, as given by sources and destination parameter."""
-        transfer_data: List[TransferData] = [
-            TransferData(source=source.as_posix(), destination=destination.as_posix())
-            for source, destination in sources_and_destinations.items()
-        ]
-        transfer_request: TransferPayload = TransferPayload(files_to_transfer=transfer_data)
+    def archive_folders(self, sources_and_destinations: List[TransferData]) -> int:
```
ChrOertlin, line 189: Function name is misleading with the new implementation as now it returns a job id in addition to starting the transfer request.

cg/meta/archive/ddn_dataflow.py
```diff
@@ -186,34 +186,39 @@ def auth_header(self) -> Dict[str, str]:
...
+        """Archives all folders provided, to their corresponding destination,
+        as given by sources and destination parameter."""
```
ChrOertlin, line 191:
```suggestion
        """Archives all folders provided, to their corresponding destination,
        as given by sources and destination parameter.
        Returns the job id associated to the transfer request."""
```

cg/meta/archive/ddn_dataflow.py
```diff
@@ -186,34 +186,39 @@ def auth_header(self) -> Dict[str, str]:
...
             url=urljoin(base=self.url, url=DataflowEndpoints.ARCHIVE_FILES),
         )
+        if response.ok:
+            return response.json()["job_id"]
+        else:
+            raise IOError(response.text)
 
-    def retrieve_folders(self, sources_and_destinations: Dict[Path, Path]) -> bool:
+    def retrieve_folders(self, sources_and_destinations: List[TransferData]) -> int:
         """Retrieves all folders provided, to their corresponding destination, as given by the sources and destination parameter."""
```
ChrOertlin, line 209:
```suggestion
        """Retrieves all folders provided, to their corresponding destination, as given by the sources and destination parameter. Returns a job id associated to the transfer request."""
```

tests/meta/archive/test_archiving.py
```diff
@@ -280,11 +287,9 @@ def test_archive_folders(
...
-        return_value=ok_response,
+        return_value=ok_ddn_response,
     ) as mock_request_submitter:
-        response: bool = ddn_dataflow_api.archive_folders(
-            {Path(local_directory): Path(remote_path)}
-        )
+        response: int = ddn_dataflow_api.archive_folders([transfer_data_archive])
 
     # THEN a boolean response should be returned
     assert response
```
ChrOertlin, line 295:
```suggestion
    # THEN a job id response should be returned
    assert response = ok_ddn_response{"job_id"}
```
right?

tests/meta/archive/test_archiving.py
```diff
@@ -305,25 +310,20 @@ def test_archive_folders(
...
-        return_value=ok_response,
+        APIRequest, "api_request_from_content", return_value=ok_ddn_response
     ) as mock_request_submitter:
-        response: bool = ddn_dataflow_api.retrieve_folders(
-            {Path(remote_path): Path(local_directory)}
-        )
+        response: int = ddn_dataflow_api.retrieve_folders([transfer_data_retrieve])
 
     # THEN the response returned should be true
     assert response
```
ChrOertlin, line 329:
```suggestion
    # THEN the response returned should be the job id
    assert response = ok_ddn_response{"job_id"}
```
similar to above I suppose.

ChrOertlin (approved):
Looks nice, some suggestions.

cg/meta/archive/ddn_dataflow.py
seallard, line 206: There is a `response.raise_for_status()` which raises a `HTTPError` exception if the response has a non successful response code. What do you think?

```python
        response: Response = transfer_request.post_request(
            headers=dict(self.headers, **self.auth_header),
            url=urljoin(base=self.url, url=DataflowEndpoints.ARCHIVE_FILES),
        )
        response.raise_for_status()
        return response.json()["job_id"]
```

cg/meta/archive/ddn_dataflow.py
```diff
@@ -186,34 +186,39 @@ def auth_header(self) -> Dict[str, str]:
...
         transfer_request.add_repositories(
             source_prefix=self.local_storage, destination_prefix=self.archive_repository
         )
-        return transfer_request.post_request(
+        response: Response = transfer_request.post_request(
             headers=dict(self.headers, **self.auth_header),
             url=urljoin(base=self.url, url=DataflowEndpoints.ARCHIVE_FILES),
         )
+        if response.ok:
+            return response.json()["job_id"]
```
seallard, line 204: Is it possible that the `"job_id"` key is missing and what happens then? Should we handle the `KEyError`?

islean (approved):
Looks nice!

cg/meta/archive/ddn_dataflow.py
islean: Would it make it clearer if the response was a Response object? We somehow need to get the job_id but I think it would be worse having "archive_folders_and_get_job_id" as a name than to just return the job_id or response with the current naming.

cg/meta/archive/ddn_dataflow.py
```diff
@@ -186,34 +186,39 @@ def auth_header(self) -> Dict[str, str]:
...
         transfer_request.trim_paths(attribute_to_trim=DESTINATION_ATTRIBUTE)
         transfer_request.add_repositories(
             source_prefix=self.archive_repository, destination_prefix=self.local_storage
         )
-        return transfer_request.post_request(
+        response: Response = transfer_request.post_request(
             headers=dict(self.headers, **self.auth_header),
             url=urljoin(base=self.url, url=DataflowEndpoints.RETRIEVE_FILES),
         )
+        if response.ok:
```
seallard, line 221: Consider using the `response.raise_for_status()`

cg/meta/archive/ddn_dataflow.py
islean: The only thing in the response should be a job id pretty much, given that the request was handled successfully. So I have a hard time seeing this happening, but I guess a null check could be added if we feel uncertain.

cg/meta/archive/ddn_dataflow.py
```diff
@@ -186,34 +186,39 @@ def auth_header(self) -> Dict[str, str]:
...
 
-    def retrieve_folders(self, sources_and_destinations: Dict[Path, Path]) -> bool:
+    def retrieve_folders(self, sources_and_destinations: List[TransferData]) -> int:
         """Retrieves all folders provided, to their corresponding destination, as given by the sources and destination parameter."""
-        transfer_data: List[TransferData] = [
-            TransferData(source=source.as_posix(), destination=destination.as_posix())
-            for source, destination in sources_and_destinations.items()
-        ]
-        transfer_request: TransferPayload = TransferPayload(files_to_transfer=transfer_data)
+        transfer_request: TransferPayload = TransferPayload(
```
seallard, line 210: Consider extracting a separate method for creating the `transfer_request` here. Something like:
```python
def retrieve_folders(self, sources_and_destinations: List[TransferData]) -> int:
    transfer_request: TransferRequest = self.create_transfer_request(sources_and_destinations)
     
     response: Response = transfer_request.post_request(
         headers=dict(self.headers, **self.auth_header),
         url=urljoin(base=self.url, url=DataflowEndpoints.RETRIEVE_FILES),
      )
        
      response.raise_for_status()
      return response.json()["job_id"]
```

cg/meta/archive/ddn_dataflow.py
```diff
@@ -186,34 +186,39 @@ def auth_header(self) -> Dict[str, str]:
...
         transfer_request.add_repositories(
             source_prefix=self.archive_repository, destination_prefix=self.local_storage
         )
-        return transfer_request.post_request(
+        response: Response = transfer_request.post_request(
             headers=dict(self.headers, **self.auth_header),
             url=urljoin(base=self.url, url=DataflowEndpoints.RETRIEVE_FILES),
         )
+        if response.ok:
+            return response.json()["job_id"]
```
seallard, line 222: Could the `"job_id"` key be missing here and should it be handled?

seallard (approved):
Nice clean up 🧹  Some minor questions to consider

cg/meta/archive/ddn_dataflow.py
seallard: The `response.raise_for_status()` could also be moved into the method sending the request, that might be even cleaner. That way we won't have to raise the exceptions everywhere

cg/meta/archive/ddn_dataflow.py
ChrOertlin: Yes, I think mentioning it in the docstring should be fine in this case.

Vince-janv: I think archiving a folder is what the function does. The return is just to keep track of it. I Think clarifying the return in the doc string is enough

cg/meta/archive/ddn_dataflow.py
Vince-janv: Clarified doc string 👍

cg/meta/archive/ddn_dataflow.py
Vince-janv: Added a BaseModel to parse the response

cg/meta/archive/ddn_dataflow.py
Vince-janv: Moved that part into the post request function which avoids the duplication 👍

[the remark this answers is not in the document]
Vince-janv: I think returning the job id is the clearest here. I added `raise_for_status` which @seallard suggested and it print includes `response.reason` in the error message

Vince-janv: Good idea! Implemented

cg/meta/archive/ddn_dataflow.py
Vince-janv: Clarified doc string

cg/meta/archive/ddn_dataflow.py
Vince-janv: Moved `response.raise_for_status()` into `post_request`. Will look inte extracting the creation into a method

tests/meta/archive/conftest.py
```diff
@@ -55,16 +61,24 @@ def fixture_ddn_dataflow_api(ddn_dataflow_config: DDNDataFlowConfig) -> DDNDataF
...
+
 @pytest.fixture(name="transfer_payload")
-def fixture_transfer_payload(transfer_data: TransferData) -> TransferPayload:
-    """Return a TransferPayload object containing two identical TransferData object.."""
-    return TransferPayload(files_to_transfer=[transfer_data, transfer_data.copy(deep=True)])
+def fixture_transfer_payload(transfer_data_archive: TransferData) -> TransferPayload:
+    """Return a TransferPayload object containing two identical DataFlowFileTransferData object."""
+    return TransferPayload(
+        files_to_transfer=[transfer_data_archive, transfer_data_archive.copy(deep=True)]
+    )
```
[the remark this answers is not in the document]
Vince-janv: I believe I wanted the fixture to be as accurate as possible and include more than one object. A copy was just the easiest way of doing it. Maybe remove that from the doc string? Do you find it confusing?

tests/meta/archive/test_archiving.py
```diff
@@ -25,50 +25,58 @@
...
-    local_directory: Path, transfer_data: TransferData, trimmed_local_directory: Path
+    local_directory: Path,
+    transfer_data_archive: TransferData,
+    trimmed_local_directory: Path,
 ):
     """Tests the method for trimming the source directory."""
 
     # GIVEN a source path and a destination path
 
     # WHEN creating the correctly formatted dictionary
```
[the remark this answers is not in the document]
Vince-janv: Nope, good catch. Going trough these and correcting them 👍

tests/meta/archive/test_archiving.py
```diff
@@ -25,50 +25,58 @@
...
     # THEN the repositories should be prepended to the paths
-    assert transfer_data.source == ddn_dataflow_config.local_storage + str(local_directory)
-    assert transfer_data.destination == ddn_dataflow_config.archive_repository + str(remote_path)
+    assert transfer_data_archive.source == ddn_dataflow_config.local_storage + str(local_directory)
+    assert transfer_data_archive.destination == ddn_dataflow_config.archive_repository + str(
+        remote_path
+    )
 
 
 def test_transfer_payload_dict(transfer_payload: TransferPayload):
```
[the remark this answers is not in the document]
Vince-janv: Some confusion regarding these tests. Kept the name and re-wrote the doc string for `test_transfer_payload_correct_source_root` and `test_transfer_payload_correct_destination_root`

tests/meta/archive/test_archiving.py
Vince-janv: Indeed, fixed assert and comment

tests/meta/archive/test_archiving.py
Vince-janv: Fixed assert and comment

cg/apps/housekeeper/hk.py
```diff
@@ -55,7 +55,7 @@ def create_new_bundle_and_version(self, name: str) -> Bundle:
         return new_bundle
 
     def set_to_archive(self, file: File, value: bool) -> None:
-        """Sets the 'to_archive' field of a file."""
+        """Sets the 'is_archiving_request' field of a file."""
```
seallard, line 58: This doc string is confusing to me, maybe it can be removed?

seallard: A name alternative: `mark_file_as_ready_to_archive`

cg/meta/archive/ddn_dataflow.py
```diff
@@ -86,15 +77,24 @@ def dict(self, **kwargs) -> dict:
...
+        """
+        response: Response = APIRequest.api_request_from_content(
             api_method=APIMethods.POST,
             url=url,
             headers=headers,
             json=self.dict(),
-        ).ok
+        )
+        response.raise_for_status()
+        parsed_response = TransferResponse(
```
seallard, line 92: I think Pydantic has some nice functionality for parsing the response body into a data model. Just an alternative to consider which might be easier to read, something like
```python
TransferResponse.mode_validate(response.json())
```

seallard: Also, maybe an alternative name for the model could just be `TransferJob` instead of `TransferResponse`? I think it might be confusing to include `Response` in the model name if we settle on using a pattern where we return a data model from the clients

cg/apps/housekeeper/hk.py
Vince-janv: This change was made accidentally, due to a refactoring 😅  Removing

tests/meta/archive/conftest.py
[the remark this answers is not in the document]
islean: No, the doc string was clear! I just thought it would be nice to have two different entries that can be used for different purposes. But for this PR you can keep it as is :)

cg/meta/archive/ddn_dataflow.py
Vince-janv: Implemented. It was only added in pydantic v2 though so I hade to make some other changes as well

islean:
Test ran - deployment did not break and since no new functionality which is to be used now is added, I (IO) declare that it passed.

PR #2307, opened 2023-08-08
Archiving make spring archive api agnostic

islean:
## Description

This PR aims to address some issues raised in PR #2302 and aims to make the SpringArchiveAPI agnostic of the implementation. This entails moving functionality from SpringArchiveAPI to the different ArchiveHandlers. Also, in the process some issues with the tests were addressed.

### Added

- Fixture for a FileAndSample object.
- Fixture for a trimmed local path.

### Changed

- Moved class FileAndSample to cg.meta.archive.models
- Moved the handler dictionary out of the instantiation of the SpringArchiveAPI
- Instead of a DDNDataFlowClient, you give a DataFlowConfig as input to the SpringArchiveAPI. This config is used to instantiate an arbitrary ArchiveHandler.
- Renamed "call_corresponding_archiving_method" to "archive_files".
- Moved convert_files_into_transfer_data from SpringArchiveAPI to ArchiveHandler.
-  Added support for creating file transfer data which adapts to whether you are archiving or retrieving files. This is needed as the source and destination given in the request reverses between the two.
- Renamed DDNDataFlowConfig to DataFlowConfig, which is intended to be a config which can be used to instantiate the DDN archive handler, but also any other, should we add some in the future.
- archive_folders and retrieve_folders now take a list of files and samples as input, and conversion to the correct format for the handler is done within the ArchiveHandler within this flow.
- Renamed sources_and_destinations to miria_file_data within the DDNDataFlowClient.

### Fixed

- Some typos and imports.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/archive/archive.py
+++ b/cg/meta/archive/archive.py
@@ -2,7 +2,7 @@ import logging
 from typing import Callable, Dict, List, Optional
 
 from cg.apps.housekeeper.hk import HousekeeperAPI
-from cg.constants.archiving import ArchiveLocationsInUse
+from cg.constants.archiving import ArchiveLocations
 from cg.meta.archive.ddn_dataflow import DDNDataFlowClient, MiriaFile
 from cg.meta.archive.models import ArchiveHandler, FileTransferData
 from cg.store import Store
@@ -28,7 +28,7 @@ class ArchiveModels(BaseModel):
 
 
 def filter_files_on_archive_location(
-    files_and_samples: List[FileAndSample], archive_location: ArchiveLocationsInUse
+    files_and_samples: List[FileAndSample], archive_location: ArchiveLocations
 ) -> List[FileAndSample]:
     """
     Returns a list of FileAndSample where the associated sample has a specific archive location.
@@ -54,13 +54,13 @@ class SpringArchiveAPI:
         self.ddn_client: DDNDataFlowClient = ddn_dataflow_client
         self.status_db: Store = status_db
         self.handler_map: Dict[str, ArchiveModels] = {
-            ArchiveLocationsInUse.KAROLINSKA_BUCKET: ArchiveModels(
+            ArchiveLocations.KAROLINSKA_BUCKET: ArchiveModels(
                 file_model=MiriaFile.from_file_and_sample, handler=self.ddn_client
             )
         }
 
     def call_corresponding_archiving_method(
-        self, files: List[FileTransferData], archive_location: ArchiveLocationsInUse
+        self, files: List[FileTransferData], archive_location: ArchiveLocations
     ) -> int:
         return self.handler_map[archive_location].handler.archive_folders(
             sources_and_destinations=files
@@ -88,7 +88,7 @@ class SpringArchiveAPI:
         return files_and_samples
 
     def convert_files_into_transfer_data(
-        self, files_and_samples: List[FileAndSample], archive_location: ArchiveLocationsInUse
+        self, files_and_samples: List[FileAndSample], archive_location: ArchiveLocations
     ) -> List[FileTransferData]:
         return [
             self.handler_map[archive_location].file_model(
--- a/cg/meta/archive/archive.py
+++ b/cg/meta/archive/archive.py
@@ -35,7 +35,7 @@ def filter_files_on_archive_location(
     ]
 
 
-archive_handlers: Dict[str, Callable] = {ArchiveLocations.KAROLINSKA_BUCKET: DDNDataFlowClient}
+ARCHIVE_HANDLERS: Dict[str, Callable] = {ArchiveLocations.KAROLINSKA_BUCKET: DDNDataFlowClient}
 
 
 class SpringArchiveAPI:
@@ -50,7 +50,7 @@ class SpringArchiveAPI:
         self.data_flow_config: DataFlowConfig = data_flow_config
 
     def archive_files(self, files: List[FileAndSample], archive_location: ArchiveLocations) -> int:
-        archive_handler: ArchiveHandler = archive_handlers[archive_location](self.data_flow_config)
+        archive_handler: ArchiveHandler = ARCHIVE_HANDLERS[archive_location](self.data_flow_config)
         return archive_handler.archive_folders(files_and_samples=files)
 
     def get_sample(self, file: File) -> Optional[Sample]:
--- a/cg/meta/archive/archive.py
+++ b/cg/meta/archive/archive.py
@@ -3,6 +3,7 @@ from typing import Callable, Dict, List, Optional
 
 from cg.apps.housekeeper.hk import HousekeeperAPI
 from cg.constants.archiving import ArchiveLocations
+from cg.meta.archive.ddn_dataflow import DDNDataFlowClient
 from cg.meta.archive.models import ArchiveHandler, FileAndSample
 from cg.models.cg_config import DataFlowConfig
 from cg.store import Store
--- a/cg/meta/archive/archive.py
+++ b/cg/meta/archive/archive.py
@@ -59,7 +59,7 @@ class SpringArchiveAPI:
             )
         }
 
-    def call_corresponding_archiving_function(
+    def call_corresponding_archiving_method(
         self, files: List[FileTransferData], archive_location: ArchiveLocationsInUse
     ) -> int:
         return self.handler_map[archive_location].handler.archive_folders(
@@ -87,9 +87,9 @@ class SpringArchiveAPI:
                 files_and_samples.append(FileAndSample(file=file, sample=sample))
         return files_and_samples
 
-    def convert_into_correct_model(
+    def convert_files_into_transfer_data(
         self, files_and_samples: List[FileAndSample], archive_location: ArchiveLocationsInUse
-    ) -> List[MiriaFile]:
+    ) -> List[FileTransferData]:
         return [
             self.handler_map[archive_location].file_model(
                 file=file_and_sample.file, sample=file_and_sample.sample
--- a/cg/meta/archive/archive.py
+++ b/cg/meta/archive/archive.py
@@ -1,5 +1,5 @@
 import logging
-from typing import Callable, Dict, List, Optional
+from typing import Callable, Dict, List, Optional, Type
 
 from cg.apps.housekeeper.hk import HousekeeperAPI
 from cg.constants.archiving import ArchiveLocations
@@ -35,7 +35,9 @@ def filter_files_on_archive_location(
     ]
 
 
-ARCHIVE_HANDLERS: Dict[str, Callable] = {ArchiveLocations.KAROLINSKA_BUCKET: DDNDataFlowClient}
+ARCHIVE_HANDLERS: Dict[str, Type[ArchiveHandler]] = {
+    ArchiveLocations.KAROLINSKA_BUCKET: DDNDataFlowClient
+}
 
 
 class SpringArchiveAPI:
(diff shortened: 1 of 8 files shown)


cg/meta/archive/archive.py
```diff
@@ -40,33 +35,26 @@ def filter_files_on_archive_location(
     ]
 
 
+archive_handlers: Dict[str, Callable] = {ArchiveLocations.KAROLINSKA_BUCKET: DDNDataFlowClient}
```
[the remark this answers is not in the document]
ChrOertlin: THis would be a similar structure as to what we use in cg when we made the store filters. Maybe this can be done here also. It would be easily extendable and we do not need to rely on dicts.

The class name can be changed just took the variable name that was given

```suggestion
class ArchiveHandlers(Enum):
KAROLINSKA_BUCKET: Callable = DDNDataFlowClient
```

cg/meta/archive/ddn_dataflow.py
```diff
@@ -188,23 +192,29 @@ def auth_header(self) -> Dict[str, str]:
             self._refresh_auth_token()
         return {"Authorization": f"Bearer {self.auth_token}"}
 
-    def archive_folders(self, sources_and_destinations: List[MiriaFile]) -> int:
+    def archive_folders(self, files_and_samples: List[FileAndSample]) -> int:
```
ChrOertlin, line 195: Would this only work for MiriaFiles or can this be easily extended to other File types? From the type hits it looks like it

cg/meta/archive/ddn_dataflow.py
```diff
@@ -188,23 +192,29 @@ def auth_header(self) -> Dict[str, str]:
...
-    def retrieve_folders(self, sources_and_destinations: List[MiriaFile]) -> bool:
+    def retrieve_folders(self, files_and_samples: List[FileAndSample]) -> int:
         """Retrieves all folders provided, to their corresponding destination, as given by sources
         and destination in TransferData. Returns the job ID of the retrieval task."""
+        miria_file_data: List[MiriaFile] = self.convert_into_transfer_data(
+            files_and_samples, is_archiving=False
+        )
         transfer_request: TransferPayload = self.create_transfer_request(
-            sources_and_destinations, is_archiving_request=False
+            miria_file_data=miria_file_data, is_archiving_request=False
```
ChrOertlin, line 217: same for this function. miria_file_data

cg/meta/archive/ddn_dataflow.py
```diff
@@ -213,7 +223,7 @@ def retrieve_folders(self, sources_and_destinations: List[MiriaFile]) -> bool:
         return job_id
 
     def create_transfer_request(
-        self, sources_and_destinations: List[MiriaFile], is_archiving_request: bool
+        self, miria_file_data: List[MiriaFile], is_archiving_request: bool
```
ChrOertlin, line 226: ah I see it got changed here. I don't see why `create_transfer_request` has to know the specifics of what files they transfer, i.e. MiriaFile in this case.

Can't we make a structure where we handle the different File types i.e. `MiriaFile` and what might come in the future and pass that class here? Then these functions can be more general and not tied to one specific file type.

ChrOertlin (commented):
Looks good. Maybe we can structure the file handling such as `MiriaFile` a bit more so that we can reduce limitations of the functions to only return `MiriaFiles` this would make it easier to extend in the future without needing to rewrite things.

cg/meta/archive/ddn_dataflow.py
islean: @ChrOertlin Ideally I would like to rename the whole class to miria_dataflow. I think the naming "DDNDataFlowApi" is an artifact from before it was decided that Miria was more descriptive. So ArchiveHandler has the general method "archive_folders" whereas DDNDataFlowApi has an implementation which is only intended to be used for Karolinska customers. 

The answer to your question would depend on what you mean by "work". If you provide a list of FileAndSamples, the implementation in DDNDataFlowApi will convert these to a list of MiriaFiles and send a request to store these using Miria. So it would probably execute regardless of what files and samples are provided, but it is only intended for storage of files via Miria.

Any other implementation of archive_folders would convert the files and samples to the format used for archiving there. Say FileAndSample -> OtherArchivingApiFile -> archive using OtherArchivingApi.

cg/meta/archive/ddn_dataflow.py
islean: @ChrOertlin Do you mean that we would have in ArchiveHandler a "create_transfer_request" which takes List[FileTransferData] as input, and that we implement it in the different ArchiveHandlers such as DDNDataFlowClient? 

It would not be feasible to have a "create_transfer_request" which works for all external APIs as they will most likely require different requests.

cg/meta/archive/ddn_dataflow.py
ChrOertlin: Sorry I meant to point to this part of the code. Why would we care about miriafiles here? can it not be a general file class that determines internally if it are miria files or something else?

```Python 
   miria_file_data: List[MiriaFile] = self.convert_into_transfer_data(
            files_and_samples, is_archiving=True
```

cg/meta/archive/ddn_dataflow.py
islean: @ChrOertlin Yes, precisely 😃

cg/meta/archive/ddn_dataflow.py
ChrOertlin: this comment can be disregarded from the discussion above

cg/meta/archive/archive.py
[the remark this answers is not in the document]
islean: Seems a bit cumbersome since you can not reuse enums within another enum, and for an enum to have multiple values, the easiest way would be to add "aenum" to the requirements. Would that be something we would want?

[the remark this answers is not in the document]
islean: Seems as if a common way is to use a dict in these instances if we don't want to import aenum, so I will keep as is for now. https://stackoverflow.com/questions/70566944/python-enum-keys-from-another-enum

cg/meta/archive/archive.py
```diff
@@ -40,33 +35,28 @@ def filter_files_on_archive_location(
     ]
 
 
+ARCHIVE_HANDLERS: Dict[str, Type[ArchiveHandler]] = {
```
seallard, line 38: We could encapsulate/abstract away this data structure into a separate class using the registry/factory pattern. Something like 

```python
from typing import Type, Dict

class ArchiveHandlerManager:
    def __init__(self):
        self._handlers: Dict[str, Type[ArchiveHandler]] = {
            ArchiveLocations.KAROLINSKA_BUCKET: DDNDataFlowClient
        }
    # The location could be typed as an enum
    def get_archive_handler(self, location: str) -> Type[ArchiveHandler]:
        return self._handlers.get(location)

    # If you anticipate adding more handlers in the future
    def register_handler(self, location: str, handler: Type[ArchiveHandler]):
        self._handlers[location] = handler

    # ... any other necessary methods related to the handlers
```
Probably not in scope for this PR and just an idea. Feel free to disregard

cg/meta/archive/archive.py
```diff
@@ -40,33 +35,28 @@ def filter_files_on_archive_location(
...
+        self.data_flow_config: DataFlowConfig = data_flow_config
 
-    def call_corresponding_archiving_method(
-        self, files: List[FileTransferData], archive_location: ArchiveLocations
-    ) -> int:
-        return self.handler_map[archive_location].handler.archive_folders(
-            sources_and_destinations=files
-        )
+    def archive_files(self, files: List[FileAndSample], archive_location: ArchiveLocations) -> int:
+        archive_handler: ArchiveHandler = ARCHIVE_HANDLERS[archive_location](self.data_flow_config)
```
seallard, line 55: Will the archive location always be the same for this class? If so, maybe the archive handler could be provided in the constructor instead

seallard: I'm guessing it won't always be the same though

cg/meta/archive/ddn_dataflow.py
```diff
@@ -39,9 +39,13 @@ class MiriaFile(FileTransferData):
     source: str
 
     @classmethod
-    def from_file_and_sample(cls, file: File, sample: Sample) -> "MiriaFile":
+    def from_file_and_sample(
```
seallard, line 42: This method is confusing to me, how can the sample internal id be the destination? Will a sample model with the provided id be populated with data from the file?

cg/meta/archive/ddn_dataflow.py
```diff
@@ -162,7 +166,7 @@ def _set_auth_tokens(self) -> None:
             ).model_dump(),
         )
         if not response.ok:
-            raise DdnDataflowAuthenticationError(message=response.content)
+            raise DdnDataflowAuthenticationError(message=response.text)
```
seallard, line 169: Out of scope, but we should not automatically assume it is an authentication issue for any response status other than 200

seallard: Authentication error would be 401

cg/meta/archive/ddn_dataflow.py
seallard, line 195: Out of scope, but I think it is less confusing to return the actual `TransferJob` model from these methods rather than an int

seallard (approved):
Nice! I think we can iterate some more on the design to make it easier to understand but looks good 👍  I think all of my comments are out of scope for this PR

cg/meta/archive/ddn_dataflow.py
islean: @seallard It is basically down to what kind of file structure we want in the data lake we will be storing the data in. Current suggestion is to have a very simple structure wherein we put the files within a directory which we simply call the same as the sample_internal_id. Would love to test exactly how this works once the API is up, though.

In the request, the source is where the file is currently located and destination is where we want to put it. I think for retrieving files we will probably need to think about this a bit - are we fetching a folder and putting it in the location where there previously was a file? Maybe we should drop everything after "version" in the destination string at that point. Probably no point in developing this until their API is up.

cg/meta/archive/ddn_dataflow.py
islean: Good point! Will create a small PR for this in the future.

cg/meta/archive/ddn_dataflow.py
islean: Yes, I have an intention to create a PR where I create response class extending request.Response or something similar. In general, I think we should expect a request.Response object.

islean: Or a data class modelling a successful response.

cg/meta/archive/archive.py
islean: No, it won't be the same, although there is a one-to-one correspondence between archive_location and ArchiveHandlers at the moment. So the effect is the same as giving the ArchiveHandler, it is just a matter of where the instantiation happens.

islean:
Deployment to stage succeeded. Since no functionality touched is in use, IO declares the testing a success.

cg/meta/archive/archive.py
islean: Yeah, might be an idea for the future. I will think about it!

PR #2313, opened 2023-08-09
Refactor token validation

seallard:
## Description
This PR refactors the token validation. Using `id_token.verify_oauth2_token` instead of `jwt.decode` lets us replace the logic for retrieving the google certs with library functionality.

I'm not completely sure if it is correct to use the client id as the audience, maybe it should be the base url to the api or something else. Skipping that validation for now since it was not done previously either (but something we really should do).

According to [Google guidelines](https://developers.google.com/identity/openid-connect/openid-connect#validatinganidtoken), the following aspects of a token should be verified:

- [x] Verify that the ID token is properly signed by the issuer. Google-issued tokens are signed using one of the certificates found at the URI specified in the jwks_uri metadata value of the [Discovery document](https://developers.google.com/identity/openid-connect/openid-connect#discovery). 
- [x] Verify that the value of the iss claim in the ID token is equal to https://accounts.google.com or accounts.google.com.
- [ ] Verify that the value of the aud claim in the ID token is equal to your app's client ID.
- [x] Verify that the expiry time (exp claim) of the ID token has not passed.


seallard:
> Probably a bit beyond me to approve or request changes but curious still how this will be tested? I mean, have we been able to reproduce a problem which this solves?

The problem is the way we validate jwt tokens in cg does not follow googles guidelines. I guess I could try to create an exploit for it, not sure though - it might be the case that this does not make any difference in practice.

cg/server/api.py
```diff
@@ -30,54 +36,25 @@
...
+    user_data = id_token.verify_oauth2_token(
+        id_token=token, request=google_requests.Request(), audience=client_id
+    )
+    issuer: str = user_data["iss"]
+    valid_issuers = ["accounts.google.com", "https://accounts.google.com"]
 
-    if not certs:
-        certs = fetch_and_cache_google_oauth2_certificates()
+    if issuer not in valid_issuers:
+        raise ValueError("Invalid issuer")
```
islean, line 55: I think this check is already performed in verify_oath2_token: https://google-auth.readthedocs.io/en/stable/_modules/google/oauth2/id_token.html#verify_oauth2_token

seallard: Great! Nice 😄

islean:
Also - my guess is that the provided certificate for jwt.decode ends up performing the same validation as here? But it is more convoluted as it is currently written so even if that is the case, this would be a nice change :)

seallard:
> Also - my guess is that the provided certificate for jwt.decode ends up performing the same validation as here? But it is more convoluted as it is currently written so even if that is the case, this would be a nice change :)

Yeah, I could not figure out whether it checked the issuer. There was a parameter for validating the audience but it was not used previously used

Vince-janv (approved):
Nice to use libraries for as much as possible for this stuff 👍 Do you think we should still implement some cacheing feature as suggested [here](https://google-auth.readthedocs.io/en/stable/reference/google.oauth2.id_token.html), to avoid fetching the certificate every time?

seallard:
> Nice to use libraries for as much as possible for this stuff 👍 Do you think we should still implement some cacheing feature as suggested [here](https://google-auth.readthedocs.io/en/stable/reference/google.oauth2.id_token.html), to avoid fetching the certificate every time?

Yes, nice catch!

seallard:
Deployed to stage, possible to sign in and use endpoints from order portal.

PR #2156, opened 2023-06-22
change(rnafusion) refactor rnafusion tower and nextflow classes

fevac:
## Description

Related to:
- https://github.com/Clinical-Genomics/cg/issues/2044 
- https://github.com/Clinical-Genomics/cg/issues/1819 
- https://github.com/Clinical-Genomics/cg/issues/1959

### Changed

- Adds a common class for nf-core analysis from which rnafusion and taxprofiler will inherits. This is done to avoid code redundancy.
- Refactor NextflowAnalysisAPI and TowerAnalysisAPI as executors instead of analysis and clean up their methods
- Removed unused CLI options
- Group tower variables in cg.yalm

### How to prepare for test

- [ ] Ssh to relevant server (depending on type of change)
- [ ] Use stage: `us`
- [ ] Paxa the environment: `paxa`
- [ ] Install on stage (example for Hasta):
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b [THIS-BRANCH-NAME] -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...
- [ ] Take a screenshot and attach or copy/paste the output.

## Review

- [ ] Tests executed by
- [ ] "Merge and deploy" approved by
Thanks for filling in who performed the code review and the test!

### This [version](https://semver.org/) is a

- [ ] **MAJOR** - when you make incompatible API changes
- [ ] **MINOR** - when you add functionality in a backwards compatible manner
- [ ] **PATCH** - when you make backwards compatible bug fixes or documentation/instructions

## Implementation Plan

- [ ] Document in ...
- [ ] Deploy this branch on ...
- [ ] Inform to ...

PR #2318, opened 2023-08-10
refactor get parameters from model

fevac:
## Description

Part of https://github.com/Clinical-Genomics/cg/pull/2156

Add BaseModel for rnafusion parameters

### Changed

- Removed `get_default_parameters ` in favour of BaseModel
- Code cleanup

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b parameter_datamodel -a
    ```

### How to test

- [x] `cg workflow rnafusion config-case <case_id>`
- [x] `cg workflow rnafusion run <case_id>`
- [x] `cg workflow rnafusion store <case_id>`
- [x] `cg workflow taxprofiler config-case <case_id>`

### Expected test outcome

- [x] Commands run successfully

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/models/rnafusion/rnafusion_sample.py
```diff
@@ -15,5 +15,5 @@ class RnafusionSample(NextflowSample):
     @validator("strandedness", always=True)
     def valid_value_strandedness(cls, value) -> str:
         """Verify that the strandedness value is accepted."""
-        assert value in RNAFUSION_ACCEPTED_STRANDEDNESS
+        assert value in Strandedness.accepted_values()
         return "Strandedness value not valid"
```
ivadym, line 19: We're missing the benefits of pydantic models here to perform the validation. This validator is unnecessary, you could just type hint `strandedness` and obtain the same results:

```
strandedness: Strandedness
```

cg/constants/constants.py
```diff
@@ -199,3 +200,15 @@ class MetaApis:
...
+class Strandedness(str, Enum):
+    """Strandedness types."""
+
+    FORWARD: str = "forward"
+    REVERSE: str = "reverse"
+    UNSTRANDED: str = "unstranded"
+
+    @classmethod
+    def accepted_values(cls) -> set:
+        return {stranded_type.value for stranded_type in cls}
```
ivadym, line 214: You want to use `set(Strandedness)` and make the most of the `StrEnum` classes.

cg/meta/workflow/nf_handlers.py
```diff
@@ -21,11 +21,13 @@ class NfBaseHandler:
     def write_nextflow_yaml(
         cls,
         content: Dict[str, Any],
-        file_path: str,
+        file_path: Path,
```
ivadym, line 24:
```suggestion
        file_path: Path
```

cg/meta/workflow/nf_handlers.py
```diff
@@ -21,11 +21,13 @@ class NfBaseHandler:
...
     ) -> None:
         """Write Nextflow file with non-quoted booleans and quoted strings."""
         with open(file_path, "w") as outfile:
             for key, value in content.items():
+                if isinstance(value, Path):
```
ivadym, line 29: Could you rename `key` and `values` to something more specific to this case? 👼

cg/meta/workflow/rnafusion.py
```diff
@@ -129,58 +129,21 @@ def write_samplesheet(self, case_id: str, strandedness: str, dry_run: bool = Fal
...
-    def write_trailblazer_config(self, case_id: str, tower_id: str) -> None:
-        """Write Tower IDs to a .YAML file used as the trailblazer config."""
-        config_path = self.get_trailblazer_config_path(case_id=case_id)
-        LOG.info(f"Writing Tower ID to {config_path.as_posix()}")
-        WriteFile.write_file_from_content(
-            content={case_id: [tower_id]},
-            file_format=FileFormat.YAML,
-            file_path=config_path,
-        )
+    def get_pipeline_parameters(self, case_id: str, genomes_base: Optional[Path] = None) -> dict:
```
ivadym, line 132: I think it would be better if this method returns `RnafusionParameters`. If you end up needing a dictionary somewhere else, you can convert it there. It's always easier to work with data models than dicts

ivadym (approved):
Nicely done 👍 

I have just a couple of comments regarding the data models

cg/models/rnafusion/rnafusion_sample.py
fevac: removed!

cg/constants/constants.py
fevac: Removed `accepted_values`

cg/meta/workflow/nf_handlers.py
fevac: I cannot come up with anything better than key and value. Depending on the content to write the keys could be parameters ids or something else, but this function is quite general. It writes a yaml block that works with nextflow. Unless you have a specific suggestion I think I will leave it like that

cg/meta/workflow/rnafusion.py
fevac: I think this method should return something general that can be use in other pipelines (e.g. taxprofiler). I will make then a parent class (PipelineParameters) and return that

cg/meta/workflow/nf_handlers.py
ivadym: I see, then I would say it's ok if it's not possible to provide accurate information 👍

cg/meta/workflow/rnafusion.py
ivadym: Yes, that's a good idea, a data model with common fields between pipelines 💯

ChrOertlin (approved):
Just a comment on using the yaml IO controller in cg/io/yaml.py to write yaml files. You can extend existing functionality there. In future, others might have use of it. :)

fevac:
I've refactored rnafusion models to a single file. I found it easier to follow. Have a look and I can revert back if you don't like it

cg/meta/workflow/rnafusion.py
fevac: done!

cg/meta/workflow/nf_handlers.py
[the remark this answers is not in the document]
fevac: I moved this method as a function in io/yaml.py

ivadym (approved):
Looks good to me 👍 

The only thing confusing to me is that I would have a hard time finding these fixtures. Give the nextflow folder and that certain models used in the rnafusion file are sourced from there, I'm wondering if there's a specific reason of not adding these models to the nextflow folder?

I also think the problem may be that we are not following a specific folder structure here 🤔  I would look into the rnafusion folder for these models but basically because balsamic models is structured in that way.

fevac:
I was thinking that the nextflow folder should be deleted and the models there should also be part of the nf_analysis. I can add it in this PR too. Those models are related to the nf_analysis not nextflow itself. Would that make sense to you?

ivadym:
> I was thinking that the nextflow folder should be deleted and the models there should also be part of the nf_analysis. I can add it in this PR too. Those models are related to the nf_analysis not nextflow itself. Would that make sense to you?

I see. If you want to follow the current structure I think it should be:

```
models
├── balsamic
├── mip
├── rnafusion
│   ├── analysis.py or rnafusion.py
├── analysis.py
└── nf_analysis.py
```

And add the models from the nextflow folder to `nf_analysis.py`. 

That said, if you don't like the current structure and want to change for `nextflow` pipelines, it I think this is the time ⏳

fevac:
well I don't understand why there is a workflow subfolder with only one of the workflows. To me the current structure if a mix of structures. IMO if there is a workflow subfolder these workflows should inside there. 

I also don't like having a single model per file. Sometimes this might be good but in the case of the rnafusion (for example) I think it makes it more confusing. But I guess it's a matter of personal preference?

ivadym:
> well I don't understand why there is a workflow subfolder with only one of the workflows. To me the current structure if a mix of structures. IMO if there is a workflow subfolder these workflows should inside there.
> 
> I also don't like having a single model per file. Sometimes this might be good but in the case of the rnafusion (for example) I think it makes it more confusing. But I guess it's a matter of personal preference?

Considering that only mutant adheres to that structure, I would say that's the outlier. On top of that, the model that mutant has there does not belong to workflow/analysis, but more to the config case part. The rest of the pipelines have that distinction inside their folders (`mip_analysis.py` vs `mip_config.py`).

fevac:
alright, I'm convinced. I'll move it to the structure you suggest but keeping it into a single file

fevac:
- Deployed to stage
image


- Deployed to prod
image

PR #2320, opened 2023-08-11
fix(add samples to flow cell if it already exists)

ChrOertlin:
## Description

This PR changes the behaviour of the `store_flow_cell_data_in_status_db` so that samples can be added to an existing flow cell.

This behaviour is wanted to avoid the need to delete flow cells in the associated samples need to be altered in cases of re-demultiplexing of flow cells.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/demultiplex/status_db_storage_functions.py
+++ b/cg/meta/demultiplex/status_db_storage_functions.py
@@ -44,7 +44,7 @@ def store_flow_cell_data_in_status_db(
         sample_sheet_path=parsed_flow_cell.sample_sheet_path,
         flow_cell_sample_type=parsed_flow_cell.sample_type,
     )
-    flow_cell = add_samples_to_flow_cell_in_status_db(
+    flow_cell: Flowcell = add_samples_to_flow_cell_in_status_db(
         flow_cell=flow_cell,
         sample_internal_ids=sample_internal_ids,
         store=store,
--- a/cg/meta/demultiplex/status_db_storage_functions.py
+++ b/cg/meta/demultiplex/status_db_storage_functions.py
@@ -29,7 +29,7 @@ def store_flow_cell_data_in_status_db(
     Create flow cell from the parsed and validated flow cell data.
     And add the samples on the flow cell to the model.
     """
-    flow_cell: Flowcell = store.get_flow_cell_by_name(flow_cell_name=parsed_flow_cell.id)
+    flow_cell: Optional[Flowcell] = store.get_flow_cell_by_name(flow_cell_name=parsed_flow_cell.id)
     if not flow_cell:
         flow_cell: Flowcell = Flowcell(
             name=parsed_flow_cell.id,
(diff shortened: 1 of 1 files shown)


ChrOertlin:
# Samples get added to flow cell
image

cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -25,28 +25,33 @@
...
-        store.session.commit()
         LOG.info(f"Flow cell added to status db: {parsed_flow_cell.id}.")
     else:
-        LOG.info(f"Flow cell already exists in status db: {parsed_flow_cell.id}. Skipping.")
+        LOG.info(f"Flow cell already exists in status db: {parsed_flow_cell.id}.")
+    sample_internal_ids = get_sample_internal_ids_from_sample_sheet(
+        sample_sheet_path=parsed_flow_cell.sample_sheet_path,
+        flow_cell_sample_type=parsed_flow_cell.sample_type,
+    )
+    flow_cell: Flowcell = add_samples_to_flow_cell_in_status_db(
```
seallard, line 47: It is not necessary to return the flow cell object from `add_samples_to_flow_cell_in_status_db`, but you decide if it is more readable

seallard: We already have a reference to the flow cell object that is mutated in `add_samples_to_flow_cell_in_status_db` which means we can just commit the changes to the object using that reference

cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -25,28 +25,33 @@
 def store_flow_cell_data_in_status_db(
     parsed_flow_cell: FlowCellDirectoryData, store: Store
 ) -> None:
-    """Create flow cell from the parsed and validated flow cell data."""
-    if not store.get_flow_cell_by_name(flow_cell_name=parsed_flow_cell.id):
+    """
+    Create flow cell from the parsed and validated flow cell data.
+    And add the samples on the flow cell to the model.
+    """
+    flow_cell: Flowcell = store.get_flow_cell_by_name(flow_cell_name=parsed_flow_cell.id)
```
seallard, line 32: type hint should be Optional[Flowcell] as to not lie

seallard: Also consider skipping the named parameter if black introduces line breaks, since it is only a single parameter it does not really add any value. Alternatively one could also create a variable above `flow_cell_name: str = parsed_flow_cell.id` and use it in the call

seallard (approved):
Nice 👍

PR #2323, opened 2023-08-11
Fix existence check for sample on flow cell

seallard:
## Description
This PR fixes the existence check whether a sample already has been added to a flow cell. Previously the check used actual sample objects which does not work unless we override the __hash__ and __eq__ methods on the Sample model. That could be a cleaner approach, not sure.

### Fixed
- Existence check of sample on flow cell in new post processing flow


seallard:
Actually, when adding a test case it seems like duplicates are not allowed anyway in the original version. Seems like it already works as intended somehow 😅

Issue #2030, opened 2023-05-12
Explicitly referring to DRAGEN software usage


ChrOertlin:
## Description

In some instances we are using DRAGEN as a constant in the CG codebase. These include constants that should point to a more specific software that is run on dragen. One such example is the BclConverter class. Here we have the options "bcl2fastq" and "dragen". However, dragen runs the "bclconvert" software, which is the more appropriate constant to use.

This issue is also carried over in some models that carry "Dragen" in the name but likely should point to a more specific software.

Being more explicit might help us identify potential issues faster.

One examlpe can be found in this PR:
https://github.com/Clinical-Genomics/cg/pull/2019

## Suggested solution

Find the usages of Dragen and if applicable be explicit of the underlying software that is used when we parse information. 

## This can be closed when

*Describe what needs to be done for this issue to be closed*

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

henrikstranneheim:
Perhaps let dragen be the key and the value "the software" used on the dragen. If there is merit keeping it around

islean:
My current suggestion is that at least in the mentioned bcl converter class, the naming could be similar to

<img width="498" alt="image" src="https://github.com/Clinical-Genomics/cg/assets/40887124/714c76ce-3ba2-426e-a66e-1a9d16a7d741">

where the only difference to previously is "DRAGEN: str = "dragen"" -> "DRAGEN_BCLCONVERT: str = "dragen_bcl_convert""

ChrOertlin:
On hold for now

PR #2328, opened 2023-08-14
feat(dynamic setting of bcl converter)

ChrOertlin:
## Description

Set the bcl converter in the flow cell directory data dynamically based on sequencer type if no bcl converter is specified. If the sequencer type cannot be resolved it defaults to bcl2fastq.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/models/demultiplex/flow_cell.py
```diff
@@ -126,6 +126,28 @@ def sequencer_type(
         """Return the sequencer type."""
         return sequencer_types[self.machine_name]
 
+    def set_bcl_converter(self, bcl_converter: str) -> str:
```
henrikstranneheim, line 129:
```suggestion
    def get_bcl_converter(self, bcl_converter: str) -> str:
```

cg/models/demultiplex/flow_cell.py
```diff
@@ -126,6 +126,28 @@ def sequencer_type(
...
+        """
+        Return the bcl converter to use.
+        Tries to fetch the bcl converter from the sequencer type if not provided.
+        """
+        if bcl_converter:
+            return bcl_converter
+        return self.get_bcl_converter_by_sequencer()
```
henrikstranneheim, line 136:
```suggestion
        return bcl_converter or self.get_bcl_converter_by_sequencer()
```
I think you can use a ternary operator here

cg/models/demultiplex/flow_cell.py
```diff
@@ -126,6 +126,28 @@ def sequencer_type(
         """Return the sequencer type."""
         return sequencer_types[self.machine_name]
 
+    def set_bcl_converter(self, bcl_converter: str) -> str:
+        """
+        Return the bcl converter to use.
+        Tries to fetch the bcl converter from the sequencer type if not provided.
```
henrikstranneheim, line 132:
```suggestion
        Return the BCL converter to use.
        Tries to get the BCL converter from the sequencer type if not provided.
```

cg/models/demultiplex/flow_cell.py
```diff
@@ -126,6 +126,28 @@ def sequencer_type(
...
+        """
+        if bcl_converter:
+            return bcl_converter
+        return self.get_bcl_converter_by_sequencer()
+
+    def get_bcl_converter_by_sequencer(
+        self,
+    ) -> str:
+        """Return the bcl converter based on sequencer."""
```
henrikstranneheim, line 141:
```suggestion
        """Return the BCL converter based on sequencer."""
```

cg/models/demultiplex/flow_cell.py
```diff
@@ -126,6 +126,28 @@ def sequencer_type(
...
+        try:
+            if self.sequencer_type in [Sequencers.HISEQGA, Sequencers.HISEQX]:
+                return BclConverter.BCL2FASTQ
+        except KeyError:
+            raise FlowCellError(
+                f"Could not determine bcl converter from sequencer type for machine name {self.machine_name}"
```
henrikstranneheim, line 147:
```suggestion
                f"Could not determine BCL converter from sequencer type for sequencing machine name {self.machine_name}"
```

cg/models/demultiplex/flow_cell.py
henrikstranneheim: I think it would be nice to also add the `Raise` to the doc str

henrikstranneheim (approved):
Nice 👍 !

cg/models/demultiplex/flow_cell.py
```diff
@@ -126,6 +126,28 @@ def sequencer_type(
...
+            )
+        return BclConverter.DRAGEN
```
diitaz93, line 149: Wouldn't it be better if you check first if it is hiseq, then check if is novaseq and if not, raise an error? I think this error-catching may not be necessary.

diitaz93 (approved):
Looks good, but I am wondering if the error catching is necessary or you could use just if-else

cg/models/demultiplex/flow_cell.py
ChrOertlin: Yes, looking at it know the raised error is a bit pointless.

tests/models/demultiplexing/test_flowcell_model.py
```diff
@@ -134,3 +135,52 @@ def test_get_sample_model_novaseq_x(novaseq_x_flow_cell: FlowCellDirectoryData):
...
+def test_flow_cell_directory_data_with_novaseq_flow_cell_directory(
+    flow_cell_directory_name_demultiplexed_with_bcl_convert: str, bcl_converter=BclConverter.DRAGEN
+):
+    """Test that the bcl converter is set to dragen when prodiving a novaseq flow cell directory."""
+    # GIVEN a Bcl2Fastq flow cell directory
+
+    # WHEN instantiating a flow cell object
+    flow_cell = FlowCellDirectoryData(
+        flow_cell_path=Path(flow_cell_directory_name_demultiplexed_with_bcl_convert),
+        bcl_converter=bcl_converter,
```
islean, line 182: I interpret this test as wanting to show that if the flow cell was demultiplexed with BCL convert, BCL convert is set automatically from the flow cell dir name? If so, should we not omit setting the bcl_converter to BclConverter.DRAGEN here?

islean (approved):
Looks good! I just have the hunch that one test tests something slightly different than intended.

cg/models/demultiplex/flow_cell.py
```diff
@@ -126,6 +126,21 @@ def sequencer_type(
         """Return the sequencer type."""
         return sequencer_types[self.machine_name]
 
+    def get_bcl_converter(self, bcl_converter: str) -> str:
+        """
+        Return the BCL converter to use.
+        Tries to get the BCL converter from the sequencer type if not provided.
+        """
+        return bcl_converter or self.get_bcl_converter_by_sequencer()
```
seallard, line 134: Oh, is the bcl conversion software used really dependent on the sequencer type? Neat!

cg/models/demultiplex/flow_cell.py
```diff
@@ -33,18 +33,18 @@
 class FlowCellDirectoryData:
     """Class to collect information about flow cell directories and their particular files."""
 
-    def __init__(self, flow_cell_path: Path, bcl_converter: Optional[str] = "bcl2fastq"):
+    def __init__(self, flow_cell_path: Path, bcl_converter: Optional[str] = None):
```
seallard, line 36: I guess we should remove this argument then! Maybe better in a separate PR?

cg/models/demultiplex/flow_cell.py
```diff
@@ -126,6 +126,21 @@ def sequencer_type(
...
+        Tries to get the BCL converter from the sequencer type if not provided.
+        """
+        return bcl_converter or self.get_bcl_converter_by_sequencer()
+
+    def get_bcl_converter_by_sequencer(
+        self,
+    ) -> str:
+        """Return the BCL converter based on sequencer."""
+        if self.sequencer_type in [Sequencers.NOVASEQ, Sequencers.NOVASEQX]:
+            return BclConverter.DRAGEN
```
seallard, line 141: We have two values for this conversion software: `BclConverter.DRAGEN` and `BclConverter.BCLCONVERT`. I think the new post processing flow relies on `BCLCONVERT` to be set. Might be worth double checking if this introduces some conflict.

I think we should remove one of the enum values so it is not duplicate because it can cause bugs and confusion. Might be out of scope for this PR though!

seallard (approved):
Nice 🚀  Some smaller fixes to apply to not cause confusion. I really like embedding this logic in the actual `FlowCellDirectoryData` model!

The places which instantiate `FlowCellDirectoryData` need to be updated to reflect this change as well.

cg/models/demultiplex/flow_cell.py
```diff
@@ -126,6 +126,21 @@ def sequencer_type(
         """Return the sequencer type."""
         return sequencer_types[self.machine_name]
 
+    def get_bcl_converter(self, bcl_converter: str) -> str:
```
seallard, line 129: I don't like that we still pass the `bcl_converter` as an argument - it is confusing and I think it is cleaner if we just rely on retrieving it based on the sequencing machine type.

Maybe this involves more work to ensure only `BclConverter.BCLCONVERT` enum is used and not `BclConverter.DRAGEN` as to not introduce bugs. I guess that might explain why you chose to keep the `bcl_converter` argument - but it should be removed. Might have to be in a separate PR, you decide :)

cg/models/demultiplex/flow_cell.py
```diff
@@ -126,6 +126,21 @@ def sequencer_type(
...
+        """
+        Return the BCL converter to use.
+        Tries to get the BCL converter from the sequencer type if not provided.
+        """
+        return bcl_converter or self.get_bcl_converter_by_sequencer()
+
+    def get_bcl_converter_by_sequencer(
+        self,
+    ) -> str:
+        """Return the BCL converter based on sequencer."""
```
diitaz93, line 139:
```suggestion
        """Return the BCL converter based on the sequencer."""
```

cg/models/demultiplex/flow_cell.py
islean: When I looked at this some time ago, it seemed as if the software BCL convert did not care whether it is run on Dragen or not, i.e. it can be run on Dragen but it does not need to be that way (although maybe that is how it is set up at cg). So to me, BclConverter.DRAGEN represents something different than BCL Convert. Outside of the scope of this pr but I would say that the enums are not duplicated per se. 

Link for some documentation https://support.illumina.com.cn/bulletins/2020/10/upgrading-from-bcl2fastq-to-bcl-convert.html "The Illumina BCL Convert software is a standalone local Linux application"

cg/models/demultiplex/flow_cell.py
diitaz93: I am unsure if we would still want to manually demultiplex a novaseq flowcell with Bcl2fastq. If that is the case, `bcl_converter` would still be needed. I left that possibility open when writing the sample sheet code, but I am not sure how often we use this and if it is needed. I agree that we should remove it if we don't need it.

cg/models/demultiplex/flow_cell.py
diitaz93: In the context of cg, BclConvert is always run in Dragen machines. There is no Bclconvert run in hasta. I I am not mistaken, the duplicate constant was indtroduced to slowly migrate from the constant `DRAGEN` to `BCLConvert`, which is more correct. I agree that it is time now to replace DRAGEN by BCLConvert, and that definitely has to be done in a different PR

cg/models/demultiplex/flow_cell.py
diitaz93: I am not entirely sure, but if it is, we can remove the bclconvert parameter. @karlnyr may know

cg/models/demultiplex/flow_cell.py
seallard: Yeah, it should not be called dragen. That is the name of the hardware platform. Not the bcl conversion software. So we should remove it from this enum and only use `BCLCONVERT`

seallard: I guess the easiest would have been to just rename the dragen field to bclconvert

tests/cli/demultiplex/test_create_sample_sheet.py
```diff
@@ -62,7 +62,9 @@ def test_create_bcl2fastq_sample_sheet(
     mocker,
 ):
     # GIVEN a flowcell directory with some run parameters
-    flowcell: FlowCellDirectoryData = FlowCellDirectoryData(flow_cell_working_directory)
+    flowcell: FlowCellDirectoryData = FlowCellDirectoryData(
+        flow_cell_working_directory, bcl_converter=BclConverter.BCL2FASTQ
```
diitaz93, line 66: function with more than two parameters
```suggestion
        flow_cell_path=flow_cell_working_directory, bcl_converter=BclConverter.BCL2FASTQ
```

tests/models/demultiplexing/test_flowcell_model.py
```diff
@@ -134,3 +135,52 @@ def test_get_sample_model_novaseq_x(novaseq_x_flow_cell: FlowCellDirectoryData):
...
+    flow_cell = FlowCellDirectoryData(
+        flow_cell_path=Path(flow_cell_directory_name_demultiplexed_with_bcl2fastq)
+    )
+
+    # THEN it sets the converter to blc2fastq
+    assert flow_cell.bcl_converter == BclConverter.BCL2FASTQ
+
+
+def test_flow_cell_directory_data_with_set_bcl_converter(
+    flow_cell_directory_name_demultiplexed_with_bcl2fastq: str, bcl_converter=BclConverter.DRAGEN
```
diitaz93, line 156: Maybe remove the bcl_convert parameter and write the constant directly below

tests/models/demultiplexing/test_flowcell_model.py
```diff
@@ -134,3 +135,52 @@ def test_get_sample_model_novaseq_x(novaseq_x_flow_cell: FlowCellDirectoryData):
...
+    # GIVEN the bcl converter is set to dragen
+
+    # WHEN instantiating a flow cell object
+    flow_cell = FlowCellDirectoryData(
+        flow_cell_path=Path(flow_cell_directory_name_demultiplexed_with_bcl2fastq),
+        bcl_converter=bcl_converter,
+    )
+
+    # THEN the bcl converter is dragen
+    assert flow_cell.bcl_converter == bcl_converter
```
diitaz93, line 170: so like this, following my previous comment
```suggestion
    assert flow_cell.bcl_converter == BclConverter.DRAGEN
```

tests/models/demultiplexing/test_flowcell_model.py
diitaz93: agree

cg/models/demultiplex/flow_cell.py
ChrOertlin: So, the FlowCellDirectory data is used in both the Demultiplex and DemultiplexPostProcessing API. 

They are used in a different capacity in these two places where the BCL converter parameter has an impact. 

It can be so that we want to demultiplex a novaseq flow cell with bcl2fastq (for reproducability) and this would not be possible if we were to remove the bcl converter option here since it would always default to "dragen" or "bclConvert" for that matter.

cg/models/demultiplex/flow_cell.py
ChrOertlin: See explanation above, tldr might be needed for reproducability for novaseq flowcells demultiplexed with bcl2fastq or hiseqx with bclconvert.

cg/models/demultiplex/flow_cell.py
ChrOertlin: I have an open issue regarding this. My aim was to do this once we have a more stable codebase around the demultiplexing and demux post processing. So yes, but not in PR.

cg/models/demultiplex/flow_cell.py
ChrOertlin: explained above. Reproducability reasons if those exist. Need input from Karl on this.

cg/models/demultiplex/flow_cell.py
diitaz93: I think we should leave it then

cg/models/demultiplex/flow_cell.py
diitaz93: see #2030. So should we write that the issue is blocked by something? @ChrOertlin

ChrOertlin:
# cg demultiplex finish 

cg demultiplex flow-cell 201203_A00689_0200_AHVKJCDRXX --dry-run
/home/proj/stage/bin/miniconda3/envs/S_main/lib/python3.7/site-packages/sqlalchemy/orm/relationships.py:2231: SAWarning: Cascade settings "delete, merge, save-update" should not be combined with a viewonly=True relationship.   This configuration will raise an error in version 1.4.  Note that in versions prior to 1.4, these cascade settings may still produce a mutating effect even though this relationship is marked as viewonly=True.
  "viewonly=True." % (", ".join(sorted(non_viewonly)))
Running cg demultiplex.
Running cg demultiplex flow cell, using bcl2fastq
Set environment to stage
setting flow cell id to 201203_A00689_0200_AHVKJCDRXX
setting out dir to /home/proj/stage/demultiplexed-runs
Check if demultiplexing is possible for HVKJCDRXX
Check if flow cell is ready for demultiplexing
Check if sequencing is done
Sequencing is not completed for flow cell HVKJCDRXX
Check if sample sheet exists

tests/models/demultiplexing/test_flowcell_model.py
ChrOertlin: fixed

ChrOertlin:
stage:

Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/christian.oertlin
done.
Log deploy... done.
cg, version 40.9.5

prod:
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/christian.oertlin
done.
Log deploy... done.
cg, version 40.9.5

Issue #2331, opened 2023-08-14
sql duplicate entry error in upload-auto-mip-dna.sh


beatrizsavinhas:
**Describe the bug**

We have received CRON alerts for `upload-auto-mip-dna.sh` with the following error:

> sqlalchemy.exc.IntegrityError: (pymysql.err.IntegrityError) (1062, "Duplicate entry '<CASE_ID>/2023-08-13/delivery-report.html' for key 'file.path'")
> [SQL: UPDATE file SET path=%(path)s WHERE file.id = %(file_id)s]
> [parameters: {'path': '<CASE_ID>/2023-08-13/delivery-report.html', 'file_id': 4388941}]

This seems to stop the upload of other cases as well:

> sqlalchemy.exc.InvalidRequestError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (pymysql.err.IntegrityError) (1062, "Duplicate entry '<CASE_ID>/2023-08-13/delivery-report.html' for key 'file.path'")
> [SQL: UPDATE file SET path=%(path)s WHERE [file.id](http://file.id/) = %(file_id)s]
> [parameters: {'path': '<CASE_ID>/2023-08-13/delivery-report.html', 'file_id': 4391889}]

The affected cases later upload without a problem.
The first error seems to have popped up on 20230703.

Spreadsheet to keep track of affected cases: https://docs.google.com/spreadsheets/d/1wdo9DX_Av6hJyZH4Bx03E6zxmhlO3kye46QZAPRE6gY/edit?usp=sharing

---

> 2023-08-14 09:40:19 [hasta.scilifelab.se](http://hasta.scilifelab.se/) cg.meta.report.report_api[194874] INFO Adding a new delivery report to housekeeper for <CASE_ID>
> 2023-08-14 09:40:19 [hasta.scilifelab.se](http://hasta.scilifelab.se/) housekeeper.store.api.handlers.read[194874] INFO Fetching tag with name: <CASE_ID>
> 2023-08-14 09:40:19 [hasta.scilifelab.se](http://hasta.scilifelab.se/) housekeeper.store.api.handlers.read[194874] INFO Fetching tag with name: delivery-report
> 2023-08-14 09:40:19 [hasta.scilifelab.se](http://hasta.scilifelab.se/) housekeeper.store.api.handlers.read[194874] INFO Fetching tag with name: <CASE_ID>
[...]

PR #2332, opened 2023-08-15
Fix already included file call in MIP DNA auto upload.

islean:
## Description

Addresses issue #2331. Since a file is included in Housekeeper by default now, auto uploads fail due to a missed update where an add is followed by an include.

### Fixed

- Removed include_file call in add_delivery_report_to_hk

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b fix-already-included-file -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


ivadym (approved):
Well done, nice debugging!!! 💯 

Do you think we should completely get rid of the `include_file()` method too? This combination of using `add_file` first and then `include_file` is also seen in functions like `add_scout_config_to_hk`, for instance.

islean:
> Well done, nice debugging!!! 💯
> 
> Do you think we should completely get rid of the `include_file()` method too? This combination of using `add_file` first and then `include_file` is also seen in functions like `add_scout_config_to_hk`, for instance.

Good catch! Might be some more instances in the flow that I have missed, will have to test it! I don't think we should completely remove `include_file()` yet, though. I think there are a number of entries which perhaps have not been included since the change was done quite recently. Also, `add_file()` has as option to keep the full path. If used, I guess include file could be of use at some later stage.

islean (commented):
The Housekeeper change was done in the CLI, which this does not use. So the fix will probably need some other action.

Issue #2350, opened 2023-08-18
Inconsistent values for q30 metrics


seallard:
## Description
In the sequencing metrics table, we have inconsistent data in the column called `sample_base_fraction_passing_q30`. Most of the entries are reported as percentages, but all new ones are reported as fractions. The discrepancy arose due to the q30 scores ported from cgstats being stored as percentages.

## Suggested solution
Store the new q30 scores as percentages as well and transform existing fractions to percentages.

1. Create revision renaming the column to `sample_base_percentage_passing_q30` or maybe just `base_percentage_passing_q30`.
2. Update logic to populate the column with percentages instead of fractions.
3. Update NIP q30 threshold to percentage.
4. Update q30 filtering in store.
5. Update lims usage of metrics.
6. Transform existing fractions to percentages.

## This can be closed when
- The q30 scores are stored and used assuming a single format

seallard:
Given that there might be current and future dependencies on this value assuming it is reported as a percentage, maybe the least invasive/error prone approach would be to store it as a percentage. What do you think?

ChrOertlin:
From what I see we use the bases passing QC in the following places:
cg/apps/sequencing_metrics_parser/parsers/bcl_convert.py - parses the q30 values directly from file which are in fractions
cg/constants/sequencing.py - Enum class where Q30thresholds are stored as INTs (i.e. percentages)
cg/store/filters/status_metrics_filters.py - Checking if somehting passes a q30 thresholds, takes the sequencing constants above and converts them to fractions
cg/constants/nipt.py - Constant Q30 threshold stored as fraction

We use the Q30_THRESHOLD to determine whether a flow cell should be stored in CGstats currently and in NIPT to check whether to upload. Is this right @karlnyr 

# approach 1

1. Transform all fractions to percentages in SampleLaneSequencingMetrics
2. Change the column name in the SampleLaneSequencingMetrics table to display percentage
3. Transfrom the  parsed fractions %Q30 in cg/apps/sequencing_metrics_parser/parsers/bcl_convert.py to percentages (possibly also %Q30 in cg/apps/sequencing_metrics_parser/parsers/bcl_convert.py)
4. Alter the cg/store/filters/status_metrics_filters.py  to not transform the Q30 threshold used therein
5. Alter the NIPT Q30 Threshold from fraction to percentage cg/constants/nipt.py 

# approach 2: 

1. When CGstats is deprecated alter cg/constants/sequencing.py  to fractions
[...]

PR #2352, opened 2023-08-18
Patch q30 metrics

seallard:
## Description
This PR ensures that q30 values are stored and used as percentages throughout cg. Resolves https://github.com/Clinical-Genomics/cg/issues/2350.

### Fixed
- Rename q30 metrics column
- Patch logic related to q30 thresholds to treat the metrics as percentages


cg/constants/nipt.py
```diff
@@ -1 +1 @@
-Q30_THRESHOLD = 0.88
+Q30_THRESHOLD = 88
```
seallard, line 1: This seems like a bad idea, it is used across many files. We should keep it as is and only adjust it for the specific comparison against the metrics table q30 values

seallard: Only used for calls to a single function which has been patched

alembic/versions/5cb3ce4c3e39_rename_q30_metrics_column.py
```diff
@@ -0,0 +1,32 @@
...
+down_revision = "201b16c45366"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.alter_column(
+        "sample_lane_sequencing_metrics",
+        "sample_base_fraction_passing_q30",
+        new_column_name="sample_base_percentage_passing_q30",
```
ChrOertlin, line 23:
```suggestion
        new_column_name="base_percentage_passing_q30",
```
I think this could be informative enough, the rows are samples in the table.

seallard: I thought that too, maybe better in a separate PR where we rename the other fields as well?

ChrOertlin: yes sounds great! It is nothing super urgent anyway.

ChrOertlin (approved):
Looks good, you changed the parsing, nipt, storing fastq in hk, the filter and the table name. I think that was all on the list! Nice!

Vince-janv (approved):
Nice, looks like you got all bases covered 🔥

diitaz93 (approved):
Looks good 👍

seallard:
Patched data in stage: 
```
UPDATE sample_lane_sequencing_metrics
SET sample_base_percentage_passing_q30 = sample_base_percentage_passing_q30 * 100
WHERE created_at > '2023-07-18'
AND sample_base_percentage_passing_q30 <= 1;
```

seallard:
Deployed to hasta and applied revision. 
Deploy to vms fails.

seallard:
Patched data in prod:
```
UPDATE sample_lane_sequencing_metrics
SET sample_base_percentage_passing_q30 = sample_base_percentage_passing_q30 * 100
WHERE created_at > '2023-07-18'
AND sample_base_percentage_passing_q30 <= 1;
```
Which updated 6982 entries.

Validated with
```
SELECT *
FROM sample_lane_sequencing_metrics
WHERE created_at > '2023-07-18'
AND sample_base_percentage_passing_q30 <= 1;
```

Which returned 13 entries.

PR #2358, opened 2023-08-24
Update to pydantic v2 in apps/orderform

islean:
## Description

This PR updates apps/orderform to use pydantic v2. The main difference is in how the fields in the base models are validated. There are two ways, one that I could not get working until now, using methods decorated with field_validator instead of validator as before. This would be the closest to how it looked before migration but has the drawback that one can not see the validators by just looking at the field declaration.

The second option would be the one currently implemented, where the validators are moved out of the base model and instead annotated in conjunction with the field declaration. It is further from how it looked before but has the benefit of visualising in a much clearer way which validators that are applied to a field and in what order.

The behaviour and validations should work completely the same, though, and it would only be a visual change.

Also, behaviour of optionals in base models have changed and defaults are required for Optional values, see https://docs.pydantic.dev/latest/migration/#required-optional-and-nullable-fields.

Examples below although the Annotated way really comes into its own when there are larger data classes such as the ExcelSample class in the Files changed.

```Python
from pydantic import BaseModel, Field, field_validator, AfterValidator
from typing_extensions import Annotated


# VERSION 1 using the field validator property

class DummyModel(BaseModel):
    field_with_alias: str

    @field_validator("field_with_alias", mode='after')
    @classmethod
    def validate_field(cls, value):
        assert value == "Expected value"


# VERSION 2 where the validator is a function and the field is annotated with the validator. 
# I like that you can see immediately that field_with_alias will be validated with 'validate_field'.

def validate_field(value):
    if value != "Expected value":
        raise ValueError("Not the expected value!")


class OtherDummyModel(BaseModel):
    field_with_alias: Annotated[str, AfterValidator(validate_field)]

```

### Changed

- Basemodels and parsers in apps/orderform
- Validators are now outside of the base class and used as annotated validators à la https://docs.pydantic.dev/latest/usage/validators/#annotated-validators.

### How to prepare for test

- [x] Use stage: `us`
- [x] Install on cg-vm1
### How to test

- [x] Upload orders in the old order portal.

### Expected test outcome

- [x] Check that the orders are parsed properly

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/models/orders/json_sample.py
```diff
@@ -1,33 +1,32 @@
...
-from pydantic.v1 import constr, validator
+from pydantic import BeforeValidator, constr
+from typing_extensions import Annotated
+
+
+def join_list(value: Any):
+    return "".join(value) if isinstance(value, list) else value
+
+
+def convert_well(value: str):
```
ChrOertlin, line 13: Can we add a docstring what it is we are converting more specifcally? What should the layout be of a "converted_well" ?

cg/models/orders/json_sample.py
```diff
@@ -1,33 +1,32 @@
...
+from typing import Any, List, Optional
 
 from cg.constants import DataDelivery, Pipeline
 from cg.models.orders.sample_base import OrderSample
-from pydantic.v1 import constr, validator
+from pydantic import BeforeValidator, constr
+from typing_extensions import Annotated
+
+
+def join_list(value: Any):
```
ChrOertlin, line 9: What kind of lists are we joining? Maybe clarify in docstring

cg/models/orders/excel_sample.py
```diff
@@ -1,139 +1,147 @@
...
+    value = value.strip()
+    return REV_SEX_MAP.get(value, "unknown")
+
+
+def convert_to_lower(value: Optional[str]):
+    return value.lower()
+
+
+def convert_to_priority(value: Optional[str]):
+    return "priority" if value == "förtur" else value
```
ChrOertlin, line 68:
```suggestion
    return "priority" if value == "förtur" else value
```

I think we would want an english named value here. Can we create a constant for "förtur" which is its english translation?

cg/models/orders/excel_sample.py
```diff
@@ -1,139 +1,147 @@
...
+    if str_value.replace(".", "").isnumeric():
+        return str_value
+    raise AttributeError(f"Order contains non-numeric value '{value}'")
+
+
+def validate_parent(value: str):
+    return None if value == "0.0" else value
+
+
+def validate_source(value: Optional[str]):
```
ChrOertlin, line 50:
```suggestion
def validate_source(source: Optional[str]):
```

cg/models/orders/excel_sample.py
```diff
@@ -1,139 +1,147 @@
...
+    if value not in SOURCE_TYPES:
+        raise ValueError(f"'{value}' is not a valid source")
+    return value
+
+
+def convert_sex(value: Optional[str]):
```
ChrOertlin, line 56:
```suggestion
def convert_sex(sex: Optional[str]):
```

cg/models/orders/excel_sample.py
islean: I don't know if there is a functioning translator for arbitrary strings? And I think "förtur" is the only Swedish word that seems to be supported when filling out certain forms, allowed values are in the PriorityEnum, i.e. "research, standard, priority, express or clinical_trials". So if none of those are specified we throw an error (unless "förtur" is specified which is converted to priority). But a constant should be doable.

cg/models/orders/excel_sample.py
```diff
@@ -1,139 +1,147 @@
 from typing import List, Optional
-from pydantic.v1 import Field, validator
 
 from cg.constants.orderforms import REV_SEX_MAP, SOURCE_TYPES
 from cg.models.orders.sample_base import OrderSample
+from pydantic import AfterValidator, BeforeValidator, Field
+from typing_extensions import Annotated
+
+
+def parse_panels(value: str):
```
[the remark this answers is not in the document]
islean: Nice! With field_validator I think we are coerced into using value, but with the annotated way we can probably rename the argument :)

ChrOertlin (approved):
👍

cg/apps/orderform/excel_orderform_parser.py
```diff
@@ -221,7 +213,8 @@ def parse_orderform(self, excel_path: str) -> None:
         if not raw_samples:
             raise OrderFormError("orderform doesn't contain any samples")
 
-        self.samples: List[ExcelSample] = parse_obj_as(List[ExcelSample], raw_samples)
+        excel_sample_list_validator = TypeAdapter(List[ExcelSample])
```
seallard, line 216: From the docs, it seems like `TypeAdapter` is used to parse and validate objects which are not instantiated from a `BaseModel`. Is that right? So the use case is for when we have non-pydantic models that we want to validate?

https://docs.pydantic.dev/2.0/migration/#introduction-of-typeadapter

islean: I think it should be fine. They specifically list List[SomeModel] as a candidate to validate with TypeAdapter, which seems to be intended for all classes that need some validation:

"You may have types that are not BaseModels that you want to validate data against. Or you may want to validate a List[SomeModel], or dump it to JSON.

For use cases like this, Pydantic provides TypeAdapter, which can be used for type validation, serialization, and JSON schema generation without creating a BaseModel."

https://docs.pydantic.dev/latest/usage/type_adapter/

cg/models/orders/excel_sample.py
```diff
@@ -1,139 +1,148 @@
...
+    return REV_SEX_MAP.get(sex, "unknown")
+
+
+def convert_to_lower(value: Optional[str]):
+    return value.lower()
+
+
+def convert_to_priority(priority: Optional[str]):
+    """Translates the Swedish 'förtur' to 'priority' if specified in the order."""
+    return PriorityEnum.priority if priority == "förtur" else priority
```
seallard, line 69: The old function used `priority.lower()`, consider keeping the old logic unless it does not matter

seallard: Oh, you extracted it to an after validator

seallard (approved):
Nice 👍  The diff was a bit hard to read due to the restructuring but looks good

islean:
Tests worked, orderforms were filled in correctly:

Screenshot 2023-08-31 at 09 39 28
Screenshot 2023-08-31 at 09 39 41

Issue #2362, opened 2023-08-25
Finding fastq files patterns for BCLConvert pattern without the moving of files


karlnyr:
## Description

With the deprecation of the demultiplexing repo we are not moving files anymore, we need to support a new file path pattern to find fastq files.

## Suggested solution

Support the following pattern:

```Python
bcl_convert_pattern = f"Unaligned*/*/Sample_{sample_internal_id}_*{FileExtensions.FASTQ}{FileExtensions.GZIP}"
```

## This can be closed when

We can parse flow cells that has been demultiplexed with BCLConvert without any files being moved.

## Blocked by

*NA*

karlnyr:
Directory with the structure

```Command
/home/proj/production/demultiplexed-runs/230822_A00621_0892_AHGTHYDSX7/
```

seallard:
Refinement: 
- Should we match sample fastqs with and without the `Sample_` prefix?
- Should we match sample fastqs with and without a trailing sequence?

henrikstranneheim:
@seallard Fixed?

seallard:
> @seallard Fixed?

@ChrOertlin did you take over this?

PR #2363, opened 2023-08-25
#2362: Patch new flow cell dir structure

seallard:
## Description
After removing the demux repo (?) we no longer restructure flow cells. We need to handle the pattern described in https://github.com/Clinical-Genomics/cg/issues/2362 to find the fastq files in a flow cell directory.

### Fixed
- Support new flow cell dir structure in old and new post processing flow

### DIFF
--- a/cg/apps/cgstats/stats.py
+++ b/cg/apps/cgstats/stats.py
@@ -151,7 +151,7 @@ class StatsAPI(alchy.Manager):
         alt_pattern = "*{}/Unaligned*/Project_*/Sample_{}_*/*.fastq.gz"
 
         # The default structure for flow cells demultiplexed with bclconvert
-        bcl_convert_pattern = "*{}/Unaligned*/*/Sample_{}_*/*.fastq.gz"
+        bcl_convert_pattern = "*{}/Unaligned*/*/{}_*/*.fastq.gz"
 
         for fastq_pattern in (base_pattern, alt_pattern, bcl_convert_pattern):
             pattern = fastq_pattern.format(flowcell, sample_obj.samplename)
--- a/cg/apps/cgstats/stats.py
+++ b/cg/apps/cgstats/stats.py
@@ -142,7 +142,7 @@ class StatsAPI(alchy.Manager):
         return flow_cell_reads_and_q30_summary
 
     def fastqs(self, flowcell: str, sample_obj: Sample) -> Iterator[Path]:
-        """Fetch FASTQ files for a sample."""
+        """Retrieve all fastq files for a specific sample in a flow cell directory."""
 
         # The default structure for flow cells demultiplexed with bcl2fastq
         base_pattern = "*{}/Unaligned*/Project_*/Sample_{}/*.fastq.gz"
--- a/cg/apps/cgstats/stats.py
+++ b/cg/apps/cgstats/stats.py
@@ -145,6 +145,8 @@ class StatsAPI(alchy.Manager):
         """Fetch FASTQ files for a sample."""
         base_pattern = "*{}/Unaligned*/Project_*/Sample_{}/*.fastq.gz"
         alt_pattern = "*{}/Unaligned*/Project_*/Sample_{}_*/*.fastq.gz"
-        for fastq_pattern in (base_pattern, alt_pattern):
+        bcl_convert_pattern = "*{}/Unaligned*/*/Sample_{}_*/*.fastq.gz"
+
+        for fastq_pattern in (base_pattern, alt_pattern, bcl_convert_pattern):
             pattern = fastq_pattern.format(flowcell, sample_obj.samplename)
             yield from self.root_dir.glob(pattern)
--- a/cg/apps/cgstats/stats.py
+++ b/cg/apps/cgstats/stats.py
@@ -143,8 +143,14 @@ class StatsAPI(alchy.Manager):
 
     def fastqs(self, flowcell: str, sample_obj: Sample) -> Iterator[Path]:
         """Fetch FASTQ files for a sample."""
+
+        # The default structure for flow cells demultiplexed with bcl2fastq
         base_pattern = "*{}/Unaligned*/Project_*/Sample_{}/*.fastq.gz"
+
+        # Alternative structure for flow cells demultiplexed with bcl2fastq and where the sample has a trailing sequence
         alt_pattern = "*{}/Unaligned*/Project_*/Sample_{}_*/*.fastq.gz"
+
+        # The default structure for flow cells demultiplexed with bclconvert
         bcl_convert_pattern = "*{}/Unaligned*/*/Sample_{}_*/*.fastq.gz"
 
         for fastq_pattern in (base_pattern, alt_pattern, bcl_convert_pattern):
--- a/cg/apps/cgstats/stats.py
+++ b/cg/apps/cgstats/stats.py
@@ -147,7 +147,7 @@ class StatsAPI(alchy.Manager):
         # The default structure for flow cells demultiplexed with bcl2fastq
         base_pattern = "*{}/Unaligned*/Project_*/Sample_{}/*.fastq.gz"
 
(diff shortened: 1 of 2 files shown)


cg/apps/cgstats/stats.py
```diff
@@ -145,6 +145,8 @@ def fastqs(self, flowcell: str, sample_obj: Sample) -> Iterator[Path]:
         """Fetch FASTQ files for a sample."""
         base_pattern = "*{}/Unaligned*/Project_*/Sample_{}/*.fastq.gz"
         alt_pattern = "*{}/Unaligned*/Project_*/Sample_{}_*/*.fastq.gz"
-        for fastq_pattern in (base_pattern, alt_pattern):
+        bcl_convert_pattern = "*{}/Unaligned*/*/Sample_{}_*/*.fastq.gz"
```
seallard, line 148: We never added the pattern for novaseqx pattern in the old flow. I guess that is fine if we ever manage to delete the old flow 😅

seallard, line 148: Should we have a pattern for matching fastq files without a trailing sequence in the new structure as well?

diitaz93: So what this does is introduces another folder structure type in the list of valid structures and if it matches with it assigns a bcl_converter value?

diitaz93: I don't have knowledge to answer that

seallard: Oh, now I understand. So the function applies all the patterns and return any file path which match any of the patterns

cg/apps/cgstats/stats.py
```diff
@@ -145,6 +145,8 @@ def fastqs(self, flowcell: str, sample_obj: Sample) -> Iterator[Path]:
         """Fetch FASTQ files for a sample."""
```
ChrOertlin, line 145: I think it would be helpful to identify what pattern is used and needed by what and update the docstring.
It will make our lives easier in future when we hopefully restructure this.

seallard: I agree, might be out of scope for this PR though. Do you have any ideas for a good doc string?

seallard: I think the doc string should not leak too much of the implementation details. But maybe something like
```
Retrieve all fastq files for a specific sample in a flow cell directory.
```

ChrOertlin: Something along the lines of:

root_pattern is used for the flat output structure for the novaseqx sequencer types
unaligned pattern is used for the .... structure  for the xx sequencer types 
etc. 

I think as soon as we have implemented the new flow properly we can revisit the patterns and remove them. So I guess we either do it now or later. Up to you I suppose

seallard: Alternatively, we could refactor this function later to take in a parameter indicating which bcl software was used to make the logic clearer in the function

seallard: Yeah, I can try to add comments about that in the function

seallard: Added some comments to clarify what the different patterns represent, check whether they make sense - I'm not familiar with the different patterns

seallard: Updated the doc strings to be clearer

seallard: I think these changes are out of scope of this PR but might as well do it now 👍

cg/apps/cgstats/stats.py
ChrOertlin: I guess this would make sense if we encounter such patterns for fastq files. Maybe we can refine the issue given a bit further on what the expected fastq files can be currently. Then we can tailer our patterns specifically to that.

cg/meta/demultiplex/utils.py
```diff
@@ -38,12 +38,23 @@ def get_lane_from_sample_fastq(sample_fastq_path: Path) -> int:
 def get_sample_fastqs_from_flow_cell(
```
seallard, line 38: This function should probably be split into several smaller ones. Like `fetch_fastq_files_for_bcl_convert_flow_cell` and yada yada. Something to do later

cg/apps/cgstats/stats.py
seallard: I think @karlnyr has the answer 🙏

cg/meta/demultiplex/utils.py
```diff
@@ -38,12 +38,23 @@ def get_lane_from_sample_fastq(sample_fastq_path: Path) -> int:
...
+
+    # The default structure for flow cells demultiplexed with bcl2fastq
     unaligned_pattern = f"Unaligned*/Project_*/Sample_{sample_internal_id}/*{FileExtensions.FASTQ}{FileExtensions.GZIP}"
+
+    # Alternative structure for flow cells demultiplexed with bcl2fastq where the sample fastq files have a trailing sequence
     unaligned_alt_pattern = f"Unaligned*/Project_*/Sample_{sample_internal_id}_*/*{FileExtensions.FASTQ}{FileExtensions.GZIP}"
 
-    for pattern in [root_pattern, unaligned_pattern, unaligned_alt_pattern]:
+    # The default structure for flow cells demultiplexed with bclconvert
+    bcl_convert_pattern = (
```
seallard, line 53: Consider adding a test case for this flow cell directory structure

cg/apps/cgstats/stats.py
```diff
@@ -142,9 +142,17 @@ def flow_cell_reads_and_q30_summary(self, flow_cell_name: str) -> Dict[str, Unio
...
+
+        # The default structure for flow cells demultiplexed with bcl2fastq
         base_pattern = "*{}/Unaligned*/Project_*/Sample_{}/*.fastq.gz"
+
+        # Alternative structure for flow cells demultiplexed with bcl2fastq where the sample fastq files have a trailing sequence
         alt_pattern = "*{}/Unaligned*/Project_*/Sample_{}_*/*.fastq.gz"
-        for fastq_pattern in (base_pattern, alt_pattern):
+
+        # The default structure for flow cells demultiplexed with bclconvert
+        bcl_convert_pattern = "*{}/Unaligned*/*/Sample_{}_*/*.fastq.gz"
```
Vince-janv, line 154: Looking at the example @karlnyr provided I think the pattern will not catch them. Here's an example path: `/home/proj/production/demultiplexed-runs/230822_A00621_0892_AHGTHYDSX7/Unaligned/151250/ACC12542A6_S14_L001_R1_001.fastq.gz`. @karlnyr is this what we want to support?

karlnyr: yes - that is what I want to support

cg/meta/demultiplex/utils.py
seallard: Do in separate PR, no existing tests

cg/apps/cgstats/stats.py
seallard: Adjusted the pattern, should  match the provided example path now

cg/apps/cgstats/stats.py
seallard: Not relevant based on the example file path provided in the issue

cg/apps/cgstats/stats.py
Vince-janv: This looks good 🔥

Vince-janv (approved):
Hopefully this is the last pattern we need to add 🤞

islean:
Are we sure it matches? Tried Path(Kalles_path).match(pattern) and it returned false but maybe I am using the functions wrong

islean:
I.e. this returns false
```Python
print(PurePath("/home/proj/production/demultiplexed-runs/230822_A00621_0892_AHGTHYDSX7/Unaligned/151250/ACC12542A6_S14_L001_R1_001.fastq.gz").match(path_pattern="*{}/Unaligned*/*/{}_*/*.fastq.gz"))
```

PR #2364, opened 2023-08-25
Patch path to sample sheet in hk

diitaz93:
## Description

The path to the sample sheet used in the function `store_flow_cell_data` (called when finishing all flow cells) refers to a no longer present file. 

The fix is to replace it with the path to the sample sheet in Housekeeper

### Added

-

### Changed

- The referring to the sample sheet in the demux_runs directory to the sample sheet in Housekeeper

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b fix-sample-sheet-finish -a
    ```

### How to test

- [ ] Do `cg demultiplex finish temporary`

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -40,8 +42,12 @@ def store_flow_cell_data_in_status_db(
         LOG.info(f"Flow cell added to status db: {parsed_flow_cell.id}.")
     else:
         LOG.info(f"Flow cell already exists in status db: {parsed_flow_cell.id}.")
+    sample_sheet: File = get_sample_sheets_from_latest_version(
+        flow_cell_id=parsed_flow_cell.id,
+        hk_api=hk_api,
+    )[0]
```
diitaz93, line 48: I am thinking of what happens if the sample sheet is not in HK (unlikely). This would give a not-so meaningful Housekeeper exception. Should I make a try-except statement around the sample sheet fetching?

ChrOertlin: Yes I think that makes sense here.

ChrOertlin (approved):
Adding a try except with an informative message would be good. Otherwise looks good. It should fix the error of looking for the sample sheet in the wrong place.

islean (approved):
Looks good 🔥

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -141,7 +140,9 @@ def finish_all_flow_cells_temp(self) -> bool:
 
     def store_flow_cell_data(self, parsed_flow_cell: FlowCellDirectoryData) -> None:
         """Store data from the flow cell directory in status db and housekeeper."""
-        store_flow_cell_data_in_status_db(parsed_flow_cell=parsed_flow_cell, store=self.status_db)
+        store_flow_cell_data_in_status_db(
```
Vince-janv, line 143: I don't like involving housekeeper in a function related to status_db. Maybe you could fetch the samplesheet early in the finish command set it to an attribute of parsed_flow_cell?

cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -40,8 +43,16 @@ def store_flow_cell_data_in_status_db(
         LOG.info(f"Flow cell added to status db: {parsed_flow_cell.id}.")
     else:
         LOG.info(f"Flow cell already exists in status db: {parsed_flow_cell.id}.")
+
+    try:
+        sample_sheet: File = get_sample_sheets_from_latest_version(
```
Vince-janv, line 48: Then this part would not be needed

Vince-janv (approved):
I have a comment regarding where the sample_sheet is fetched, but since this needs to be done ASAP I think you can address it in small PR later

cg/meta/demultiplex/status_db_storage_functions.py
diitaz93: added

cg/meta/demultiplex/demux_post_processing.py
diitaz93: fixed

diitaz93:
Tested on stage: `cg demultiplex finish temporary-all` finished successfully for the desired flow cells. Omitting output.

diitaz93:
Deployed to stage and prod

PR #2379, opened 2023-08-28
Migrate trailblazer pydantic models

seallard:
## Description
This PR migrates the Trailblazer Pydantic model to V2. Part of project https://github.com/Clinical-Genomics/project-planning/issues/514. 

The `case_id` is not using the pattern me and @islean settled on (`Annotated` with a validator in a separate file). This is because it inherits the `family` value, probably a temporary patch made in the past with the idea of renaming the field. The new way of accessing previously validated fields uses the `FieldValidationInfo` attribute ([docs](https://docs.pydantic.dev/2.0/migration/#changes-to-validators-allowed-signatures)).

### Fixed
- Replace deprecated `parse_obj` with `model_validate`
- Replace deprecated `validator` decorator with `Annotated` and `field_validator`

### DIFF
--- a/cg/apps/tb/models.py
+++ b/cg/apps/tb/models.py
@@ -1,6 +1,8 @@
+import datetime as dt
+from pathlib import Path
 from typing import Optional
 
-from pydantic import AfterValidator, BaseModel, ConfigDict, FieldValidationInfo, field_validator
+from pydantic import BaseModel, BeforeValidator, ConfigDict, FieldValidationInfo, field_validator
 from typing_extensions import Annotated
 
 from cg.apps.tb.validators import parse_str_to_datetime, parse_str_to_path
@@ -17,13 +19,13 @@ class TrailblazerAnalysis(BaseModel):
         return info.data.get("family")
 
     version: Optional[str] = None
-    logged_at: Annotated[Optional[str], AfterValidator(parse_str_to_datetime)]
-    started_at: Annotated[Optional[str], AfterValidator(parse_str_to_datetime)]
-    completed_at: Annotated[Optional[str], AfterValidator(parse_str_to_datetime)]
+    logged_at: Annotated[Optional[dt.datetime], BeforeValidator(parse_str_to_datetime)]
+    started_at: Annotated[Optional[dt.datetime], BeforeValidator(parse_str_to_datetime)]
+    completed_at: Annotated[Optional[dt.datetime], BeforeValidator(parse_str_to_datetime)]
     status: Optional[str] = None
     priority: Optional[str] = None
-    out_dir: Annotated[Optional[str], AfterValidator(parse_str_to_path)]
-    config_path: Annotated[Optional[str], AfterValidator(parse_str_to_path)]
+    out_dir: Annotated[Optional[Path], BeforeValidator(parse_str_to_path)]
+    config_path: Annotated[Optional[Path], BeforeValidator(parse_str_to_path)]
     comment: Optional[str] = None
     is_deleted: Optional[bool] = None
     is_visible: Optional[bool] = None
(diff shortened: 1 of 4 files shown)


islean:
Still slightly on the fence regarding which validation scheme to use. I think in general, the annotated way is cleaner, but it seems as if quite some functionality goes away since (as the documentation also states) the validation concerns a type rather than the base model in question. In other words, model validators and field validators which take other fields into account can not be used that way. 

Although the annotated way is cleaner - should we aim to keep it consistent by instead using the decorated way? Or are the identified cases where Annotated doesn't work too few to warrant the switch?

cg/apps/tb/models.py
```diff
@@ -1,49 +1,41 @@
...
-    config_path: Optional[str]
-    comment: Optional[str]
-    is_deleted: Optional[bool]
-    is_visible: Optional[bool]
-    type: Optional[str]
-    user_id: Optional[int]
+    case_id: Optional[str] = None
+
+    @field_validator("case_id")
+    @classmethod
```
islean, line 15: Nice! Found out in a previous PR that if the ordering of the decorators are switched, then the validation never takes place. So this is the correct way!

cg/apps/tb/models.py
```diff
@@ -1,49 +1,41 @@
...
+    def inherit_family_value(cls, value: str, info: FieldValidationInfo) -> str:
+        return info.data.get("family")
+
+    version: Optional[str] = None
+    logged_at: Annotated[Optional[str], AfterValidator(parse_str_to_datetime)]
+    started_at: Annotated[Optional[str], AfterValidator(parse_str_to_datetime)]
+    completed_at: Annotated[Optional[str], AfterValidator(parse_str_to_datetime)]
```
islean, line 22: I feel like it would make more sense to have both started_at and completed_at type hinted as `Optional[datetime.datetime]` and use a BeforeValidator instead. But given that this was not how it was done before, perhaps that should be avoided in this PR. Won't they always be returned as datetime.datetime when accessed though?

cg/apps/tb/models.py
```diff
@@ -1,49 +1,41 @@
...
+    status: Optional[str] = None
+    priority: Optional[str] = None
+    out_dir: Annotated[Optional[str], AfterValidator(parse_str_to_path)]
+    config_path: Annotated[Optional[str], AfterValidator(parse_str_to_path)]
```
islean, line 26: Same thing with out_dir and config_path where I think it would make more sense to have a BeforeValidator converting the values to Paths and type hinting them as such.

islean (approved):
Looks nice! Some comments that are potentially out of scope for this PR.

seallard:
> Still slightly on the fence regarding which validation scheme to use. I think in general, the annotated way is cleaner, but it seems as if quite some functionality goes away since (as the documentation also states) the validation concerns a type rather than the base model in question. In other words, model validators and field validators which take other fields into account can not be used that way.
> 
> Although the annotated way is cleaner - should we aim to keep it consistent by instead using the decorated way? Or are the identified cases where Annotated doesn't work too few to warrant the switch?

I'm still not too familiar with Pydantic or the changes introduced in the new version. I think these edge cases are rare and maybe it is OK to stick to the `Annotated` pattern as much as possible? I have not refactored so many models yet though

cg/apps/tb/models.py
seallard: Good to know, was not aware it breaks

cg/apps/tb/models.py
seallard: Yeah, it is true that they will be returned as datetimes when accessed. I think having a different type hint is confusing and maybe this is something that should be refactored

cg/apps/tb/models.py
seallard: Nice 👍

seallard:
TODO:

Deployed to stage:
```
```

Tested with `cg workflow rnafusion store <case_id>` which instantiates the model.

diitaz93 (approved):
Great! 👍

PR #2384, opened 2023-08-29
fix(rename fastq files in new demux post processing flow)

ChrOertlin:
## Description

This PR aims to address the issue that fastq files paths were are non unique across flow cells in housekeeper when using the new demultiplexing post processing flow.

This PR introduces a change to the new demultiplexing post processing flow, it now appends the flow cell name to the fastq files for a given sample.

It renames the provided fastq files on the server to the new fastq file name with the appended flow cell name.

intended new name: `<parent_dir>/<flow_cell_name>_<sample_fastq_and_extensions>`

### Added

- function `append_flow_cell_name_to_fastq_file_path`. This function appends the flow cell name to a fastq file path.

- function `rename_file` this function renames a given file to a provided file name

- function `is_pattern_in_file` this function checks if a string is in a file name returns a bool

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do `cg demultiplex finish temporary` Check that the fastq files the samples are renamed, check that the housekeepr entries have the flow cell name appended to the fastq files for a sample bundle
### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/demultiplex/housekeeper_storage_functions.py
+++ b/cg/meta/demultiplex/housekeeper_storage_functions.py
@@ -17,7 +17,7 @@ from cg.meta.demultiplex.utils import (
     get_q30_threshold,
     get_sample_fastqs_from_flow_cell,
     get_sample_sheet_path,
-    append_flow_cell_name_to_fastq_file_path,
+    add_flow_cell_name_to_fastq_file_path,
 )
 from cg.models.demultiplex.flow_cell import FlowCellDirectoryData
 from cg.store import Store
@@ -89,13 +89,15 @@ def add_sample_fastq_files_to_housekeeper(
             continue
 
         for sample_fastq_path in sample_fastq_paths:
-            fastq_file_path_appended: Path = append_flow_cell_name_to_fastq_file_path(
+            renamed_fastq_file_path: Path = add_flow_cell_name_to_fastq_file_path(
                 fastq_file_path=sample_fastq_path, flow_cell_name=flow_cell.id
             )
-            rename_file(file_path=sample_fastq_path, renamed_file_path=fastq_file_path_appended)
+            if not renamed_fastq_file_path.exists():
+                rename_file(file_path=sample_fastq_path, renamed_file_path=renamed_fastq_file_path)
+                sample_fastq_path = renamed_fastq_file_path
             store_fastq_path_in_housekeeper(
                 sample_internal_id=sample_internal_id,
-                sample_fastq_path=fastq_file_path_appended,
+                sample_fastq_path=sample_fastq_path,
                 flow_cell=flow_cell,
                 hk_api=hk_api,
                 store=store,
--- a/cg/meta/demultiplex/housekeeper_storage_functions.py
+++ b/cg/meta/demultiplex/housekeeper_storage_functions.py
@@ -17,11 +17,11 @@ from cg.meta.demultiplex.utils import (
     get_q30_threshold,
     get_sample_fastqs_from_flow_cell,
     get_sample_sheet_path,
-    add_flow_cell_name_to_fastq_file_path,
+    rename_fastq_file_if_needed,
 )
 from cg.models.demultiplex.flow_cell import FlowCellDirectoryData
 from cg.store import Store
-from cg.utils.files import get_files_matching_pattern, rename_file
+from cg.utils.files import get_files_matching_pattern
 
 LOG = logging.getLogger(__name__)
 
@@ -89,12 +89,9 @@ def add_sample_fastq_files_to_housekeeper(
             continue
 
         for sample_fastq_path in sample_fastq_paths:
-            renamed_fastq_file_path: Path = add_flow_cell_name_to_fastq_file_path(
+            sample_fastq_path: Path = rename_fastq_file_if_needed(
                 fastq_file_path=sample_fastq_path, flow_cell_name=flow_cell.id
             )
-            if not renamed_fastq_file_path.exists():
-                rename_file(file_path=sample_fastq_path, renamed_file_path=renamed_fastq_file_path)
-                sample_fastq_path = renamed_fastq_file_path
             store_fastq_path_in_housekeeper(
                 sample_internal_id=sample_internal_id,
                 sample_fastq_path=sample_fastq_path,
--- a/cg/meta/demultiplex/housekeeper_storage_functions.py
+++ b/cg/meta/demultiplex/housekeeper_storage_functions.py
@@ -17,6 +17,8 @@ from cg.meta.demultiplex.utils import (
     get_q30_threshold,
     get_sample_fastqs_from_flow_cell,
     get_sample_sheet_path,
+    append_flow_cell_name_to_fastq_file_path,
+    rename_fastq_file,
 )
 from cg.models.demultiplex.flow_cell import FlowCellDirectoryData
 from cg.store import Store
@@ -88,9 +90,15 @@ def add_sample_fastq_files_to_housekeeper(
             continue
 
         for sample_fastq_path in sample_fastq_paths:
+            fastq_file_path_appended: Path = append_flow_cell_name_to_fastq_file_path(
+                fastq_file_path=sample_fastq_path, flow_cell_name=flow_cell.id
+            )
+            rename_fastq_file(
+                fastq_file_path=sample_fastq_path, renamed_fastq_file_path=fastq_file_path_appended
+            )
             store_fastq_path_in_housekeeper(
                 sample_internal_id=sample_internal_id,
-                sample_fastq_path=sample_fastq_path,
+                sample_fastq_path=fastq_file_path_appended,
                 flow_cell=flow_cell,
                 hk_api=hk_api,
                 store=store,
--- a/cg/meta/demultiplex/housekeeper_storage_functions.py
+++ b/cg/meta/demultiplex/housekeeper_storage_functions.py
@@ -18,11 +18,10 @@ from cg.meta.demultiplex.utils import (
     get_sample_fastqs_from_flow_cell,
     get_sample_sheet_path,
     append_flow_cell_name_to_fastq_file_path,
-    rename_fastq_file,
 )
 from cg.models.demultiplex.flow_cell import FlowCellDirectoryData
 from cg.store import Store
-from cg.utils.files import get_files_matching_pattern
+from cg.utils.files import get_files_matching_pattern, rename_file
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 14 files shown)


karlnyr (approved):
Nice! Would appreciate that you did a functional test of a flow cell as well :)

ChrOertlin:
> Nice! Would appreciate that you did a functional test of a flow cell as well :)

We do have a functional test of the whole post processing flow in `tests/meta/demultiplexing/test_demultiplexing_post_processing.py`

cg/meta/demultiplex/utils.py
```diff
@@ -95,3 +95,8 @@ def parse_flow_cell_directory_data(
 ) -> FlowCellDirectoryData:
     """Return flow cell data from the flow cell directory."""
     return FlowCellDirectoryData(flow_cell_path=flow_cell_directory, bcl_converter=bcl_converter)
+
+
+def append_flow_cell_name_to_fastq_file_path(fastq_file_path: Path, flow_cell_name: str) -> Path:
```
Vince-janv, line 100:
```suggestion
def add_flow_cell_name_to_fastq_file_path(fastq_file_path: Path, flow_cell_name: str) -> Path:
```
Can we change this to add? I append just makes me think of lists

seallard, line 100: Might be worth checking whether the flow cell name already is in the fastq file path. We could run into issues where we rename the file multiple times. It is called idempotence, we want to be able to run the flow multiple times and have the same result

seallard, line 100: Append is a little misleading since that usually means attaching to the end, maybe prepend or add is less confusing

cg/utils/files.py
```diff
@@ -27,3 +31,14 @@ def get_files_matching_pattern(directory: Path, pattern: str) -> List[Path]:
...
+
+
+def rename_file(file_path: Path, renamed_file_path: Path) -> None:
+    """Rename the given fastq file path."""
+    if renamed_file_path.exists():
+        LOG.debug(
+            f"File {renamed_file_path} already exists. Skipping renaming of {renamed_file_path}."
+        )
+        return
+    shutil.move(src=file_path, dst=renamed_file_path)
```
Vince-janv, line 43: I think os.rename is more appropriate. I think shutil is more for large collections of files

cg/meta/demultiplex/housekeeper_storage_functions.py
```diff
@@ -88,9 +89,13 @@ def add_sample_fastq_files_to_housekeeper(
             continue
 
         for sample_fastq_path in sample_fastq_paths:
+            fastq_file_path_appended: Path = append_flow_cell_name_to_fastq_file_path(
```
Vince-janv, line 92: Same here regarding the use of appended. How about `modified_fastq_path` or `modified_fastq_file_path`?

tests/utils/test_files.py
```diff
@@ -9,3 +9,46 @@ def test_get_file_in_directory(nested_directory_with_file: Path, some_file: str)
...
+
+    # WHEN renaming the file
+    rename_file(file_path=file_path, renamed_file_path=renamed_file_path)
+
+    # THEN the renamed file path should exist and a debug message should be logged
+    assert renamed_file_path.exists()
+    assert (
+        f"File {renamed_file_path} already exists. Skipping renaming of {renamed_file_path}."
+        in caplog.text
+    )
```
seallard, line 54: Consider adding a test case for when the file we attempt to rename does not exist

seallard: It could just assert that a `FileNotFound` exception is thrown if that is the expected behaviour

Vince-janv (approved):
Nice 👍

cg/meta/demultiplex/utils.py
seallard: Maybe this is already done below 👍

seallard (approved):
Looks good 👍 

EDIT: I think there might be a bug in the case if the fastq file already has been renamed

cg/utils/files.py
seallard: Another option is to use Pathlib
```python
from pathlib import Path
src = Path("src.txt")
dst = Path("dst.txt")
src.rename(dst)
```

cg/meta/demultiplex/housekeeper_storage_functions.py
```diff
@@ -88,9 +89,13 @@ def add_sample_fastq_files_to_housekeeper(
...
+                fastq_file_path=sample_fastq_path, flow_cell_name=flow_cell.id
+            )
+            rename_file(file_path=sample_fastq_path, renamed_file_path=fastq_file_path_appended)
             store_fastq_path_in_housekeeper(
                 sample_internal_id=sample_internal_id,
-                sample_fastq_path=sample_fastq_path,
+                sample_fastq_path=fastq_file_path_appended,
```
seallard, line 98: This is a bug, if the file was previously renamed already we should not use the path returned from `append_flow_cell_name_to_fastq_file_path`

tests/utils/test_files.py
seallard, line 54: Consider adding a test case for what happens if a fastq file already has been renamed

cg/utils/files.py
```diff
@@ -27,3 +31,14 @@ def get_files_matching_pattern(directory: Path, pattern: str) -> List[Path]:
 def get_all_files_in_dir(base_path: Path) -> Set[Path]:
     """Get a set of all files relative to the given base path."""
     return {file.relative_to(base_path) for file in base_path.rglob("*") if file.is_file()}
+
+
+def rename_file(file_path: Path, renamed_file_path: Path) -> None:
+    """Rename the given fastq file path."""
+    if renamed_file_path.exists():
```
seallard, line 38: This might be a single responsibility violation, consider moving the existence check to before this function is called in the main logic

cg/meta/demultiplex/utils.py
ChrOertlin: fixed

ChrOertlin: fixed

ChrOertlin: fixed

cg/meta/demultiplex/housekeeper_storage_functions.py
ChrOertlin: fixed used `renamed` pre-fix instead of `modified` to stay in line with the rest of the function body

cg/utils/files.py
ChrOertlin: Since this function is for generic use I will keep `shutil` since as it is more flexible and allows us to use the function in more places. From what I found:

1 OS module might fail to move a file if the source and destination path is on different file systems or drive.
But shutil.move will not fail in this kind of cases.
2 shutil.move checks if the source and destination path are on the same file system or not. But os.rename does not check, thus it fails sometimes.
3 After checking the source and destination path, if it is found that they are not in the same file system, shutil.move will copy the file first to the destination. Then it will delete the file from the source file. Thus we can say shutil.move is a smarter method to move a file in Python when the source and destination path are not on the same drive or file system.
4 shutil.move works on high-level functions, while os.rename works on lower-level functions.

Do you agree?

Vince-janv: I think the cases you are referring are edge cases, for which we probably want it to fail (at least with the current usage). I vote using Path as it is a simple solution and a library most are familiar with.

ChrOertlin: Changed to using path.rename()

tests/utils/test_files.py
ChrOertlin: fixed

cg/meta/demultiplex/housekeeper_storage_functions.py
ChrOertlin: fixed

ChrOertlin:
# Test premise

# Flow cell HGW2VDRX3

## already processed once - changed a few file names to remove flow cell pre fix from them:
image

## After removing files from databases

### flow cell not in statusdb

image

### No entries in sampleLaneSequencing metrics
image

### No housekeeper files are present
image

cg/utils/files.py
```diff
@@ -27,3 +30,14 @@ def get_files_matching_pattern(directory: Path, pattern: str) -> List[Path]:
...
     return {file.relative_to(base_path) for file in base_path.rglob("*") if file.is_file()}
+
+
+def rename_file(file_path: Path, renamed_file_path: Path) -> None:
+    """Rename the given file path."""
+    file_path.rename(renamed_file_path)
+    LOG.debug(f"Renamed {file_path} to {renamed_file_path}.")
+
+
+def is_pattern_in_file_path(file_path: Path, pattern: str) -> bool:
```
seallard, line 41: If I just saw this function name, I would assume that it checks whether the parent directories contain the pattern as well. Not just the actual file name. Minor detail though, feel free to skip

cg/meta/demultiplex/housekeeper_storage_functions.py
```diff
@@ -88,6 +89,12 @@ def add_sample_fastq_files_to_housekeeper(
             continue
 
         for sample_fastq_path in sample_fastq_paths:
+            renamed_fastq_file_path: Path = add_flow_cell_name_to_fastq_file_path(
+                fastq_file_path=sample_fastq_path, flow_cell_name=flow_cell.id
+            )
+            if not renamed_fastq_file_path.exists():
+                rename_file(file_path=sample_fastq_path, renamed_file_path=renamed_fastq_file_path)
+                sample_fastq_path = renamed_fastq_file_path
```
seallard, line 97: This logic could be flattened
```python
    for sample_fastq_path in sample_fastq_paths:
        sample_fastq_path = rename_file_if_needed(sample_fastq_path, flow_cell.id)
        store_fastq_path_in_housekeeper(
            sample_internal_id=sample_internal_id, sample_fastq_path=sample_fastq_path,
        )


def rename_file_if_needed(file_path: Path, flow_cell_name: str) -> Path:
    renamed_file_path = add_flow_cell_name_to_fastq_file_path(file_path, flow_cell_name)
    if file_path != renamed_file_path:
        rename_file(file_path, renamed_file_path)
    return renamed_file_path
```
Or something, feel free to disregard

ChrOertlin:
# cg demultiplex finish temporary exited with success

## files are renamed
image
image


## files are in hk
image
image

## files that had flow cell in name are not renamed
image

cg/utils/files.py
ChrOertlin: You are right, will make it more explicit.

ChrOertlin:
stage:
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/proj/stage/demultiplexed-runs/230824_A00187_0994_BHGW2VDRX3/Unaligned/Project_887803/Sample_ACC12589A10
done.
Log deploy... done.
cg, version 44.0.2

Prod: 
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/proj/stage/demultiplexed-runs/230824_A00187_0994_BHGW2VDRX3/Unaligned/Project_887803/Sample_ACC12589A10
done.
Log deploy... done.
cg, version 44.0.2

PR #2392, opened 2023-08-30
Migrate index model

seallard:
## Description
Migrate Index model to Pydantic v2.

### Fixed
- Migrate Index model to Pydantic v2.


henrikstranneheim (approved):
No change to the models? Easy enough 👍

seallard:
> No change to the models? Easy enough 👍

Nope! Only a single model present which did not need any updating
```python
class Index(BaseModel):
    """Class that represents an index."""
    name: str
    sequence: str
```

PR #2393, opened 2023-08-30
fix(correct path to novaseqx fastq files)

ChrOertlin:
## Description

the searching for fastq files for fastq files coming from a novaseqx machine was not set for when demultiplexing happened onboard the dragen, which causes the flow to not find the fastq files to add them to housekeeper.

this pr updates with a`dragen_pattern` to the path where fastq files are stored for a demultiplexed novaseqx run on hasta.

new_pattern:
`/demultiplexed-runs/flow-cell-run-name/BCLConvert/fastq/`

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean (approved):
Looks good! My guess is that the fixtures were renamed to match the actual structure on Hasta?

tests/meta/demultiplex/conftest.py
```diff
@@ -401,6 +408,22 @@ def fixture_flow_cell_directory_name_demultiplexed_with_bcl_convert_flat(
     return f"230505_A00689_0804_B{flow_cell_name_demultiplexed_with_bcl_convert}"
 
 
+@pytest.fixture(
+    name="flow_cell_directory_name_demultiplexed_with_bcl_convert_on_sequencer", scope="session"
```
seallard, line 412: This is not how pytest fixtures are commonly used (using the fixture prefix and renaming them with the name parameter).

There is also no need to change the scope of the fixture from the default to session unless it is expensive to setup.

seallard: This might be an opportunity for us to change how we write our fixtures. What do you think @Vince-janv, @diitaz93, @henrikstranneheim, @islean, @ChrOertlin?

seallard (approved):
Nice 👍 Maybe we can reach a conclusion regarding how we want to write our fixtures, feel free to merge though - we can continue the discussion after in this PR anyway

tests/meta/demultiplex/conftest.py
Vince-janv: I'm in favour of skipping fixture in the function name, and also skipping the `name` parameter. I don't really have an opinion on usage of the `scope` parameter.

ChrOertlin:
# test complete

`cg demultiplex flow-cell` is now able to pick up files in `<flow_cell_run_dir>/BCLConvert/fastq`

tests/meta/demultiplex/conftest.py
ChrOertlin: I agree, the use of fixture_name and fixuture names almost repeating the exact same is not really DRY. 
In cases where we might shadow other variables one can change the name.

Regarding the scope - I will need to read up on it. But I am in favour of doing what is most efficient/ best practice in that instance.

tests/meta/demultiplex/test_demux_post_processing.py
```diff
@@ -518,15 +518,6 @@ def test_post_processing_of_flow_cell(
     # GIVEN a directory with a flow cell demultiplexed with BCL Convert
     demux_post_processing_api.demux_api.demultiplexed_runs_dir = tmp_demultiplexed_runs_directory
 
-    # GIVEN that a sample sheet exists in the flow cell run directory
-    path = Path(
```
diitaz93, line 522: nice deletion

tests/meta/demultiplex/test_demux_post_processing.py
```diff
@@ -536,6 +527,10 @@ def test_post_processing_of_flow_cell(
         hk_api=demux_post_processing_api.hk_api,
     )
 
+    # THEN the sample sheet is in housekeeper
```
diitaz93, line 530: Remove the THEN as it is part of the GIVEN
```suggestion

```

ChrOertlin:
stage:
Logging deploy ...
Getting deployer... done.
done.
Log deploy... done.
cg, version 45.0.8
prod:
repository is clean
Logging deploy ...
Getting deployer... done.
done.
Log deploy... done.
cg, version 45.0.8

tests/meta/demultiplex/conftest.py
islean: I agree as well! A bit unsure of how fixtures work but seems to me as good suggestions @seallard

seallard: I think the practice of prefixing a fixture with `fixture_` comes from the instances where we directly import them. pytest actually picks up any fixtures annotated with the `fixture` decorator, so this is not necessary.

For our database fixtures, it requires some thought how to write and scope them. Attached a general summary about fixture scopes below generated by GPT:

<details>
  <summary><strong>Fixture Scopes</strong></summary>

  ### Use `function` scope when:
  - The fixture sets up state that should be isolated between tests.
  - The fixture performs actions like database transactions that need to be rolled back.
  - The setup and teardown actions are quick, and there's no performance concern.
  
  ### Use `class` scope when:
  - Multiple test methods within the same class can share the fixture without affecting each other.
  - You want to optimize performance by reducing setup/teardown for tests in the same class.
  
  ### Use `module` scope when:
  - The fixture data or state can safely be shared across multiple test functions in the same module.
  - Initialization of the fixture is resource-intensive, and you want to minimize that overhead.
  
  ### Use `session` scope when:
  - The fixture data or state is immutable or can be shared across all tests in the session.
  - The fixture initializes a resource that is expensive to create, like a database connection or a web server.
  
  ### Considerations:
  - **Resource Cleanup**: Be cautious about resource leaks when using broader scopes like `module` or `session`. Make sure to provide teardown logic if necessary.
  - **Test Isolation**: Broader scopes can lead to issues with test isolation. Make sure tests are still independent despite sharing fixtures.
  - **Debugging**: Tests that share fixtures at broader scopes can be harder to debug when they fail, as you may need to consider the shared state.
  - **Trade-off**: Changing the scope of a fixture is a trade-off between test isolation and performance. Choose the one that best suits your specific testing needs.
</details>

PR #2397, opened 2023-08-31
Patch analysis type

seallard:
## Description
When migrating a Pydantic model, some tests started failing due to this model returning an enum field instead of the enum value. The diff should be pretty clear.

The `analysis_type` property still does not make sense since it seems to assume it is `WHOLE_EXOME_SEQUENCING` if it is not `WHOLE_TRANSCRIPTOME_SEQUENCING` or `WHOLE_GENOME_SEQUENCING`.

### Fixed
- Fix analysis_type return type on Application model

### DIFF
--- a/cg/store/models.py
+++ b/cg/store/models.py
@@ -131,7 +131,7 @@ class Application(Model):
     @property
     def analysis_type(self):
         if self.prep_category == PrepCategory.WHOLE_TRANSCRIPTOME_SEQUENCING.value:
-            return self.prep_category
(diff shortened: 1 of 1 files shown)


cg/store/models.py
```diff
@@ -131,7 +131,7 @@ def expected_reads(self):
     @property
     def analysis_type(self):
```
diitaz93, line 132: return type hint would be nice

diitaz93 (approved):
👍

PR #2398, opened 2023-08-31
Migrate scout models

seallard:
## Description
This PR migrates the scout models to Pydantic V2.
- Replace deprecated `validator` with `Annotated`
- Replace deprecated `Config` with `model_config`
- Initialise fields marked as Optional to None
- Patch fixture

It does not really make sense to have a not null check for the fields which are required but also initialised as None. But a lot of tests broke when removing it.

### Fixed
- Migrate scout models to Pydantic v2 


cg/models/scout/scout_load_config.py
```diff
@@ -100,38 +98,25 @@ class ScoutLoadConfig(BaseModel):
...
+    vcf_cancer: Annotated[str, BeforeValidator(field_not_none)] = None
     vcf_cancer_sv: Optional[str] = None
     vcf_cancer_research: Optional[str] = None
     vcf_cancer_sv_research: Optional[str] = None
     samples: List[ScoutCancerIndividual] = []
 
-    @validator("vcf_cancer")
-    def check_mandatory_files(cls, vcf):
-        if vcf is None:
-            raise ValueError("Vcf can not be none")
```
seallard, line 124: The validator no longer raises this exception which means the behaviour has been changed

diitaz93:
Great PR description 💯

diitaz93 (approved):
Great 💯

cg/models/scout/validators.py
```diff
@@ -0,0 +1,3 @@
+def field_not_none(value: str):
+    assert value is not None
```
henrikstranneheim, line 2: We should avoid asserting in production code. Pydantic also warns about this.

henrikstranneheim (changes requested):
Looks great 👍! Just want the to replace the assert

cg/models/scout/validators.py
islean: Also - can we just not avoid setting a default value for the field it validates?

islean (approved):
Looks good! Although it feels weird having a validator for doing a null check? Can we not just avoid setting defaults for those fields?

cg/models/scout/validators.py
seallard: That is very true, I'll remove the validator. It does not make sense

islean:
Just a thought - can be good to double check that we don't see the same problem as with the order parsing with coercing to strings.

cg/models/scout/validators.py
seallard: Well, removing the validator caused a bunch of tests to break. Not opening that can of worms today. Replaced the assert.

seallard: The issue is that other models inherit from the `ScoutIndividual` class. It seems like the None validation is only run when `ScoutIndividual` is instantiated directly. So if for instance `sample_id` is marked as required, it suddenly becomes required in all the models inheriting from the base class.

I think this inheritance pattern is not so good here. It causes confusion and weird behaviour when refactoring. If data is not needed in subclasses, the subclasses has no business inheriting from the class in the first place

seallard:
> Looks good! Although it feels weird having a validator for doing a null check? Can we not just avoid setting defaults for those fields?

Yes, I thought so too but tests started failing when removing it. Leaving this PR for now, the behaviour does not make sense to me

seallard:
Deployed to Hasta stage and prod.

PR #2399, opened 2023-09-01
Fix volume type

seallard:
## Description
This PR fixes an issue for the parsing of a volume for a sample in a order.

### Fixed
-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/models/orders/sample_base.py
+++ b/cg/models/orders/sample_base.py
@@ -119,6 +119,6 @@ class OrderSample(BaseModel):
     tumour: bool = False
     tumour_purity: Optional[int] = None
     verified_organism: Optional[bool] = None
-    volume: Optional[str] = None
+    volume: Optional[int] = None
     well_position: Optional[str] = None
     well_position_rml: Optional[str] = None
--- a/cg/models/orders/sample_base.py
+++ b/cg/models/orders/sample_base.py
@@ -119,6 +119,6 @@ class OrderSample(BaseModel):
     tumour: bool = False
     tumour_purity: Optional[int] = None
     verified_organism: Optional[bool] = None
-    volume: Optional[int] = None
+    volume: Annotated[Optional[int], BeforeValidator(lambda x: str(x))] = None
     well_position: Optional[str] = None
     well_position_rml: Optional[str] = None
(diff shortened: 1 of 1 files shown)


diitaz93:
Im not 100% sure this will solve the issue

diitaz93:
what if we add a validator that turns ints into strings or vice versa?

diitaz93 (approved):
Let's give it a try as it is urgent for the customer, but I say this needs further investigation

islean (approved):
Should probably work! What I did for another field (but I must have missed this one) is to use a BeforeValidator(lambda x: str(x)) to coerce the int to a string

diitaz93:
> Should probably work! What I did for another field (but I must have missed this one) is to use a BeforeValidator(lambda x: str(x)) to coerce the int to a string

That would be better, right?

islean:
> > Should probably work! What I did for another field (but I must have missed this one) is to use a BeforeValidator(lambda x: str(x)) to coerce the int to a string
> 
> That would be better, right?

Potentially, if the volume is not necessarily an int

diitaz93:
there is a validation in teh excel order like this:
```python
volume: Annotated[str, AfterValidator(numeric_value)] = Field(None, alias="UDF/Volume (uL)")
```
but is not present in the jason sample. Should we do it too?

seallard:
> there is a validation in teh excel order like this:
> 
> ```python
> volume: Annotated[str, AfterValidator(numeric_value)] = Field(None, alias="UDF/Volume (uL)")
> ```
> 
> but is not present in the jason sample. Should we do it too?

Yes, I think @islean is adding it now

Vince-janv:
Nice patch. Testing should be quite easy given that she sent us the json file 👍

islean:
> there is a validation in the excel sample like this:
> 
> ```python
> volume: Annotated[str, AfterValidator(numeric_value)] = Field(None, alias="UDF/Volume (uL)")
> ```
> 
> but is not present in the jason sample. Should we do it too?

An AfterValidator will not work in this case since that would be performed after Pydantics type check (I think), but I added a before validator with string coercion now

PR #2367, opened 2023-08-25
refactor(nf_analysis) deliverables as model

fevac:
### Changed

- Refactor deliverables for rnafusion

### How to prepare for test

- [x] Ssh to relevant server (depending on type of change)
- [x] Use stage: `us`
- [x] Paxa the environment: `paxa`
- [x] Install on stage (example for Hasta):
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b deliverables_as_model -a
    ```

### How to test

- [x] `cg workflow rnafusion store <case_id>`

### Expected test outcome

- [x] Check that deliverable file is correctly created
- [x] Check that store is completed

## Review

- [ ] Tests executed by
- [ ] "Merge and deploy" approved by
Thanks for filling in who performed the code review and the test!

### This [version](https://semver.org/) is a

- [ ] **MAJOR** - when you make incompatible API changes
- [ ] **MINOR** - when you add functionality in a backwards compatible manner
- [ ] **PATCH** - when you make backwards compatible bug fixes or documentation/instructions

## Implementation Plan

- [ ] Document in ...
- [ ] Deploy this branch on ...
- [ ] Inform to ...

PR #2405, opened 2023-09-01
Migrate mutant sample config model

seallard:
## Description
This PR migrates the `MutantSampleConfig` model to Pydantic V2.

- Replace deprecated `validator` with `Annotated`

### Fixed
- Migrate MutantSampleConfig to Pydantic V2.


cg/models/workflow/validators.py
```diff
@@ -0,0 +1,2 @@
+def get_first_word(value: str) -> str:
```
henrikstranneheim, line 1:
```suggestion
def get_first_word(sentance: str) -> str:
```
Seems not specific to Mutant - should we create a general Pydantic validators module? aving the same discussion here with @fevac @islean https://github.com/Clinical-Genomics/cg/pull/2367#discussion_r1312628865)

henrikstranneheim (approved):
👍

cg/models/workflow/validators.py
islean: Yeah, I think it is worth having a discussion about it, although it might be worth structuring a bit more. Either by using sub directories or keeping quite strict in ordering of the validators in the file (to the extent possible).

I would also add that this could probably be an AfterValidator. It would probably make more sense to get a parsing error for wrong type than getting an error stating that "split" does not work for the input.

islean (approved):
Looks good! I have a slight preference for an AfterValidator, but in practice it will most likely not make any difference.

cg/models/workflow/validators.py
seallard: Sure, that could make sense. But we should only move stuff that is used elsewhere. In practice, I think most of the "general" validators will only get used in a single model anyway and then it makes more sense to keep it close to the model

seallard: Out of scope for this PR, lets create it once we stumble across a validator that is used elsewhere

PR #2412, opened 2023-09-04
Migrate cg config models to Pydantic v2

seallard:
## Description
This PR migrates the `cg_config` models to Pydantic v2.
- Update imports
- Replace deprecated `Config` with `model_config`
- Initialise optional fields with None

### Fixed
- Migrate config models to Pydantic v2.


Vince-janv (approved):
This one is important to test, but to me everything looks good 👍

cg/models/cg_config.py
```diff
@@ -206,7 +206,7 @@ class GisaidConfig(CommonAppConfig):
 class DataDeliveryConfig(BaseModel):
     destination_path: str
     covid_destination_path: str
-    covid_source_path = str
```
seallard, line 209: I assume this was a typo. The field had to be marked as optional since tests started failing after just correcting the type hint

karlnyr: I would not be so sure about it - I would look further into the data delivery of sars-cov-2 to make sure that it is in fact correct

seallard: I'm 100 % it is a typo, type hints are assigned with `: str` not `= str` and it makes no sense to assign initialise it as `str`. But it might be the case that `covid_source_path` should not be optional, but that is in fact the case previously.

seallard: If it should not be optional, it is new behaviour. Creating an issue for it and we can refine it in a refinement session.

seallard: https://github.com/Clinical-Genomics/cg/issues/2415

cg/models/cg_config.py
```diff
@@ -302,10 +302,10 @@ class CGConfig(BaseModel):
...
     # These are meta APIs that gets instantiated in the code
     meta_apis: dict = {}
-
-    class Config:
-        arbitrary_types_allowed = True
-        fields = {
+    model_config = ConfigDict(
+        arbitrary_types_allowed=True,
+        fields={
+            "cg_stats_api_": "cg_stats_api",
```
diitaz93, line 308: do we still have the cg_stats_api?

diitaz93 (approved):
Looks good!

islean (approved):
I would probably add "coerce_numbers_to_strings" or whatever the setting is called so as to not run into similar problems as we've had with other migrations.

seallard:
stale

PR #2425, opened 2023-09-05
Remove unused fixtures

seallard:
## Description
Remove unused fixtures found with `pytest --dead-fixtures`

### Fixed
- Remove unused fixtures

### DIFF
--- a/tests/apps/demultiplex/conftest.py
+++ b/tests/apps/demultiplex/conftest.py
@@ -17,30 +17,6 @@ from cg.constants.demultiplexing import BclConverter, SampleSheetNovaSeq6000Sect
 from cg.models.demultiplex.flow_cell import FlowCellDirectoryData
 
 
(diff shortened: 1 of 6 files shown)


diitaz93 (approved):
Looks good on my side 👍 I would still wait for the approval of @ChrOertlin and @sofstam before merging as they may be using some of these fixtures in their open PRs.

seallard:
> Looks good on my side 👍 I would still wait for the approval of @ChrOertlin and @sofstam before merging as they may be using some of these fixtures in their open PRs.

Hm, sure. I think its fine though, worst case the fixtures can retrieved from the diff in this commit for example

> Looks good on my side 👍 I would still wait for the approval of @ChrOertlin and @sofstam before merging as they may be using some of these fixtures in their open PRs.

From what I can see, they don't use any of these. Worst case the fixtures can be retrieved from any previous version of the repo

Issue #2419, opened 2023-09-05
Addition of Undetermined FASTQs to non-pooled samples


Vince-janv:
## Description

On old flow cells some samples were run alone on a lane (currently samples are always pooled and separated by index). For these samples the undetermined FASTQ file were added to the sample in housekeeper since we could guarantee that the reads came from that sample.  

This logic was handled [here](https://github.com/Clinical-Genomics/demultiplexing/blob/9680e28fc0d406320c863dd98a37d19db3d840db/scripts/hiseqx/xdemuxtiles.batch#L128-L139C11) in the demultiplexing repo.

For these old samples to get all their files when re-demultiplexing we need to implement this logic.
## Suggested solution

TBD

## This can be closed when

Undetermined FASTQ files are added to non-pooled samples

henrikstranneheim:
@seallard Could you work on this? If refinement is needed please coordinate with @Vince-janv and @karlnyr .

seallard:
@Vince-janv @karlnyr Do we want to generate sequencing metrics based on these undetermined fastqs as well? Should the read count for the samples be updated with the undetermined fastqs assigned to it?

Vince-janv:
I've looked trough this and it seems like the following was done previously. If the sample was alone on a lane the metrics assigned to the sample, and the metrics for the undetermined reads were baked into one unaligned entry. 

I fear we might have to do the same. I don't think we can have just add the files to HK without updating the reads, and I don't think we can have the reads not be the sum of the sequencing_metrics entries. I hope this is not that hard to implement 😬

seallard:
The required data seems to be available in the generated metrics files. So it should be doable, might just need to update the sequencing metrics calculator module.

For the sequencing metrics model we need

- Sample Total Reads In Lane
- Sample Base Percentage Passing Q30 
- Sample Base Mean Quality Score

These are calculated from the statistics files generated by the demultiplexing software. They are reported for bclconvert

| Lane | SampleID    | Sample_Project | index | index2 | ReadNumber |      Yield      |  YieldQ30   | QualityScoreSum | Mean Quality Score (PF) | % Q30 |
|------|-------------|----------------|-------|--------|------------|-----------------|-------------|-----------------|------------------------|-------|
| 2    | Undetermined| Undetermined   |       |        |     1      |  2196984528     | 1886452123  | 75448828497     |          34.34         |  0.86 |
[...]

PR #2427, opened 2023-09-06
Add undetermined fastqs

seallard:
## Description
This PR implements https://github.com/Clinical-Genomics/cg/issues/2419, storing undetermined fastqs for non-pooled samples.

The intended functionality is to (for each flow cell being processed):
1. Retrieve all samples that are not pooled (alone in a lane)
2. Retrieve any undetermined fastq files associated with each lane containing a single sample
3. Assign the undetermined fastqs to each corresponding non pooled sample and store them in housekeeper

This functionality should only run for old flow cells which are re-demultiplexed. All newer ones always have pooled samples.
 
Merging this without calling the new functionality, will implement new requirements in separate PR.

### Added
- Storing of undetermined fastqs for non pooled samples


cg/models/demultiplex/flow_cell.py
```diff
@@ -86,7 +86,7 @@ def sample_sheet_path(self) -> Path:
     def set_sample_sheet_path_hk(self, hk_path: Path):
         self._sample_sheet_path_hk = hk_path
 
-    def get_sample_sheet_path_hk(self) -> Optional[Path]:
```
seallard, line 89: This type hint was not correct since an exception is raised if the path has not been set for the sample sheet

cg/meta/demultiplex/utils.py
```diff
@@ -175,3 +177,21 @@ def rename_fastq_file_if_needed(fastq_file_path: Path, flow_cell_name: str) -> P
...
+    sample_type = flow_cell.sample_type
+    sample_sheet: SampleSheet = get_sample_sheet_from_file(
+        infile=sample_sheet_path, flow_cell_sample_type=sample_type
+    )
+    return sample_sheet
+
+
+def get_undetermined_fastqs(lane: int, flow_cell_path: Path) -> List[Path]:
+    """Get the undetermined fastq files for a specific lane on a flow cell."""
+    unaligned_pattern = f"Undetermined*_L00{lane}_*{FileExtensions.FASTQ}{FileExtensions.GZIP}"
```
seallard, line 193: Not sure if undetermined fastq file names always are formatted like this

cg/apps/demultiplex/sample_sheet/models.py
```diff
@@ -76,6 +78,15 @@ class FlowCellSampleNovaSeq6000Dragen(FlowCellSampleNovaSeq6000):
 class SampleSheet(BaseModel):
     samples: List[FlowCellSample]
 
+    def get_non_pooled_samples(self) -> List[FlowCellSample]:
```
seallard, line 81: Do we want to keep logic outside of our data models? This could go in a utils module as well. We don't have any outspoken design choice regarding this. 

We use Pydantic for most of our domain models, so I think it makes sense to have logic in them as well. It encapsulates logic nicely

tests/meta/demultiplex/test_utils.py
```diff
@@ -293,3 +294,44 @@ def test_add_flow_cell_name_to_fastq_file_path_when_flow_cell_name_already_in_na
...
+    lane = 1
+    expected_files = [
+        Path(tmp_path, "Undetermined_L001_R1.fastq.gz"),
+        Path(tmp_path, "Undetermined_L001_R2.fastq.gz"),
+    ]
+    for f in expected_files:
+        f.touch()
+
+    # WHEN retrieving the undetermined fastqs for the lane
+    result = get_undetermined_fastqs(lane=lane, flow_cell_path=tmp_path)
```
henrikstranneheim, line 334:
```suggestion
    result = get_undetermined_fastqs(lane=1, flow_cell_path=tmp_path)
```

henrikstranneheim (approved):
Well done 👍

cg/apps/demultiplex/sample_sheet/models.py
henrikstranneheim: I think this is well placed 👍

cg/meta/demultiplex/utils.py
```diff
@@ -175,3 +177,21 @@ def rename_fastq_file_if_needed(fastq_file_path: Path, flow_cell_name: str) -> P
     if fastq_file_path != renamed_fastq_file_path:
         rename_file(file_path=fastq_file_path, renamed_file_path=renamed_fastq_file_path)
     return renamed_fastq_file_path
+
+
+def get_sample_sheet(flow_cell: FlowCellDirectoryData) -> SampleSheet:
```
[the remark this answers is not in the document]
seallard: Does `Return sample sheet associated with flowcell` work? Or do we want to explain that it is retrieved from housekeeper? I think the caller should not really care where it is retrieved from, as long as it is clear what is returned

cg/meta/demultiplex/housekeeper_storage_functions.py
```diff
@@ -43,6 +46,40 @@ def store_flow_cell_data_in_housekeeper(
...
+        for fastq_path in undetermined_fastqs:
+            add_fastq_file_to_housekeeper(
+                sample_id=sample.sample_id,
+                flow_cell_id=flow_cell.id,
+                fastq_path=fastq_path,
+                hk_api=hk_api,
+            )
+
+
+def add_fastq_file_to_housekeeper(
```
Vince-janv, line 73: Do we not have a function for this when adding the normal FASTQ files?

cg/apps/demultiplex/sample_sheet/models.py
```diff
@@ -76,6 +78,15 @@ class FlowCellSampleNovaSeq6000Dragen(FlowCellSampleNovaSeq6000):
...
+        """Return samples that are sequenced solo in their lane."""
+        lane_samples = defaultdict(list)
```
Vince-janv, line 83: Cool 😎

Vince-janv (approved):
Well done, this looks great ⭐

cg/meta/demultiplex/housekeeper_storage_functions.py
seallard: That function also does a check against the sequencing metrics table whether the fastq file should be tracked. Probably a SRP violation, the check should be extracted. Can be refactored in a separate PR to limit the scope of this one

Vince-janv: Makes sense 👍

cg/meta/demultiplex/utils.py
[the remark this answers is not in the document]
diitaz93: Is there already an issue for this refactoring?

diitaz93 (commented):
Looks good! I we can discuss all comments in the refinement session

cg/meta/demultiplex/utils.py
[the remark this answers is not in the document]
henrikstranneheim: `Return sample sheet associated with flowcell` - perfect 👍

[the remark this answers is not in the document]
henrikstranneheim: Ju*st add a space in flowcell

cg/apps/demultiplex/sample_sheet/models.py
```diff
@@ -76,6 +78,15 @@ class FlowCellSampleNovaSeq6000Dragen(FlowCellSampleNovaSeq6000):
...
+        for sample in self.samples:
+            lane: int = sample.lane
+            lane_samples[lane].append(sample)
```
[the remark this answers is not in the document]
seallard: I think duplication of this kind is fine since it is just three lines of code. It might end up more confusing calling the helper function. What do you think?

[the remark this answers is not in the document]
diitaz93: I think the usages can be fixed so that the function is defined only once. But that is out of scope. I will create an issue and create a tag `demultiplex` to group all the OOS ideas regarding refactoring the demultiplex module

seallard:
I'll merge this and continue with the new requirements in a separate PR. The new functionality is not called anywhere

PR #2432, opened 2023-09-07
Fix(bcl2fastq) Add new location bcl2fastq stats file

Vince-janv:
## Description
After recent changes (???) to the demultiplexing, the output structure for bcl2fastq has changed and no longer fits the one we're expecting when parsing the sequencing metrics. Example flow cells `HK3K5BCX2`, `HY2N7DSXX`. For these no metrics are found, and no error is raised, leading to no metrics or FASTQ file added to status-db and housekeeper respectively

### Changed

- We are now looking for any file which path ends with `Stats/Stats.json`

### Fixed
- Now correctly raises FileNotFoundError if no file is found

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b debug-bcl2fastq
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


Vince-janv:
### The file was now found, metrics were added and files added to housekeeper

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
```diff
@@ -88,17 +89,16 @@ def discard_index_sequence(sample_id_with_index: str) -> str:
...
-        if pattern.match(subdir):
-            stats_json_path = (
-                demultiplex_result_directory
-                / subdir
-                / BCL2FASTQ_METRICS_DIRECTORY_NAME
-                / BCL2FASTQ_METRICS_FILE_NAME
-            )
-            if stats_json_path.is_file():
+
+    for root, dirs, files in os.walk(demultiplex_result_directory):
```
seallard, line 93: dirs can be called `_` since we are not using it

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
```diff
@@ -88,17 +89,16 @@ def discard_index_sequence(sample_id_with_index: str) -> str:
...
+        if BCL2FASTQ_METRICS_FILE_NAME in files:
+            stats_json_path = Path(root, BCL2FASTQ_METRICS_FILE_NAME)
+            if root.endswith(BCL2FASTQ_METRICS_DIRECTORY_NAME):
```
seallard, line 96: Might be more intuitive to switch the location of the if statements (check dir name first and then file name)

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
```diff
@@ -88,17 +89,16 @@ def discard_index_sequence(sample_id_with_index: str) -> str:
...
                 stats_json_paths.append(stats_json_path)
 
+    if not stats_json_paths:
+        raise FileNotFoundError(
```
seallard, line 100: Did we settle on exceptions being included in the doc string?

seallard (approved):
Great 👍

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
islean, line 96: Perhaps cleaner if stats_json_path was instantiated after both ifs, since it only seems to be used if both are true.

islean (approved):
Looks good! I guess that there is introduced risk from having a more lenient check on the path structure?

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
ChrOertlin, line 93:
```suggestion
    for root, dirs, files in os.walk(demultiplex_result_directory):
```
We have a function for this in /cg/utils/files/ 
that one is tested and we should use the implementation of it here. 
`get_file_in_directory()`

ChrOertlin (approved):
Looks good! Make sure to re-use the already implemented function in cg/utils/files.py

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
Vince-janv: I think it should be included. Adding it 👍

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
Vince-janv: Good call! Unfortunately we want all examples of the Stats.json file and `get_file_in_directory` seems to only return the first match

PR #2434, opened 2023-09-07
Move retrieval of sample sheets into housekeeper api

seallard:
## Description
This PR moves a function used to retrieve sample sheets for a flow cell from a utils module into the housekeeper api.
Many of the functions in this module should probably be moved into the housekeeper api

### Fixed
- Move housekeeper function into housekeeper api

### DIFF
--- a/cg/apps/housekeeper/hk.py
+++ b/cg/apps/housekeeper/hk.py
@@ -482,3 +482,18 @@ class HousekeeperAPI:
             raise ValueError(f"No Archive entry found for file with id {file_id}.")
         self._store.update_retrieval_task_id(archive=archive, retrieval_task_id=retrieval_task_id)
         self.commit()
+
+    def get_sample_sheets_from_latest_version(self, flow_cell_id: str) -> List[File]:
+        """Returns the files tagged with 'samplesheet' or 'archived_sample_sheet' for the given bundle."""
+        try:
+            sheets_with_normal_tag: List[File] = self.get_files_from_latest_version(
+                bundle_name=flow_cell_id, tags=[flow_cell_id, SequencingFileTag.SAMPLE_SHEET]
+            ).all()
+            sheets_with_archive_tag: List[File] = self.get_files_from_latest_version(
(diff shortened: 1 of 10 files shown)


cg/apps/housekeeper/hk.py
```diff
@@ -482,3 +482,18 @@ def set_archive_retrieval_task_id(self, file_id: int, retrieval_task_id: int) ->
             raise ValueError(f"No Archive entry found for file with id {file_id}.")
         self._store.update_retrieval_task_id(archive=archive, retrieval_task_id=retrieval_task_id)
         self.commit()
+
+    def get_sample_sheets_from_latest_version(self, flow_cell_id: str) -> List[File]:
```
islean, line 486: Are we ever expecting more than one sample sheet from a version? Else it might be nice to have this method returning an Optional[File] instead.

islean (approved):
Looks nice! Similar question to the other PR moving functions from this file into hk api - is the goal to move everything which only does checks towards HK to hk api?

seallard:
> Looks nice! Similar question to the other PR moving functions from this file into hk api - is the goal to move everything which only does checks towards HK to hk api?

That was my thought, it seems intuitive to keep that logic there. I'll wait for some more reviews to see what the others think as well. Do you think it makes sense?

cg/apps/housekeeper/hk.py
diitaz93: It may return the sample sheet and the archived sample sheet if the flow cell has both, which will be deprecated

diitaz93 (approved):
Good idea 💡

diitaz93:
> > Looks nice! Similar question to the other PR moving functions from this file into hk api - is the goal to move everything which only does checks towards HK to hk api?
> 
> That was my thought, it seems intuitive to keep that logic there. I'll wait for some more reviews to see what the others think as well. Do you think it makes sense?

yes! I agree

seallard:
Too many merge conflicts to resolve, this can be done in a separate PR

PR #2435, opened 2023-09-07
Move file existence check into hk api

seallard:
## Description
This PR moves a function which checks whether a file has been added to a bundle into the housekeeper api.
Many of the functions in this module should probably be moved into the housekeeper api.

### Fixed
- Move housekeeper functionality into housekeeper api

### DIFF
--- a/cg/apps/housekeeper/hk.py
+++ b/cg/apps/housekeeper/hk.py
@@ -482,3 +482,11 @@ class HousekeeperAPI:
             raise ValueError(f"No Archive entry found for file with id {file_id}.")
         self._store.update_retrieval_task_id(archive=archive, retrieval_task_id=retrieval_task_id)
         self.commit()
+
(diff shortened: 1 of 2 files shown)


islean (approved):
I like it although it makes me wonder if the goal is to move more functions from housekeeper_storage_functions to hk api? Are we moving e.g. add_file_to_bundle_if_non_existent and add_tags_if_non_existent in future PRs as well?

seallard:
> I like it although it makes me wonder if the goal is to move more functions from housekeeper_storage_functions to hk api? Are we moving e.g. add_file_to_bundle_if_non_existent and add_tags_if_non_existent in future PRs as well?

Thats my idea. I did not want to move everything in one PR since it might end up a bit messy. If everyone thinks it is a good idea I think we should move it

diitaz93 (approved):
👍

seallard:
Moving all functions in a separate PR

PR #2437, opened 2023-09-07
Remove verbose doc strings

seallard:
## Description
Found some overly verbose doc strings.

### Fixed
- Shorten verbose doc strings


seallard:
I wrote these doc strings, they are not very helpful.

PR #2440, opened 2023-09-07
Migrate lims models to Pydantic v2

seallard:
## Description
This PR migrates the LIMS models to Pydantic v2.
- Update imports
- Initialise optional fields as None
- Replace deprecated `validator` decorator
- Replace deprecated `parse_obj` with `model_validate`
- Ensure non string values typed as strings are coerced to strings

### Fixed
- Migrate LIMS models to Pydantic v2

### DIFF
--- a/cg/models/lims/sample.py
+++ b/cg/models/lims/sample.py
@@ -1,6 +1,6 @@
 from typing import Optional
 
-from pydantic.v1 import BaseModel, validator
+from pydantic import field_validator, BaseModel
 from typing_extensions import Literal
 
 from cg.constants import Priority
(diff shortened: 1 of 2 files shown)


cg/models/lims/sample.py
```diff
@@ -10,64 +10,65 @@
...
     require_qc_ok: bool = False
-    rml_plate_name: Optional[str]
-    selection_criteria: Optional[str]
+    rml_plate_name: Optional[str] = None
+    selection_criteria: Optional[str] = None
     sex: Literal["M", "F", "unknown"] = "unknown"
     skip_reception_control: Optional[bool] = None
     source: str = "NA"
-    tissue_block_size: Optional[str]
+    tissue_block_size: Optional[str] = None
```
islean, line 48: Is this always given as a string or do we need to coerce it here as well?

cg/models/lims/sample.py
```diff
@@ -10,64 +10,65 @@
...
-    well_position_rml: Optional[str]
-    verified_organism: Optional[bool]
+    tumour_purity: Annotated[Optional[str], BeforeValidator(lambda v: str(v))] = None
+    volume: Annotated[Optional[str], BeforeValidator(lambda v: str(v))] = None
+    well_position_rml: Optional[str] = None
+    verified_organism: Optional[bool] = None
 
-    @validator("sex", pre=True)
+    @field_validator("sex", mode="before")
+    @classmethod
```
islean, line 56: Why are we not using the Annotated pattern for this validator?

cg/models/lims/sample.py
```diff
@@ -10,64 +10,65 @@
...
-    well_position: Optional[str]
-    index_sequence: Optional[str]
-    udfs: Optional[Udf]
+    container_name: Optional[str] = None
+    well_position: Optional[str] = None
+    index_sequence: Optional[str] = None
+    udfs: Optional[Udf] = None
 
     @classmethod
     def parse_obj(cls, obj: dict):
```
islean, line 70: Consider a type hint since parse_obj is not the clearest of names (although LimsSample.parse_obj() might be I guess.

tests/meta/orders/test_meta_orders_lims.py
```diff
@@ -139,7 +139,7 @@ def test_to_lims_balsamic(balsamic_order_to_submit, project):
     container_names = {sample.container_name for sample in samples if sample.container_name}
 
     # ... and pick out relevant UDFs
-    first_sample = samples[0].dict()
+    first_sample = samples[0].model_dump()
```
islean, line 142: A bit out of scope but why would we need this as a dict? Can't we just use first_sample.name and first_sample.udfs and so on?

islean (approved):
Would love to see the field_validator following the Annotated pattern but great job! 🔥

tests/meta/orders/test_meta_orders_lims.py
seallard: I have no idea 😅 Don't want to change the behaviour though

islean: Fair enough!

diitaz93 (approved):
Looks good to me 👍

seallard:
stale

PR #2441, opened 2023-09-07
patch(Excel parsing )Replace spaces with underscores for priority when parsing Excel sheets

Vince-janv:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/models/orders/validators/excel_sample_validators.py
+++ b/cg/models/orders/validators/excel_sample_validators.py
@@ -62,6 +62,10 @@ def convert_to_lower(value: Optional[str]):
     return value.lower()
 
 
+def replace_spaces_with_underscores(value: Optional[str]):
+    return value.replace(" ", "_")
+
+
 def convert_to_priority(priority: Optional[str]):
     """Translates the Swedish 'förtur' to 'priority' if specified in the order."""
     return PriorityEnum.priority if priority == "förtur" else priority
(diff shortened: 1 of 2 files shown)


cg/models/orders/validators/excel_sample_validators.py
```diff
@@ -62,6 +62,10 @@ def convert_to_lower(value: Optional[str]):
     return value.lower()
 
 
+def replace_spaces_with_underscores(value: Optional[str]):
+    return value.replace(" ", "_")
```
islean, line 66: Don't know if we should keep the inputs to these validators as Optionals given that they most likely will break with no given input. But on the other hand, maybe we want them to break?

islean (approved):
Looks good! Do we know if this bug was introduced with the pydantic migration?

cg/models/orders/validators/excel_sample_validators.py
henrikstranneheim: I think we should remove optional

henrikstranneheim (approved):
👍

seallard (approved):
👍 Will resolve https://github.com/Clinical-Genomics/cg/issues/2452

cg/models/orders/validators/excel_sample_validators.py
Vince-janv: Done 👍

PR #2445, opened 2023-09-08
Refactor logic for retrieving sample ids from sample sheet

seallard:
## Description
The logic for extracting the sample ids from a sample sheet belongs on the `SampleSheet` model.
This PR does not contain any functional changes, it only moves logic from one spot to another.

### Fixed
- Extract logic for retrieving sample ids to the `SampleSheet` object

### DIFF
--- a/tests/apps/demultiplex/test_read_sample_sheet.py
+++ b/tests/apps/demultiplex/test_read_sample_sheet.py
@@ -2,6 +2,7 @@ import logging
 from typing import List, Dict, Type
 import pytest
 from pathlib import Path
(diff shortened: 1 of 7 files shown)


tests/apps/demultiplex/test_read_sample_sheet.py
diitaz93, line 1: To be consistent, the test functions `test_is_valid_sample_internal_id` and `test_is_valid_sample_internal_id_invalid_sample_internal_ids` should be moved to a new script `test_validate.py`  in this same directory.

diitaz93 (approved):
Great idea 👍 Consider moving the tests for consistency

PR #2448, opened 2023-09-08
Refactor sample sheet property

seallard:
## Description
This PR refactors how the sample sheet path property is set and accessed on the flow cell directory data object since it was a bit complicated (getters and setters for attributes are not commonly used in Python). The `sample_sheet_path` is now provided as an optional argument to the constructor.

Changes:
- Pass `sample_sheet_path` to `FlowCellDirectoryData` constructor
- Rename existing `sample_sheet_path` property to `run_dir_sample_sheet_path`
- Remove `get_sample_sheet_path_hk` and `set_sample_sheet_path_hk` on `FlowCellDirectoryData`
- Replace `get_sample_internal_ids_from_sample_sheet` with `sample_sheet` property
- Add helper function `get_sample_sheet_path` for retrieving sample sheet path from housekeeper
- Remove unnecessary helper function `parse_flow_cell_directory_data`
- Refactor tests

### Fixed
- Pass sample sheet path to flow cell directory data constructor

### DIFF
--- a/cg/meta/demultiplex/demux_post_processing.py
+++ b/cg/meta/demultiplex/demux_post_processing.py
@@ -14,7 +14,7 @@ from cg.constants.demultiplexing import DemultiplexingDirsAndFiles
 from cg.exc import FlowCellError, MissingFilesError
 from cg.meta.demultiplex import files
 from cg.meta.demultiplex.housekeeper_storage_functions import (
-    get_sample_sheets_from_latest_version,
+    get_sample_sheet_path_hk,
     store_flow_cell_data_in_housekeeper,
 )
 from cg.meta.demultiplex.status_db_storage_functions import (
@@ -24,7 +24,7 @@ from cg.meta.demultiplex.status_db_storage_functions import (
 )
 from cg.meta.demultiplex.utils import (
     create_delivery_file_in_flow_cell_directory,
-    parse_flow_cell_directory_data,
+    get_flow_cell_id,
 )
 from cg.meta.demultiplex.validation import is_flow_cell_ready_for_postprocessing
 from cg.meta.transfer import TransferFlowCell
@@ -99,22 +99,14 @@ class DemuxPostProcessingAPI:
 
         LOG.info(f"Finish flow cell {flow_cell_directory_name}")
 
-        flow_cell_out_directory: Path = Path(
-            self.demux_api.demultiplexed_runs_dir, flow_cell_directory_name
-        )
+        flow_cell_out_directory: Path = self.get_flow_cell_path(flow_cell_directory_name)
+        sample_sheet_path: Path = self.get_sample_sheet_path(flow_cell_directory_name)
 
-        parsed_flow_cell: FlowCellDirectoryData = parse_flow_cell_directory_data(
-            flow_cell_directory=flow_cell_out_directory,
+        parsed_flow_cell = FlowCellDirectoryData(
+            flow_cell_path=flow_cell_out_directory,
             bcl_converter=bcl_converter,
+            sample_sheet_path=sample_sheet_path,
         )
-
-        sample_sheet_path: Path = Path(
-            get_sample_sheets_from_latest_version(
-                flow_cell_id=parsed_flow_cell.id, hk_api=self.hk_api
-            )[0].full_path
-        )
-        parsed_flow_cell.set_sample_sheet_path_hk(hk_path=sample_sheet_path)
-        LOG.debug("Set path for Housekeeper sample sheet in flow cell")
 
         try:
             is_flow_cell_ready_for_postprocessing(
@@ -171,6 +163,13 @@ class DemuxPostProcessingAPI:
                 LOG.debug(f"Found directory {flow_cell_dir}")
                 demultiplex_flow_cells.append(flow_cell_dir)
         return demultiplex_flow_cells
+
+    def get_sample_sheet_path(self, flow_cell_dir_name: str) -> Path:
+        flow_cell_id: str = get_flow_cell_id(flow_cell_dir_name)
+        return get_sample_sheet_path_hk(flow_cell_id=flow_cell_id, hk_api=self.hk_api)
+
+    def get_flow_cell_path(self, flow_cell_dir_name: str) -> Path:
+        return Path(self.demux_api.demultiplexed_runs_dir, flow_cell_dir_name)
 
 
 class DemuxPostProcessingHiseqXAPI(DemuxPostProcessingAPI):
--- a/cg/meta/demultiplex/demux_post_processing.py
+++ b/cg/meta/demultiplex/demux_post_processing.py
@@ -99,7 +99,9 @@ class DemuxPostProcessingAPI:
 
         LOG.info(f"Finish flow cell {flow_cell_directory_name}")
 
-        flow_cell_out_directory: Path = self.get_flow_cell_path(flow_cell_directory_name)
+        flow_cell_out_directory: Path = self.get_demultiplexed_runs_flow_cell_path(
+            flow_cell_directory_name
+        )
         sample_sheet_path: Path = self.get_sample_sheet_path(flow_cell_directory_name)
 
         parsed_flow_cell = FlowCellDirectoryData(
@@ -168,7 +170,7 @@ class DemuxPostProcessingAPI:
         flow_cell_id: str = get_flow_cell_id(flow_cell_dir_name)
         return get_sample_sheet_path_hk(flow_cell_id=flow_cell_id, hk_api=self.hk_api)
 
-    def get_flow_cell_path(self, flow_cell_dir_name: str) -> Path:
+    def get_demultiplexed_runs_flow_cell_path(self, flow_cell_dir_name: str) -> Path:
         return Path(self.demux_api.demultiplexed_runs_dir, flow_cell_dir_name)
 
 
--- a/cg/meta/demultiplex/housekeeper_storage_functions.py
+++ b/cg/meta/demultiplex/housekeeper_storage_functions.py
@@ -5,9 +5,6 @@ from typing import List, Optional
 
 from housekeeper.store.models import File, Version
 
-from cg.apps.demultiplex.sample_sheet.read_sample_sheet import (
-    get_sample_internal_ids_from_sample_sheet,
-)
 from cg.apps.housekeeper.hk import HousekeeperAPI
 from cg.constants.housekeeper_tags import SequencingFileTag
 from cg.constants.sequencing import Sequencers
@@ -72,11 +69,7 @@ def add_sample_fastq_files_to_housekeeper(
     flow_cell: FlowCellDirectoryData, hk_api: HousekeeperAPI, store: Store
 ) -> None:
     """Add sample fastq files from flow cell to Housekeeper."""
-    sample_internal_ids: List[str] = get_sample_internal_ids_from_sample_sheet(
-        sample_sheet_path=flow_cell.sample_sheet_path,
-        flow_cell_sample_type=flow_cell.sample_type,
-    )
-
+    sample_internal_ids: List[str] = flow_cell.sample_sheet.get_sample_ids()
     for sample_internal_id in sample_internal_ids:
         sample_fastq_paths: Optional[List[Path]] = get_sample_fastqs_from_flow_cell(
             flow_cell_directory=flow_cell.path, sample_internal_id=sample_internal_id
--- a/cg/meta/demultiplex/housekeeper_storage_functions.py
+++ b/cg/meta/demultiplex/housekeeper_storage_functions.py
@@ -73,7 +73,7 @@ def add_sample_fastq_files_to_housekeeper(
 ) -> None:
     """Add sample fastq files from flow cell to Housekeeper."""
     sample_internal_ids: List[str] = get_sample_internal_ids_from_sample_sheet(
-        sample_sheet_path=flow_cell.get_sample_sheet_path_hk(),
+        sample_sheet_path=flow_cell.sample_sheet_path,
         flow_cell_sample_type=flow_cell.sample_type,
     )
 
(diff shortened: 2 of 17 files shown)


cg/models/demultiplex/flow_cell.py
```diff
@@ -75,21 +81,28 @@ def full_name(self) -> str:
...
+            self._sample_sheet = self._read_sample_sheet()
+        return self._sample_sheet
+
+    def _read_sample_sheet(self) -> SampleSheet:
+        if not self.sample_sheet_path_hk:
+            raise FlowCellError("Sample sheet path not set")
+        return get_sample_sheet_from_file(
+            infile=self.sample_sheet_path_hk,
+            flow_cell_sample_type=self.sample_type,
+        )
```
seallard, line 97: It might make more sense to pass the actual `SampleSheet` object to the constructor of the `FlowCellDirectoryData` class instead. I think the class becomes easier to test that way and it would not contain logic for parsing a sample sheet any longer

cg/meta/demultiplex/housekeeper_storage_functions.py
```diff
@@ -72,11 +69,7 @@ def add_sample_fastq_files_to_housekeeper(
     flow_cell: FlowCellDirectoryData, hk_api: HousekeeperAPI, store: Store
 ) -> None:
     """Add sample fastq files from flow cell to Housekeeper."""
-    sample_internal_ids: List[str] = get_sample_internal_ids_from_sample_sheet(
-        sample_sheet_path=flow_cell.get_sample_sheet_path_hk(),
-        flow_cell_sample_type=flow_cell.sample_type,
-    )
-
+    sample_internal_ids: List[str] = flow_cell.sample_sheet.get_sample_ids()
```
ChrOertlin, line 72:
```suggestion
    sample_internal_ids: Set[str] = flow_cell.sample_sheet.get_sample_ids()
```
The sample sheet has duplicated sample internal ids, might be safer to get a set.

cg/meta/demultiplex/housekeeper_storage_functions.py
```diff
@@ -238,3 +231,13 @@ def get_sample_sheets_from_latest_version(flow_cell_id: str, hk_api: Housekeeper
     except HousekeeperBundleVersionMissingError:
         sample_sheet_files: List = []
     return sample_sheet_files
+
+
+def get_sample_sheet_path_hk(flow_cell_id: str, hk_api: HousekeeperAPI) -> Path:
+    """Returns the path to the sample sheet for the given bundle."""
+    sample_sheet_files: List[File] = get_sample_sheets_from_latest_version(
```
ChrOertlin, line 238:
```suggestion
    sample_sheet_files: List[File] = get_sample_sheets_from_latest_version(
```
Why would this be a list? I think we do not allow multiple sample sheets for a flow cell. or is the function a more generic one. Likely out of scope but good to think about.

cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -37,10 +34,7 @@ def store_flow_cell_data_in_status_db(
     else:
         LOG.info(f"Flow cell already exists in status db: {parsed_flow_cell.id}.")
 
-    sample_internal_ids = get_sample_internal_ids_from_sample_sheet(
-        sample_sheet_path=parsed_flow_cell.get_sample_sheet_path_hk(),
-        flow_cell_sample_type=parsed_flow_cell.sample_type,
-    )
+    sample_internal_ids = parsed_flow_cell.sample_sheet.get_sample_ids()
```
ChrOertlin, line 37:
```suggestion
    sample_internal_ids: List[str] = parsed_flow_cell.sample_sheet.get_sample_ids()
```

cg/models/demultiplex/flow_cell.py
```diff
@@ -45,7 +50,8 @@ def __init__(self, flow_cell_path: Path, bcl_converter: Optional[str] = None):
         self.position: Literal["A", "B"] = "A"
         self.parse_flow_cell_dir_name()
         self.bcl_converter: Optional[str] = self.get_bcl_converter(bcl_converter)
-        self._sample_sheet_path_hk: Optional[Path] = None
+        self.sample_sheet_path_hk: Optional[Path] = sample_sheet_path
```
ChrOertlin, line 53:
```suggestion
        self.sample_sheet_path: Optional[Path] = sample_sheet_path
```
Not sure whether we should care if it is from housekeeper here.

cg/models/demultiplex/flow_cell.py
```diff
@@ -75,21 +81,28 @@ def full_name(self) -> str:
...
+    @property
+    def sample_sheet(self) -> SampleSheet:
+        """Return sample sheet object."""
+        if not self._sample_sheet:
+            self._sample_sheet = self._read_sample_sheet()
+        return self._sample_sheet
+
+    def _read_sample_sheet(self) -> SampleSheet:
+        if not self.sample_sheet_path_hk:
+            raise FlowCellError("Sample sheet path not set")
```
ChrOertlin, line 93:
```suggestion
        if not self.sample_sheet_path_hk:
            raise FlowCellError("Sample sheet path not set")
```
Would this be needed if we set the `self.sample_sheet_path_hk` to required instead of optional? 
I think with the current flow there always has to be a sample sheet anyway.

cg/models/demultiplex/flow_cell.py
ChrOertlin: I think that would move the reading/ parsing of the sample sheet to other places where we call the `FlowCellDirectoryData`? Not sure if that would require more changes overall.

tests/apps/demultiplex/test_read_sample_sheet.py
```diff
@@ -183,14 +183,15 @@ def test_get_sample_internal_ids_from_sample_sheet(
...
 ):
     """Test that getting sample internal ids from a sample sheet returns a unique list of strings."""
-    # GIVEN a path to a sample sheet with only valid samples
-
-    # WHEN getting the valid sample internal ids
-    sample_internal_ids: List[str] = get_sample_internal_ids_from_sample_sheet(
-        sample_sheet_path=novaseq6000_bcl_convert_sample_sheet_path,
+    # GIVEN a sample sheet with only valid samples
+    sample_sheet: SampleSheet = get_sample_sheet_from_file(
+        infile=novaseq6000_bcl_convert_sample_sheet_path,
```
ChrOertlin, line 188:
```suggestion
        sample_sheet_path=novaseq6000_bcl_convert_sample_sheet_path,
```
out of scope likely

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -172,6 +164,13 @@ def get_all_demultiplexed_flow_cell_dirs(self) -> List[Path]:
                 demultiplex_flow_cells.append(flow_cell_dir)
         return demultiplex_flow_cells
 
+    def get_sample_sheet_path(self, flow_cell_dir_name: str) -> Path:
+        flow_cell_id: str = get_flow_cell_id(flow_cell_dir_name)
+        return get_sample_sheet_path_hk(flow_cell_id=flow_cell_id, hk_api=self.hk_api)
+
+    def get_flow_cell_path(self, flow_cell_dir_name: str) -> Path:
```
ChrOertlin, line 171:
```suggestion
    def get_demultiplexed_runs_flow_cell_path(self, flow_cell_dir_name: str) -> Path:
```

As to not confuse it with the `flow_cells` i.e. flow_cell_runs directory

ChrOertlin (approved):
Looks good. I like making the sample sheet integral to the `FlowCellDirectoryData` since flow cells should always have a sample sheet anyway.

cg/models/demultiplex/flow_cell.py
seallard: I thought about that too, it would not be necessary and it seems better. But `FlowCellDirectoryData` is instantiated in other parts of the code where we don't have the `sample_sheet_path`. So might require additional refactoring. 

What do you think of doing it in a separate PR?

cg/meta/demultiplex/housekeeper_storage_functions.py
seallard: True, I did not write the function so I'm not sure. Something we can refactor

ChrOertlin: another PR sounds good.

diitaz93:
I agree that the `FlowCellDirectoryData` always needs a sample sheet and that the current implementation is very confusing with the calls to Housekeeper. However, I feel that adding the sample sheet path as parameter is not making the implementation cleaner and it opens the possibility of easily changing the 'official' sample sheet for the `FlowCellDirectoryData`. 
I feel that the sample sheet path should be something intrinsic to the `FlowCelDirectoryData` and should not be assigned from the outside. I am thinking that maybe it could be a good idea to raise an error when creating a `FlowCellDirectoryData` instance if there is no sample sheet in Housekeeper and keep that sample sheet as the official one. However I am thinking of multiple downsides of this idea, one of them being that we use `FlowCellDirectoryData` instances initialised from demultiplex run dirs instead of flow cell run dirs. Maybe we can bring this up for refinement next session? What do you think? @seallard @ChrOertlin

ChrOertlin:
> I agree that the `FlowCellDirectoryData` always needs a sample sheet and that the current implementation is very confusing with the calls to Housekeeper. However, I feel that adding the sample sheet path as parameter is not making the implementation cleaner and it opens the possibility of easily changing the 'official' sample sheet for the `FlowCellDirectoryData`. I feel that the sample sheet path should be something intrinsic to the `FlowCelDirectoryData` and should not be assigned from the outside. I am thinking that maybe it could be a good idea to raise an error when creating a `FlowCellDirectoryData` instance if there is no sample sheet in Housekeeper and keep that sample sheet as the official one. However I am thinking of multiple downsides of this idea, one of them being that we use `FlowCellDirectoryData` instances initialised from demultiplex run dirs instead of flow cell run dirs. Maybe we can bring this up for refinement next session? What do you think? @seallard @ChrOertlin

I agree with that passing a sample sheet in the constructor should likely be avoided. We decided to keep the samplesheet centralised in housekeeper as to have a common file to use in all processes downstream of sample sheet creation (i.e. demultiplexing, post processing). So if it is not there, something went wrong and should prompt investigation which can be done by throwing an error indeed.

seallard:
This refactoring is pretty much identical to the previous one, it just removes the getter and setter on the `FlowCellDirectoryData` object since those are not commonly used in Python and it instead accepts the path in the constructor. Not sure I follow your reasoning @ChrOertlin @diitaz93

seallard:
> I agree that the `FlowCellDirectoryData` always needs a sample sheet and that the current implementation is very confusing with the calls to Housekeeper. However, I feel that adding the sample sheet path as parameter is not making the implementation cleaner and it opens the possibility of easily changing the 'official' sample sheet for the `FlowCellDirectoryData`. I feel that the sample sheet path should be something intrinsic to the `FlowCelDirectoryData` and should not be assigned from the outside. I am thinking that maybe it could be a good idea to raise an error when creating a `FlowCellDirectoryData` instance if there is no sample sheet in Housekeeper and keep that sample sheet as the official one. However I am thinking of multiple downsides of this idea, one of them being that we use `FlowCellDirectoryData` instances initialised from demultiplex run dirs instead of flow cell run dirs. Maybe we can bring this up for refinement next session? What do you think? @seallard @ChrOertlin

It is currently assigned from the outside already with the `set_sample_sheet_path_hk`

```python
    def set_sample_sheet_path_hk(self, hk_path: Path):
        self._sample_sheet_path_hk = hk_path
```

cg/models/demultiplex/flow_cell.py
seallard: I don't think it would since it is an optional argument and we only provide it in the post processing flow

seallard: We can decide on it later

cg/meta/demultiplex/housekeeper_storage_functions.py
seallard: The function actually returns a list of unique ids already, maybe the function should be renamed to `get_unique_sample_ids` or something

diitaz93:
> This refactoring is pretty much identical to the previous one, it just removes the getter and setter on the `FlowCellDirectoryData` object since those are not commonly used in Python and it instead accepts the path in the constructor. Not sure I follow your reasoning @ChrOertlin @diitaz93

My point is that the refactoring should be that a FlowCellDirectoryData should have a sample sheet upon creation otherwise raise an error. The sample sheet path should always have the same structure and should be an internal construction, not a path coming from the outside. Right now is set from the outside but it should not be like that.

seallard:
> > This refactoring is pretty much identical to the previous one, it just removes the getter and setter on the `FlowCellDirectoryData` object since those are not commonly used in Python and it instead accepts the path in the constructor. Not sure I follow your reasoning @ChrOertlin @diitaz93
> 
> My point is that the refactoring should be that a FlowCellDirectoryData should have a sample sheet upon creation otherwise raise an error. The sample sheet path should always have the same structure and should be an internal construction, not a path coming from the outside. Right now is set from the outside but it should not be like that.

Ah, ok! So you want to move the call against housekeeper to inside the `FlowCellDirectoryData`?

cg/meta/demultiplex/housekeeper_storage_functions.py
seallard: Updated the docstring of `get_sample_ids` to clarify it is a list of unique ids.

cg/models/demultiplex/flow_cell.py
seallard: The name clashes with 
```
    @property
    def sample_sheet_path(self) -> Path:
        """
        Return sample sheet path.
        """
        return Path(self.path, DemultiplexingDirsAndFiles.SAMPLE_SHEET_FILE_NAME.value)
```
I renamed this one to `sample_sheet_run_dir_path`

tests/apps/demultiplex/test_read_sample_sheet.py
seallard: I'll do it together with the other refactoring in a separate PR 👍

Issue #2449, opened 2023-09-11
Incorrect hamming distance calculation and missing OverRideCycles


karlnyr:
## Description

Below is a stack trace from the demultiplexing with DRAGEN:

```Command
ERROR: Sample Sheet Error: Insufficient hamming distance for 1-mismatch-tolerance between AGGTTATAGT and AAGTTCTAGT at common value AAGTTATAGT in i5 index of lane 1
ERROR: Sample Sheet Error: hamming distance errors occurred in the Sample Sheet
ERROR: Assertion failed in ../src/host/dragen_api/file_io/bcl/process_bcl_directory.cpp line 1062 -- false -- ERROR: Exiting due to hamming distance errors2023-09-10T08:30:29Z thread 197828 ERROR: Assertion failed in ../src/host/dragen_api/file_io/bcl/process_bcl_directory.cpp line 1062 -- false -- 2023-09-10T08:30:29Z thread 197828 ERROR: Exiting due to hamming distance errors
Dumping diagnostics....
DRAGEN replay file saved to /var/log/dragen/dragen_replay_1694334629117_197828.json
sh: pidof: command not found
DRAGEN registers saved to /var/log/dragen/dragen_info_1694334629117_197828.log
Hang diagnostic saved to /var/log/dragen/hang_diag_1694334629117_197828.txt
pstack saved to /var/log/dragen/pstack_1694334629449_197828.log

Fatal exception: 2023-09-10T08:30:29Z thread 197828 ERROR: Assertion failed in ../src/host/dragen_api/file_io/bcl/process_bcl_directory.cpp line 1062 -- false -- 2023-09-10T08:30
:29Z thread 197828 ERROR: Exiting due to hamming distance errors

Fatal error: 2023-09-10T08:30:29Z thread 197828 ERROR: Assertion failed in ../src/host/dragen_api/file_io/bcl/process_bcl_directory.cpp line 1062 -- false -- 2023-09-10T08:30:29Z
 thread 197828 ERROR: Exiting due to hamming distance errors
```

This implies that the allowed barcode mistmatches is incorrectly set in the sample sheet. Furthermore, we are not correctly adding the OverRideCycles command into the sample sheet.
 
## Suggested solution

I suggest that we try to implement logic from LIMS to include OverRideCycles and the correct calculations for BarcodeMismatches.

## This can be closed when

We can demultiplex `HJWJFDSX7` with a samplesheet generated by `cg`

## Blocked by

[...]

PR #2454, opened 2023-09-11
sample sheet - override cycles, barcode mismatches and hamming distance fixes

diitaz93:
## Description

This PR addresses three things described in #2449 :
1. Fix the barcode mismatch assigning for NovaSeq6000, as it was set to 1 (default) for all cases. The problem was that the correct barcode mismatch assigning was only done for NovaSeqX.
2. Implement the assigning of OverrideCycles for BCLConvert sample sheets.
3. Fix the hamming distance calculation based on the discussion in https://github.com/Clinical-Genomics/cg_lims/issues/392

### Added

- Functions to calculate the hamming distance independently for index1 (i7) and index2 (i5) according to what we agreed upon on https://github.com/Clinical-Genomics/cg_lims/issues/392
- Tests for the new hamming distance calculators
- Functions to update barcode mismatch values for each sample in each SampleSheetCreator class
- Functions to calculate override cycles for each sample in each SampleSheetCreator class
- Attribute `OverrideCycles` to FlowCellSampleBCLConvert` model

### Changed

- Splitted the function `adapt_samples` into `update_indexes_for_samples` and `update_barcode_mismatch_values_for_samples`
- Renamed function `adapt_indexes_for_sample` to `pad_and_reverse_complement_sample_indexes`

### Fixed

- Removed if statement that excluded NovaSeq6000 barcode mismatches to be updated
- Modified the checks for padding to only pass when the samples are from a Bcl2fastq flow cell
- Removed repeated fixtures and updated usages

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b sample-sheet-override-and-barcode -a
    ```

### How to test

- [x] Delete a sample sheet from a flow cell on stage and from HK
- [x] `cg -l DEBUG demultiplex samplesheet create <flow-cell-name> --dry-run``
- [x] Do this for BCLConvert and Bcl2fastq

### Expected test outcome

- [x] Verify that the sample sheet was created correctly

## Review

- [x] Tests executed by SD
- [x] "Merge and deploy" approved by VJ

## Implementation Plan

- [ ] Deploy this branch on 
  - [ ] stage
  - [ ] production


cg/apps/demultiplex/sample_sheet/index.py
```diff
@@ -142,14 +146,22 @@ def pad_index_two(index_string: str, reverse_complement: bool) -> str:
     return index_string + INDEX_TWO_PAD_SEQUENCE
 
 
-def get_hamming_distance_for_indexes(sequence_1: str, sequence_2: str) -> int:
-    """Get the hamming distance between two index sequences.
+def get_hamming_distance_index_1(sequence_1: str, sequence_2: str) -> int:
+    """Get the hamming distance between two index 1 sequences.
     In the case that one sequence is longer than the other, the distance is calculated between
     the shortest sequence and the first segment of equal length of the longest sequence."""
     limit: int = min(len(sequence_1), len(sequence_2))
```
ChrOertlin, line 153:
```suggestion
    limit: int = get_sequence_limit(sequence_1=sequence_1, sequence_2=sequence_2)
  
def get_sequence_limit(sequence_1: int, sequence_2: int) -> int:
    return min(len(sequence_1), len(sequence_2))
```

cg/apps/demultiplex/sample_sheet/index.py
```diff
@@ -142,14 +146,22 @@ def pad_index_two(index_string: str, reverse_complement: bool) -> str:
...
     return get_hamming_distance(str_1=sequence_1[:limit], str_2=sequence_2[:limit])
 
 
+def get_hamming_distance_index_2(sequence_1: str, sequence_2: str) -> int:
+    """Get the hamming distance between two index 2 sequences.
+    In the case that one sequence is longer than the other, the distance is calculated between
+    the shortest sequence and the last segment of equal length of the longest sequence."""
+    limit: int = min(len(sequence_1), len(sequence_2))
```
ChrOertlin, line 161:
```suggestion
    limit: int = [min(len(sequence_1), len(sequence_2))](limit: int = get_sequence_limit(sequence_1=sequence_1, sequence_2=sequence_2))
```

cg/apps/demultiplex/sample_sheet/index.py
ChrOertlin: With next comment - keep it dry

cg/apps/demultiplex/sample_sheet/index.py
```diff
@@ -142,14 +146,22 @@ def pad_index_two(index_string: str, reverse_complement: bool) -> str:
...
+    return get_hamming_distance(str_1=sequence_1[-limit:], str_2=sequence_2[-limit:])
```
ChrOertlin, line 162: Not sure if this is needed to know but we could clarify why we flip and sign the limit here compared to index_1?

cg/apps/demultiplex/sample_sheet/index.py
```diff
@@ -194,7 +202,9 @@ def adapt_indexes_for_sample(
     3. Assigns the indexes to the sample attributes index1 and index2."""
     index1, index2 = get_index_pair(sample=sample)
     index_length = len(index1)
-    if is_padding_needed(index_cycles=index_cycles, sample_index_length=index_length):
+    if type(sample) == FlowCellSampleBcl2Fastq and is_padding_needed(
```
ChrOertlin, line 205:
```suggestion
    if isInstance(sample, FlowCellSampleBcl2Fastq) and is_padding_needed(
```
Would this work too? I think `type` fetches the class object

ChrOertlin (commented):
Looks good, don't know enough to approve.

diitaz93:
## Tests on stage
Test for NovaSeq6000 BCLConvert
```shell
Sample sheet passed validation
[Header]
FileFormatVersion,2
RunName,H7FYHDRX2
InstrumentPlatform,NovaSeq6000
IndexOrientation,Forward
[Reads]
Read1Cycles,51
Read2Cycles,51
Index1Cycles,8
Index2Cycles,8
[BCLConvert_Settings]
SoftwareVersion,4.1.7
FastqCompressionFormat,gzip
[BCLConvert_Data]
Lane,Sample_ID,Index,Index2,OverrideCycles,AdapterRead1,AdapterRead2,BarcodeMismatchesIndex1,BarcodeMismatchesIndex2
1,ACC10641A24,GATCAACA,TAACATAG,Y51;I8;I8;Y51;,,,1,1
etc.....
```

cg/apps/demultiplex/sample_sheet/index.py
diitaz93: fixed

cg/apps/demultiplex/sample_sheet/index.py
diitaz93: The explanation is quite long to include in the docstring I think. For more info https://github.com/Clinical-Genomics/cg/issues/2449

Vince-janv (changes requested):
Nice work! I might have found a bug though. Please have a look 😊

cg/apps/demultiplex/sample_sheet/index.py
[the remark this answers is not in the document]
diitaz93: I was thinking of `index1` and `index2` but that would be confusing with the other nomenclature of index1 = i7 and index2 = 15

ChrOertlin: the code: `limit: int = min(len(sequence_1), len(sequence_2))` is repeated below. It can be put in a function which can be called in both places.

tests/apps/demultiplex/test_sample_sheet_creator.py
```diff
@@ -136,3 +137,21 @@ def test_remove_unwanted_samples_no_dual_index(
...
+
+    assert all(sample.override_cycles == "" for sample in samples)
+
+    # WHEN adding the correct values of override samples
+    bcl_convert_sample_sheet_creator.process_samples_for_sample_sheet()
+    samples: List[FlowCellSampleBCLConvert] = bcl_convert_sample_sheet_creator.lims_samples
+
+    # THEN the Override Cycles attribute is added to all samples
+    pattern: str = r"^Y.*I.*I.*Y.*$"
+    assert all(re.match(pattern, sample.override_cycles) for sample in samples)
```
islean, line 157: I would ideally like to see a test that checks if the correct cycles are added. This would also catch the potential bug that @Vince-janv have spotted. Might take some effort with the setup though.

cg/apps/demultiplex/sample_sheet/sample_sheet_creator.py
```diff
@@ -196,8 +214,43 @@ def __init__(
...
+                    + str(flow_cell_index_len - sample_index_len)
+                    + ";"
+                )
+                index2_str = (
+                    "N"
+                    + str(flow_cell_index_len - sample_index_len)
+                    + "I"
+                    + str(sample_index_len)
+                    + ";"
+                )
```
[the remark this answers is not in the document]
diitaz93: Of course, fixed

tests/conftest.py
```diff
@@ -1388,7 +1388,7 @@ def lims_novaseq_samples_file(raw_lims_sample_dir: Path) -> Path:
     return Path(raw_lims_sample_dir, "raw_samplesheet_novaseq.json")
 
 
-@pytest.fixture(name="lims_novaseq_samples_raw")
+@pytest.fixture()
```
islean, line 1391:
```suggestion
@pytest.fixture
```
for consistency

islean (commented):
Looks nice but I think Vincent's comment needs to be addressed as I agree it could be a bug.

tests/apps/demultiplex/test_sample_sheet_creator.py
diitaz93: You are totally right. And this has to do with being more strict with tests. I will give it a try

cg/apps/demultiplex/sample_sheet/index.py
```diff
@@ -184,7 +192,7 @@ def update_barcode_mismatch_values_for_sample(
             sample_to_update.barcode_mismatches_2 = 0
 
 
-def adapt_indexes_for_sample(
+def pad_and_reverse_complement_sample_indexes(
```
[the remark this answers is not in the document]
islean: Hmm good point - not really. I am probably slightly in favour of having the somewhat vaguer "adapt" indexes or something similar and then having smaller methods called "convert_to_reverse_complement" and "pad_indexes" or something like that. It seems clearer to me now than when I first saw it, though. Might have been thrown off by "reverse_complement" being used as a verb (or is it a verb as well?).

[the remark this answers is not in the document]
diitaz93: yeah, in this context reverse complement is used as a verb. But then yes, is more confusing than I thought 😅  I will consider renaming

cg/apps/demultiplex/sample_sheet/models.py
```diff
@@ -49,6 +49,7 @@ class FlowCellSampleBCLConvert(FlowCellSample):
     sample_id: str = Field(..., alias=SampleSheetBCLConvertSections.Data.SAMPLE_INTERNAL_ID.value)
     index: str = Field(..., alias=SampleSheetBCLConvertSections.Data.INDEX_1.value)
     index2: str = Field("", alias=SampleSheetBCLConvertSections.Data.INDEX_2.value)
+    override_cycles: str = Field("", alias="OverrideCycles")
```
Karl-Svard, line 52:
```suggestion
    override_cycles: str = Field("", alias=SampleSheetBCLConvertSections.Data.OVERRIDE_CYCLES.value)
```

cg/apps/demultiplex/sample_sheet/index.py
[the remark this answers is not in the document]
islean: I am okay with you disregarding this, it is only a slight preference and if you prefer this naming, feel free to keep it 😄

cg/apps/demultiplex/sample_sheet/models.py
diitaz93: fixed

cg/apps/demultiplex/sample_sheet/index.py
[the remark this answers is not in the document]
diitaz93: We can think of a better name later, when we refactor the whole demux code 😊

tests/apps/demultiplex/test_sample_sheet_creator.py
diitaz93: Take a look now @islean

cg/apps/demultiplex/sample_sheet/index.py
```diff
@@ -184,39 +196,38 @@ def update_barcode_mismatch_values_for_sample(
...
+    """Updates the values to the fields index1 and index 2 of samples."""
     index_cycles: int = run_parameters.index_length
     for sample in samples:
-        if run_parameters.sequencer == Sequencers.NOVASEQX:
-            update_barcode_mismatch_values_for_sample(
-                sample_to_update=sample, samples_to_compare_to=samples
-            )
-        adapt_indexes_for_sample(
+        pad_and_reverse_complement_sample_indexes(
             sample=sample,
```
karlnyr, line 231: Is this function not just obsolete? Why not just use pad_and_reverse_complemet_sample_indexes and put the for loop in there?

cg/apps/demultiplex/sample_sheet/sample_sheet_creator.py
```diff
@@ -112,17 +128,17 @@ def process_samples_for_sample_sheet(self) -> None:
...
-            )
-            adapt_samples(
+            LOG.info(f"Adapting index and barcode mismatch values for samples in lane {lane}")
+            self.update_barcode_mismatch_values_for_samples(samples_in_lane)
+            update_indexes_for_samples(
                 samples=samples_in_lane,
                 run_parameters=self.run_parameters,
                 reverse_complement=reverse_complement,
             )
+        self.add_override_cycles_to_samples()
```
karlnyr, line 141: here, I think that semantics is everything. We need to first add the override cycles, then we need to pad_revcomp indexes and after that we need to update the barcode mismatches. Override cycles will tell us how many index cycles the sample has. Padding should then be done afterwards, even though we currently pad based on the sample type. The last thing we want to do is to update the barcode mismatch values, once again, the way all our logic plays together now wont allow us to pad or alter the index length, but if we did, then this would also be a bug. I don't know how comprehensive the test is, so I would simply put it in the semantic order in which it could prevent bugs to begin with.

tests/apps/demultiplex/test_index.py
```diff
@@ -308,5 +308,37 @@ def test_get_hamming_distance_index_different_prefixes():
...
+    """Test that the hamming distance between indexes with different suffixes is greater than zero."""
+    # GIVEN two index_2 sequences with the same prefixes but different suffixes
+    sequence_1: str = "GATTACA"
+    sequence_2: str = "GATTACAXX"
+
+    # WHEN getting the hamming distance between them in any order
+
+    # THEN the distance is greater than zero
+    assert get_hamming_distance_index_2(sequence_1=sequence_1, sequence_2=sequence_2) > 0
+    assert get_hamming_distance_index_2(sequence_1=sequence_2, sequence_2=sequence_1) > 0
```
karlnyr, line 344: Any reason to not assert this to be the exact number that you expect it to be? I feel like this is something we cannot accept be ANYTHING but 2. Goes for the tests above as well.

cg/apps/demultiplex/sample_sheet/index.py
diitaz93: for SRP and clarity, I thought the inner function should do the update for one sample and the outer function loop over all functions. This way the functions are easier to test. Do you have any reason why we wanted it as one big function? I think `pad_and_reverse_complement_sample_indexes` is already a bit complex on its own

cg/apps/demultiplex/sample_sheet/sample_sheet_creator.py
diitaz93: Agree, fxed

tests/apps/demultiplex/test_index.py
diitaz93: In this case, the distance is actually not 2 but 6. As this is an index2 comparison, it will always compare aligning from the tail. Making the expected distance explicitly equal to 6 in my opinion, is not that meaningful as you as a reviewer would have to manually verify that aligning by the tails gives 6, while the most important thing is to show that the function is not aligning from the head, otherwise it would be zero. I can put it as 6 if you think it would be clearer but honestly, I think the number is not important.

diitaz93:
## New tests on stage deliver correct output for both sample sheets

Vince-janv (approved):
My comment have been adressed 👍 Please get approval of @karlnyr before merging

cg/apps/demultiplex/sample_sheet/index.py
diitaz93: done

cg/apps/demultiplex/sample_sheet/sample_sheet_creator.py
```diff
@@ -183,21 +208,48 @@ def get_data_section_header_and_columns(self) -> List[List[str]]:
...
+            index2_str: str = f"I{str(self.run_parameters.get_index_2_cycles())};"
+            sample_index_len: int = len(get_index_pair(sample)[0])
+            if sample_index_len < flow_cell_index_len:
+                index1_str = (
+                    f"I{str(sample_index_len)}N{str(flow_cell_index_len - sample_index_len)};"
+                )
+                index2_str = (
+                    f"N{str(flow_cell_index_len - sample_index_len)}I{str(sample_index_len)};"
+                )
+            sample.override_cycles = read1_str + index1_str + index2_str + read2_str
```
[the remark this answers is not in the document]
diitaz93: agree, fixed

tests/apps/demultiplex/test_index.py
diitaz93: Ok I refactored the test to be clearer

diitaz93:
## Deployed to stage:
```shell
repository is clean
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/js.diazboada
done.
Log deploy... done.
cg, version 49.0.3
θ63° 1d [js.diazboada@hasta:~] [S_base] 3m30s
```
## Deployed to production:
```shell
repository is clean
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/js.diazboada
done.
Log deploy... done.
cg, version 49.0.3
θ63° 1d [js.diazboada@hasta:~] [P_base] 3m30s
```

Issue #2456, opened 2023-09-11
Broken fluffly CLI command


seallard:
## Description
`cg workflow fluffy create-samplesheet <case_id>` is broken (how?).

The command does not actually create a sample sheet, but rather a config file which is a modified version of a sample sheet. It outputs a `SampleSheet_{starlims_id}.csv` in the fluffy root directory (`fluffy_root/<case_id>`) containing data from the sample sheet, status db and lims.

## This can be closed when
`cg workflow fluffy create-samplesheet` can be run

seallard:
Reproducing the issue in stage

Create sample sheet for `secret_case_id`:
```cg demultiplex create samplesheet <secret_flow_cell_dir_name```

seallard:
Example of sample sheet which works: `/home/proj/production/flow_cells/230907_A00621_0903_AHJT27DRX3/SampleSheet.csv`

Example of new sample sheet which breaks:
`/home/proj/production/housekeeper-bundles/HJT27DRX3/2023-09-07/SampleSheet_dragen.csv`

seallard:
The generated "sample sheet"/config file should be a csv with the following columns:

- FCID
- Lane
- Sample_ID
- SampleRef
- index
- index2
- SampleName
- Control
- Recipe
- Operator
- Sample_Project
- Exclude
- Library_nM
- SequencingDate

seallard:
Completed in https://github.com/Clinical-Genomics/cg/pull/2468

PR #2458, opened 2023-09-11
#2456: Fix fluffy config generation wip

seallard:
## Description
This PR adresses https://github.com/Clinical-Genomics/cg/issues/2456. The error message indicates that the header was not identified correctly. The logic could probably use a complete refactoring and also an overview of which fields actually are used in the fluffy repo.

The bug has not been reproduced in stage, so I can't test it. Trying to reproduce it with an automated test instead, digging for a sample sheet which causes the error.

### Fixed
- Fix extraction of header in fluffy sample sheet generation


ChrOertlin (approved):
Looks good! This should make the header line count identification work for both sample sheet versions.

seallard:
This will fix the reported error, but the actual issue is that fluffy relies on the bcl2fastq sample sheet structure. So the output fluffy sample sheet need to follow the bcl2fastq structure.

henrikstranneheim (approved):
Nice 👍

henrikstranneheim:
> This will fix the reported error, but the actual issue is that fluffy relies on the bcl2fastq sample sheet structure. So the output fluffy sample sheet need to follow the bcl2fastq structure.

@J35P312 It would be great to decouple Fluffy from this. Do you think it is worth changing this for the current version of Fluffy or do we wait for the NF-version?

seallard:
> > This will fix the reported error, but the actual issue is that fluffy relies on the bcl2fastq sample sheet structure. So the output fluffy sample sheet need to follow the bcl2fastq structure.
> 
> @J35P312 It would be great to decouple Fluffy from this. Do you think it is worth changing this for the current version of Fluffy or do we wait for the NF-version?

According to @karlnyr we should just do a quick fix and refactor later. So I'm closing this in favour of resolving https://github.com/Clinical-Genomics/cg/issues/2465

J35P312:
Perfect! Good luck =P.

In theory we could perhaps even skip the samplesheet.... The samplesheet is used to transfer various data from the orderportal into fluffy (concentration and index for instance). This data is not used by fluffy, but Visulized in Statina or printed in the final output CSV. Maybe we could provide this data to Statina directly? Then we can drop the samplesheet in some future fluffy release.

henrikstranneheim:
@J35P312 Sounds good then we can create a separate file for this in HK instead and not couple this to Fluffy.

Issue #2459, opened 2023-09-11
Unused fields on sample sheet sample types


seallard:
## Description
Many of the fields on the flow cell sample sheet types (`FlowCellSampleBcl2Fastq` and `FlowCellSampleBcl2Fastq`) are not used anywhere. These add bloat and are likely to cause issues unnecessarily, we do not need to attempt to parse them if they are not used anywhere.

## Suggested solution
Determine if the fields are used anywhere, if not - remove them.

## This can be closed when
No unused fields for the samples in the sample sheet sample models exist.

seallard:
The demultiplexing software expects these fields to exist. So we need to keep them to be able to validate that the sample sheets are formatted correctly.

PR #2460, opened 2023-09-11
Remove unused sample sheet sample fields

seallard:
## Description
Resolves https://github.com/Clinical-Genomics/cg/issues/2459.
This PR removes some of the fields which are not referenced anywhere in cg for samples on the sample sheet model. To make the sample sheet parsing less brittle, it is better to only parse the needed information.

Should these remain? 
- sample_ref
- sample_name
- operator
- recipe
- control
- adapter_read_1
- adapter_read_2

### Fixed
- Remove some unused fields on the sample sheet sampel models

Currently working on an issue with fluffy sample sheets https://github.com/Clinical-Genomics/cg/issues/2456 where it might turn out that some of these are needed - so will hold off on merging until that issue is resolved.


ChrOertlin:
I think the demultiplexing software expects sample sheets to be in a specific format for them to work. That includes some of the columns that would be deleted in this PR.
If I am not mistaken, when we create sample sheets we validate the sample sheets to have a specific structure for the demultiplexing software (i.e. bcl2fastq or bclconvert). So then we would need those columns. @karlnyr , @diitaz93. Is this right?

seallard:
> I think the demultiplexing software expects sample sheets to be in a specific format for them to work. That includes some of the columns that would be deleted in this PR. If I am not mistaken, when we create sample sheets we validate the sample sheets to have a specific structure for the demultiplexing software (i.e. bcl2fastq or bclconvert). So then we would need those columns. @karlnyr , @diitaz93. Is this right?

Ok, so it might be necessary to keep these fields in order for the validation to work correctly?

ChrOertlin:
> 

I think so, yes. Here is the documentation for the BCL convert sample sheet as an example ( https://support.illumina.com/content/dam/illumina-support/documents/documentation/software_documentation/bcl_convert/bcl-convert-v3-7-5-software-guide-1000000163594-00.pdf)

diitaz93 (changes requested):
Many columns of the sample sheet are not used, and they are even empty for all samples. But the Illumina software expects these fields to be there otherwise it would raise an error.

diitaz93:
> > I think the demultiplexing software expects sample sheets to be in a specific format for them to work. That includes some of the columns that would be deleted in this PR. If I am not mistaken, when we create sample sheets we validate the sample sheets to have a specific structure for the demultiplexing software (i.e. bcl2fastq or bclconvert). So then we would need those columns. @karlnyr , @diitaz93. Is this right?
> 
> Ok, so it might be necessary to keep these fields in order for the validation to work correctly?

Exactly. Unfortunately we will have to keep those fields

seallard:
> > > I think the demultiplexing software expects sample sheets to be in a specific format for them to work. That includes some of the columns that would be deleted in this PR. If I am not mistaken, when we create sample sheets we validate the sample sheets to have a specific structure for the demultiplexing software (i.e. bcl2fastq or bclconvert). So then we would need those columns. @karlnyr , @diitaz93. Is this right?
> > 
> > 
> > Ok, so it might be necessary to keep these fields in order for the validation to work correctly?
> 
> Exactly. Unfortunately we will have to keep those fields

Good 👍

Issue #2098, opened 2023-06-02
Sync cg with database structure


islean:
## Description

When using auto generate and letting alembic sync the database with the python models in the genotype-api repo, many discrepancies were discovered, in particular with regards to indices. This proposes a potential risk to performance as the upgrade script would have dropped many indices if no manual action was taken. A smaller risk would be functionality dropping if there are larger discrepancies than just indices missing. Syncing for genotype-api was done upon discovery.

A similar syncing should be performed in cg.

## Suggested solution

Compare the Python models with the actual database structure and sync the two by deciding which version should be kept for any discrepancy found.

## This can be closed when

Python models agree with the database.

islean:
This is the auto-generated upgrade script:

def upgrade():
    # ### commands auto generated by Alembic - please adjust! ###
    op.alter_column("analysis", "family_id", existing_type=mysql.INTEGER(), nullable=True)
    op.alter_column(
        "application",
        "prep_category",
        existing_type=mysql.ENUM("cov", "mic", "rml", "tgs", "wes", "wgs", "wts"),
        nullable=False,
    )
    op.alter_column(
        "application", "min_sequencing_depth", existing_type=mysql.INTEGER(), nullable=False
    )
    op.alter_column(
        "family",
        "priority",
        existing_type=mysql.ENUM("research", "standard", "priority", "express", "clinical_trials"),
        nullable=False,
    )
    op.alter_column("invoice", "customer_id", existing_type=mysql.INTEGER(), nullable=False)
    op.alter_column("invoice", "comment", existing_type=mysql.TEXT(), nullable=True)
    op.drop_index("customer_invoice_pk", table_name="invoice")
    op.create_foreign_key(None, "invoice", "customer", ["customer_id"], ["id"])
    op.create_foreign_key(None, "pool", "invoice", ["invoice_id"], ["id"])
    op.alter_column(
        "sample",
        "priority",
        existing_type=mysql.ENUM("research", "standard", "priority", "express", "clinical_trials"),
        nullable=False,
    )
    op.create_foreign_key(None, "sample", "invoice", ["invoice_id"], ["id"])
    op.drop_column("sample", "beaconized_at")
    op.alter_column(
        "sample_lane_sequencing_metrics",
        "flow_cell_name",
        existing_type=mysql.VARCHAR(charset="latin1", collation="latin1_swedish_ci", length=32),
        nullable=False,
    )
    op.alter_column(
        "sample_lane_sequencing_metrics",
[...]

PR #2462, opened 2023-09-12
Sync cg and statusDB

islean:
## Description

Addresses issue #2098. Main differences will be that sample.beaconized_at is dropped and the adding of some foreign keys to the actual db, where the Python models indicate that it should be so.

### Added

- Revision

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b sync-cg-and-statusdb -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [x] Tests executed by IO
- [x] "Merge and deploy" approved by HS, SA

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
Test fails with following error:

```
sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (3780, "Referencing column 'invoice_id' and referenced column 'id' in foreign key constraint 'pool_ibfk_3' are incompatible.")
[SQL: ALTER TABLE pool ADD FOREIGN KEY(invoice_id) REFERENCES invoice (id)]
(Background on this error at: http://sqlalche.me/e/13/e3q8)
```

islean:
Tests fail with:

```
sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (1091, "Can't DROP 'customer_invoice_pk'; check that column/key exists")
[SQL: 
DROP INDEX customer_invoice_pk ON invoice]
```

islean:
Currently seeing a strange behaviour caused by there being 1467 samples with invoice_id = 0. There is no entry in the invoice table with id 0, rendering a foreign key from sample invoice impossible.

islean:
Latest entry with invoice_id = 0 was created in 2018. Safest way to proceed might be to create a dummy entry in the invoice table with id 0 so that we can keep track of these samples. Once they have an invoice, we should be able to create the foreign key though.

henrikstranneheim:
Sounds good as long as it easy to understand that this is a dummy entry.

islean:
Added dummy invoice in stage:

image

islean:
Still something wrong:

`[SQL: ALTER TABLE invoice MODIFY customer_id INTEGER NOT NULL]`

islean:
Tying it to cust000 at least made the deployment work:

image

islean:
Deployment to cg-vm1 went well and all tables are seemingly updated. E.g. sample.invoice_id now is a foreign key to invoice.id:

image

islean:
Downgrade worked:

image

henrikstranneheim (approved):
LGTM, but this Alembic script does a lot. It would be great with some more eyes on it @Clinical-Genomics/sysdev

alembic/versions/936887038ba8_sync_cg_and_statusdb.py
```diff
@@ -0,0 +1,121 @@
...
+    op.alter_column(
+        "family",
+        "priority",
+        existing_type=mysql.ENUM("research", "standard", "priority", "express", "clinical_trials"),
+        nullable=False,
+    )
+    op.alter_column("invoice", "customer_id", existing_type=mysql.INTEGER(), nullable=False)
+    op.alter_column("invoice", "comment", existing_type=mysql.TEXT(), nullable=True)
+
+    op.drop_index("customer_invoice_pk", table_name="invoice")
```
seallard, line 36: How come we are dropping this index?

seallard (approved):
Looks good 👍 
What if some of the values currently stored in the nullable columns are null? Maybe we need some default value or modify the data before it is possible to apply this migration in prod

alembic/versions/936887038ba8_sync_cg_and_statusdb.py
islean: It is basically a conversion with the line below. In the Python model we have this as a foreign key but in the actual database we have it as an index. Foreign keys are automatically indexed, though, so the line beneath should ensure that we have an index and a foreign key on the column. 

At the moment there would be no constraint to going into the db manually and an entry in the invoice with a customer_id which does not exist.

islean:
> Looks good 👍 What if some of the values currently stored in the nullable columns are null? Maybe we need some default value or modify the data before it is possible to apply this migration in prod

I would be surprised given that the tests have been on a copy of production and I have tried to check it but I agree, this is a concern. It won't be that the columns with null will be dropped, but given that alembic currently assumes a non-transactional DDL it could be a bit cumbersome resetting a broken alembic deployment, since some but not all changes have gone through. This would break the downgrade script since the alembic version has not been updated in the db table, and even if it had been, it would assume all changes have been made.

What do we think - would it be enough to check manually before deployment that all those tables fulfill the non-nullable requirements or would we want to do a for-loop à la "set "dummy_value" where null"? @Clinical-Genomics/sysdev ?

ChrOertlin:
> > Looks good 👍 What if some of the values currently stored in the nullable columns are null? Maybe we need some default value or modify the data before it is possible to apply this migration in prod
> 
> I would be surprised given that the tests have been on a copy of production and I have tried to check it but I agree, this is a concern. It won't be that the columns with null will be dropped, but given that alembic currently assumes a non-transactional DDL it could be a bit cumbersome resetting a broken alembic deployment, since some but not all changes have gone through. This would break the downgrade script since the alembic version has not been updated in the db table, and even if it had been, it would assume all changes have been made.
> 
> What do we think - would it be enough to check manually before deployment that all those tables fulfill the non-nullable requirements or would we want to do a for-loop à la "set "dummy_value" where null"? @Clinical-Genomics/sysdev ?

Can we get an overview that contains:
1. the extent to which these null values exist?
2. What are these attributes for which null values exist
3. Is there a way to fetch the data in question from some place else and fill in the proper values, if they exist?

Maybe this could be an attempt to make the database more complete?

islean:
> > > Looks good 👍 What if some of the values currently stored in the nullable columns are null? Maybe we need some default value or modify the data before it is possible to apply this migration in prod
> > 
> > 
> > I would be surprised given that the tests have been on a copy of production and I have tried to check it but I agree, this is a concern. It won't be that the columns with null will be dropped, but given that alembic currently assumes a non-transactional DDL it could be a bit cumbersome resetting a broken alembic deployment, since some but not all changes have gone through. This would break the downgrade script since the alembic version has not been updated in the db table, and even if it had been, it would assume all changes have been made.
> > What do we think - would it be enough to check manually before deployment that all those tables fulfill the non-nullable requirements or would we want to do a for-loop à la "set "dummy_value" where null"? @Clinical-Genomics/sysdev ?
> 
> Can we get an overview that contains:
> 
>     1. the extent to which these null values exist?
> 
>     2. What are these attributes for which null values exist
> 
>     3. Is there a way to fetch the data in question from some place else and fill in the proper values, if they exist?
> 
> 
> Maybe this could be an attempt to make the database more complete?

I have double checked when I did it on stage and when going through the autorevision to see what changes should be done to it. Also, Iíf there had been any such entries, the deployment to stage would have failed. So the extent unless any new entries have been malformed is zero. And since there are restrictions in the Python models, anything added via the CLI or via the web interface will already have these restrictions.

islean:
It would be more a question of "If we have missed anything - should we have a backup plan in the alembic script?" Of course this would have some problems, such as needing a decision on what to do with the values during the downgrade. 

I am in favour of believing that a deployment to stage immediately after copying prod to stage is enough of a security to forego this.

islean:
Deployment successful after adding constraint on Python model make Analysis.family_id non-nullable.

image

islean:
Deployment to cg-vm1 successful as well, although for some reason you need to rebuild the docker image sometimes in order to deploy it successfully.

islean:
Downgrade successful: 

image

alembic/versions/936887038ba8_sync_cg_and_statusdb.py
diitaz93: I will resolve this conversation.

diitaz93:
Deployed to stage, deployment failed in production due to an SQLAlchemy error

diitaz93:
Deployment done in fix #2490

Issue #2465, opened 2023-09-12
Always format the fluffy sample sheet the same way


seallard:
## Description
fluffy relies on the fluffy sample sheet generated in cg to contain these fields:

- FCID
- Lane
- Sample_ID
- SampleRef
- index
- index2
- SampleName
- Control
- Recipe
- Operator
- Sample_Project
- Exclude
- Library_nM
- SequencingDate

Due to some changes in the existing sample sheet generation logic, the generation of fluffy sample sheets fail and outputs the wrong format.

## Suggested solution
Adjust the fluffy sample sheet/config generation so it conforms to the structure above, no matter if the sample sheet it uses was generated for bcl2fastq or bclconvert.

## This can be closed when
- The `cg workflow fluffy create-samplesheet` command works
- The generated sample sheet follows the structure above and can be used in the fluffy pipeline

PR #2468, opened 2023-09-12
Fix fluffy config formatting

seallard:
## Description
This PR adresses https://github.com/Clinical-Genomics/cg/issues/2465.

The intent is to ensure that
1. `cg workflow fluffy create-samplesheet` works
2. The generated fluffy sample sheet always is structured in the same way

My first patch https://github.com/Clinical-Genomics/cg/pull/2458 was apparently insufficient due to not fixing 2.
The bug has not been reproduced in stage.

### Fixed
- Patch generation of fluffy sample sheets
- Add test reproducing bug with `cg workflow fluffy create-samplesheet`

### How to test

- [ ] Do `cg workflow fluffy create-samplesheet`


cg/apps/demultiplex/sample_sheet/read_sample_sheet.py
```diff
@@ -50,6 +57,20 @@ def get_sample_sheet_from_file(
...
+        return FlowCellSampleBcl2Fastq
+    return FlowCellSampleBCLConvert
+
+
+def get_bcl_converter_by_sequencer(sequencer_type: str) -> str:
+    """Return the BCL converter based on the sequencer."""
+    if sequencer_type in [Sequencers.NOVASEQ, Sequencers.NOVASEQX]:
+        return BclConverter.DRAGEN
+    return BclConverter.BCL2FASTQ
+
```
seallard, line 72: Not a fan of determining the sample sheet sample type like this, ideas are welcome 😅

cg/meta/workflow/fluffy.py
```diff
@@ -267,23 +227,25 @@ def make_sample_sheet(self, case_id: str, dry_run: bool) -> None:
...
-            "Writing modified csv from %s to %s",
-            sample_sheet_housekeeper_path,
-            sample_sheet_workdir_path,
+        flow_cell: Flowcell = self.status_db.get_latest_flow_cell_on_case(case_id)
+        sample_sheet_housekeeper_path: Path = self.get_sample_sheet_housekeeper_path(flow_cell.name)
+        flow_cell_sample_type = get_sample_type_from_sequencer_type(flow_cell.sequencer_type)
+
+        sample_sheet: SampleSheet = get_sample_sheet_from_file(
+            infile=sample_sheet_housekeeper_path,
+            flow_cell_sample_type=flow_cell_sample_type,
```
seallard, line 236: Maybe the `flow_cell_sample_type` can be made into an optional argument so we can hide away the logic for determining the sample type in the `get_sample_sheet_from_file` instead?

Vince-janv (approved):
Nice 💯

cg/meta/workflow/fluffy.py
```diff
@@ -20,6 +28,51 @@
...
+    control: str = "Control"
+    recipe: str = "Recipe"
+    operator: str = "Operator"
+    sample_project: str = "SampleProject"
+    exclude: str = "Exclude"
+    library_nM: str = "Library_nM"
+    sequencing_date: str = "SequencingDate"
+
+
+class FluffySampleSheetEntry(BaseModel):
```
diitaz93, line 45: to make it consistent with the other sample sheets
```suggestion
class FluffySampleSheetSample(BaseModel):
```

ChrOertlin (approved):
Nice refactoring of the sample sheet creation!!

cg/meta/workflow/fluffy.py
```diff
@@ -20,6 +28,51 @@
...
+    recipe: Optional[str] = "R1"
+    operator: Optional[str] = "script"
+    sample_project: str
+    exclude: bool
+    library_nM: float
+    sequencing_date: dt.date
+
+
+class FluffySampleSheet(BaseModel):
+    entries: List[FluffySampleSheetEntry]
```
diitaz93, line 63:
```suggestion
    samples: List[FluffySampleSheetSample]
```

cg/meta/workflow/fluffy.py
```diff
@@ -20,6 +28,51 @@
...
+
+    def write_sample_sheet(self, out_path: Path) -> None:
+        LOG.info("Writing fluffy sample sheet to {sample_sheet_out_path}")
```
diitaz93: Is this supposed to be a formatted string? Where is the sample_sheet_path?

cg/apps/demultiplex/sample_sheet/read_sample_sheet.py
islean: If this is the sole purpose, then I guess you could skip the middle step and just return FlowCellSampleBcl2FastQ if sequencer_type in [Sequencers.NOVASEQ, Sequencers.NOVASEQX] else FlowCellSampleBCLConvert?

Maybe not making it much clearer but I am uncertain if the middle step helps.

islean: But in general I guess it should be possible from the sample sheet.

seallard: Yes, I think so too - would be nice to maybe encapsulate it in the `get_sample_sheet_from_file` function

cg/meta/workflow/fluffy.py
```diff
@@ -20,6 +28,51 @@
...
+    exclude: str = "Exclude"
+    library_nM: str = "Library_nM"
+    sequencing_date: str = "SequencingDate"
+
+
+class FluffySampleSheetEntry(BaseModel):
+    flow_cell_id: str
+    lane: int
+    sample_internal_id: str
+    sample_reference: str = "hg19"
```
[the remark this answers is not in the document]
islean: image
There is a difference in logic between optional and not so double check which one we want. Optional in this case refers not to the data parsed when creating the model, but rather how we use a FluffySampleSheetEntry after it has been created.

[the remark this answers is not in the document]
Vince-janv: ooh nice, table! Me like ⭐

cg/meta/workflow/fluffy.py
diitaz93: What I see is that is needed for validating the sample sheet. What I am thinking is that maybe you need specific functions for the Fluffy sample sheet and not use the ones from the normal sample sheets

tests/cli/workflow/fluffy/test_cli_create_samplesheet.py
```diff
@@ -77,9 +78,60 @@ def test_create_samplesheet_success(
...
+    # THEN command terminates successfully
+    assert result.exit_code == EXIT_SUCCESS
+
+    # THEN newly generated SampleSheet file can be found on disk
+    assert fluffy_analysis_api.get_sample_sheet_path(fluffy_case_id_existing).exists()
+
+
+def test_create_fluffy_samplesheet_from_bcl_convert_sample_sheet(
+    cli_runner: CliRunner,
+    fluffy_case_id_existing,
```
diitaz93, line 107: type hint

tests/cli/workflow/fluffy/test_cli_create_samplesheet.py
```diff
@@ -58,6 +58,7 @@ def test_create_samplesheet_success(
     fluffy_case_id_existing,
     fluffy_context: CGConfig,
     samplesheet_path,
```
diitaz93, line 60: type hints here

cg/meta/workflow/fluffy.py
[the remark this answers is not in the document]
seallard: Nice! As far as I can see it is not even used in fluffy. But probably better to use the default value than None. Lets keep it as is.

diitaz93 (approved):
Good work 👍

cg/meta/workflow/fluffy.py
seallard: Lets do it in a separate PR

cg/meta/workflow/fluffy.py
```diff
@@ -20,6 +28,51 @@
...
+
+class FluffySampleSheet(BaseModel):
+    entries: List[FluffySampleSheetEntry]
+
+    def write_sample_sheet(self, out_path: Path) -> None:
+        LOG.info("Writing fluffy sample sheet to {sample_sheet_out_path}")
+        with out_path.open("w") as outfile:
+            outfile.write(",".join([header.value for header in FluffySampleSheetHeaders]) + "\n")
+            for entry in self.entries:
+                outfile.write(",".join(str(value) for value in entry.model_dump().values()) + "\n")
```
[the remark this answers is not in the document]
islean: Should be possible with WriteCsv if a list is created [header, entry.model_dump().values()] or something similar, I think.

cg/meta/workflow/fluffy.py
```diff
@@ -135,121 +181,32 @@ def get_sample_control_status(self, sample_id: str) -> bool:
...
-                new_column=column,
-                to_add=sample_sheet_df[sample_id_column_alias].apply(value),
+        sample_sheet: SampleSheet,
+        flow_cell_id: str,
+    ) -> FluffySampleSheet:
+        fluffy_sample_sheet_rows = []
+
+        for sample in sample_sheet.samples:
+            sample_id: str = sample.sample_id
+            db_sample: Sample = self.status_db.get_sample_by_internal_id(sample_id)
```
islean, line 193: What is the difference between sample and db_sample?

cg/apps/demultiplex/sample_sheet/read_sample_sheet.py
seallard: I'll keep the current logic as is and open a separate PR to refactor it (maybe just avoiding the duplication or moving the type determination). I don't think this PR should make any changes to the logic used in the post processing flow

cg/meta/workflow/fluffy.py
seallard: Renamed it to `FluffySample`

cg/meta/workflow/fluffy.py
seallard: Renamed 👍

cg/meta/workflow/fluffy.py
```diff
@@ -20,6 +28,51 @@
 LOG = logging.getLogger(__name__)
 
 
+class FluffySampleSheetHeaders(Enum):
```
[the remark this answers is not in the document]
seallard: I prefer keeping them as is since they are only used here
```python
headers = [header.value for header in FluffySampleSheetHeaders]
```
And all caps variables are stressful haha 😅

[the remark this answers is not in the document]
seallard: Apparently we will most likely remove this config completely https://github.com/Clinical-Genomics/cg/pull/2458#issuecomment-1715524346 so I don't think the constants are likely to be used elsewhere

cg/meta/workflow/fluffy.py
seallard: Ah, missed your comment. The fluffy "sample sheet" (more like a pipe line config) uses data from the existing sample sheet for some reason. If we had refined this issue further and also increased the scope, we probably would have ended up removing this completely

seallard:
Deployed to stage. Since the command is already broken in prod I'll just deploy this and test it live 👍  Reproducing data in stage is too much of a hassle

seallard:
Deployed to prod and tested. Generated sample sheet seems ok according to @karlnyr

PR #2479, opened 2023-09-13
Replace deprecated parse_file

seallard:
## Description
This PR replaces a deprecated function call on a Pydantic model (noticed a warning in my editor).

This syntax was deprecated in Pydantic 2
```python
model.parse_file(json_path)
```

And should be replaced with `model_validate`, see [docs](https://docs.pydantic.dev/2.0/migration/#changes-to-pydanticbasemodel). It was only used in one place in the code base.

### Fixed
- Replace deprecated Pydantic json parsing

### DIFF
--- a/cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
+++ b/cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
@@ -1,4 +1,3 @@
-import json
 import logging
 import os
 from pathlib import Path
--- a/cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
+++ b/cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
@@ -37,9 +37,8 @@ def parse_bcl2fastq_raw_tile_metrics(
 
     for stats_json_path in stats_json_paths:
         LOG.debug(f"Parsing stats.json file {stats_json_path}")
-        sequencing_metrics = Bcl2FastqSampleLaneTileMetrics.model_validate_json(
-            stats_json_path.open()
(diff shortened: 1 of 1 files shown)


henrikstranneheim (approved):
👍

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
```diff
@@ -36,7 +37,8 @@ def parse_bcl2fastq_raw_tile_metrics(
 
     for stats_json_path in stats_json_paths:
         LOG.debug(f"Parsing stats.json file {stats_json_path}")
-        sequencing_metrics = Bcl2FastqSampleLaneTileMetrics.parse_file(stats_json_path)
+        data = read_json(stats_json_path)
+        sequencing_metrics = Bcl2FastqSampleLaneTileMetrics.model_validate(data)
```
islean, line 41: I think you don't really need to open and read the json, given that there is a model_validate_json function in pydantic v2

```suggestion
        sequencing_metrics = Bcl2FastqSampleLaneTileMetrics.model_validate_json(stats_json_path.open())
```

islean (approved):
Looks nice!

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
seallard: Nice 👍

seallard: Hm, it broke some tests doing it like that. I'll keep it as is since it works

PR #2480, opened 2023-09-13
Create metrics for non pooled undetermined reads from bcl2fastq

seallard:
## Description
This PR adds logic for parsing and creating metrics for undetermined non pooled reads from a bcl2fastq demultiplexed flow cell. The new logic is not called anywhere yet (except for in tests), it will be hooked up to the post processing flow in a separate PR.

Part of https://github.com/Clinical-Genomics/cg/issues/2419. The new function which will be called outside of this module in the post processing flow is:

```python
def create_undetermined_sequencing_metrics_for_flow_cell(
    flow_cell_directory: Path,
    bcl_converter: str,
    non_pooled_lanes_and_samples: List[Tuple[int, str]],
) -> List[SampleLaneSequencingMetrics]:
    """Return sequencing metrics for undetermined reads in the specified non-pooled lanes."""

    if bcl_converter == BclConverter.BCL2FASTQ:
        return create_bcl2fastq_undetermined_metrics(
            bcl2fastq_flow_cell_path=flow_cell_directory,
            non_pooled_lane_sample_pairs=non_pooled_lanes_and_samples,
        )
```

Some refactoring (renaming model fields, docstrings etc.)

### Fixed
- Add logic for parsing the undetermined reads from a bcl2fastq run
- Add tests for new logic
- Give bcl2fastq model fields more descriptive names


cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
```diff
@@ -104,3 +109,73 @@ def get_bcl2fastq_stats_paths(demultiplex_result_directory: Path) -> List[Path]:
...
+                continue
+
+            lane: int = conversion_result.lane_number
+            flow_cell_name: str = tile_metric.flow_cell_name
+
+            if lane not in sample_lane_metrics:
+                sample_lane_metrics[lane] = create_default_sample_lane_metrics(
+                    flow_cell_name=flow_cell_name,
+                    lane=lane,
+                    sample_id="undetermined",
```
seallard, line 168: The sample id will be possible to assign for lanes containing a single sample. I think that logic should be elsewhere. Let me know if you have any ideas

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
```diff
@@ -15,37 +18,46 @@
...
+
+    return tile_metrics
+
 
-    return tile_sequencing_metrics
+def update_lane_metrics_with_demux_data(
+    sample_lane_metrics: SampleLaneMetrics, demux_result: DemuxResult
+):
+    reads: int = demux_result.number_reads
+    sample_lane_metrics.total_reads += reads * 2
```
ChrOertlin, line 47:
```suggestion
    sample_lane_metrics.total_reads += get_total_reads()
```
Add a function with a docstring explaining why the multiplication is needed for this. So that it is easily traceable for other in future.

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
```diff
@@ -104,3 +109,73 @@ def get_bcl2fastq_stats_paths(demultiplex_result_directory: Path) -> List[Path]:
         )
 
     return stats_json_paths
+
+
+def create_default_sample_lane_metrics(
```
ChrOertlin, line 114:
```suggestion
def create_sample_lane_metrics(
```
I think the default adds confusion. Is this used to just initiliase a SampleLaneMetrics for undetermined reads? If so this should be clarified I think.

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
ChrOertlin: We should assign the sample internal id as soon as possible in the flow I suppose. Where do we get it from? Is it the sample sheet and look if there is a sample on a single lane?

tests/apps/sequencing_metrics_parser/parsers/test_bcl2fastq_parser.py
ChrOertlin, line 1: We need a use case test with a sample sheet and undetermined files that replicates the actual thing I think.
I.e. have a flow cell with an undetermined files, a sample sheet with a sample only in one lane etc. and verify the correct information is added

ChrOertlin (commented):
Looks good. I think we need an integration test for the whole flow of adding undetermined samples to the database.

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
seallard: Exactly, we get them from the sample sheet 👍

tests/apps/sequencing_metrics_parser/parsers/test_bcl2fastq_parser.py
seallard: Yes, I have added a test that parses the undetermined reads. This is only part of the functionality required to close the ticket, breaking it up in smaller PR:s

ChrOertlin (approved):

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
ChrOertlin: I think we should set the sample internal id here then already, what do others think?

seallard: Yes, I'm leaning towards this as well

seallard: Done below in `get_non_pooled_undetermined_metrics_and_assign_sample_ids` instead.

cg/apps/sequencing_metrics_parser/api.py
```diff
@@ -16,8 +17,20 @@ def create_sample_lane_sequencing_metrics_for_flow_cell(
     """Parse the sequencing metrics data for the correct demultiplexing software into the sequencing statistics model."""
     if bcl_converter == BclConverter.BCLCONVERT or bcl_converter == BclConverter.DRAGEN:
         return create_sample_lane_sequencing_metrics_from_bcl_convert_metrics_for_flow_cell(
-            flow_cell_dir=flow_cell_directory
+            flow_cell_directory
+        )
+    return create_bcl2fastq_metrics(flow_cell_directory)
+
+
+def create_undetermined_sequencing_metrics_for_flow_cell(
```
seallard, line 25: This is the entry point for the new logic

cg/apps/sequencing_metrics_parser/models/bcl2fastq_metrics.py
```diff
@@ -1,56 +1,54 @@
...
-    number_reads: int = Field(..., alias="NumberReads", ge=0)
-    yield_: int = Field(..., alias="Yield", ge=0)
-    read_metrics: List[ReadMetric] = Field(..., alias="ReadMetrics")
+    tile_sample_reads: int = Field(..., alias="NumberReads", ge=0)
+    tile_sample_yield: int = Field(..., alias="Yield", ge=0)
+    tile_sample_read_metrics: List[ReadMetric] = Field(..., alias="ReadMetrics")
+
+
+class Undetermined(BaseModel):
+    """Metrics for reads without a sample id within a tile of a lane."""
```
islean, line 23: Maybe UndeterminedReads or something? I would have a hard time understanding what Undetermined represents.

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
```diff
@@ -16,93 +19,175 @@
...
-                        sample_total_yield_q30_in_lane=0,
-                        sample_total_quality_score_in_lane=0,
                     )
-                # Double the total reads since they are reported in pairs
-                metrics[sample_lane_key].sample_total_reads_in_lane += demux_result.number_reads * 2
-                metrics[sample_lane_key].sample_total_yield_in_lane += demux_result.yield_
-                metrics[sample_lane_key].sample_total_yield_q30_in_lane += sum(
-                    [read_metric.yield_q30 for read_metric in demux_result.read_metrics]
+                update_lane_metrics_with_demux_data(
+                    lane_metrics=metrics[metric_key], demux_result=demux_result
```
islean, line 115: So it does not matter if we are updating or creating from scratch? Update_lane_metrics_with_demux_data will work either way?

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
```diff
@@ -15,37 +18,46 @@
...
 
-    return tile_sequencing_metrics
+def update_lane_metrics_with_demux_data(
+    sample_lane_metrics: SampleLaneMetrics, demux_result: DemuxResult
+):
+    reads: int = demux_result.number_reads
+    sample_lane_metrics.total_reads += reads * 2
 
+    yield_: int = demux_result.yield_
+    sample_lane_metrics.total_yield += yield_
```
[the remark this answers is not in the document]
ChrOertlin: yes basically the same as for the `total_reads()`

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
```diff
@@ -16,93 +19,175 @@
 LOG = logging.getLogger(__name__)
 
 
-def parse_bcl2fastq_sequencing_metrics(flow_cell_dir: Path) -> List[Bcl2FastqSampleLaneMetrics]:
-    """Parse metrics for a flow cell demultiplexed with Bcl2fastq."""
-    tile_sequencing_metrics: List[
-        Bcl2FastqSampleLaneTileMetrics
-    ] = parse_bcl2fastq_raw_tile_metrics(flow_cell_dir)
+def scale_paired_reads_to_total_reads(paired_reads: int) -> int:
```
ChrOertlin, line 22:
```suggestion
def scale_paired_reads_to_total_reads(paired_reads: int) -> int:
     """Multiply the read count of R1 by two to represent the total amount of reads for both read pairs R1 and R2"""
```

cg/apps/sequencing_metrics_parser/models/bcl2fastq_metrics.py
seallard: Did not realise these models can be renamed as longs as they are aliased correctly 👍 Yes, something like that sounds better

seallard: Renamed to `UndeterminedTileReads`

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
seallard: Yes, it will work either way since there is always an initialised metrics object to update.

seallard: Renamed it to `update_lane_metrics_with_tile_reads`

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
seallard: Not sure if the caller should care about the details, but added this 👍

cg/apps/sequencing_metrics_parser/parsers/bcl2fastq.py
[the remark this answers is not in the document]
seallard: I think the new version is clear enough:
```python
    tile_total_yield: int = undetermined_tile_metrics.tile_total_yield
    lane_metrics.total_yield += tile_total_yield
```
Since there is no transformation of the data I don't see a reason for a separate function

seallard:
Deployed to stage

Vince-janv:
Well done! Was late with the review, but had nothing to comment on 🎖️

Issue #2278, opened 2023-07-24
Add `cg downsample` subcommand


fevac:
## Description

Add `cg downsample` subcommand that performs all steps required for downsampling an existing sample via cg.
- Start with `cg downsample sample`
- Once implemented, add `cg downsample case` that links over the samples  (note: for cases with several linked samples we might want to include the possibility to downsample different number of reads for different apptags)

Required parameters:
- sample_id: sample_id
- read-number or read-pairs-number : amount of reads to subsample

Optional:
- sample_name: name to give to subsample sample. If not given it should be created based on the original name or id + a suffix with the number of reads (e.g. *-75M)
- customer

Steps to perform:

1. Check if fastq are decompress. If not:
- Decompress samples: `cg decompress sample $sample_id`
- Store samples: `cg store sample $sample_id`
2. Get fastq file directory
3. Subsample: `sbatch --mem 100GB -n 2 -t 04:00:00 -A production -e $run_directory/down-%j.err -o $run_directory/down-%j.log /PATH2/downsample.sh $in_fastq_directory $run_directory $number_reads_to_downsample $total_reads`
4. Tracking in trailblazer status and kick after completion
5. Create the new sample adding the downsample information
- `cg add sample -d $number_reads_to_downsample -s $status -a  $apptag $customer  $down_sample_id`
- `cg set sample --skip-lims -y -kv from_sample $sample_id $down_sample_id`
- number of reads should also be updated in statusdb
6. Store in housekeeper
- `housekeeper add bundle $down_sample_id`
- `housekeeper include $down_sample_id`
- `housekeeper add file -b $down_sample_id -t fastq $down_fastq_r1`
- `housekeeper add file -b $down_sample_id -t fastq $down_fastq_r2`
7. Print new downsample ID in stdout

## Suggested solution

*Write here if there is a suggested way to solve the issue*

## This can be closed when

*Describe what needs to be done for this issue to be closed*

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

beatrizsavinhas:
I had some notes on the current downsampling process that might be useful here:

* keep case name matching original

* update reads on status DB so that fastq files can be uploaded automatically

* generate delivery reports https://github.com/Clinical-Genomics/cg/issues/1842

fevac:
Very good points!  

Regarding the first one, do you mean keeping same case name and creating new case ID? What about sample names, should they also be the same?

[...]

PR #2488, opened 2023-09-18
feat(cg downsample)

ChrOertlin:
## Description

Implementation of `cg down sample` as described in: #2278

The class `DownSampleMetaData` handles the data needed to construct the statusDB models, housekeeper entries and the down sample call.
In addition, the class performs pre-flight checks:
1. Does the original sample exist
2. Does the original case exist
3. Does the original sample have enough reads to down sample to the wanted depth
4. Does the down samples sample already exist in statusDB
5. Does the down sampled case already exist in statusDB (although it looks like this check will not work as adding a case always generates a unique case name).
6. Gets the parent directory of the lastest version of the sample bundle to extract the fastq files from
7. Gets a working directory currently set to `/home/proj/<ENV>/downsample/<down_sampled_sample>`

call to: `cg downsample samples -c case_id -i sample_id reads`
The Class `DownsampleAPI`
1. Instantiates the `DownsampleData` class for a sample
2. Checks if decompression is needed
3. Runs decompression if needed, then exits. The command needs to be re-run.
4. Commits the down sampled models to statusDB
5. Starts the down sampling job

call to: `cg downsample store <downsampled_sample_id>`
1. Creates a down sampled housekeeper bundle
2. Adds the fastq files to housekeeper with tags: `fastq`, `downsampled_sample_id`

The class `DownSampleWorkflow`
Manages the slurm script and job processes

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
-
```Shell
bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b 2278-cg-downsample
```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/workflow/down_sample/downsample.py
ChrOertlin, line 1: I will clean out this file a bit, as with the current implementation the getters and setters are not needed. All needed attributes will be provided and cannot be set manually

cg/cli/downsample.py
```diff
@@ -0,0 +1,81 @@
...
+
+@click.group()
+def downsample_cmd():
+    """Downsample reads in a sample."""
+
+
+@downsample_cmd.command("samples", help="Downsample reads in one or multiple samples in a case.")
+@click.option(
+    "-c",
+    "--case-internal-id",
```
fevac, line 24: do we want to rename this?

cg/cli/downsample.py
```diff
@@ -0,0 +1,81 @@
...
+    required=True,
+    help="Case identifier used in statusdb, e.g. supersonicturtle",
```
fevac, line 26:
```suggestion
    help="Case identifier used in statusdb from which case information will be trasnferred, e.g. supersonicturtle",
```

cg/models/downsample/downsample_meta_data.py
```diff
@@ -0,0 +1,167 @@
...
+        return Path(
+            self.config.housekeeper_api.get_file_from_latest_version(
+                bundle_name=self.original_sample.internal_id, tags={SequencingFileTag.FASTQ}
+            ).path
+        ).parent
+
+    @property
+    def fastq_file_output_directory(self):
+        """Get the output directory for the down sampled sample."""
+        return Path("home", "proj", "production", "downsample", self.downsampled_sample.internal_id)
```
fevac, line 143: I don't think this path should be hard coded. I think this should be part of cg.yalm in servers

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,166 @@
...
+        """Down sample a sample."""
+        if self.is_decompression_needed(self.downsample_meta_data.downsampled_case):
+            self.start_decompression(self.downsample_meta_data.downsampled_sample)
+            return
+        self.add_downsampled_sample_case_to_statusdb()
+        self.add_downsampled_sample_to_housekeeper()
+        self.start_downsample_job(
+            original_sample=self.downsample_meta_data.original_sample,
+            sample_to_downsample=self.downsample_meta_data.downsampled_sample,
+        )
```
fevac, line 166: I guess we want to include this in trailblazer too

fevac (commented):
Great job so far! I started having a look and have few comments. I will have another look to check details when your PR is ready for review

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,166 @@
...
+        self.status_db.session.add_commit(sample_case_link)
+        LOG.info(f"Related sample {sample.internal_id} to {case.internal_id}")
+
+    def add_downsampled_sample_case_to_statusdb(self) -> None:
+        """
+        Add the down sampled sample and case to statusDB and generate the sample case link.
+            -Asserts that both the original sample and case exists in statusDB.
+            -Asserts that the down sampled sample and case does not already exist in statusDB.
+            -Asserts that the original sample has more reads than the number to which it is down sampled.
```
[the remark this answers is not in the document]
ChrOertlin: No indeed, this has been moved to the DownSampleMetaData. This function only commits to statusdb now. I will change the docstring to reflect this

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,166 @@
...
+                tags=[SequencingFileTag.FASTQ],
+            )
+
+    def downsample_sample(self) -> None:
+        """Down sample a sample."""
+        if self.is_decompression_needed(self.downsample_meta_data.downsampled_case):
+            self.start_decompression(self.downsample_meta_data.downsampled_sample)
+            return
+        self.add_downsampled_sample_case_to_statusdb()
+        self.add_downsampled_sample_to_housekeeper()
```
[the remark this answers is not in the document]
ChrOertlin: Yes, indeed, this has to be moved. I need to have a look at how I can track slurm jobs in trailblazer and let them run sequentially.

cg/cli/downsample.py
ChrOertlin: At some point we tried had an effort to be more explicit in what "ids" one can refer to. 
case_id is ambiguous as it can refer to the `database_row` which is `case.id`. 
`case.internal_id` is the pet name. 

The same goes for `sample_id` and `sample_internal_id`

cg/cli/downsample.py
```diff
@@ -0,0 +1,81 @@
...
+@click.option(
+    "-sr",
+    "--sample_internal_id_reads",
+    required=True,
+    multiple=True,
+    help="Identifier used in statusdb, e.g. ACC1234567 and the number of reads to down sample to in millions separated by ;."
+    "e.g. ACC1234567;30",
+)
+@DRY_RUN
+def downsample_sample(
```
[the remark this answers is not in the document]
ChrOertlin: what does the `case-name` refer to here? Is it the given pet name? Or the name provided by the users?
`case.internal_id` or `case.name`

karlnyr (commented):
looks like a nice draft, some work to be done but a good start 👍

cg/cli/downsample.py
```diff
@@ -0,0 +1,81 @@
...
+    help="Case identifier used in statusdb, e.g. supersonicturtle",
+)
+@click.option(
+    "-sr",
+    "--sample_internal_id_reads",
+    required=True,
+    multiple=True,
+    help="Identifier used in statusdb, e.g. ACC1234567 and the number of reads to down sample to in millions separated by ;."
+    "e.g. ACC1234567;30",
+)
```
Vince-janv, line 35: I think this should be two options

cg/cli/downsample.py
```diff
@@ -0,0 +1,81 @@
...
+@DRY_RUN
+def downsample_sample(
+    context: CGConfig, case_internal_id: str, sample_internal_id_reads: List[str], dry_run: bool
```
Vince-janv, line 38: why is there a case_id in the downsample_sample command?

cg/cli/downsample.py
```diff
@@ -0,0 +1,81 @@
...
+    help="Case identifier used in statusdb, e.g. subsonicrabbit",
+)
+@click.option(
+    "-r",
+    "--number_of_reads",
+    required=True,
+    help="Number of reads to down sample to in millions, e.g. 30",
+)
+@DRY_RUN
+def downsample_case(context: CGConfig, case_internal_id: str, number_of_reads: int, dry_run: bool):
```
Vince-janv, line 67: Is this functionality requested by the users?

cg/meta/workflow/downsample/downsample.py
```diff
@@ -0,0 +1,163 @@
...
+        self.input_fastq_dir: str = input_fastq_dir
+        self.original_sample: Sample = original_sample
+        self.output_fastq_dir: str = output_fastq_dir
+        self.slurm_api: SlurmAPI = SlurmAPI()
+        self._account: str
+        self._memory: str
+        self._number_tasks: str
+        self._time: str
+        self._quality_of_service: str
+        self._email: str
```
Vince-janv, line 41: can these not be grouped together somehow?

cg/models/downsample/downsample_meta_data.py
```diff
@@ -0,0 +1,167 @@
...
+        self.sample_internal_id: str = sample_internal_id
+        self.number_of_reads: int = number_of_reads
+        self.case_internal_id: str = case_internal_id
+        self.original_sample: Sample = self.get_sample_to_downsample()
+        self.original_case: Family = self.get_case_to_downsample()
+        self.has_enough_reads: bool = self.has_enough_reads_to_downsample()
+        self.downsampled_sample: Sample = self._generate_statusdb_downsampled_sample_record()
+        self.downsampled_case: Family = self._generate_statusdb_downsampled_case()
+        self.create_down_sampling_working_directory()
+        LOG.info(f"Pre-flight checks completed for {self.sample_internal_id}")
```
Vince-janv, line 27: I don't think all this should be done in the instantiation of the model. What is the reasoning for putting it there?

cg/models/downsample/downsample_meta_data.py
```diff
@@ -0,0 +1,167 @@
...
+
+from cg.constants import Priority, SequencingFileTag
+from cg.models.cg_config import CGConfig
+from cg.store.models import ApplicationVersion, Family, Sample
+
+LOG = logging.getLogger(__name__)
+
+
+class DownsampleMetaData:
+    def __init__(
```
Vince-janv, line 13: I like the idea of checking for requirements before submitting the job!

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,166 @@
...
+    def __init__(
+        self, config: CGConfig, sample_reads: str, case_internal_id: str, dry_run: bool = False
+    ):
+        """Initialize the API."""
+        super().__init__(config)
+        self.config = config
+        self.sample_reads: str = sample_reads
+        self.case_internal_id: str = case_internal_id
+        self.dry_run: bool = dry_run
+        self.downsample_meta_data: DownsampleMetaData = self.get_meta_data()
```
Vince-janv, line 27: Same here regarding performing the actual function of the class in the instantiation

Vince-janv (commented):
Nice work! I think it could be beneficial to very clearly divide the structure into, first, prep in the database, and then secondly submit the actual downsampling command.

cg/cli/downsample.py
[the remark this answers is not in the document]
ChrOertlin: The number gets multiplied by a million so it would be 0.1 * 1_000_000

cg/cli/downsample.py
ChrOertlin: Yes it is outlined in the issue. Give a case and a number of reads and downsample eveyrthing in that case to the number of reads

cg/cli/downsample.py
ChrOertlin: A sample cannot exist without a case for an analysis for which the down sampled sample will be used. Outlined in the issue

cg/models/downsample/downsample_meta_data.py
ChrOertlin: To me it makes the code more compact. Since the class organises itself upon initialising it. It would avoid needed to add extra code to call and update. Why do you think it should not be done like this?

cg/models/downsample/downsample_meta_data.py
ChrOertlin: noted with a comment for now!

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+class DownSampleAPI(MetaAPI):
+    def __init__(
+        self, config: CGConfig, sample_reads: str, case_internal_id: str, dry_run: bool = False
+    ):
+        """Initialize the API."""
+        super().__init__(config)
+        self.config = config
+        self.sample_reads: str = sample_reads
```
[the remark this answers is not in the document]
ChrOertlin: fixed the input now so this is deprecated

cg/apps/downsample/downsample.py
ChrOertlin: As discussed in the office, if this class cannot be initialised the api should exit.

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,166 @@
...
+        self.case_internal_id: str = case_internal_id
+        self.dry_run: bool = dry_run
+        self.downsample_meta_data: DownsampleMetaData = self.get_meta_data()
+
+    @staticmethod
+    def parse_sample_reads_input(
+        sample_reads: str, case_internal_id: str, config: CGConfig
+    ) -> DownsampleMetaData:
+        """Parse the sample reads input into the SampleToDownSample model."""
```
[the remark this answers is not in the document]
ChrOertlin: deprecated

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,166 @@
...
+        number_of_reads: int
+        sample_internal_id, number_of_reads = sample_reads.split(";")
+        return DownsampleMetaData(
+            config=config,
+            sample_internal_id=sample_internal_id,
+            number_of_reads=number_of_reads,
+            case_internal_id=case_internal_id,
+        )
+
+    def get_meta_data(self) -> DownsampleMetaData:
```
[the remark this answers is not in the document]
ChrOertlin: fixed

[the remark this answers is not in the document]
ChrOertlin: refactored the code now. This is needed to initialise the data class

ChrOertlin:
# cases and sample are created in statusdb
image

image

image

ChrOertlin:
# downsampling submitted and finished

alembic/versions/2023_10_30_b6f00cc615cf_sync_db.py
ChrOertlin, line 1: isort stuff

ChrOertlin:
# Successfully added fastq files to housekeeper

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class DownsampleAPI(MetaAPI):
+    def __init__(
+        self,
+        config: CGConfig,
+        sample_id: str,
+        number_of_reads: float,
+        case_id: str,
```
seallard, line 26: I have a suggestion regarding the design pattern used:

Currently, the API is instantiated with specific details for each sample/case (e.g., sample_id, number_of_reads, case_id). This approach ties each instance of the API to a particular sample/case.

To enhance the API's usability and maintainability, I recommend decoupling the instance creation from specific samples/cases. Instead, we could:

1. Initialize DownsampleAPI with only configuration and shared resources, not tied to any sample or case.
2. Modify methods like downsample_sample to accept sample_id, number_of_reads, case_id, etc., as parameters. This way, one API instance can flexibly handle various samples and cases.
3. Ensure the API methods remain stateless regarding sample/case details, relying solely on the parameters provided to them.

This approach would make the API more general and versatile, simplifying the handling of different downsample tasks with a single, reusable API instance. Concretely, it would allow us to follow the pattern of only passing API:s via the context and also minimising the amount of logic in the CLI layer.

cg/meta/workflow/downsample/downsample.py
```diff
@@ -0,0 +1,121 @@
...
+        return self._account
+
+    @property
+    def email(self) -> str:
+        return self._email
+
+    @property
+    def bin_downsample(self) -> str:
+        """The path to the downsample.sh script on Hasta."""
+        return Path("home", "proj", "production", "bin", "downsample.sh").name
```
seallard, line 53: I think this will only return the name of the script, probably want to extract it to a config as well

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+        sample_case_link: FamilySample = self.status_db.relate_sample(
+            family=case,
+            sample=sample,
+            status=self.downsample_data.sample_status(sample=sample),
+        )
+        self.status_db.session.add(sample_case_link)
+        self.status_db.session.commit()
+        LOG.info(f"Related sample {sample.internal_id} to {case.internal_id}")
+
+    def add_downsampled_sample_case_to_statusdb(self) -> None:
```
seallard, line 109: Some naming ideas, I think `add_downsampled_case_to_statusdb` rolls better off the tongue or `store_downsampled_case`. Does not really matter, but if you like it better 👍

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+            status_db=self.status_db, sample_id=downsampled_sample.internal_id
+        ):
+            raise ValueError(f"Sample {downsampled_sample.internal_id} already exists in StatusDB.")
+        LOG.info(
+            f"New downsampled sample created: {downsampled_sample.internal_id} from {downsampled_sample.from_sample}"
+            f"Application tag set to: {downsampled_sample.application_version.application.tag}"
+            f"Customer set to: {downsampled_sample.customer}"
+        )
+        if not self.dry_run:
+            self.status_db.session.add(downsampled_sample)
```
seallard, line 71: Consider wrapping this in a try/except and rollback the transaction if an error occurs. I think we need a general pattern for this though and it should probably not be handled throughout the codebase. So disregard this 😅

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+            self.status_db.session.commit()
```
seallard, line 72: Consider wrapping this in a try/except and rollback the transaction if an error occurs. I think we need a general pattern for this though and it should probably not be handled throughout the codebase. So disregard this 😅  Just a good thing to be aware of

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+            LOG.info(f"Added {downsampled_sample.name} to StatusDB.")
+            return downsampled_sample
+        return downsampled_sample
+
+    def add_downsampled_case_to_statusdb(self) -> Family | None:
+        """
+        Add a down sampled case entry to StatusDB.
+        """
+        downsampled_case: Family = self.downsample_data.downsampled_case
+        if case_exists_in_statusdb(status_db=self.status_db, case_name=downsampled_case.name):
```
seallard, line 82: This method belongs on the store class, right?

seallard (commented):
I'll continue on Monday! Looks good 🚀

islean (commented):
Great job!

cg/models/downsample/downsample_data.py
```diff
@@ -0,0 +1,161 @@
...
+        hk_api: HousekeeperAPI,
+        sample_id: str,
+        number_of_reads: float,
+        case_id: str,
+        out_dir: Path,
+    ):
+        """Initialize the downsample data and perform integrity checks.
+        Raises:
+            ValueError
+            FileExistsError
```
diitaz93, line 28: I think we should write error raised only in function where the error is originated and not in the functions where the error is propagated, but I am not sure. I am just thinking that if we propagate the error in the docstrings we will end up with functions raising many errors.

cg/models/downsample/downsample_data.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def get_sample_to_downsample(self) -> Sample:
+        """Check if a sample exists in StatusDB."""
+        sample: Sample = self.status_db.get_sample_by_internal_id(self.sample_id)
+        if not sample:
+            raise ValueError(f"Sample {self.sample_id} not found in StatusDB.")
+        return sample
+
+    def get_case_to_downsample(self) -> Family:
+        """Check if a case exists in StatusDB."""
```
diitaz93, line 67: Here there should be the error warning in the docstring

diitaz93 (commented):
I have so much left to review, will finish later

cg/models/downsample/downsample_data.py
```diff
@@ -0,0 +1,161 @@
...
+    @property
+    def downsampled_case_name(
+        self,
+    ) -> str:
+        """Return a case name with _downsampled appended."""
+        new_name: str = f"{self.case_id}_downsampled"
+        return new_name
+
+    def get_sample_to_downsample(self) -> Sample:
+        """Check if a sample exists in StatusDB."""
```
diitaz93, line 60: Add error raising in the docstring

cg/models/downsample/downsample_data.py
```diff
@@ -0,0 +1,161 @@
...
+            raise ValueError(f"Case {self.case_id} not found in StatusDB.")
+        return case
+
+    def _generate_statusdb_downsampled_sample_record(
+        self,
+    ) -> Sample:
+        """
+        Generate a downsampled sample record for StatusDB.
+        The new sample contains the original sample internal id and meta data.
+        Raises:
```
diitaz93, line 79: Again, I think this error raising should not be here but in the function where the error is originated, but this is something that we could discuss later

cg/models/downsample/downsample_data.py
```diff
@@ -0,0 +1,161 @@
...
+            name=self.downsampled_case_name,
+            panels=self.original_case.panels,
+            priority=self.original_case.priority,
+            ticket=self.original_case.latest_ticket,
+        )
+        downsampled_case.customer = self.original_case.customer
+        return downsampled_case
+
+    def has_enough_reads_to_downsample(self) -> bool:
+        """Check if the sample has enough reads to downsample."""
```
diitaz93, line 116: add error raising

cg/models/downsample/downsample_data.py
```diff
@@ -0,0 +1,161 @@
...
+            data_delivery=self.original_case.data_delivery,
+            name=self.downsampled_case_name,
+            panels=self.original_case.panels,
+            priority=self.original_case.priority,
+            ticket=self.original_case.latest_ticket,
+        )
+        downsampled_case.customer = self.original_case.customer
+        return downsampled_case
+
+    def has_enough_reads_to_downsample(self) -> bool:
```
diitaz93, line 115: If it raises an error instead of returning false when it does not have enough reads, I suggest renaming this function to validate_has_enough_reads, and making it return None

cg/models/downsample/downsample_data.py
```diff
@@ -0,0 +1,161 @@
...
+            bundle_name=self.original_sample.internal_id,
+        ).full_path
+
+    @property
+    def fastq_file_output_directory(self):
+        """Get the output directory for the downsampled sample."""
+        return Path(self.out_dir, self.downsampled_sample.internal_id)
+
+    @staticmethod
+    def sample_status(sample: Sample) -> str:
```
diitaz93, line 140:
```suggestion
def get_sample_status(sample: Sample) -> str:
```

cg/models/downsample/downsample_data.py
```diff
@@ -0,0 +1,161 @@
...
+        return self.housekeeper_api.get_latest_bundle_version(
+            bundle_name=self.original_sample.internal_id,
+        ).full_path
+
+    @property
+    def fastq_file_output_directory(self):
+        """Get the output directory for the downsampled sample."""
+        return Path(self.out_dir, self.downsampled_sample.internal_id)
+
+    @staticmethod
```
diitaz93, line 139: Remembering what we have discussed about favouring utils modules to static methods, It seems that these static functions are more suitable for a statusdb utils module, but feel free to disregard

cg/models/downsample/downsample_data.py
```diff
@@ -0,0 +1,161 @@
...
+        """Return the application for a sample."""
+        return sample.application_version.application.tag
+
+    def get_application_version(self, sample: Sample) -> ApplicationVersion:
+        """Return the application version for a sample."""
+        application_tag: str = self.get_application_tag(sample)
+        return self.status_db.get_current_application_version_by_tag(application_tag)
+
+    def create_down_sampling_working_directory(self) -> Path:
+        """Create a working directory for the downsample job."""
```
diitaz93, line 155: add error raising to docstring

cg/apps/downsample/downsample.py
seallard: Yeah, I think it is a bit more common - API:s usually provide a general set of methods.
Sure! I think the pattern I described is what generally applied throughout the cg already. We could document it in a design decision

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+        sample_id: str,
+        number_of_reads: float,
+        case_id: str,
+        dry_run: bool = False,
+    ):
+        """Initialize the API."""
+        super().__init__(config)
+        self.config: CGConfig = config
+        self.status_db: Store = config.status_db
+        self.hk_api: HousekeeperAPI = config.housekeeper_api
```
seallard, line 33: These can be removed since they are defined in the base class `MetaAPI` already

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+            LOG.info(f"Case with name {downsampled_case.name} already exists in StatusDB.")
+            return
+        if not self.dry_run:
+            self.status_db.session.add(downsampled_case)
+            self.status_db.session.commit()
+            LOG.info(f"New down sampled case created: {downsampled_case.internal_id}")
+            return downsampled_case
+        return downsampled_case
+
+    def _link_downsampled_sample_to_case(self, sample: Sample, case: Family) -> None:
```
seallard, line 92: This method likely belongs on the store class? We should settle on a clearer structure in which layer what kind of logic belongs. Feel free to keep it here though

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+        return downsample_work_flow.write_and_submit_sbatch_script()
+
+    def downsample_sample(self) -> int | None:
+        """Down sample a sample."""
+        if self.is_decompression_needed():
+            self.start_decompression(self.downsample_data.original_sample)
+            return
+        LOG.debug("No Decompression needed.")
+        self.add_downsampled_sample_case_to_statusdb()
```
seallard, line 162: I'm not a big fan of having database names in methods, but does not really matter (which database it writes to could be considered an implementation detail that is irrelevant to the caller). Feel free to keep it though, we do it everywhere

cg/apps/downsample/utils.py
```diff
@@ -0,0 +1,67 @@
...
+from cg.apps.housekeeper.hk import HousekeeperAPI
+from cg.constants import SequencingFileTag
+from cg.store import Store
+from cg.store.models import Family, Sample
+from cg.utils.files import get_files_matching_pattern
+
+LOG = logging.getLogger(__name__)
+
+
+def case_exists_in_statusdb(status_db: Store, case_name: str) -> bool:
```
seallard, line 14: Probably belongs on the store class

cg/apps/downsample/utils.py
```diff
@@ -0,0 +1,67 @@
...
+    """Check if a case exists in StatusDB."""
+    case: Family = status_db.get_case_by_name(case_name)
+    if case:
+        return True
+    return False
+
+
+def sample_exists_in_statusdb(status_db: Store, sample_id: str) -> bool:
```
seallard, line 22: Probably belongs on the store class

cg/models/downsample/downsample_data.py
```diff
@@ -0,0 +1,161 @@
...
+from cg.utils.calculations import multiply_by_million
+
+LOG = logging.getLogger(__name__)
+
+
+class DownsampleData:
+    def __init__(
+        self,
+        status_db: Store,
+        hk_api: HousekeeperAPI,
```
seallard, line 19: It is a little counter intuitive to me that a class called `DownsampleData` require these parameters, it sounds like a pure data class. It works though, feel free to disregard :+1: Or maybe you think there is a better name for it

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+        Decompression is needed if there are no files with fastq tag found for the sample in housekeeper.
+        """
+        LOG.debug("Checking if decompression is needed.")
+        is_decompression_needed: bool = False
+        if not self.housekeeper_api.get_files(
+            bundle=self.downsample_data.original_sample.internal_id,
+            tags=[SequencingFileTag.FASTQ],
+        ):
+            is_decompression_needed = True
+        return is_decompression_needed
```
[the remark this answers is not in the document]
ChrOertlin: That is based on case, but I need it for a sample only.

tests/models/downsample/test_down_sample_meta_data.py
```diff
@@ -0,0 +1,49 @@
...
+
+def test_downsample_meta_data_pass_pre_flight(
+    store_with_case_and_sample_with_reads: Store,
+    downsample_hk_api: HousekeeperAPI,
+    downsample_case_internal_id: str,
+    downsample_sample_internal_id_1: str,
+    number_of_reads_in_millions: int,
+    tmp_path_factory,
+):
+    """Test that the pre-flight checks pass when initialising a DownsampleMetaData class."""
```
seallard, line 17: Not clear to me what pre-flight checks mean in this context

seallard (approved):
Wow, great work! :100: 
Tests look good, solid effort! Some methods might belong elsewhere and you could shuffle the overall structure a bit if you think it would be an improvement.

cg/apps/downsample/downsample.py
seallard: I think this is out of scope, better to stick to the existing non-existent transaction handling

cg/apps/downsample/downsample.py
ChrOertlin: I like the approach you describe. I will give it a go!

cg/apps/downsample/downsample.py
ChrOertlin: That is true, will fix it

cg/apps/downsample/downsample.py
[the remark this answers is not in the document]
islean: Oh right! Then the docstring should probably be updated. But there is still a point in getting the compression objects for the sample and returning in a similar manner as in that method, since I think the suggested approach might need to support ongoing decompressions. In an ideal world, maybe the is_spring_decompression_needed method should be refactored so that we can stay DRY, but that is probably out of scope and can be done at a later time. 

Return statement for the aforementioned decompression method:
```Python
return any(
            not self.crunchy_api.is_compression_pending(compression_object)
            and not compression_object.pair_exists()
            for compression_object in compression_objects
        )
```

cg/apps/downsample/utils.py
```diff
@@ -0,0 +1,67 @@
...
+        sample_id=sample_id,
+        housekeeper_api=housekeeper_api,
+        fastq_file_output_directory=fastq_file_output_directory,
+    )
+
+
+def add_downsampled_fastq_files_to_housekeeper(
+    fastq_file_output_directory: str, sample_id: str, housekeeper_api: HousekeeperAPI
+) -> None:
+    """Add down sampled fastq files to housekeeper."""
```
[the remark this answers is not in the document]
islean: Yes, although I think the naming of the method might be more important to consider.

cg/models/downsample/downsample_data.py
islean: Good point for discussion! I lean towards wanting some indication that errors can be propagated by this method, since for a user of this method, as little of the implementation as possible should be needed. But I don't know what the general consensus is.

cg/constants/demultiplexing.py
```diff
@@ -1,10 +1,10 @@
 """Constants related to demultiplexing."""
+from enum import Enum, StrEnum
```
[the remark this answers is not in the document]
ChrOertlin: I have it on 5.12 and ran it this came out

cg/apps/downsample/downsample.py
ChrOertlin: The statusdb functiion is `relate_sample` which is used in here. This function extends that a bit by performing the database commit.
Any ideas on how to do it better? I was thinking of adding the commit to the existing function but that might break other stuff so I did it like it is now.

cg/cli/downsample.py
```diff
@@ -0,0 +1,78 @@
...
+def downsample():
+    """Downsample reads in a sample."""
+
+
+@downsample.command("samples", help="Downsample reads in one or multiple samples in a case.")
+@click.option(
+    "-c",
+    "--case-id",
+    required=True,
+    help="Case identifier used in statusdb, e.g. supersonicturtle. The case information wil be transferred.",
```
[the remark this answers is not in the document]
ChrOertlin: No. :)

tests/apps/downsample/test_downsample.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # THEN a downsampled sample is added to the store
+    assert downsample_api.status_db.get_sample_by_internal_id(
+        downsample_api.downsample_data.downsampled_sample.internal_id
+    )
+
+    # WHEN adding the sample again
+
+    # THEN a ValueError is raised
+    with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
ChrOertlin: A case can hold multiple samples, so a case existing already might not need to throw an error. While samples have to be unique

cg/apps/downsample/downsample.py
ChrOertlin: @seallard  have a look, I tried to implement a sample agnostic approach to the API. What do you think? Overall I can see the benefits of maintainability and reusability.

cg/apps/downsample/downsample.py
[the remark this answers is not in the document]
ChrOertlin: Yes this was the initial method I used, but here I ran into errors when t would want to decompress the whole case instead of just the sample.

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+
+    def start_decompression(self, sample: Sample) -> None:
+        """Start decompression of spring compressed fastq files for the given sample."""
+        LOG.info(f"Decompression for {sample.internal_id} is needed.")
+        self.prepare_fastq_api.compress_api.decompress_spring(sample.internal_id)
+        LOG.info("Re-run down sample command when decompression is done.")
```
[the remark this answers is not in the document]
ChrOertlin: No, we don't Will refactor

tests/apps/downsample/test_downsample.py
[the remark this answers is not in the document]
islean: Ah, makes sense!

cg/apps/downsample/downsample.py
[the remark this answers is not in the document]
ChrOertlin: added more extensive checks

tests/models/downsample/test_down_sample_meta_data.py
ChrOertlin: renamed

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,169 @@
...
+            f"Customer set to: {downsampled_sample.customer}"
+        )
+        if not self.dry_run:
+            self.status_db.session.add(downsampled_sample)
+            self.status_db.session.commit()
+            LOG.info(f"Added {downsampled_sample.name} to StatusDB.")
+            return downsampled_sample
+        return downsampled_sample
+
+    def add_downsampled_case_to_statusdb(self) -> Family | None:
```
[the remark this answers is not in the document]
ChrOertlin: No, it just needs to commit.

ChrOertlin:
Re-tested functionality after addressing comments - tests passed.

cg/apps/downsample/utils.py
```diff
@@ -0,0 +1,57 @@
...
+        FileNotFoundError
+    """
+    fastq_file_paths: list[Path] = get_files_matching_pattern(
+        directory=Path(fastq_file_output_directory),
+        pattern=f"*{sample_id.split(sep='_')[0]}*.{SequencingFileTag.FASTQ}.gz",
+    )
+    if not fastq_file_paths:
+        raise FileNotFoundError(
+            f"No fastq files found with pattern: *{sample_id}*.{SequencingFileTag.FASTQ}.gz"
+        )
```
karlnyr, line 45: Since we dont have any tracking of the downsampling - what hinders the user from running this command mid downsampling? Can we not create a very primitive way of finding if the downsampling is done or ongoing?

cg/models/downsample/downsample_data.py
karlnyr: I guess one could have one being the collector of data "CollectDownsampleData" and then the other being as Sebastian suggested a data class "DownsampleData"

cg/cli/downsample.py
```diff
@@ -0,0 +1,74 @@
...
```
karlnyr, line 27: So - the case id given here does not need to be connected to the samples. My thoughts then are - is it correct to say that the case is a downsampled version of the old one if it is not the same? I think that we should refer to adding the `<case_id>_downsampled` name to the case, and instead just give it a new pet name. The samples themselves that are connected to the case keep that information.

cg/cli/downsample.py
```diff
@@ -0,0 +1,74 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+@click.group()
+def downsample():
+    """Downsample reads in a sample."""
+
+
+@downsample.command("samples", help="Downsample reads in one or multiple samples in a case.")
```
karlnyr, line 22: In addition to the comment below - I think that you can give an elaborate full use case with the help of this command - one being for one sample to downsample, and also for multiple ones to downsample. Alternatively, just write that you can only supply one sample and give an example of a multiple sample case instead

cg/cli/downsample.py
```diff
@@ -0,0 +1,74 @@
...
+    help="Case identifier used in statusdb, e.g. supersonicturtle. The case information wil be transferred.",
+)
+@click.option(
+    "-i",
+    "--input-data",
+    required=True,
+    nargs=2,
+    multiple=True,
+    help="Identifier used in statusdb, e.g. ACC1234567 and the number of reads to down sample to in millions separated by a space"
+    " e.g. ACC1234567 30. Multiple inputs can be provided.",
```
karlnyr, line 36: I do not think it is clear here that the user can supply `0.001` as a possible value for the number of reads.

karlnyr (commented):
This looks very nice! I have some questions about some logic, and also think that we can be a bit more verbose in this command for our new bioinformaticians that may use it in the future.

cg/cli/downsample.py
ChrOertlin: I added a use case that also indicates that one can supply values below 1

cg/apps/downsample/downsample.py
```diff
@@ -0,0 +1,172 @@
...
+            for compression_object in compression_objects
+        )
+
+    def get_sample_compression_objects(self, sample_id: str) -> list[CompressionData]:
+        compression_objects: list[CompressionData] = []
+        version_obj: Version = self.prepare_fastq_api.compress_api.hk_api.get_latest_bundle_version(
+            bundle_name=sample_id
+        )
+        compression_objects.extend(files.get_spring_paths(version_obj))
+        return compression_objects
```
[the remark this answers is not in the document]
ChrOertlin: Yes these ought to be moved away from here. Will fix it.

cg/apps/downsample/utils.py
ChrOertlin: Nothing hinders them. We would need to fetch the sbatch job and see if its pending. I could not see functionality for this in the SlurmAPI. 

I added a note to the store command that one should check that the downsample command is finished before running it for now.

ChrOertlin:
Stage: 

Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/christian.oertlin
done.
Log deploy... done.
cg, version 52.0.0

Prod:

Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/christian.oertlin
done.
Log deploy... done.
cg, version 52.0.0

henrikstranneheim:
@ChrOertlin Great work 🎉 ! Do we need to add an automated clean of the downsampled dir?

PR #2491, opened 2023-09-19
Patch sample sheet sample id property

seallard:
## Description
The sample id attribute on the `FlowCellSample` is not always a sample id, but might include a trailing sequence which should be discarded.

### Added
- Method on sample sheet model for retrieving the sample id

### Fixed
- Fluffy sample sheet generation uses the formatted and validated sample id

### DIFF
--- a/cg/apps/demultiplex/sample_sheet/models.py
+++ b/cg/apps/demultiplex/sample_sheet/models.py
@@ -1,6 +1,6 @@
 import logging
 from collections import defaultdict
-from typing import List
+from typing import List, Optional
 
 from pydantic import BaseModel, ConfigDict, Extra, Field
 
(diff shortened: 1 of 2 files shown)


diitaz93 (approved):
Looks good, I think it would still need a patch in the path list for the current bug to be fixed but you can merge this

seallard:
> Looks good, I think it would still need a patch in the path list for the current bug to be fixed but you can merge this

Maybe it would make more sense to format the sample ids when the sample sheet is created instead? We will probably run into other bugs by accidentally using the `sample_id` property

seallard:
Replaced by https://github.com/Clinical-Genomics/cg/pull/2528

PR #2496, opened 2023-09-20
Remove alchy from requirements

seallard:
## Description
There are no references to Alchy in cg, so we can remove it as a dependency.

### Fixed
- Remove alchy as a dependency


henrikstranneheim (approved):
👍

seallard:
Deployed and tested in stage.

Issue #2497, opened 2023-09-20
Migrate SQLAlchemy from 1.3 to 2.0


seallard:
## Description
A new major version has been released for the ORM we use. Upgrading it in cg also means that we have to upgrade it in housekeeper since the packages are tightly coupled. Part of project https://github.com/Clinical-Genomics/project-planning/issues/474

- See new [features](https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html)
- See [migration guide](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html)

## Suggested solution
The recommended approach is an iterative migration approach, transitioning first to 1.4 which has most of the new features but also supports the old. These are then removed in version 2.0.

### Migration plan
The steps in the migration are described [here](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#x-2-x-migration-overview).

- [x] [First Prerequisite, step one - A Working 1.3 Application](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#first-prerequisite-step-one-a-working-1-3-application)
- [x] [First Prerequisite, step two - A Working 1.4 Application](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#first-prerequisite-step-two-a-working-1-4-application)
- [x] [Migration to 2.0 Step One - Python 3 only (Python 3.7 minimum for 2.0 compatibility)](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-to-2-0-step-one-python-3-only-python-3-7-minimum-for-2-0-compatibility)
- [x] [Migration to 2.0 Step Two - Turn on RemovedIn20Warnings](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-to-2-0-step-two-turn-on-removedin20warnings)
- [x] [Migration to 2.0 Step Three - Resolve all RemovedIn20Warnings](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-to-2-0-step-three-resolve-all-removedin20warnings)
- [x] [Migration to 2.0 Step Four - Use the future flag on Engine](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-to-2-0-step-four-use-the-future-flag-on-engine)
- [x] [Migration to 2.0 Step Five - Use the future flag on Session](https://docs.sqlalchemy.org/en/20/changelog/migration_20.html#migration-to-2-0-step-five-use-the-future-flag-on-session)
[...]

PR #2499, opened 2023-09-20
Upgrade sqlalchemy to 1.4

seallard:
## Description
This PR does a minimal migration of cg from sqlalchemy 1.3 to 1.4 ([docs](https://docs.sqlalchemy.org/en/20/changelog/migration_14.html)), the transition version which contains most of the new functionality in 2.0 but still supports the old api. Part of project https://github.com/Clinical-Genomics/cg/issues/2497.

I think the biggest risk in the changes is the [new auto begin behaviour](https://docs.sqlalchemy.org/en/20/changelog/migration_14.html#session-features-new-autobegin-behavior). 

Before, a transaction would be automatically initiated when a session was created and a new one would be created after commit and rollback was called. Now, it is only initiated when add is called. Reviewing the code base, it seems like we always initiate a transaction with a `session.add(...)` before either doing a rollback or commit. If we did not, the rollback and commit methods would be no-ops, meaning that they do nothing. 

### Fixed
- Minimal migration of sqlalchemy to the latest 1.4.x version

### Testing
- Verify that some commonly used flows work and write data as expected to status db and housekeeper.

The post processing of a flow cell interacts a lot with the db:s, should be a good testing point.


seallard:
Tests start failing when setting the sqlalchemy version to 1.4. Most likely due to a change in how sessions behave https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#session-features-new-autobegin-behavior

seallard:
The `test_get_sample_coverage` and `test_get_sample_metadata` are failing due to changes in the auto flush behaviour in 1.4.

The tests set the internal id of the created sample to another id which already exists in the database. With 1.4, this change is auto flushed when another database object is accessed - causing the unique constraint on the sample id to be violated and the test to fail.

tests/meta/report/test_mip_dna_api.py
```diff
@@ -1,14 +1,14 @@
...
 
 def test_get_sample_metadata(
-    report_api_mip_dna, mip_analysis_api, case_mip_dna, sample_store, helpers
+    report_api_mip_dna, mip_analysis_api, case_mip_dna, sample_store, helpers: StoreHelpers
 ):
     """Tests sample metadata extraction."""
 
     # GIVEN a mock sample and the latest metadata
-    sample = helpers.add_sample(sample_store)
-    sample.internal_id = "ADM1"
```
seallard, line 11: With version 1.4, this update of the internal id causes a unique constraint error:
```
E       sqlalchemy.exc.IntegrityError: (raised as a result of Query-invoked autoflush; consider using a session.no_autoflush block if this flush is occurring prematurely)
E       (sqlite3.IntegrityError) UNIQUE constraint failed: sample.internal_id
E       [SQL: UPDATE sample SET internal_id=?, reads=? WHERE sample.id = ?]
E       [parameters: ('ADM1', None, 13)]
E       (Background on this error at: http://sqlalche.me/e/14/gkpj)
```
I'm not sure why it does not cause an error with version 1.3, but it seems like the auto flush behaviour differs somehow. Or maybe it is related to the new auto begin behaviour.

requirements.txt
```diff
@@ -1,6 +1,6 @@
 # database
 alembic
-SQLAlchemy<1.4
+SQLAlchemy~=1.4.0
```
seallard, line 3: This syntax ensures the latest 1.4.x version is picked up

diitaz93 (approved):
Looks good 👍

henrikstranneheim (approved):
👍

seallard:
Deployed to Hasta and vm stage. Data is written as expected for some common CLI commands. Orders can be placed.

PR #2511, opened 2023-09-22
fix(FASTQ-upload) Prevent all FASTQ uploads from failing if one fails

Vince-janv:
## Description
If the upload of FASTQ files fails on one case it will raise an error and stop the loop.

This PR catches any exceptions raised and continues the loop. But raises a click.abort error in the end if any case failed.

### Fixed

- One FASTQ upload fail no longer blocks other from being uploaded

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b catch-fastq-upload-errors
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/cli/upload/clinical_delivery.py
```diff
@@ -83,12 +84,12 @@ def upload_clinical_delivery(context: click.Context, case_id: str, dry_run: bool
...
-
+    exit_code: int = EXIT_SUCCESS
     status_db: Store = context.obj.status_db
     trailblazer_api: TrailblazerAPI = context.obj.trailblazer_api
     for analysis_obj in status_db.get_analyses_to_upload(pipeline=Pipeline.FASTQ):
         if analysis_obj.family.analyses[0].uploaded_at:
-            LOG.warning(
+            LOG.debug(
                 "Newer analysis already uploaded for %s, skipping",
                 analysis_obj.family.internal_id,
```
diitaz93, line 94:
```suggestion
            LOG.debug(
                f"Newer analysis already uploaded for {analysis_obj.family.internal_id}, skipping"
```

cg/cli/upload/clinical_delivery.py
```diff
@@ -104,14 +105,24 @@ def auto_fastq(context: click.Context, dry_run: bool):
                 )
                 analysis_obj.uploaded_at = dt.datetime.now()
             else:
-                LOG.warning(
+                LOG.debug(
                     "Upload to clinical-delivery for %s has already started, skipping",
```
diitaz93, line 109: same here

cg/cli/upload/clinical_delivery.py
```diff
@@ -104,14 +105,24 @@ def auto_fastq(context: click.Context, dry_run: bool):
...
                     analysis_obj.family.internal_id,
                 )
             continue
         case: Family = analysis_obj.family
         LOG.info("Uploading family: %s", case.internal_id)
```
diitaz93, line 114: f string

cg/cli/upload/clinical_delivery.py
```diff
@@ -104,14 +105,24 @@ def auto_fastq(context: click.Context, dry_run: bool):
...
+        except Exception as error:
+            LOG.error(f"Upload of case {case.internal_id} failed")
+            LOG.error(error)
+            exit_code: int = EXIT_FAIL
+            continue
+
         if not dry_run:
             status_db.session.commit()
+
+        if exit_code:
```
diitaz93, line 127: Im wondering if it should check for fail as `exit_code` is defined above as  `EXIT_SUCCESS`
```suggestion
        if exit_code == EXIT_FAIL:
```

diitaz93 (approved):
Looks good 👍

cg/cli/upload/clinical_delivery.py
```diff
@@ -104,14 +105,24 @@ def auto_fastq(context: click.Context, dry_run: bool):
...
+        try:
+            context.invoke(upload_clinical_delivery, case_id=case.internal_id, dry_run=dry_run)
+        except Exception as error:
+            LOG.error(f"Upload of case {case.internal_id} failed")
+            LOG.error(error)
+            exit_code: int = EXIT_FAIL
+            continue
+
         if not dry_run:
             status_db.session.commit()
```
seallard, line 125: This is interesting, it assumes that no commits are made outside of the CLI command layer (not true in general, maybe in this case). Probably a pattern we would like to implement as to tie the session/transaction life cycle to the duration of the CLI command

cg/cli/upload/clinical_delivery.py
seallard: Any non zero exit code indicates an error occurred, but maybe it is better to be explicit

cg/cli/upload/clinical_delivery.py
```diff
@@ -104,14 +105,24 @@ def auto_fastq(context: click.Context, dry_run: bool):
...
                     "Upload to clinical-delivery for %s has already started, skipping",
                     analysis_obj.family.internal_id,
                 )
             continue
         case: Family = analysis_obj.family
         LOG.info("Uploading family: %s", case.internal_id)
         analysis_obj.upload_started_at = dt.datetime.now()
-        context.invoke(upload_clinical_delivery, case_id=case.internal_id, dry_run=dry_run)
+        try:
+            context.invoke(upload_clinical_delivery, case_id=case.internal_id, dry_run=dry_run)
```
seallard, line 117: Not a fan of this pattern where we call other CLI commands, very much out of scope though 😅

seallard (approved):
Should work as intended 👍

cg/cli/upload/clinical_delivery.py
```diff
@@ -104,14 +105,24 @@ def auto_fastq(context: click.Context, dry_run: bool):
...
+        except Exception as error:
+            LOG.error(f"Upload of case {case.internal_id} failed")
+            LOG.error(error)
+            exit_code: int = EXIT_FAIL
+            continue
```
islean, line 122: Would we want some more explicit errors to catch?

cg/cli/upload/clinical_delivery.py
islean, line 125: I would also like to add that with this general commit, we could introduce errors with this change, if we commit changes for a failed upload. I suggest we either do a more specific error handling and do bot continue regardless of the Exception raised, or that we create a transaction for each case and either do multiple commits, or rollback transactions that fail and keep the "large" commit.

islean (changes requested):
Somewhat worried about problems with database changes committed for failed uploads.

cg/cli/upload/clinical_delivery.py
Vince-janv: Changed to `if exit_code != EXIT_SUCCESS:`

cg/cli/upload/clinical_delivery.py
Vince-janv: Yep, it's the basis upon which cg is built 😢 If we can remove it in every module we re-write I'll be happy.

cg/cli/upload/clinical_delivery.py
Vince-janv: I don't think we want to. I see no type of error where we would not want to try with the next case 🤔  Do you see any potential bugs here?

cg/cli/upload/clinical_delivery.py
Vince-janv: Looking at the command we are calling, there are no changes to any ORM objects. But it's true that it the pattern is very prone to bugs. SA suggested we review it when looking at the session management with the latest SQL-alchemy version

cg/cli/upload/clinical_delivery.py
islean: Not really, I was thinking that maybe it is unnecessary to continue (and that an error should be raised) if we get a connection dropped error or something. But feel free to keep it as is.

cg/cli/upload/clinical_delivery.py
islean: Okay! Then I consider this comment to be out of scope :)

islean (approved):
Nice work! I think the pattern we use is somewhat error prone, but that is out of scope for this PR.

Vince-janv:
Tested in stage.
The command did not exit after the first fail

Vince-janv:
Deployed in production

PR #2519, opened 2023-09-26
Fix complicated comparison code smell

seallard:
## Description
Missed a flagged code smell where a comparison was needlessly complicated.

### Fixed
- Simplify conditional flagged as code smell


Vince-janv (approved):
👍

PR #2521, opened 2023-09-26
Remove function for retrieving sample ids

seallard:
## Description
This PR replaces a function which wrapped the `get_sample_ids` function on the sample sheet with a direct call to it.
This function
```python
def get_sample_internal_ids_from_sample_sheet(
    sample_sheet_path: Path, flow_cell_sample_type: Type[FlowCellSample]
) -> List[str]:
    """Return the sample internal ids for samples in the sample sheet."""
    sample_sheet = get_sample_sheet_from_file(
        infile=sample_sheet_path, flow_cell_sample_type=flow_cell_sample_type
    )
    return sample_sheet.get_sample_ids()
```
is always called in a context where we have a sample sheet object available. It is not necessary to reparse the entire sample sheet each time we want to retrieve the sample ids. We can just call `sample_sheet.get_sample_ids()` directly.

### Fixed
- Replace unnecessary function for retrieving sample ids from a sample sheet

### DIFF
--- a/cg/apps/demultiplex/sample_sheet/read_sample_sheet.py
+++ b/cg/apps/demultiplex/sample_sheet/read_sample_sheet.py
@@ -123,13 +123,3 @@ def get_samples_by_lane(
             sample_by_lane[sample.lane] = []
         sample_by_lane[sample.lane].append(sample)
     return sample_by_lane
-
-
-def get_sample_internal_ids_from_sample_sheet(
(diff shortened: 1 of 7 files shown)


henrikstranneheim (approved):
Great! 👍

diitaz93 (approved):
Great simplification 👍

ChrOertlin (approved):
Nice!

seallard:
Deployed to stage. Could not set up suitable flow cell for testing, but existing tests cover the refactored functionality.

PR #2524, opened 2023-09-26
fix(sars-cov2 excel order) Properly parse the primer field

Vince-janv:
## Description
Customers reported that the Primer field was not properly parsed when they uploaded orderforms.

This PR changes the primer alias from UDF/Primer to UDF/primer to match the header in the excel sheet

### Fixed

- Changed primer alias from UDF/Primer to UDF/primer

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/models/orders/excel_sample.py
+++ b/cg/models/orders/excel_sample.py
@@ -1,5 +1,8 @@
 from typing import List, Optional
 
+from pydantic import AfterValidator, BeforeValidator, Field
+from typing_extensions import Annotated
+
 from cg.models.orders.sample_base import OrderSample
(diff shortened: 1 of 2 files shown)


seallard (approved):
👍 
Out of scope for this PR but it seems like it is possible to make aliases case insensitive with something called [alias generators](https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.alias_generator) in Pydantic.

EDIT: nvm, read some more. Alias generators would not solve this issue

Vince-janv:
Sars-cov2 excel order was successfully parsed in stage

PR #2528, opened 2023-09-26
Add validators to sample id

seallard:
## Description
This PR adds two validators to the sample_id on the sample sheet model.

- Removes trailing index.
- Enforces LIMS ID format (3 letters + ≥3 digits).

This means that any non valid sample id will cause a validation error when the sample sheet is parsed. As far as I know, there is no reason to allow invalid sample ids in sample sheets. Maybe if it is a really old sample sheet?

Having the validation directly on the `sample_id` property on the sample sheet domain model is useful because it provides guarantees for any access of the property in the code base that the sample id has a valid value.

### Added
- Validation of sample id on sample sheet samples


ChrOertlin (approved):
Awesome! This is much better!

cg/apps/demultiplex/sample_sheet/models.py
```diff
@@ -14,12 +17,18 @@
 
 LOG = logging.getLogger(__name__)
 
+SampleId = Annotated[
```
henrikstranneheim, line 20: Should this be in a `cg/apps/demultiplex/sample_sheet/validators.py` and imported here? It feels a bit out of place defined directly in this module

henrikstranneheim (approved):
Nice 👍

seallard:
Deployed to stage. Could not setup suitable test data.

PR #2537, opened 2023-09-29
Add tests for the validators for ExcelSample

islean:
## Description

When migrating to pydantic v2, several bugs have unfortunately been brought to production, highlighting the need for better unit tests for the validators. This PR adds (late) tests to the ExcelSample class to ensure that no more bugs show up, and to catch any introduced by future PRs. 

One validator also sees a change, the numeric_value validator, since the validator was easily bypassed. Previously both 32.3.1 and 2.0ABC would have been seen as valid inputs.

### Added

- Tests for the Excel order form validators

### Changed

- Validator for numeric_values is stricter
- Created constant for the aliases used in ExcelSample.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/models/orders/validators/excel_sample_validators.py
```diff
@@ -33,11 +33,19 @@ def validate_data_analysis(data_analysis):
...
     if not value:
         return None
-    str_value = value.rsplit(".0")[0]
-    if str_value.replace(".", "").isnumeric():
-        return str_value
+    if value.replace(".", "").isnumeric():
+        try:
+            float_value = float(value)
+            if float_value.is_integer():
+                return value.split(".")[0]
```
seallard, line 44: Would this not return the first decimal value? Is that what we want?

cg/models/orders/validators/excel_sample_validators.py
```diff
@@ -33,11 +33,19 @@ def validate_data_analysis(data_analysis):
...
+            return value
+        except ValueError:
+            if value.isnumeric():
+                return value
```
seallard, line 48: This validator is confusing to me, do you think it can be made clearer somehow? Here are some suggestions, feel free to disregard:

```python
from typing import Optional

def numeric_value(value: Optional[str]) -> Optional[str]:
    """Validates that the given string can be cast to either an integer or a float.
    Also converts floats of the type x.00 to x."""
    
    if value is None or value == '':
        return None # I would expect this to raise a value error, but seems like the field is optional?
    
    try:
        float_value = float(value)
    except ValueError:
        return None  # Cannot convert to a float, return None
    
    if float_value.is_integer():
        return str(int(float_value))  # Return as integer string if it's an integer
    
    return str(float_value)  # Return as float string

```

seallard: Not sure about catching the exceptions though and whether it makes sense to do that in a validator

seallard (approved):
Great 👍

cg/models/orders/validators/excel_sample_validators.py
islean: I think that looks much better! Might be worth raising the AttributeError if we can not parse the string as a float as to not change the behaviour, though.

cg/models/orders/validators/excel_sample_validators.py
```diff
@@ -33,12 +33,17 @@ def validate_data_analysis(data_analysis):
 
 
 def numeric_value(value: Optional[str]):
```
Vince-janv, line 35: can we get a return type hint?

Vince-janv: So the function basically checks whether the given value is a valid float or int? It does not perform any type conversion?

Vince-janv (approved):
Looks good 👌

cg/models/orders/validators/excel_sample_validators.py
islean: Yeah, I do not intend to change any output types in this PR, I just wanted to patch some potential bypasses. So it will still return an Optional[str]. I'll add a type hint

islean:
Deployed to stage and production by IO 2023-10-04

PR #2540, opened 2023-10-02
refactor(FlowCellDirectoryData)

ChrOertlin:
## Description

NOTE: the functions are just moved at this point to illustrate my suggestion, so no need to comment on functionality per se. Just the overall idea.

Proposal for a refactoring of the `FlowCellDirectoryData`

Currently the `FlowCellDirectoryData` holds information and functions for files used for example in `DemultiplexingAPI` or `SampleSheetCreator`. 

The `FlowCellDirectoryData` use is more widespread than that. We could refactor the `FlowCellDirectoryData` to only hold common `flow cell` information. This would reduce the class to `directory parsing`, `bcl converter` and `sample sheet`.

Functionality such such `is_flow_cell_ready` and relevant function would be extracted to the `DemultiplexingAPI` where this information is relevant. Similarly with the `run paramters` file for the sample sheet creation.
 

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


diitaz93:
Looking forward to when this is done 👀

cg/apps/demultiplex/demultiplex_api.py
```diff
@@ -186,6 +191,76 @@ def is_demultiplexing_possible(self, flow_cell: FlowCellDirectoryData) -> bool:
...
+
+    def is_copy_completed(self) -> bool:
+        """Check if copy of flow cell is done.
+        This is indicated by that the file CopyComplete.txt exists.
+        """
+        LOG.info("Check if copy of data from sequence instrument is ready")
+        return self.copy_complete_path.exists()
+
+
+def is_flow_cell_ready(self) -> bool:
```
diitaz93, line 248: I guess you missed an indent here?

seallard, line 248: Not sure where this function belongs, but it probably should accept the path to the flow cell directory

diitaz93:
I like the idea to distribute the functionality of this class to other classes. I think that moving some to the SampleSheet class is a good idea. For the demux API I would have to take a closer look. I was thinking more of having another class called DemuxRun, being a parallel of FlowCellDirectoryData and representing the directory of the demultiplex run. Also, I am thinking that we need to think carefully about the adding of the sample sheet to Housekeeper because right now it is done in the FlowCellDirectoryData quite entangled with the sample sheet class. In summary, I like the idea but I would need some time to think if I agree with the implementation details

ChrOertlin:
_"I was thinking more of having another class called DemuxRun, being a parallel of FlowCellDirectoryData and representing the directory of the demultiplex run."_

I was trying to avoid this scenario by the class `FlowCellDirectoryData` just representing the parsed directory with the commonly used features of it by `DemultiplexingAPI`, `CleanFlowCellAPI`, `DemultiplexingPostProcessingAPI` etc. Then have the functionality specific to each API in that respective one. This would decouple some things while centralising core functionality of these APIs

seallard:
The `FlowCellDirectoryData` class is violating the single responsibility principle so I like reducing the number of things it does and moving the more complex logic to more suitable spots 👍

diitaz93:
> I was trying to avoid this scenario by the class `FlowCellDirectoryData` just representing the parsed directory with the commonly used features of it by `DemultiplexingAPI`, `CleanFlowCellAPI`, `DemultiplexingPostProcessingAPI` etc. Then have the functionality specific to each API in that respective one. This would decouple some things while centralising core functionality of these APIs

Makes sense

> `FlowCellDirectoryData` should only retrieve a samplesheet path. As it stands now, if a samplesheet is not in housekeeper then a `FlowCell` should be invalid.

Ok, so what we discussed on Friday of making it an attributed initialised inside the `__init__()` method?

ChrOertlin:
> > I was trying to avoid this scenario by the class `FlowCellDirectoryData` just representing the parsed directory with the commonly used features of it by `DemultiplexingAPI`, `CleanFlowCellAPI`, `DemultiplexingPostProcessingAPI` etc. Then have the functionality specific to each API in that respective one. This would decouple some things while centralising core functionality of these APIs
> 
> Makes sense
> 
> > `FlowCellDirectoryData` should only retrieve a samplesheet path. As it stands now, if a samplesheet is not in housekeeper then a `FlowCell` should be invalid.
> 
> Ok, so what we discussed on Friday of making it an attributed initialised inside the `__init__()` method?

Yes, that sounds good to me!

ChrOertlin:
> he `FlowCellDirectoryData` class is violating the single responsibility principle so I like reducing the number of things it does and moving the more complex logic to suitable spots 👍 I think it makes sense to let this class mainly parse and hold flow cell directory data. It might make sense to keep some methods related to the state of the directory in the class though

When it comes to the state of the directory it becomes a bit messy since it can be both used for a `flow_cells` and `demultiplexed_runs` directory. Depending on the context those states differ. Therefore if for example an API requires a file to be present, the API should perform the check in my opinion. 

I think it is fair to also propose a name change to perhaps; `ParsedFlowCell` or something in that direction.

Requirements for a `ParsedFlowCell`:
- The directory is properly names
- It has a sample sheet in housekeeper
- It holds the bcl converter as an attribute

I think those are the `core` things we require from a flow cell in any other workflow they are used in.

seallard:
Another idea would be to merge the `FlowCell` and `FlowCellDirectoryData` models, to just merge the data in the `FlowCellDirectoryData` model into the database model.

islean:
I like the idea of decoupling this a bit! I am not too familiar with the FlowCellDirectoryData class but it seems to be doing a bit too much. I would just like to throw into the mix that we could the proposed, stripped FlowCellDirectoryData class as a parent class, which we extend in child classes which are more specialized to the flow. Would it make sense to have e.g. a DemultiplexFlowCellDirectory(FlowCellDirectoryData) class which adds `is_copy_completed` and other methods currently proposed to move to the DemultiplexAPI?

ChrOertlin:
> I like the idea of decoupling this a bit! I am not too familiar with the FlowCellDirectoryData class but it seems to be doing a bit too much. I would just like to throw into the mix that we could have the proposed, stripped FlowCellDirectoryData class as a parent class, which we extend in child classes which are more specialized to the flow. Would it make sense to have e.g. a DemultiplexFlowCellDirectory(FlowCellDirectoryData) class which adds `is_copy_completed` etc rather than move them to the Demultiplex API?

I discussed this with @Vince-janv a bit. This could work also, although it would introduce juggling multiple different child classes for things that are related to specific workflows like demultiplexing, encryption etc. Why not let that be handled by the workflow it is needed for?

karlnyr:
Is there a reason to why we cannot have flow cell directory data which is a mother class to the FlowCellRunsData and DemultiplexedRunsData? Would implement more realistic information about what is inside the different data structures as well because they are not equivalent, the only thing they really share is the directory name.

cg/models/demultiplex/flow_cell.py
```diff
@@ -87,26 +90,6 @@ def get_sample_sheet_path_hk(self) -> Path:
             raise FlowCellError("Attribute _sample_sheet_path_hk has not been assigned yet")
         return self._sample_sheet_path_hk
 
-    @property
-    def run_parameters_path(self) -> Path:
-        """Return path to run parameters file."""
-        return Path(self.path, DemultiplexingDirsAndFiles.RUN_PARAMETERS)
-
-    @property
-    def run_parameters(self) -> RunParameters:
```
henrikstranneheim, line 96: Does these more belong to how the flow cell was processed than downstream demux (and potential other processes)?

islean:
> > I like the idea of decoupling this a bit! I am not too familiar with the FlowCellDirectoryData class but it seems to be doing a bit too much. I would just like to throw into the mix that we could have the proposed, stripped FlowCellDirectoryData class as a parent class, which we extend in child classes which are more specialized to the flow. Would it make sense to have e.g. a DemultiplexFlowCellDirectory(FlowCellDirectoryData) class which adds `is_copy_completed` etc rather than move them to the Demultiplex API?
> 
> I discussed this with @Vince-janv a bit. This could work also, although it would introduce juggling multiple different child classes for things that are related to specific workflows like demultiplexing, encryption etc. Why not let that be handled by the workflow it is needed for?

I don't think I see the clash here. As I see it, the child classes would be a part of the workflow, but decouple the Demultiplex API a bit, making it a bit closer to adhering to the SRP.

henrikstranneheim (commented):
Great initiative 👍

ChrOertlin:
I see the point @karlnyr, @islean and @Vince-janv  were trying to make with the child classes now. I alter this Pr in that spirit.

ChrOertlin:
I updated the proposal to incorporate feedback of child classes now. Have a look at the cg/model/flow_cell/flow_cell.py file in particular.

cg/models/flow_cell/flow_cell.py
```diff
@@ -250,52 +188,102 @@ def get_sample_sheet(self) -> SampleSheet:
...
+
+class DemultiplexedFlowCell(FlowCellDirectoryData):
+    """This class represents a demultiplexed flow cell in the demultiplexed runs folder."""
+
+    @property
+    def is_demultiplexing_complete(self) -> bool:
+        return Path(self.path, DemultiplexingDirsAndFiles.DEMUX_COMPLETE).exists()
+
+
+class CleanFLowCell(FlowCellDirectoryData):
```
islean, line 280:
```suggestion
class CleanFlowCell(FlowCellDirectoryData):
```

islean:
This feels much more intuitive to me! 🔥

cg/apps/demultiplex/demultiplex_api.py
```diff
@@ -186,6 +193,76 @@ def is_demultiplexing_possible(self, flow_cell: FlowCellDirectoryData) -> bool:
...
+    def has_demultiplexing_started_on_sequencer(self) -> bool:
+        """Check if demultiplexing has started on the NovaSeqX machine."""
+        latest_analysis: Path = get_latest_analysis_path(self.path)
+        if not latest_analysis:
+            return False
+        return Path(
+            latest_analysis, DemultiplexingDirsAndFiles.DATA, DemultiplexingDirsAndFiles.BCL_CONVERT
+        ).exists()
+
+    def is_sequencing_done(self) -> bool:
```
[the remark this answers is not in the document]
ChrOertlin: have a look at the new cg/models/flow_cell/flow_cell.py in here. @Vince-janv , @karlnyr and @islean  suggested a parent- child class setup that I outlined there now. This would solve the cluttering and non specificity of the FlowCellDirectoryData, while keeping the usability in multiple places how it is used now.

cg/models/flow_cell/flow_cell.py
```diff
@@ -250,52 +191,102 @@ def get_sample_sheet(self) -> SampleSheet:
...
-        """Check if copy of Hiseq X flow cell is done."""
-        LOG.info("Check if copy of data from Hiseq X sequence instrument is ready")
-        return self.hiseq_x_copy_complete_path.exists()
+class SequencedFlowCellData(FlowCellDirectoryData):
+    """This class represents a sequenced flow cell in the flow cells folder."""
+
+    @property
+    def rta_complete_path(self) -> Path:
+        """Return RTAComplete path."""
+        return Path(self.path, DemultiplexingDirsAndFiles.RTACOMPLETE)
```
henrikstranneheim, line 204:
```suggestion
        return Path(self.path, SequencingDirsAndFiles.RTACOMPLETE)
```

cg/models/flow_cell/flow_cell.py
```diff
@@ -250,52 +191,102 @@ def get_sample_sheet(self) -> SampleSheet:
...
+
+    @property
+    def copy_complete_path(self) -> Path:
+        """Return copy complete path."""
+        return Path(self.path, DemultiplexingDirsAndFiles.COPY_COMPLETE)
```
henrikstranneheim, line 209:
```suggestion
        return Path(self.path, SequencingDirsAndFiles.COPY_COMPLETE)
```

cg/models/flow_cell/flow_cell.py
```diff
@@ -250,52 +191,102 @@ def get_sample_sheet(self) -> SampleSheet:
...
+
+    @property
+    def demultiplexing_started_path(self) -> Path:
```
henrikstranneheim, line 212: Does this one belong here or in demux?

cg/models/flow_cell/flow_cell.py
```diff
@@ -250,52 +191,102 @@ def get_sample_sheet(self) -> SampleSheet:
...
+                if self.sequencer_type == Sequencers.NOVASEQX
+                else RunParametersNovaSeq6000(run_parameters_path=self.run_parameters_path)
+            )
+        return self._run_parameters
 
-    def is_hiseq_x_delivery_started(self) -> bool:
-        """Check if delivery of Hiseq X flow cell is started."""
-        LOG.info("Check if delivery of data from Hiseq X sequence instrument is ready")
-        return self.hiseq_x_delivery_started_path.exists()
+    def has_demultiplexing_started_locally(self) -> bool:
```
henrikstranneheim, line 231: Needed here?

cg/models/flow_cell/flow_cell.py
```diff
@@ -250,52 +191,102 @@ def get_sample_sheet(self) -> SampleSheet:
...
+
+class DemultiplexedFlowCellData(FlowCellDirectoryData):
+    """This class represents a demultiplexed flow cell in the demultiplexed runs folder."""
+
+    @property
+    def is_demultiplexing_complete(self) -> bool:
+        return Path(self.path, DemultiplexingDirsAndFiles.DEMUX_COMPLETE).exists()
+
+
+class CleanFlowCellData(FlowCellDirectoryData):
```
henrikstranneheim, line 283: Will these hold something otherwise we can skip them

henrikstranneheim (commented):
Nice - I think this is a great improvement

cg/models/flow_cell/flow_cell.py
ChrOertlin: Yes, they will hold the information on used in the cleaning of flow cells. I will check to what extent. For now this is more  to see if we all agree on this change. Then I will implement it!

ChrOertlin:
closing this in favour of avoiding the many conflicts. Will start in other PRs focussing on single APIs things at a time.

PR #2553, opened 2023-10-03
Fix deprecation warnings

seallard:
## Description
This PR fixes all Pydantic deprecation warnings for models already migrated to v2 but still using methods from v1.

Replace calls to
- `parse_obj` with `model_validate`
- `dict` with `model_dump`
- `Extra.ignore` with `'ignore'`

### Fixed
- Pydantic deprecation warnings

### DIFF
--- a/cg/apps/demultiplex/sample_sheet/models.py
+++ b/cg/apps/demultiplex/sample_sheet/models.py
@@ -2,7 +2,7 @@ import logging
 from collections import defaultdict
 from typing import List, Tuple
 
-from pydantic import BaseModel, ConfigDict, Extra, Field
+from pydantic import BaseModel, ConfigDict, Field
 
 from cg.apps.demultiplex.sample_sheet.validators import SampleId
 from cg.constants.constants import GenomeVersion
@@ -21,7 +21,7 @@ class FlowCellSample(BaseModel):
     sample_id: SampleId
     index: str
     index2: str = ""
-    model_config = ConfigDict(populate_by_name=True, extra=Extra.ignore)
+    model_config = ConfigDict(populate_by_name=True, extra='ignore')
 
 
 class FlowCellSampleBcl2Fastq(FlowCellSample):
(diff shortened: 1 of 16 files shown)


cg/apps/demultiplex/sample_sheet/models.py
```diff
@@ -21,7 +21,7 @@ class FlowCellSample(BaseModel):
     sample_id: SampleId
     index: str
     index2: str = ""
-    model_config = ConfigDict(populate_by_name=True, extra=Extra.ignore)
+    model_config = ConfigDict(populate_by_name=True, extra="ignore")
```
islean, line 24: Would we want to create a constant for "ignore?

cg/server/api.py
```diff
@@ -489,7 +489,7 @@ def parse_orderform():
         error_message = error.message if hasattr(error, "message") else str(error)
         http_error_response = http.HTTPStatus.INTERNAL_SERVER_ERROR
     else:
-        return jsonify(**parsed_order.dict())
+        return jsonify(**parsed_order.model_dump())
```
islean, line 492: What does jsonify do? Does it do the same as `return parsed_order.model_dump_json()`?

islean (approved):
Looks good!

cg/apps/demultiplex/sample_sheet/models.py
seallard: Maybe if it is used elsewhere, but it is not now so I think it is fine

cg/server/api.py
seallard: What is jsonify?

seallard: Oh, did not even notice it. Hm, no idea. Reading docs...
```
This function wraps dumps to add a few enhancements that make
life easier. It turns the JSON output into a ~flask.Response
```

seallard: So I don't think it returns the same as `model_dump_json()`

islean: Oh, not the same then! model_dump_json() gives a string I think.

islean: Or maybe? All examples seems to give a string formatted as the class...

seallard: Might be the case! Going to leave it for now, seems a little out of scope

Issue #2008, opened 2023-05-05
Database connection dropped


seallard:
## Description
Occasionally we get 500 errors due to the database connection being dropped. We should look into handling this more gracefully. Here is an example log with the error:

```
10.0.2.100 - - [05/May/2023:05:41:34 +0200] "GET /admin/organism/ HTTP/1.0" 500 0 "-" "-"
(Background on this error at: http://sqlalche.me/e/13/e3q8)
[parameters: {'email_1': 'sebastian.allard@scilifelab.se', 'param_1': 1}]
 LIMIT %(param_1)s]
WHERE user.email = %(email_1)s
FROM user
[SQL: SELECT user.id AS user_id, user.name AS user_name, user.email AS user_email, user.is_admin AS user_is_admin, user.order_portal_login AS user_>
sqlalchemy.exc.OperationalError: (pymysql.err.OperationalError) (2006, "MySQL server has gone away (BrokenPipeError(32, 'Broken pipe'))")
    CR.CR_SERVER_GONE_ERROR, "MySQL server has gone away (%r)" % (e,)
  File "/usr/local/lib/python3.7/site-packages/pymysql/connections.py", line 764, in _write_bytes
    self._write_bytes(packet)
  File "/usr/local/lib/python3.7/site-packages/pymysql/connections.py", line 818, in _execute_command
    self._execute_command(COMMAND.COM_QUERY, sql)
  File "/usr/local/lib/python3.7/site-packages/pymysql/connections.py", line 548, in query
    conn.query(q)
  File "/usr/local/lib/python3.7/site-packages/pymysql/cursors.py", line 325, in _query
    result = self._query(query)
  File "/usr/local/lib/python3.7/site-packages/pymysql/cursors.py", line 158, in execute
    cursor.execute(statement, parameters)
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 608, in do_execute
    cursor, statement, parameters, context
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1277, in _execute_context
    raise exception
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/util/compat.py", line 182, in raise_
    sqlalchemy_exception, with_traceback=exc_info[2], from_=e
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1511, in _handle_dbapi_exception
    e, statement, parameters, cursor, context
[...]

PR #2554, opened 2023-10-04
Avoid stale connections

seallard:
## Description
This PR addresses issue #2008 where the database connection seems to be dropped intermittently. It adds pessimistic disconnect handling, see [docs](https://docs.sqlalchemy.org/en/20/core/pooling.html#dealing-with-disconnects).

"The approach adds a small bit of overhead to the connection checkout process, however is otherwise the most simple and reliable approach to completely eliminating database errors due to stale pooled connections. "

The feature is available in sqlalchemy 2.0 as well.

### Fixed
- Avoid stale database connections


henrikstranneheim (approved):
👍

Vince-janv (approved):
Great addition! 👌

seallard:
Deployed and tested on cg-vm1.

PR #2556, opened 2023-10-04
Upgrade core packages

seallard:
## Description
This PR upgrades the versions of these packages:
- Flask 1.1.4 -> 3.0.0 [change log](https://flask.palletsprojects.com/en/3.0.x/changes/)
- Flask-Dance 5.1.0 -> 7.0.0 [change log](https://github.com/singingwolfboy/flask-dance/blob/main/CHANGELOG.rst)
- Click 6.7 -> 8.1.7 [change log](https://click.palletsprojects.com/en/8.1.x/changes/)

Each change log was read and changes made accordingly (for any breaking changes).

It is necessary to upgrade all of them at once since they lock each other. Upgrading these packages has the upside of also letting us upgrade Python, they all support Python 3.12.

## Upgrading flask 1.1.4 -> 3.0.0 
- Replace `from flask.json import JSONEncoder` with `from json import JSONEncoder`, [removed in 2.3.0](https://flask.palletsprojects.com/en/2.3.x/changes/#version-2-3-0).
- Replace `json_encoder` with `json_provider_class` on app, [removed in 2.3.0](https://flask.palletsprojects.com/en/3.0.x/changes/#version-2-3-0).
- Replace `file_name` argument with `path` in `send_from_directory`, [renamed in 2.2.0](https://flask.palletsprojects.com/en/3.0.x/changes/#version-2-2-0).
- Remove env variable for deprecated attribute `templates_auto_reload`, [removed in 2.3.0](https://flask.palletsprojects.com/en/3.0.x/changes/#version-2-3-0).

Notes:
- `SESSION_COOKIE_DOMAIN` is no longer set by default, see [this](https://github.com/pallets/flask/issues/5051) and [change log for 2.3.0](https://flask.palletsprojects.com/en/3.0.x/changes/#version-2-3-0). It does not seem to matter though.

## Upgrading flask-dance 5.1.0 -> 7.0.0
No breaking changes beyond requiring a more recent version of Flask, see [change log](https://github.com/singingwolfboy/flask-dance/blob/main/CHANGELOG.rst).

## Upgrading Click 6.7 -> 8.1.7 
No relevant breaking changes, see [change log](https://click.palletsprojects.com/en/8.1.x/changes/).

### Fixed
- Upgrade flask, flask-dance and click to the latest versions.

### This [version](https://semver.org/) is a
- [ ] **PATCH** - when you make backwards compatible bug fixes or documentation/instructions

## Implementation Plan
[...]

PR #2565, opened 2023-10-06
Upgrade to Python 3.9

seallard:
## Description
This PR upgrades cg to use Python 3.9. See [migration guide](https://docs.python.org/3/whatsnew/3.9.html#porting-to-python-3-9). We are not affected by any of the breaking changes. All tests pass locally.

Blocked by https://github.com/Clinical-Genomics/cg/pull/2556 since the current flask version we use (1.3) only officially supports Python 3.8. Also need to upgrade housekeeper to 3.9 before merging since they are executed in the same conda environment as cg.

## Testing
- Upgrade the Python version used in `S_cg`
- Deploy this branch to stage in Hasta and on the vm.
- Verify that the cli works and orders can be placed

### Fixed
- Upgrade to Python 3.9 from 3.8


seallard:
Blocked by https://github.com/Clinical-Genomics/cg/pull/2556 since the current flask version we use (1.3) only officially supports Python 3.8.

henrikstranneheim (approved):
👍

seallard:
Unlocking the genologic package, there does not seem to be any breaking changes in the latest release to adress

seallard:
Cli works, can place order.

PR #2568, opened 2023-10-06
Upgrade to Python 3.11

seallard:
## Description
This PR upgrades cg to Python 3.11. We are not affected by the [breaking changes](https://docs.python.org/3/whatsnew/3.11.html#removed). All of our packages support Python 3.11.

### Fixed
- Upgrade to Python 3.11


henrikstranneheim (approved):
Nice 👍

PR #2571, opened 2023-10-09
Lock genologics package

seallard:
## Description
The new major release of the genologics package breaks placing orders (at least sars cov). Locking it for now.
See https://github.com/Clinical-Genomics/cg/issues/2572.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -41,5 +41,5 @@ typing_extensions
 urllib3
 
 # apps
-genologics
+genologics==0.3.18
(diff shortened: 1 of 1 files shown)


Vince-janv (approved):
Almost expected something like this 😅

seallard:
> Almost expected something like this 😅

The issue persists in stage though, even when the package has been locked to the previous release 😢 
So now I'm not sure what to do, will dig more after lunch

PR #2579, opened 2023-10-09
Migrate delivery report models to Pydantic v2

ivadym:
### Changed
- Migrate Delivery report models to Pydantic v2

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### Test generation of delivery reports for different workflow

- [x] Balsamic report
- [x] MIP-DNA report
- [x] Rnafusion report
- [x] Upload to Scout

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/models/report/validators.py
```diff
@@ -1,100 +1,77 @@
...
+    """Return string representation of a float value."""
     return str(round(float(value), PRECISION)) if value or isinstance(value, float) else NA_FIELD
 
 
-def validate_percentage(value: Union[float, str]) -> str:
-    """Returns a processed float value as a percentage."""
-    return validate_float(float(value) * 100) if value else NA_FIELD
+def get_float_as_percentage(value: Optional[float]) -> str:
+    """Return string percentage representation of a float value."""
+    return get_float_as_string(value * 100) if value or isinstance(value, float) else NA_FIELD
```
islean, line 42: What is the intention with "or isinstance(value, float)"? Given lazy loading, I would guess that the only value passed to "isinstance(value,float)" is when value == None?

cg/models/report/validators.py
```diff
@@ -1,100 +1,77 @@
...
 
 
-def validate_date(date: datetime) -> str:
-    """Returns the date part (year, month, day) from a datetime object."""
+def get_date_as_string(date: Optional[datetime]) -> str:
+    """Return the date string representation (year, month, day) of a datetime object."""
     return str(date.date()) if date else NA_FIELD
```
islean, line 47: Do we know that we always receive a datetime object? I would guess so, since the functionality hasn't really changed. Otherwise we could get problems with date.date() I guess.

cg/models/report/validators.py
```diff
@@ -1,100 +1,77 @@
...
+    """Return a report validated file name."""
     return Path(file_path).name if file_path and Path(file_path).is_file() else NA_FIELD
 
 
-def validate_gender(value: str) -> str:
-    """Formats the provided gender."""
-    return validate_empty_field(REPORT_GENDER.get(value))
+def get_gender_as_string(gender: Optional[Gender]) -> str:
+    """Return a report adapted gender."""
+    return get_report_string(REPORT_GENDER.get(gender)) if gender else NA_FIELD
```
islean, line 62:
```suggestion
    return get_report_string(REPORT_GENDER.get(gender))
```
The get_report_string already returns NA_FIELD right?

cg/models/report/metadata.py
```diff
@@ -17,15 +19,10 @@ class SampleMetadataModel(BaseModel):
...
     """
 
-    million_read_pairs: Union[None, float, str]
-    duplicates: Union[None, float, str]
-
-    _float_values = validator("million_read_pairs", "duplicates", always=True, allow_reuse=True)(
-        validate_float
-    )
+    million_read_pairs: Annotated[Optional[str], BeforeValidator(get_float_as_string)] = NA_FIELD
+    duplicates: Annotated[Optional[str], BeforeValidator(get_float_as_string)] = NA_FIELD
```
islean, line 25: Do we need "Optional[str]" anymore if we set "N/A" with a BeforeValidator?

islean (commented):
I will come back and look through this a second time tomorrow! But seems as if it will do what you want. Generally, I think that you can probably remove many "Optional"s, since BeforeValidators act before any type check is done, i.e. if None is converted to "N/A" by a BeforeValidator, you do not need the Optional. 

I think we could probably make the code easier by setting `model_config = ConfigDict(coerce_numbers_to_str=True)` and then using model validator to set any None values to "N/A". But even that is true, I don't think this is the PR for that.

cg/models/report/validators.py
ivadym: Just to take into consideration the possibility of zero input values. It's unlikely that we'll get them, as it would mean ignoring some QC checks, but I suppose it's a good practice to account for the majority of possible scenarios

cg/models/report/validators.py
ivadym: Good point! 👍 For some reason I though that get(None) will raise an error

cg/models/report/validators.py
```diff
@@ -1,100 +1,77 @@
...
-        values["type"] = validate_balsamic_analysis_type(values["type"])
-    return values
+    return get_report_string(prep_category)
+
+
+def get_analysis_type_as_string(analysis_type: Optional[str], info: ValidationInfo) -> str:
+    """Return the analysis type as an accepted string value for the delivery report."""
+    if analysis_type and Pipeline.BALSAMIC in info.data.get("pipeline"):
+        analysis_type: str = BALSAMIC_ANALYSIS_TYPE.get(analysis_type)
+    return get_report_string(analysis_type)
```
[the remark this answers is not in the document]
ivadym: Yes! This behaviour has been retained from v1. The only thing to keep in mind is the order of definition of these fields

cg/models/report/validators.py
```diff
@@ -1,100 +1,77 @@
...
+from cg.constants.constants import Pipeline, PrepCategory
+from cg.constants.subject import Gender
 from cg.models.orders.constants import OrderType
 
 LOG = logging.getLogger(__name__)
 
 
-def validate_empty_field(value: Union[int, str]) -> str:
-    """Formats an empty value to be included in the report as N/A."""
+def get_report_string(value: Any) -> str:
```
seallard, line 23: I guess this one can be removed by adding string coercion as @islean described

seallard (approved):
Looks good! Would be nice to avoid the explicit string conversion validators

cg/models/report/validators.py
ivadym: I'm not sure if we can, because we'll still need to add validators to transform `None` to a report valid empty field `N/A`

The problem is that we could have both `None` and empty values for these field depending on different analysis

ivadym:
Deployed together with: https://github.com/Clinical-Genomics/cg/pull/2616

PR #2584, opened 2023-10-10
Extract database initialisation

seallard:
## Description
This PR extracts the database initialisation to a separate module `database.py`, which is the common and recommended way of doing it, instead of keeping it in the `Store` class. See [session maker docs](https://docs.sqlalchemy.org/en/20/orm/session_basics.html#when-do-i-make-a-sessionmaker) and [scoped_session docs](https://docs.sqlalchemy.org/en/20/orm/contextual.html#unitofwork-contextual).

### Fixed
- Extract db initialisation to separate module


cg/cli/base.py
```diff
@@ -65,22 +65,20 @@ def base(
...
-        status_db.drop_all()
+        drop_all_tables()
     elif existing_tables:
         LOG.error("Database already exists, use '--reset'")
         raise click.Abort
 
-    status_db.create_all()
-    LOG.info("Success! New tables: %s", ", ".join(status_db.engine.table_names()))
+    create_all_tables()
+    LOG.info("Success! New tables: %s", ", ".join(get_tables()))
```
ChrOertlin, line 81:
```suggestion
    LOG.info("Success! New tables: %s", ", ".join(get_tables()))
```
can we make this an f string? or out of scope?

cg/cli/base.py
```diff
@@ -65,22 +65,20 @@ def base(
 @base.command()
 @click.option("--reset", is_flag=True, help="reset database before setting up tables")
 @click.option("--force", is_flag=True, help="bypass manual confirmations")
-@click.pass_obj
-def init(context: CGConfig, reset: bool, force: bool):
+def init(reset: bool, force: bool):
```
ChrOertlin, line 68:
```suggestion
def init_statusdb(reset: bool, force: bool):
```
ignore if out of scope

cg/store/database.py
```diff
@@ -0,0 +1,44 @@
...
+from cg.store.models import Model
+
+SESSION: Optional[Session] = None
+ENGINE = None
+
+
+def initialise_database(db_uri: str):
+    global SESSION, ENGINE
+    ENGINE = create_engine(db_uri, pool_pre_ping=True)
+    session_factory = sessionmaker(ENGINE)
```
ChrOertlin, line 16: would this not throw errors if the connection cannot be established?

cg/store/database.py
```diff
@@ -0,0 +1,44 @@
...
+
+def get_engine():
+    if not ENGINE:
+        raise CgError("Database not initialised")
+    return ENGINE
+
+
+def create_all_tables() -> None:
+    session: Session = get_session()
+    Model.metadata.create_all(bind=session.get_bind())
```
ChrOertlin, line 34: just because I am unfamiliar with session handling. This take the global variable from `initialise_database` can it happen that global variable is overwritten unexpectedly? My question would it be better to be more explicit and pass a session here, or does it not matter?

cg/store/database.py
```diff
@@ -0,0 +1,44 @@
...
+    global SESSION, ENGINE
+    ENGINE = create_engine(db_uri, pool_pre_ping=True)
+    session_factory = sessionmaker(ENGINE)
+    SESSION = scoped_session(session_factory)
+
+
+def get_session() -> Session:
+    if not SESSION:
+        raise CgError("Database not initialised")
+    return SESSION()
```
ChrOertlin, line 23:
```suggestion
    return SESSION
```

cg/store/database.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def get_engine():
```
ChrOertlin, line 26:
```suggestion
def get_engine():
```
type hint

cg/store/database.py
```diff
@@ -0,0 +1,44 @@
...
+
+from cg.exc import CgError
+from cg.store.models import Model
+
+SESSION: Optional[Session] = None
+ENGINE = None
+
+
+def initialise_database(db_uri: str):
+    global SESSION, ENGINE
```
ChrOertlin, line 14: would be nice with a docstring clarifying the global variables

ChrOertlin (approved):
Looks good, I like the extraction to a separate module

cg/store/database.py
ChrOertlin: I just see this is the same as in https://github.com/Clinical-Genomics/housekeeper/pull/151 

is the `()` wanted? It is a bit confusing maybe a docstring is warranted here

cg/cli/base.py
seallard: Leaving it for now

cg/store/database.py
seallard: Yes, `SESSION` is a factory - see [session docs](https://docs.sqlalchemy.org/en/20/orm/contextual.html#contextual-thread-local-sessions)

cg/store/database.py
seallard: It is a ` MockConnection | Any` but I cannot import it. I think it is ok to omit (not that I have an option) since this function is only used inside this module.

cg/store/database.py
seallard: I don't think it matters in this case, but we will likely need to make a choice later for other methods using the session whether we want to pass it as a parameter or just retrieve a global `scoped_session`

cg/store/database.py
seallard: Not sure, cannot find any docs for it 🤔 But I would guess so. It has the same behaviour as before at least.

cg/store/database.py
ChrOertlin: maybe we can clarify that in a docstring then so it is documented for the reader.

henrikstranneheim (approved):
Much cleaner 👍

seallard:
Deployed to web stage, status db views and orders can be placed.
Deployed to hasta stage, cli works.

cg/store/api/core.py
```diff
@@ -24,25 +21,13 @@ class CoreHandler(
...
-
-    def create_all(self):
-        """Create all tables in the database."""
-        Model.metadata.create_all(bind=self.session.get_bind())
-
-    def drop_all(self):
-        """Drop all tables in the database."""
-        Model.metadata.drop_all(bind=self.session.get_bind())
+    def __init__(self):
+        self.session = get_session()
```
seallard, line 32: This will be refactored at a later point. We are using a scoped session, it does not really matter how it is passed. I think it is confusing to have it as an attribute, but lets keep it for now.

seallard:
Deployed to stage and prod

Issue #2541, opened 2023-10-02
Re-implement DeleteDemuxAPI


Vince-janv:
## Description

Before post-processing a demultiplexed flow cell we want to make sure the flow cell is correctly removed from all two (formerly three) databases.

## Suggested solution

- [x] Replace the removed cg-stats parts with cleaning of the SequencingMetrics table. #2202
- [x] Make sure that the cleaning of FASTQ, SPRING and SPRING-metadata files are cleaned correctly
- [x] Add the deletion to the beginning of the post-processing command https://github.com/Clinical-Genomics/cg/pull/2593

## This can be closed when

- [ ] Re-demultiplexing can be performed even if the flow cell has data in the databases
- [ ] Data is not duplicated when re-demultiplexing a flow cell

## Blocked by

https://github.com/Clinical-Genomics/cg/issues/2131

Vince-janv:
This is done 🥳

PR #2144, opened 2023-06-19
Fix(demulitplexing) Remove Housekeeper removal step when demuxing

Vince-janv:
## Description
Linked to [this issue](https://github.com/Clinical-Genomics/cg/issues/2131). This PR changes the demultiplexing commands, so as to not remove FASTQ and SPRING files from housekeeper before demultiplexing. As a consequence if you want to replace the FASTQ files with new ones from a second demultiplexing you will need to manually run the command `cg demultiplex delete-flow-cell` with the `--housekeeper` flag.

### Changed

- `cg demultiplex all` and `cg demultiplex flow-cell` no longer deletes housekeeper files.

### How to prepare for test

- [ ] Ssh to relevant server (depending on type of change)
- [ ] Use stage: `us`
- [ ] Paxa the environment: `paxa`
- [ ] Install on stage (example for Hasta):
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_main -t cg -b [THIS-BRANCH-NAME] -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...
- [ ] Take a screenshot and attach or copy/paste the output.

## Review

- [ ] Tests executed by
- [ ] "Merge and deploy" approved by
Thanks for filling in who performed the code review and the test!

### This [version](https://semver.org/) is a

- [x] **MINOR** - when you add functionality in a backwards compatible manner

## Implementation Plan

- [ ] Document in ...
- [ ] Deploy this branch on ...
- [ ] Inform to ...

PR #2593, opened 2023-10-16
Delete old flow cell before adding re-demultiplexed flow cell

diitaz93:
## Description
Part of the tasks in #2541. 
In #2144 the automatic deletion of a flow cell from Housekeeper and StatusDB was removed, having to explicitly delete a flow cell before re-demultiplexing. This PR re-incorporates the removed deletion but not in the demultiplex command but in the post-processing, more specifically in the `finish_flow_cell` command.

### Added

- Instantiation of the `DeleteDemuxAPI` and call to its command `delete_flow_cell` inside the command `finish_flow_cell`.

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/demultiplex/delete_demultiplex_api.py
+++ b/cg/meta/demultiplex/delete_demultiplex_api.py
@@ -52,7 +52,7 @@ class DeleteDemuxAPI:
 
     @property
     def status_db_presence(self) -> bool:
-        """Update about the presence of given flow cell in status_db"""
+        """Update about the presence of given flow cell in status_db."""
         return bool(self.status_db.get_flow_cell_by_name(flow_cell_name=self.flow_cell_name))
 
     @staticmethod
--- a/cg/meta/demultiplex/demux_post_processing.py
+++ b/cg/meta/demultiplex/demux_post_processing.py
@@ -5,6 +5,7 @@ from typing import Optional
 
 from cg.apps.housekeeper.hk import HousekeeperAPI
 from cg.exc import FlowCellError, MissingFilesError
+from cg.meta.demultiplex.delete_demultiplex_api import DeleteDemuxAPI
 from cg.meta.demultiplex.housekeeper_storage_functions import (
     store_flow_cell_data_in_housekeeper,
 )
@@ -26,6 +27,7 @@ class DemuxPostProcessingAPI:
     """Post demultiplexing API class."""
 
     def __init__(self, config: CGConfig) -> None:
+        self.config: CGConfig = config
         self.flow_cells_dir: Path = Path(config.flow_cells_dir)
         self.demultiplexed_runs_dir: Path = Path(config.demultiplexed_flow_cells_dir)
         self.status_db: Store = config.status_db
@@ -82,6 +84,19 @@ class DemuxPostProcessingAPI:
         except (FlowCellError, MissingFilesError) as e:
             LOG.warning(f"Flow cell {flow_cell_directory_name} will be skipped: {e}")
             return
+
+        delete_demux_api = DeleteDemuxAPI(
+            config=self.config, dry_run=self.dry_run, flow_cell_name=flow_cell_directory_name
+        )
+        if delete_demux_api.status_db_presence:
+            delete_demux_api.delete_flow_cell(
+                demultiplexing_dir=False,
+                run_dir=False,
+                housekeeper=True,
+                init_files=False,
+                status_db=True,
+                sample_lane_sequencing_metrics=True,
+            )
 
         try:
             self.store_flow_cell_data(flow_cell)


islean:
Nice! Seems to already have support for filtering on the flow cell tag!

ChrOertlin (approved):
Looks good. We should check with @karlnyr if this is the behaviour that we want in CG (i.e. deleting previous flow cell entries)

diitaz93:
> Looks good. We should check with @karlnyr if this is the behaviour that we want in CG (i.e. deleting previous flow cell entries)

Yes, I added this to the next refinement meeting so that @karlnyr can give me some advice on this

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -83,6 +85,19 @@ def finish_flow_cell(
...
+            config=self.config, dry_run=self.dry_run, flow_cell_name=flow_cell_directory_name
+        )
+        if delete_demux_api.status_db_presence:
+            delete_demux_api.delete_flow_cell(
+                demultiplexing_dir=False,
+                run_dir=False,
+                housekeeper=True,
+                init_files=False,
+                status_db=True,
+                sample_lane_sequencing_metrics=True,
```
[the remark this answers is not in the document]
diitaz93: I will take a look at it 👍. All these values are tentative for now anyways

cg/meta/demultiplex/delete_demultiplex_api.py
```diff
@@ -52,7 +52,7 @@ def get_path_for_flow_cell(
 
     @property
     def status_db_presence(self) -> bool:
-        """Update about the presence of given flow cell in status_db"""
+        """Update about the presence of given flow cell in status_db."""
         return bool(self.status_db.get_flow_cell_by_name(flow_cell_name=self.flow_cell_name))
```
Vince-janv, line 56: I think the name of this property should be `flow_cell_exists_in_status_db` for readability. Also the a doc string like: `"""Returns wether a flow cell exists in status_db."""`

Vince-janv (commented):
I like having the delete call here rather than before demultiplexing.

Here are some thoughts though:

- [x] Simplify `delete_demux_api.delete_flow_cell` to only perform what we think is useful
- [x] Since the call is moved after demultiplexing, it should not remove any directories.
- [ ] It should not remove the sample sheet
- [ ] Do we need to check active samples? (more hassle than worth)

cg/meta/demultiplex/delete_demultiplex_api.py
diitaz93: fixed

diitaz93:
> I like having the delete call here rather than before demultiplexing.
> 
> Here are some thoughts though:
> 
> * [ ]  Simplify `delete_demux_api.delete_flow_cell` to only perform what we think is useful
Wait for input from prod. This function is also called in the CLI command `cg demultiplex delete-flow-cell` and receives as parameters all of the desired locations for deletion.
> * [ ]  Since the call is moved after demultiplexing, it should not remove any directories.
so `demultiplexing_dir=False` and `run_dir=False`
> * [ ]  It should not remove the sample sheet
This one is hard, we will have to decouple the sample sheet and Housekeeper deletion if I understand correctly
> * [ ]  Due to the cascade setting between metrics and flow cell we do not explicitly need to remove metrics.
I guess this is for when you want to delete the metrics but not the flow cell?
> * [ ]  Do we need to check active samples? (more hassle than worth)
We do check active samples when we call `delete_flow_cell`

cg/cli/demultiplex/demux.py
```diff
@@ -121,66 +120,6 @@ def demultiplex_flow_cell(
...
-)
-@click.option(
-    "--status-db",
-    is_flag=True,
-    help="Delete flow cell in status-db, if passed all other entries are also deleted",
-)
-@click.option("--yes", is_flag=True, help="Pass yes to click confirm")
-@DRY_RUN
-@click.pass_obj
-def delete_flow_cell(
```
Vince-janv, line 150: I'm not sure this is ever used ATM

cg/meta/demultiplex/housekeeper_storage_functions.py
```diff
@@ -170,3 +173,23 @@ def add_sample_sheet_path_to_housekeeper(
         LOG.error(
             f"Sample sheet for flow cell {flow_cell_name} in {flow_cell_directory} was not found, error: {e}"
         )
+
+
+def delete_sequencing_data_from_housekeeper(flow_cell: str, hk_api: HousekeeperAPI) -> None:
```
diitaz93, line 178: Is it the id right?
```suggestion
def delete_sequencing_data_from_housekeeper(flow_cell_id: str, hk_api: HousekeeperAPI) -> None:
```

cg/meta/demultiplex/housekeeper_storage_functions.py
```diff
@@ -170,3 +173,23 @@ def add_sample_sheet_path_to_housekeeper(
...
+        {SequencingFileTag.SPRING, flow_cell},
+        {SequencingFileTag.SPRING_METADATA, flow_cell},
+    ]
+    for tags in tag_combinations:
+        housekeeper_files: Iterable[File] = hk_api.files(tags=tags)
+        for housekeeper_file in housekeeper_files:
+            hk_api.delete_file(file_id=housekeeper_file.id)
+
+
+def delete_sequencing_logs_from_housekeeper(flow_cell: str, hk_api: HousekeeperAPI) -> None:
```
diitaz93, line 190:
```suggestion
def delete_sequencing_logs_from_housekeeper(flow_cell_id: str, hk_api: HousekeeperAPI) -> None:
    """Delete sequencing log files associated with a flow cell from Housekeeper."""
```

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -125,3 +131,9 @@ def get_all_demultiplexed_flow_cell_dirs(self) -> list[Path]:
                 LOG.debug(f"Found directory {flow_cell_dir}")
                 demultiplex_flow_cells.append(flow_cell_dir)
         return demultiplex_flow_cells
+
+    def delete_flow_cell_data(self, parsed_flow_cell: FlowCellDirectoryData) -> None:
+        """Delete flow cell data from status db and housekeeper."""
+        delete_sequencing_metrics_from_statusdb(parsed_flow_cell.id, self.status_db)
+        delete_sequencing_data_from_housekeeper(parsed_flow_cell.id, self.hk_api)
+        delete_sequencing_logs_from_housekeeper(parsed_flow_cell.id, self.hk_api)
```
diitaz93, line 139: add named args as the function has more than one parameter

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -125,3 +131,9 @@ def get_all_demultiplexed_flow_cell_dirs(self) -> list[Path]:
                 LOG.debug(f"Found directory {flow_cell_dir}")
                 demultiplex_flow_cells.append(flow_cell_dir)
         return demultiplex_flow_cells
+
+    def delete_flow_cell_data(self, parsed_flow_cell: FlowCellDirectoryData) -> None:
```
diitaz93, line 135: Maybe you only need to give the id as an argument, not the complete flow cell
```suggestion
    def delete_flow_cell_data(self, parsed_flow_cell_id: str) -> None:
```

tests/meta/demultiplex/test_housekeeper_storage_functions.py
```diff
@@ -227,3 +231,93 @@ def test_store_fastq_path_in_housekeeper_correct_tags(
     file: File = populated_housekeeper_api.get_files(bundle=sample_id).first()
     expected_tags: set[str] = {SequencingFileTag.FASTQ.value, novaseq6000_flow_cell.id, sample_id}
     assert {tag.name for tag in file.tags} == expected_tags
+
+
+@pytest.mark.parametrize(
+    "file_tag",
+    [SequencingFileTag.FASTQ, SequencingFileTag.SPRING, SequencingFileTag.SPRING_METADATA],
```
diitaz93, line 238: Is parametrisation the correct approach here? I would say you need to check all three types of files every time you call the delete command. I would do a for loop over the file types with an assert statement inside in the THEN section.

cg/meta/demultiplex/demux_post_processing.py
```diff
@@ -82,6 +86,8 @@ def finish_flow_cell(
             LOG.warning(f"Flow cell {flow_cell_directory_name} will be skipped: {e}")
             return
 
+        self.delete_flow_cell_data(flow_cell)
```
diitaz93, line 89: Shouldn't this be after `store_flow_cell_data`?

cg/cli/demultiplex/demux.py
diitaz93: Shouldn't we leave this just in case? Maybe someone in prod uses it manually

cg/meta/demultiplex/demux_post_processing.py
karlnyr: The idea here is to clean before we add. @Vince-janv could there be a way to do the commit in one? What if the store fails here?

tests/cli/demultiplex/test_demultiplex_flowcell.py
```diff
@@ -216,61 +212,3 @@ def test_is_demultiplexing_complete(tmp_flow_cell_directory_bcl2fastq: Path):
...
-        in caplog.text
-    )
-    assert f"DeleteDemuxAPI-StatusDB: Would remove {bcl2fastq_flow_cell_id}" in caplog.text
-
-    assert (
-        "DeleteDemuxAPI-Hasta: Would have removed the following directory: "
-        f"{demultiplex_context.demultiplex_api.demultiplexed_runs_dir / Path(f'some_prefix_1100_{bcl2fastq_flow_cell_id}')}\n"
-        f"DeleteDemuxAPI-Hasta: Would have removed the following directory: {demultiplex_context.demultiplex_api.flow_cells_dir / Path(f'some_prefix_1100_{bcl2fastq_flow_cell_id}')}"
-    ) in caplog.text
-    assert "DeleteDemuxAPI-Init-files: Would have removed" not in caplog.text
```
karlnyr, line 276: 🤮

tests/store_helpers.py
```diff
@@ -60,6 +61,33 @@ def ensure_hk_bundle(store: HousekeeperAPI, bundle_data: dict, include: bool = F
...
+    @staticmethod
+    def quick_hk_bundle(
+        bundle_name, files: list[Path], store: HousekeeperAPI, tags: list[list[str]]
+    ):
+        """Adds a bundle to housekeeper with the given files and tags."""
+        bundle_data: dict = StoreHelpers.format_hk_bundle_dict(
+            bundle_name=bundle_name, files=files, all_tags=tags
+        )
+        return StoreHelpers.ensure_hk_bundle(store=store, bundle_data=bundle_data)
+
```
karlnyr, line 90: This is neat! This does need the order to be correct when passing the files and tags and kind of confused me when you used it in the test above as I was not certain which files got which tags. Would a more elaborate docstring help?

cg/cli/demultiplex/demux.py
karlnyr, line 1: Question, now that we delete before we ad, which is very nice, did we also adjust so that we can start with stuff being there? Rather, can we do start demultiplex and it would be able to start by force? I remember you asking about this but I am not sure if it was addressed.

karlnyr (commented):
Cool beans 🔥

cg/meta/demultiplex/housekeeper_storage_functions.py
Vince-janv: If ID is just HXXXXXX then yes 😅

cg/meta/demultiplex/demux_post_processing.py
Vince-janv: For statusDB yes, but for Housekeeper no... Right now I don't commit to status_DB before adding the new data. Maybe I should for consistency between the two databases?

tests/store_helpers.py
Vince-janv: Good point. Adding a more verbose doc string

cg/cli/demultiplex/demux.py
Vince-janv: Yes! I sneakily merged that PR yesterday 🙄 If there is no `demuxstarted.txt` in the flow cell dir we simply remove the output dir and start the demux

cg/cli/demultiplex/demux.py
Vince-janv: Can I create an issue for adding it?

Vince-janv: Question: would you want it to do the same as the automation in this PR, or even more?

cg/meta/demultiplex/demux_post_processing.py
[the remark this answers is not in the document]
karlnyr: or perhaps make the logs themselves have a timestamp?

cg/meta/demultiplex/demux_post_processing.py
karlnyr: It feels sketchy, but perhaps that could be a separate PR? This was the function previously as well

Vince-janv: Agreed that it is sketchy. Fixing it would require a re-write of the whole HouseKeeperAPI though 😬

tests/meta/demultiplex/test_housekeeper_storage_functions.py
Vince-janv: It's a valid approach but would take a lot more code. Keeping this

karlnyr (approved):
Clip it and ship it. Time for vacation 🌴

Vince-janv:
### Finishing from scratch (flow cell not added previously) also worked

Vince-janv:

PR #2596, opened 2023-10-16
Extract database module

seallard:
## Description
This PR resolves an issue with the previous refactoring. For some reason, the way we instantiated and passed around the sessions caused them to end up in an invalid state. This PR is almost identical, with the exception of using the `scoped_session` factory ([docs](https://docs.sqlalchemy.org/en/20/orm/contextual.html#implicit-method-access)) instead of explicitly instantiating the sessions ourselves.

In the end, we will likely not keep the session as an attribute on the store object and handlers since it is confusing. We are using the `scoped_session` pattern, which means we have a single global session per thread and request. The session can be retrieved directly where it is used.

### Test
Stress tested in stage

### Fixed
- session handling

### DIFF
--- a/cg/store/database.py
+++ b/cg/store/database.py
@@ -1,0 +1,61 @@
+from typing import List, Optional
+
+from sqlalchemy import create_engine, inspect
+from sqlalchemy.engine.base import Engine
+from sqlalchemy.engine.reflection import Inspector
+from sqlalchemy.orm import Session, scoped_session, sessionmaker
+
+from cg.exc import CgError
+from cg.store.models import Model
+
+SESSION: Optional[scoped_session] = None
+ENGINE: Optional[Engine] = None
+
+
+def initialize_database(db_uri: str) -> None:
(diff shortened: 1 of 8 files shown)


cg/store/database.py
```diff
@@ -0,0 +1,58 @@
...
+def get_session() -> scoped_session:
+    """Get a SQLAlchemy session with a connection to status db."""
+    if not SESSION:
+        raise CgError("Database not initialised")
+    return SESSION
+
+
+def get_scoped_session_registry() -> Optional[scoped_session]:
+    """Get the scoped session registry for status db."""
+    return SESSION
```
Vince-janv, line 32: Is there a reason for keeping this function since it will return the same as `get_session`?

Vince-janv (approved):
Nicely debugged! 👍

cg/store/database.py
seallard: The return types differ, I guess the other could be used and the exception caught. But I think I prefer this for now

PR #2598, opened 2023-10-17
Fix session merge warnings

seallard:
This PR fixes all warnings related to objects automatically being merged into a session, part of https://github.com/Clinical-Genomics/cg/issues/2497. For example
```
RemovedIn20Warning: 
"File" object is being merged into a Session along the backref cascade path for relationship "Version.files"; 
in SQLAlchemy 2.0, this reverse cascade will not take place.  Set cascade_backrefs to False in
either the relationship() or backref() function for the 2.0 behavior; or to set globally for the whole 
Session, set the future=True flag 
(Background on this error at: https://sqlalche.me/e/14/s9r1)
(Background on SQLAlchemy 2.0 at: https://sqlalche.me/e/b8d9)
    new_file.version = version
```

The warning says the following
```python
version = Version()
session.add(version)

file = File()
file.version = version  # <--- adds "file" to the session automatically, won't happen in 2.0
```
The warnings were generated by setting `export SQLALCHEMY_WARN_20=1`and running `pytest -W always`. Luckily, it seems like we did not rely on the automatic session add in that many spots.

The changes in this PR was made in the following way:
1. Generate warnings and store output in txt file `pytest -W always > out.txt`.
2. Pick an auto merge session warning.
3. Identify in which functions the warning originates.
4. Check each spot where the functions are called and verify that the object is explicitly added to the session.
5. Set `cascade_backrefs=False` on the relationship from which the warning originates to turn off the auto merge behaviour and silence it.
6. Re-run test suite and verify that the warning is gone and all tests are passing.

For some reason, `backref` was not behaving as expected. So I replaced all instances of it with `back_populates`. The difference between these is that `back_populates` needs to be explicitly defined on both models in the relationship (seems better to be explicit as well).

When we upgrade SQLAlchemy to 2.0, the `cascade_backrefs` option can be removed.

### DIFF
--- a/cg/store/database.py
+++ b/cg/store/database.py
@@ -20,7 +20,7 @@ def initialize_database(db_uri: str) -> None:
     SESSION = scoped_session(session_factory)
 
 
-def get_session() -> scoped_session:
+def get_session() -> Session:
     """Get a SQLAlchemy session with a connection to status db."""
     if not SESSION:
         raise CgError("Database not initialised")
--- a/cg/meta/orders/microbial_submitter.py
+++ b/cg/meta/orders/microbial_submitter.py
@@ -6,7 +6,14 @@ from cg.meta.orders.lims import process_lims
 from cg.meta.orders.submitter import Submitter
 from cg.models.orders.order import OrderIn
 from cg.models.orders.samples import MicrobialSample
-from cg.store.models import ApplicationVersion, Customer, Family, Organism, Sample
+from cg.store.models import (
+    ApplicationVersion,
+    Customer,
+    Family,
+    FamilySample,
+    Organism,
+    Sample,
+)
 
 
 class MicrobialSubmitter(Submitter):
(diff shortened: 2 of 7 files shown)


cg/store/database.py
```diff
@@ -20,7 +20,7 @@ def initialize_database(db_uri: str) -> None:
     SESSION = scoped_session(session_factory)
 
 
-def get_session() -> scoped_session:
+def get_session() -> Session:
```
seallard, line 23: It is a `scoped_session`, but it can be treated as a `Session` which is nice because then we get the proper syntax highlighting and auto completion as well 👍

Vince-janv (approved):
Do you see any point in me going through the warnings as well or are you confident? Everything in the code changes looks good 👍

seallard:
> Do you see any point in me going through the warnings as well or are you confident? Everything in the code changes looks good 👍

I have gone through it, but I don't feel super confident. However, there were very few instances where we actually relied on the deprecated behaviour. I think most of the previous developers were unaware of it and therefor always explicitly added objects anyway. So I think it is fine, you don't have to go through it

PR #2605, opened 2023-10-18
Patch errors

seallard:
## Description
This PR patches an issue in the `hk` module in cg which became apparent after merging https://github.com/Clinical-Genomics/housekeeper/pull/161. 

I missed one spot in the `hk` module where we relied on an implicit add, which in turned caused a bunch of tests to fail when housekeeper was updated to the most recent release.

### Fixed
- Patch due to missing add in the `hk` module

### DIFF
--- a/cg/apps/housekeeper/hk.py
+++ b/cg/apps/housekeeper/hk.py
@@ -110,6 +110,7 @@ class HousekeeperAPI:
         )
 
         new_file.version: Version = version_obj
+        self._store.session.add(new_file)
         return new_file
 
     def files(


cg/apps/housekeeper/hk.py
```diff
@@ -110,6 +110,7 @@ def add_file(
         )
 
         new_file.version: Version = version_obj
```
seallard, line 112: This used to rely on the `version_obj` being in the session and `new_file` being merged into the session as well.

diitaz93 (approved):
👍

islean (approved):
🔥

seallard:
This is kind of annoying, the pattern we have for where we add stuff is not great. Lets refine it once this migration is done.

PR #2606, opened 2023-10-18
Update validation cases

fevac:
### Changed

- Update rnafusion validation cases to avoid compression

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b add_validation_cases -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/constants/compression.py
```diff
@@ -112,21 +112,36 @@
...
+    "rareosprey",
+    "rightmoray",
+    "stablemoray",
+    "starnewt",
+    "tendergoose",
+    "truemole",
+    "truepigeon",
+    "valuedfowl",
+    "vocallocust",
+    "wantedsawfly",
```
seallard, line 144: Spontaneously, it seems like a bad idea to hard code a list with these generated id:s. Is it not possible to add a check on the case wherever the compression is done to see if it is a RNA fusion case instead?

fevac: I agree. I strongly dislike this way. 
To me, this information should be in statusDB with a column that is validation=bool or something similar. 
It's also extremely prone to be forgotten to update, like it keeps happening. 
If you have time and want to refactor this I'd be happy

seallard: Yes, I could refactor this. Seems like there is a bunch of other validation cases for other pipelines. Not sure if/how to prioritise it. Added an issue for it so it will hopefully be picked up at some point https://github.com/Clinical-Genomics/cg/issues/2607

diitaz93 (commented):
I think the correct place for this is in a txt or csv file in `cg/resources/`. This could be a good fix before the refactoring that SA proposes

fevac:
> I think the correct place for this is in a txt or csv file in `cg/resources/`, but I agree with the refactoring that SA proposes

I strongly vote for no files but a field in StatusDB

fevac:
- deployed to stage
image


- deployed to prod
image

PR #2609, opened 2023-10-18
Fix missed session warnings

seallard:
## Description
Missed a couple of warnings all originating from the `relate_sample` function which returns a `FamilySample`. The returned object is explicitly added to the session everywhere in the code base, but not in the test suite 😅  Hence the big diff.

I think it would make more sense to add the object to the session in the store layer, but changing that would require a lot of refactoring (especially in the test suite since these assume the `AddHandler` only creates the objects mostly).

### Fixed
- Missed warnings related to the deprecated auto merging into sessions


diitaz93 (approved):
👍

PR #2610, opened 2023-10-18
Containerize cg cli

seallard:
## Description
This PR adds a Dockerfile and a Github action workflow to build and push cli images to [Clinical Genomics dockerhub](https://hub.docker.com/r/clinicalgenomics/cg-cli). The image is pushed to the `clinicalgenomics/cg-cli` repository. For feature branches it is tagged with `<branch name>` and for releases it is tagged with `<version>` and `latest` (only one image will be tagged with latest). Images are built and pushed automatically on releases and pushes to feature branches.

The image can be used on Hasta like this
```
singularity run --bind /home/proj/stage:/home/proj/stage docker://clinicalgenomics/cg-cli:add-cli-container cg --config /home/proj/stage/servers/config/hasta.scilifelab.se/cg-stage.yaml add --help
```
This can be extracted to an alias, replacing the existing one for cg. The built image is tagged with the branch, like this `cg-cli:<branch-name>`.

Some benefits:
- We can test different versions of the CLI in parallel on the HPC (conflicts are possible when db migrations are involved, but should be rare)
- No need to run a deploy script on the HPC
- We cut down the execution time since the cli is bundled into one unit which improves read IO
- We can start moving away from conda
- More reproducible and portable setup

Verified that commands using the database and writing to the filesystem work as expected in the container.

### Added
- Dockerfile for running the CLI in a container
- Action to build and push the CLI image


seallard:
The cli needs to read and write to many directories on the Hasta file system. I guess the easiest way is just to mount the `home/proj/production` and `home/proj/stage` folders

seallard:
Using the pushed image on Hasta
```
[sebastian.allard@hasta:~] [base] 255 $ singularity run --bind /home/proj/stage/servers/config/hasta.scilifelab.se:/mnt docker://clinicalgenomics/cg-cli:add-cli-container cg --config /mnt/cg-stage.yaml add --help
INFO:    Converting OCI blobs to SIF format
INFO:    Starting build...
Getting image source signatures
Copying blob sha256:e67fdae3559346105027c63e7fb032bba57e62b1fe9f2da23e6fdfb56384e00b
 29.96 MiB / 29.96 MiB [====================================================] 0s
Copying blob sha256:a970ffaebef6fd07818bd49400c57484158e1665c80c51d70eac2080012d73ef
 1.03 MiB / 1.03 MiB [======================================================] 0s
Copying blob sha256:e855c2f1475c14c1b7fdeea72f25100921ec5a269ce576a713f6c00532b9a2d8
 11.44 MiB / 11.44 MiB [====================================================] 0s
Copying blob sha256:0668e0cb7eeb9ced793b8a4532206bfb3690455a5069af3a164704ba78b41026
 245 B / 245 B [============================================================] 0s
Copying blob sha256:b5251448cf317aac95e6cf71b4a19a6ed429da8a4b8ca1f0cabf0850e0429673
 3.25 MiB / 3.25 MiB [======================================================] 0s
Copying blob sha256:87dc407eb9ba6e610e9ebcfd72a6d79626875ccef9dcf32061edd1a7a98d61f4
 26.68 MiB / 26.68 MiB [====================================================] 0s
Copying blob sha256:cf6f5a96aa15c634e750d3ea1e5af1fef6f1ce6c9096fbdebeafb1092fd0ff29
 93 B / 93 B [==============================================================] 0s
Copying blob sha256:9cf33f3b315d68110cdcd6cd870d66a22765a1a9e3092f920468750fccd0070e
 24.93 MiB / 24.93 MiB [====================================================] 0s
Copying blob sha256:e682c81a33fc7246f05f9b268b6a87367152819a70455f80bc50559313057144
 99.16 MiB / 99.16 MiB [====================================================] 2s
Copying config sha256:477707184d8ea75da791f1a03ebdbf4811c5bdbd77672857799f09dfb6af0287
 7.71 KiB / 7.71 KiB [======================================================] 0s
Writing manifest to image destination
Storing signatures
INFO:    Creating SIF file...
INFO:    Build complete: /home/sebastian.allard/.singularity/cache/oci-tmp/f72a72836f252e6bf3cde24e3cbc469b58fe1334af93a12fb972c2333e2079a3/cg-cli_add-cli-container.sif
INFO:    Image cached as SIF at /home/sebastian.allard/.singularity/cache/oci-tmp/f72a72836f252e6bf3cde24e3cbc469b58fe1334af93a12fb972c2333e2079a3/cg-cli_add-cli-container.sif
Usage: cg add [OPTIONS] COMMAND [ARGS]...

  Add new things to the database.

Options:
  --help  Show this message and exit.

Commands:
  case          Add a case with the given name and associated with the...
  customer      Add a new customer with a unique internal id and name.
  external      Downloads external data from the delivery server and...
  external-hk   Adds external data to Housekeeper
  relationship  Create a link between a case id and a sample id.
  sample        Add a sample for CUSTOMER_ID with a NAME (display).
  user          Add a new user with an EMAIL (login) and a NAME (full).
```

diitaz93 (commented):
LGTM, I can't spot any obvious bugs but I lack a bit of expertise on these workflows. One question is if you would need any kind of root privileges to run and build the container.

seallard:
> LGTM, I can't spot any obvious bugs but I lack a bit of expertise on these workflows. One question is if you would need any kind of root privileges to run and build the container.

I don't think so, the container can be run with singularity on Hasta as a regular user. However, when I added a non root user in the Dockerfile stuff seems to break. Looking into it 👍

.github/workflows/build_cli_image_on_push.yml
```diff
@@ -0,0 +1,30 @@
...
+          password: ${{ secrets.DOCKER_PASSWORD }}
+
+      - name: Build and push
+        id: docker_cli_build
+        uses: docker/build-push-action@v3
+        with:
+          context: ./
+          file: ./Dockerfile.cli
+          push: true
+          tags: "clinicalgenomics/cg-cli:${{ env.branch }}"
```
Vince-janv, line 30: If we want to follow the same way we do with the other dockerfile we would have a cg-stage-cli repository which we publish to on push and then a cg-cli repository which is published to on release. But maybe we can push to the same ones we have but have `cli` as a tag? What do you think is best?

Vince-janv (commented):
Wow, you got this done so quickly! 🤩 Did you try modifying the databases via the container in stage? I'm guessing some port-mapping is needed for that?

seallard:
> Wow, you got this done so quickly! 🤩 Did you try modifying the databases via the container in stage? I'm guessing some port-mapping is needed for that?

Good point! I should try that and writing to the Hasta file system 👍

.github/workflows/build_cli_image_on_push.yml
seallard: The only difference between the stage and prod cli is the config file we point to at runtime. So I don't think it makes sense to have separate stage/prod images. I think it seems intuitive to keep the web api and cli images in separate repositories though, what do you think? It could also make sense to keep all images in one repository and tag them clearly.

Currently, I think the best approach is to use a single new repository called `cg-cli` like

Feature branches: clinicalgenomics/cg-cli:<branch-name>
Master branch: clinicalgenomics/cg-cli:<version> and clinicalgenomics/cg-cli:latest

Vince-janv: Sure I like that suggestion of a CLI repo, I don't think tags are supposed to be used the way I thought. But he same does hold true for the API though (ie no need for separate stage prod repos). Maybe we could merge those as well, while we're at it? @henrikstranneheim do you have any thoughts on this?

Vince-janv (approved):
Great work!👌

.github/workflows/build_publish_cli_image.yml
```diff
@@ -0,0 +1,58 @@
...
+            BRANCH_NAME=$(echo ${GITHUB_REF#refs/heads/} | tr "/" "-")
+            echo "::set-output name=tags::clinicalgenomics/cg-cli:$BRANCH_NAME"
+          fi
+
+      - name: Build and Push Docker Image
+        id: docker_cli_build
+        uses: docker/build-push-action@v3
+        with:
+          context: ./
+          file: ./Dockerfile.cli
```
henrikstranneheim, line 56:
```suggestion
          file: ./Dockerfile-cli
```

Dockerfile.cli
henrikstranneheim, line 1: Rename to Dockerfile-cli

henrikstranneheim (commented):
Looks great so far 👍

Dockerfile-cli
```diff
@@ -0,0 +1,16 @@
...
+    && adduser --system --ingroup nonroot nonroot
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get clean
+
+WORKDIR /app
+
+RUN git clone -b master https://github.com/Clinical-Genomics/cg.git /app
```
seallard, line 13: This won't work, we need to use the local files to pick up the changes in the feature branch

.github/workflows/build_cli_image_on_push.yml
seallard: Yes, that is how tagging is used here as well! So the conclusion is that we want to merge the `clinicalgenomics/cg` and `clinicalgenomics/cg-stage`? Does it affect this PR? I think we have to separate the images for the cg web api and cli images though since the Dockerfiles are different

henrikstranneheim (approved):
Awesome 👏🏻. Feels like the start of a paradigm shift 👍.

.github/workflows/build_cli_image_on_push.yml
Vince-janv: I agree with you @seallard. I think we should have two separate repos for CLI and API

seallard: Alright, I'm flip flopping between the options haha! I'll go for the separate `cg-cli` repo for now and we can adjust it in the future.

seallard:
This dockerfile is not usable yet in production due to the many binary paths in the config pointing to files on Hasta.

PR #2614, opened 2023-10-19
Fix relationship copy warning

seallard:
## Description
This PR fixes the following warning
```
tests/test_store_helpers.py::test_add_microbial_sample
  /Users/sebastianallard/repos/cg/cg/store/api/add.py:70: SAWarning: relationship 'Collaboration.customers' will copy column collaboration.id to column customer_collaboration.collaboration_id, which conflicts with relationship(s): 'Customer.collaborations' (copies collaboration.id to customer_collaboration.collaboration_id). If this is not the intention, consider if these relationships should be linked with back_populates, or if viewonly=True should be applied to one or more if they are read-only. For the less common case that foreign key constraints are partially overlapping, the orm.foreign() annotation can be used to isolate the columns that should be written towards.   To silence this warning, add the parameter 'overlaps="collaborations"' to the 'Collaboration.customers' relationship. (Background on this error at: https://sqlalche.me/e/14/qzyx)
    return Collaboration(internal_id=internal_id, name=name, **kwargs)
```

by explicitly setting `back_populates`. The two relationships are supposed to be opposite sides of the same relationship.

### Fixed
- Warning due to conflict between relationships


diitaz93 (approved):
💯

henrikstranneheim (approved):
👍

PR #2619, opened 2023-10-23
fix(demultiplexing )Remove demux dir before starting a new demux job

Vince-janv:
## Description
Before running demultiplexing the output directory needs to be removed in preparation for the new data.

This PR removes two pre-checks before demultiplexing and, if the command is not run with dry-run, also removes the output directory before submitting the new demultiplexing job.

With the new behaviour we will not check anything in the output directory before starting demultiplexing. We will rely on the the demuxstarted.txt file to decide wether to start demultiplexing (and removing the output dir).

### Added
- Method `prepare_output_directory` to both remove and create output directory

### Changed

- Output dir creation moved from `start_demultiplexing` to new function `prepare_output_directory`
- Method `create_demultiplexing_output_dir` now only takes one argument

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b delete-demux-dir
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


henrikstranneheim (approved):
👍

diitaz93 (approved):
👍

cg/cli/demultiplex/demux.py
```diff
@@ -59,6 +59,7 @@ def demultiplex_all(context: CGConfig, flow_cells_directory: click.Path, dry_run
             continue
 
         if not dry_run:
+            demultiplex_api.remove_demultiplexing_output_directory(flow_cell)
```
ChrOertlin, line 62: can we put this into the `demultiplex_api.start_demultiplexing(flow_cell=flow_cell)` function just before where we make the directory? 
Might even be worth considering to extract making and deleting the directory a separate function together. 
That function can check if exist, delete and create new.

Vince-janv: I'm would like to keep the deleting out of `start_demultiplexing` as it rather prepares before the start. But do you suggest extracting the directory creation out from `start_demultiplexing`?

ChrOertlin: My suggestion would: 1. make the CLI cleaner. 2. Keep logic regarding the demux dir together. The idea was to move the deletion/creation/existence check closer together instread of having one in the CLI and the other calls in the start demultiplexing. Can be outside or inside the start demux call.

Vince-janv: I created the method `prepare_output_directory` which groups together delete and create methods. Put it outside of the start command. Could you take a quick look?

ChrOertlin: Looks good to me"

Vince-janv:
Tests pass in stage

Vince-janv:
Deployed in production

PR #2624, opened 2023-10-24
Set delivery files as optional for rnafusion

fevac:
### Changed

- Files corresponding to the tags  fusioninspector, fusioninspector-html, and arriba-visualisation are now optional deliverables. If optional files are missing they are not added to the files-deliverables file.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b optional_delivery -a
    ```

### How to test

- [x] `cg workflow rnafusion report-deliver CASE_WITH_NO_FUSIONS`

### Expected test outcome

- [x] Check that deliverable-files is created and is missing non-existing files

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/workflow/nf_analysis.py
```diff
@@ -278,5 +280,8 @@ def get_deliverables_for_case(self, case_id: str) -> PipelineDeliverables:
                 file[deliverable_field] = file[deliverable_field].replace(
                     "PATHTOCASE", str(self.get_case_path(case_id=case_id))
                 )
```
islean, line 282: I am not allowed to comment on all lines apparently but a minor thing - would it be possible to have inner for-loop, i.e. lines 276-282 in its own method with a name akin to `add_case_id_and_path_to_template`? Took me a while to figure out why that was necessary and why we do it in this order 😅 Feel free to disregard if out of scope, though.

fevac:
tests are passing

cg/meta/workflow/nf_analysis.py
```diff
@@ -278,5 +280,8 @@ def get_deliverables_for_case(self, case_id: str) -> PipelineDeliverables:
...
+            if file["tag"] in optional_file_tags and not Path(file["path"]).exists():
```
diitaz93, line 283: Does the set of tags need to be included as a parameter of the function? If not, you could import the set with `from cg.constants.rnafusion import RNAFUSION_OPTIONAL_FILE_TAGS` and do
```suggestion
            if file["tag"] in RNAFUSION_OPTIONAL_FILE_TAGS and not Path(file["path"]).exists():
```
and you don't need to modify the signature of the function. Otherwise, you will have to import `RNAFUSION_OPTIONAL_FILE_TAGS` every time you use the function, while with this change you only need to import it once. Unless the tags will be different every time you call the function

islean (approved):
Looks nice!

henrikstranneheim:
Should this not be done in Hermes in the tag_map for RNAFusion using `is_mandatory`?

cg/meta/workflow/nf_analysis.py
```diff
@@ -266,7 +266,9 @@ def get_deliverables_template_content() -> list[dict]:
         """Return deliverables file template content."""
         raise NotImplementedError
 
-    def get_deliverables_for_case(self, case_id: str) -> PipelineDeliverables:
+    def get_deliverables_for_case(
```
diitaz93, line 269: Does this function. have tests? It would be good to include a test that checks that the correct files are skipped

fevac: included

cg/meta/workflow/nf_analysis.py
fevac: I will leave as it is

cg/meta/workflow/nf_analysis.py
fevac: added `_update_template_with_case_information`

fevac:
> Should this not be done in Hermes in the tag_map for RNAFusion using `is_mandatory`?

I guess ideally it would be nice to have it only in once place so it's not be redundant. Unfortunately, I'm not super familiar with Hermes and I thought it was mostly use to communicate with HK (so the step after writing the file_deliverables.yalm) 

The point was to not write a file_deliverables.yalm file containing missing files as it's quite misleading.

henrikstranneheim:
It's important that this

fevac:
@henrikstranneheim your previous message seems incomplete

fevac:
closing in favour of handling in hermes

henrikstranneheim:
Ok, not sure I understand. I think if set to is_mandory false, the file will be included if present and otherwise just ignored.

fevac:
There are two parts:
1. including files in the file_deliverables.yaml
2. including files in housekeeper

Hermes takes cares of 2

I wanted to avoid having "wrong information" (this is, including files that do not exists) in 1. But this is the part I don't know how to do without duplicating information. So we can leave 1 for now. We might be able to fix this later on once we add the nf-prov plugin in the pipeline (this will output published files)

henrikstranneheim (commented):
Sorry for the confusion I thought this was about storing, I guess that's about the delivery report...but maybe we should still use the deliverables file from Hermes as the truth of what can be delivered. However, this might require some functionality in cg.

Issue #2618, opened 2023-10-20
Broken priority value presented in Susy order summaries


Karl-Svard:
## Description

I've seen that the priority value given for samples in the Support System order summaries have been broken since the 16th.

See example:
![Screenshot 2023-10-20 at 15 32 30](https://github.com/Clinical-Genomics/cg/assets/60181709/00469e4d-8a1a-41e6-bda3-f768eeb9f21f)

The issue persists for all priority and sample types.

## Suggested solution

I suspect this will be a very quick fix. Just adding `.value` to the appropriate place.

## This can be closed when

Orders look correct again in Susy.

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

PR #2636, opened 2023-10-27
Fix enums

seallard:
## Description
This PR patches a breaking change in Python 3.11 related to enum mixins, see https://github.com/python/cpython/issues/100458. Closes https://github.com/Clinical-Genomics/cg/issues/2618.

I think this bug explains both the weird values seen in the ticketing service.

To resolve the issue, all instances of
```python
from cg.utils.enums import StrEnum
from cg.utils.enums import IntEnum
```
were replaced with
```python
from enum import StrEnum
from enum import IntEnum
```

It might be the case that some database tables need to be patched 🥲 

### Fixed
- Broken enums

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/constants/archiving.py
+++ b/cg/constants/archiving.py
@@ -1,4 +1,4 @@
-from cg.utils.enums import StrEnum
+from enum import StrEnum
 
 
 class ArchiveLocations(StrEnum):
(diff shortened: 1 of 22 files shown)


islean (approved):
🔥

seallard:
Tested https://clinical-scilifelab.supportsystem.com/scp/tickets.php?id=62788#reply

henrikstranneheim:
Good catch 🥇 . Now we should be able to remove`.value` safely for these enum classes.

PR #2637, opened 2023-10-27
Remove cascade backrefs parameter

seallard:
## Description
The behaviour in SQLAlchemy is that backrefs are not cascade loaded into the session. So we can remove the `cascade_backrefs=False` arguments from the relationships that were introduced in the migration process.

### Fixed
- Remove unnecessary parameter from SQLAlchemy relationships


henrikstranneheim (approved):
👍

PR #2642, opened 2023-10-30
Fix database model error

seallard:
## Description
When generating a revision, an error popped up
```
Alembic: sqlalchemy.exc.CompileError: VARCHAR requires a length on dialect mysql (Family.tickets)
```
This PR adds a length limit to the `Family.tickets` field which resolves this error.

Some changes detected when auto generating the revision are fixed in this PR as well. The enum fields are sensitive to ordering, which has changed for `family.action`, `family.data_delivery` and `flowcell.status`. Some of them have chaged from varchar to enum as well.

```
INFO  [alembic.autogenerate.compare] Detected NOT NULL on column 'application.min_sequencing_depth'
INFO  [alembic.autogenerate.compare] Detected type change from ENUM('analyze', 'running', 'hold') to Enum('analyze', 'hold', 'running') on 'family.action'
INFO  [alembic.autogenerate.compare] Detected type change from VARCHAR(length=64) to Enum('analysis', 'analysis-scout', 'fastq', 'fastq-scout', 'fastq_qc', 'fastq-analysis', 'fastq_qc-analysis', 'fastq-analysis-scout', 'nipt-viewer', 'no-delivery', 'scout', 'statina', name='datadelivery') on 'family.data_delivery'
INFO  [alembic.autogenerate.compare] Detected type change from ENUM('ondisk', 'processing', 'removed', 'requested', 'retrieved') to Enum('ondisk', 'removed', 'requested', 'processing', 'retrieved') on 'flowcell.status'
  Generating /Users/sebastianallard/repos/cg/alembic/versions/2023_10_30_b6f00cc615cf_sync_db.py ...  done
```

### Fixed
- Fix column error in Family model
- Sync models with schema

## Implementation Plan

- [ ] Apply migrations to stage database.
- [ ] Verify it is working by retrieving some data.
- [ ] Merge and apply to prod

### DIFF
--- a/alembic/versions/2023_10_30_b6f00cc615cf_sync_db.py
+++ b/alembic/versions/2023_10_30_b6f00cc615cf_sync_db.py
@@ -18,9 +18,6 @@ depends_on = None
 
 def upgrade():
     op.alter_column(
-        "application", "min_sequencing_depth", existing_type=mysql.INTEGER(), nullable=False
-    )
-    op.alter_column(
         "family",
         "action",
         existing_type=mysql.ENUM("analyze", "running", "hold"),
@@ -92,7 +89,4 @@ def downgrade():
         existing_type=sa.Enum("analyze", "hold", "running"),
         type_=mysql.ENUM("analyze", "running", "hold"),
         existing_nullable=True,
-    )
-    op.alter_column(
-        "application", "min_sequencing_depth", existing_type=mysql.INTEGER(), nullable=True
     )
--- a/alembic/versions/2023_10_30_b6f00cc615cf_sync_db.py
+++ b/alembic/versions/2023_10_30_b6f00cc615cf_sync_db.py
@@ -25,27 +25,6 @@ def upgrade():
         existing_nullable=True,
     )
     op.alter_column(
-        "family",
-        "data_delivery",
-        existing_type=mysql.VARCHAR(length=64),
-        type_=sa.Enum(
-            "analysis",
-            "analysis-scout",
-            "fastq",
-            "fastq-scout",
-            "fastq_qc",
-            "fastq-analysis",
-            "fastq_qc-analysis",
-            "fastq-analysis-scout",
-            "nipt-viewer",
-            "no-delivery",
-            "scout",
-            "statina",
-            name="datadelivery",
-        ),
-        existing_nullable=True,
-    )
-    op.alter_column(
         "flowcell",
         "status",
         existing_type=mysql.ENUM("ondisk", "processing", "removed", "requested", "retrieved"),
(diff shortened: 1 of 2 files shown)


alembic/versions/2023_10_30_b6f00cc615cf_sync_db.py
```diff
@@ -0,0 +1,98 @@
...
+revision = "b6f00cc615cf"
+down_revision = "392e49db40fc"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.alter_column(
+        "application", "min_sequencing_depth", existing_type=mysql.INTEGER(), nullable=False
+    )
```
islean, line 22: I think this will probably fail since there are multiple columns that are null though?

cg/store/models.py
```diff
@@ -387,7 +387,9 @@ class Family(Model, PriorityMixin):
     customer_id = Column(ForeignKey("customer.id", ondelete="CASCADE"), nullable=False)
     customer = orm.relationship(Customer, foreign_keys=[customer_id])
     data_analysis = Column(types.Enum(*list(Pipeline)))
-    data_delivery = Column(types.Enum(*list(DataDelivery)))
+    data_delivery = Column(
+        types.Enum(DataDelivery, values_callable=lambda obj: [e.value for e in obj])
+    )
```
islean, line 392: what does this difference do?

islean (approved):
Looks good! Think one part of the alembic script will crash though

cg/store/models.py
seallard: It is the same as before, but alembic was unable to autogenerate a revision for the reordering of the fields with the syntax that was used. This syntax does the same, it uses the values of the enum.

alembic/versions/2023_10_30_b6f00cc615cf_sync_db.py
seallard: Hm, yeah - there are null values in the database. The model should probably be updated to be nullable to reflect this.

seallard: I think it is out of scope for this PR, I'll create an issue for it and remove it from this revision

cg/store/models.py
seallard: Removed it

seallard:
Applied to stage
```
(cg-env) ➜  cg git:(fix-alembic) ✗ alembic --config ../alembic-stage.ini upgrade head                                     
INFO  [alembic.runtime.migration] Running upgrade 392e49db40fc -> b6f00cc615cf, Sync db
```

Issue #2558, opened 2023-10-05
Undetermined fastqs in flat bcl convert structure are not picked up


seallard:
## Description
The logic for retrieving undetermined fastq files for bcl convert currently assumes that the nested structure is always used. So if the new flat output structure is used, no undetermined fastq files will be found.

PR #2659, opened 2023-11-06
Retrieve undetermined fastqs in root for flow cells with flat output

seallard:
## Description
The old pattern does not pick up the undetermined fastq file paths for flow cells with a flat dir structure and demultiplexed with bclconvert. Closes https://github.com/Clinical-Genomics/cg/issues/2558

### Fixed
- Ensure undetermined fastq file paths are retrieved correctly for a lane, no matter flow cell output structure


cg/meta/demultiplex/utils.py
```diff
@@ -183,13 +183,19 @@ def get_sample_sheet(flow_cell: FlowCellDirectoryData) -> SampleSheet:
...
     """Get the undetermined fastq files for a specific lane on a flow cell."""
     undetermined_pattern = f"Undetermined*_L00{lane}_*{FileExtensions.FASTQ}{FileExtensions.GZIP}"
-    return get_files_matching_pattern(
-        directory=undetermined_dir_path,
+    undetermined_in_root: list[Path] = get_files_matching_pattern(
+        directory=flow_cell_path,
         pattern=undetermined_pattern,
     )
+    unaligned_path = Path(flow_cell_path, DemultiplexingDirsAndFiles.UNALIGNED_DIR_NAME)
+    undetermined_in_unaligned: list[Path] = get_files_matching_pattern(
```
ChrOertlin, line 194:
```suggestion
    unaligned_path = Path(flow_cell_path, DemultiplexingDirsAndFiles.UNALIGNED_DIR_NAME)
    undetermined_in_unaligned: list[Path] = get_files_matching_pattern(
```
I think this will be deprecated soon since bcl2fastq will be deprecated. BCLconvert does not use this.

ChrOertlin (approved):
Looks good! Not sure what to do with the `Unaligned_dir` as of now.

cg/meta/demultiplex/utils.py
seallard: Leaving it for now, but will keep it in mind for the deprecation

Issue #2658, opened 2023-11-06
Unwanted analyses starting after re-demultiplexing of old flow cells


Vince-janv:
## Description

The function `cg.store.api.status.StatusHandler.cases_to_analyze` is fetching cases we do not want to start. The query (amongst other criteria) fetches cases where a sample has a later `Reads Updated At` date than the latest Analysis date. 

When we re-demultiplex an old flow cell we may update the reads for other samples then the one we want to analyse resulting in unnecessary analyses starting. This is also confusing for production as they have to try to find the corresponding ticket.

These cases are re-uploaded to the customer causing confusion on their side as well.

### Example cases:
modernmullet
supercivet
justpug

## Suggested solution

The intention of `Reads Updated At` is to restart analyses when more data has been added compared to the existing analyses. I think this logic is good. However I think the date should not be `datetime.now()` and instead be the date where the sequencing was started i.e. 2023-11-06 in the case of `231106_A01901_0242_BXXXXXXXX`.

## This can be closed when

Only cases that we actually want to analyse are fetched by `cases_to_analyze`

seallard:
I'm not sure if the suggested solution is sufficient. Would not the problem remain since all samples will receive the flow cells new `sequenced_at` date - ensuring that they all would be fetched with `cases_to_analyze` anyway? 

That is how I understand the problem - that all samples have their `reads_updated_at` reset after re-demultiplexing when they actually have not received any new data.

It seems like a check that verifies that new reads actually have been added is necessary, and only then set the sample `reads_updated_at`. For instance comparing the new read count with the old. But this approach is problematic based on

1. demultiplexing is not deterministic
2. read count is a quantitative measure, the same read count does not mean that the actual reads are the same

Point one can be circumvented by using some error margin. Not sure about point 2.

I need to understand more about when re-demultiplexing is done and what input data is used.

Vince-janv:
> I'm not sure if the suggested solution is sufficient. Would not the problem remain since all samples will receive the flow cells new `sequenced_at` date - ensuring that they all would be fetched with `cases_to_analyze` anyway?

The flow cells `sequenced_at` date is also `datetime.now()` as you point out. I suggest using the actual date that the flow cell was run (ie disregarding the demultiplexing). So a flow cell from 2018 would have that date set.

> That is how I understand the problem - that all samples have their `reads_updated_at` set after re-demultiplexing when they actually have not received any new data.

[...]

Issue #2647, opened 2023-10-31
Reads Updated At behaviour


fevac:
## Description

- Currently some workflows automatically start when they get new reads
- The problem comes when flowcells are re-added or re-demux, then automatic start is also triggered even if there is no new reads and the analysis was already delivered
- It would be good to update the logic to avoid this from happening

This is related to: https://github.com/Clinical-Genomics/MTP-RNAFUSION/issues/45

## Suggested solution

- Maybe if the flowcell should be taken into account together with its `Sequenced at` date 
- Alternatively the reads could be checked, prior updating. If there is no new reads coming then the `Reads Updated At` date should not be modified
- Other ideas

## This can be closed when

*Describe what needs to be done for this issue to be closed*

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

seallard:
Same as https://github.com/Clinical-Genomics/cg/issues/2658 which has been resolved

PR #2660, opened 2023-11-06
Set reads_updated_at to sequencing date

seallard:
## Description
Closes https://github.com/Clinical-Genomics/cg/issues/2658.
This PR sets the sample `reads_updated_at` to the sequencing date of the flow cell to avoid always updating it when a flow cell is re-demultiplexed and triggering a re-analysis. 

Alternatively, we could remove the `reads_updated_at` field (as @islean suggested) from the Sample model and instead fetch it from the most recent flow cell on the case.

### Fixed
- Avoid triggering re-analyses unnecessarily by adjusting how the `last_sequenced_at` is set for samples
- Rename `reads_updated_at` to `last_sequenced_at`
- Refactor functions updating sample data to be more readable


islean (commented):
I think this will work but I still wonder we do not drop the `sample.sequenced_at` column and just keep it as a property in the Python model? Seems a bit unintuitive to me but maybe it is good for production to be able to see?

A bit out of scope if this is meant as a quick patch though.

cg/models/flow_cell/flow_cell.py
```diff
@@ -74,6 +74,12 @@ def full_name(self) -> str:
         """Return flow cell full name."""
         return self.path.name
 
+    @property
+    def sequenced_at(self) -> list[str]:
+        """Return the sequencing date for the flow cell."""
+        date_part: str = self.full_name[:6]
+        return datetime.datetime.strptime(date_part, "%y%m%d")
```
Vince-janv: Not sure how strptime works but it needs to work on both `231023_A00621_0927_BHXXXXXXXX` and `20231103_LH00188_0024_A2XXXXXXXX`

cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -148,6 +153,5 @@ def update_sample_read_count(sample_id: str, q30_threshold: int, store: Store) -
             f"Updating sample {sample_id} with read count {sample_read_count} and setting sequenced at."
         )
         sample.reads = sample_read_count
-        sample.reads_updated_at = datetime.datetime.now()
-    else:
+        sample.reads_updated_at = sequenced_at
```
Vince-janv, line 156: I think this field only makes sense if it is the latest sequencing date ie do not overwrite a more recent date. Maybe it would make sense to rename to column to `last_sequenced_at`

Vince-janv (approved):
Nice implementation, short and elegant!

cg/models/flow_cell/flow_cell.py
seallard: Yes, fixed that now 👍  Nice catch

cg/meta/demultiplex/status_db_storage_functions.py
seallard: I agree, I'll add a check for that and rename the field.

islean: Could we not just revert the PRs which changed this from "sequenced_at" from the beginning? Or would there be any change apart from "sequenced_at" being changed to "last_sequenced_at"? Or is the parsing of the directory name what makes the behaviour more in line with when it was sequenced?

Vince-janv: Not sure what you mean @islean. I think the naming of the column is secondary here. The more important change is that `reads_updated_at` should not be the time when the post-processing is run.

islean: Nevermind, I think I am not adding anything beneficial to the discussion! Just thought that we are approaching how it was before some changes a month or two back. But I don't think that should impact any decisions taken now.

seallard: Yeah, I think we have bounced a bit back and forth but this should hopefully be the last adjustment 🤞

seallard:
Applied revision to stage.

seallard:
Also closes https://github.com/Clinical-Genomics/cg/issues/2647

henrikstranneheim:
@seallard Was there a code smell left behind? If so can you patch it?

PR #2663, opened 2023-11-07
Patch view

seallard:
## Description
Missed to rename a parameter in the view layer which caused the `Analysis` and `FamilySample` views to break.

### Fixed
- Broken analysis and case views

### Test
Deploy to cg-vm and verify that the analysis and family sample views work and correctly link to the case.


islean (approved):
Good catch!

PR #2669, opened 2023-11-07
Rename family sample

seallard:
## Description
This PR renames the FamilySample ~table~ model to CaseSample and any references to it.

### Fixed
- Rename FamilySample
- Bonus: rename family parameter on relate_sample

### How to test

- [x] Do ensure all the views work as expected

### DIFF
--- a/cg/store/models.py
+++ b/cg/store/models.py
@@ -400,7 +400,7 @@ class Case(Model, PriorityMixin):
     tickets = Column(types.VARCHAR(128))
 
     analyses = orm.relationship(Analysis, back_populates="case", order_by="-Analysis.completed_at")
-    links = orm.relationship("FamilySample", back_populates="case")
+    links = orm.relationship("CaseSample", back_populates="case")
 
     @property
     def cohorts(self) -> list[str]:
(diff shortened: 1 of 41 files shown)


cg/store/models.py
```diff
@@ -515,7 +515,7 @@ def to_dict(self, links: bool = False, analyses: bool = False) -> dict:
         return data
 
 
-class FamilySample(Model):
+class CaseSample(Model):
     __tablename__ = "family_sample"
```
ChrOertlin, line 519:
```suggestion
    __tablename__ = "case_sample"
```

cg/store/models.py
```diff
@@ -515,7 +515,7 @@ def to_dict(self, links: bool = False, analyses: bool = False) -> dict:
...
     __table_args__ = (UniqueConstraint("family_id", "sample_id", name="_family_sample_uc"),)
```
ChrOertlin, line 520:
```suggestion
    __table_args__ = (UniqueConstraint("case_id", "sample_id", name="_case_sample_uc"),)
```

ChrOertlin (approved):
Looks good! Just a thing about database table names, if not required yet you can ignore them!

cg/store/models.py
seallard: Not remaining the table stuff yet, my PR description was a bit off

PR #2684, opened 2023-11-13
Increase case id range

seallard:
## Description
Apparently we are running out of unique id:s for our cases. This PR increases the range of possible case ids by increasing the char length for them.

### Fixed
- Increase number of possible case id:s


cg/store/api/add.py
```diff
@@ -204,14 +204,7 @@ def add_case(
...
 
-        # generate a unique case id
-        while True:
-            internal_id = petname.Generate(2, separator="")
-            if self.get_case_by_internal_id(internal_id) is None:
-                break
-            else:
-                LOG.debug(f"{internal_id} already used - trying another id")
-
+        internal_id: str = self.generate_unique_petname()
```
islean, line 207: I think this function is used for samples currently and introduces the possibility that cases share internal ids with other cases. This would probably mean a crash at a later stage due to integrity errors when adding to the DB.

I think we should have a different method for generating petnames for samples and cases (either we change some parameter or use some separator, e.g. underscores or use some other way to assign petnames to samples). If we use this method, we should have a check for clashing ids with other cases though.

islean (commented):
Nice initiative but I think this has a risk of introducing some very rare bugs which might be hard to sniff out.

cg/store/api/add.py
seallard: Ah, yeah - I missed that the function makes the check against tha sample table 👍

cg/store/api/add.py
```diff
@@ -34,10 +34,16 @@
 class AddHandler(BaseHandler):
     """Methods related to adding new data to the store."""
 
-    def generate_unique_petname(self) -> str:
+    def generate_readable_sample_id(self) -> str:
         while True:
```
ChrOertlin, line 38: Can we have a docstring here, for user comprehension, that this is not part of the typical flow to give sample ids petnames. I was extremely confused by this just now

cg/store/api/add.py
```diff
@@ -34,10 +34,16 @@
...
             random_id = petname.Generate(3, separator="")
```
henrikstranneheim, line 39:
```suggestion
            random_id: str = petname.Generate(3, separator="")
```

cg/store/api/add.py
```diff
@@ -34,10 +34,16 @@
...
-            if not self.get_sample_by_internal_id(internal_id=random_id):
+            if not self.get_sample_by_internal_id(random_id):
+                return random_id
+
+    def generate_readable_case_id(self) -> str:
+        while True:
+            random_id = petname.Generate(2, separator="", letters=10)
```
henrikstranneheim, line 45:
```suggestion
            random_id: str = petname.Generate(2, separator="", letters=10)
```

cg/store/api/add.py
seallard: I don't have any insight as to what the normal flow is, me touching this function is pretty accidental. I'm guessing sample id:s are usually created in lims? Do you have a suggestion for a dcostring?

ChrOertlin: I think the tricky part here is that it would set a petname for the sample internal id, which we apparently sometimes do in development. 

```suggestion
    def generate_readable_sample_id(self) -> str:
    """Generates a petname as sample internal id for develepment purposes. Not used in normal production flow."""
        while True:
```

Something along these lines so in future this wont make it into the flow in some way.

seallard:
I guess this is not important

cg/store/api/add.py
```diff
@@ -34,10 +34,16 @@
...
-    def generate_unique_petname(self) -> str:
+    def generate_readable_sample_id(self) -> str:
         while True:
-            random_id = petname.Generate(3, separator="")
-            if not self.get_sample_by_internal_id(internal_id=random_id):
+            random_id: str = petname.Generate(3, separator="")
+            if not self.get_sample_by_internal_id(random_id):
+                return random_id
+
+    def generate_readable_case_id(self) -> str:
```
Vince-janv, line 43: Nice to extract this into it's own function

Vince-janv (approved):
Looks good 👍 Do we have any length restrictions in the database, that we need to update?

PR #2705, opened 2023-11-23
Skip lims data retrieval if downsampled for rnafusion report

fevac:
## Description

- Skip lims data retrieval if sample is down sampled for rnafusion report generation

### Changed

- Skip lims sample metadata collection for report generation if down sampled

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b skip_lims_for_report -a
    ```

### How to test

- [x] cg generate delivery-report --force DOWNSAMPLED_CASE_ID

### Expected test outcome

- [x] Check that delivery report has generated correctly with NA values for rin and input amount

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/report/rnafusion.py
```diff
@@ -38,11 +38,20 @@ def get_sample_metadata(
...
         """Return sample metadata to include in the report."""
         sample_metrics: RnafusionQCMetrics = analysis_metadata.sample_metrics[sample.internal_id]
+
+        # Skip LIMS data collection if down sampled
+        if sample.downsampled_to:
+            input_amount = None
+            rin = None
+        else:
+            input_amount = self.lims_api.get_latest_rna_input_amount(sample_id=sample.internal_id)
+            rin = self.lims_api.get_sample_rin(sample_id=sample.internal_id)
```
ChrOertlin, line 48:
```suggestion
        
            input_amount = None
            rin = None
        if not sample.downsampled_to:
            input_amount = self.lims_api.get_latest_rna_input_amount(sample_id=sample.internal_id)
            rin = self.lims_api.get_sample_rin(sample_id=sample.internal_id)
```

ChrOertlin (approved):
👍

diitaz93 (approved):
Neat 💯

ivadym (commented):
I think this will still fail, we also have to bypass `get_lims_sample` that's used by `get_data_analysis_type` and `get_samples_data`.

I would say that the proper way of doing this would be allowing `get_lims_sample` return None when there is an HTTPError, and then change the logic downstream to be able to handle None values. In this way you would also fix it for the rest of the pipelines.

fevac:
> I would say that the proper way of doing this would be allowing `get_lims_sample` return None when there is an HTTPError, and then change the logic downstream to be able to handle None values. In this way you would also fix it for the rest of the pipelines.

I see where you're going and I partially agree. I think it's useful to have a more general way, however I would not rely on the HTTPError, as there could be errors in other ways and then we might risk bypassing the collection of lims information by mistake. I would be specific about it. So I would say the best would be to modify the methods you say, but instead of checking the HTTPError,  either:
1. adding a lims_api method that returns true or false if sample exists in lims and if false then return None
2. only returning none if sample is downsampled or external

What you think @ivadym ?

ivadym:
> > I would say that the proper way of doing this would be allowing `get_lims_sample` return None when there is an HTTPError, and then change the logic downstream to be able to handle None values. In this way you would also fix it for the rest of the pipelines.
> 
> I see where you're going and I partially agree. I think it's useful to have a more general way, however I would not rely on the HTTPError, as there could be errors in other ways and then we might risk bypassing the collection of lims information by mistake. I would be specific about it. So I would say the best would be to modify the methods you say, but instead of checking the HTTPError, either:
> 
> 1. adding a lims_api method that returns true or false if sample exists in lims and if false then return None
> 2. only returning none if sample is downsampled or external
> 
> What you think @ivadym ?

The HTTPError is what we get when there is no data in LIMS, so the first point is what I'm actually suggesting. And we have already that method (`get_lims_sample` which calls `lims_sample.sample`). Doesn't matter so much where you catch the HTTPError I would say.

ivadym:
We can also implement this in another PR and go for this partial solution for now :)

ivadym:
I don't think it matters, the upload will still fail as we have an independent logic to check for missing fields in the delivery report. And we should keep relying on that, allowing None values and failing when they are required.

fevac:
I do think it matters because then we need to rely on additional logic that is not needed if we are specific here. I think there is an important distinction between "there are not samples in lims" and "we could not retrieve samples from lims". And it can be misleading in the future and could potentially lead to errors. 

Anyways, since this partial works for rnafusion. I will merge and we can revisit a more general solution in the future

fevac:
- Deployed in stage
image

- Deployed prod:
image

Issue #2722, opened 2023-11-30
Remove required -r option from cg workflow balsamic


fevac:
## Description

Currently balsamic runs as dry-run by default, unless added `-r` flag. This is contrary to all other workflows and confusing. It would be nice to homogenise. 

If changes are made they need to be reflected in the automation and atlas documentation. 

## Suggested solution

*Write here if there is a suggested way to solve the issue*

## This can be closed when

*Describe what needs to be done for this issue to be closed*

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

seallard:
No mention of the `-r` or `--run-analysis` in the servers repo. It seems like automation runs `start-available` which in turn calls `start` which no longer takes the `--run-analysis` option.

seallard:
I'll leave removing the flag from the balsamic code base to someone more familiar with it.

PR #2723, opened 2023-11-30
Remove balsamic r flag

seallard:
## Description
This PR replaces the `-r --run-analysis` with the dry run flag. Closes https://github.com/Clinical-Genomics/cg/issues/2722.

The `-r` flag ensured that slurm jobs only were run for balsamic with the `--run-analysis` option if it was provided to the cli initiating the balsamic run. It seems like this is no longer desired, the dry run flag now replaces this behavior. So the `--run-analysis` will not be used for the balsamic run if dry run is provided.

### Fixed
- Replace the `-r` flag with just the dry run flag for balsamic


fevac (commented):
❤️  yes, this looks like what we want to me. Note that merging this has to be sync with cron/systemd

cg/cli/workflow/balsamic/base.py
```diff
@@ -213,7 +207,6 @@ def start(
         run,
         case_id=case_id,
         slurm_quality_of_service=slurm_quality_of_service,
-        run_analysis=run_analysis,
```
ivadym, line 216: Since we'll start using --dry-run instead of --r, I would remove it from the config_case invocation above to have the same behaviour (we want the config to be created but don't run the analysis).

ivadym (commented):
Nice ⚡ 💯  Left you a tiny comment so we have the same expected behaviour with the dry-run as before

ivadym (approved):
Tests look fine in Stage 🥳 👍 

We need to update automation and atlas documentation. I will do the latter in this PR: https://github.com/Clinical-Genomics/atlas/pull/2167

seallard:
> Tests look fine in Stage 🥳 👍
> 
> We need to update automation and atlas documentation. I will do the latter in this PR: [Clinical-Genomics/atlas#2167](https://github.com/Clinical-Genomics/atlas/pull/2167)

I check the servers repo and from what I can see we only use the `start available` cli command to queue balsamic jobs (which does not use the --run-analysis flag). So I do not think any update is necessary. Unless I've missed something @ivadym

ivadym:
Yes, you're right we had the `run_analysis` flag set to `True` in start available, so the logic was in CG 👍

Issue #2676, opened 2023-11-09
External samples have no reads on the sample Model and are therefore not delivered properly


Vince-janv:
## Description

When uploading files to caesar the samples are checked to see if their read count is ok before delivering. External samples have 0 reads and are therefore skipped.

Excerpt from log:

```shell
Linked 13 files for case modellouse
Sample ACC13213A1 did not receive enough reads and will not be delivered
Sample ACC13213A2 did not receive enough reads and will not be delivered
Sample ACC13213A3 did not receive enough reads and will not be delivered
```

## Suggested solution

Set the accepted reads for external apptags to 0

## This can be closed when

When external samples can be delivered

PR #2727, opened 2023-12-01
Enable delivery of external samples

seallard:
## Description
This PR enables external samples to be delivered. Closes https://github.com/Clinical-Genomics/cg/issues/2676.

The `_is_external` field on the sample model was marked as deprecated (why has it not been removed?). The property on the sample model also states

```python
    @property
    @deprecated(
        version="1.4.0",
        message="This field is deprecated, use sample.application_version.application.is_external",
    )
    def is_external(self):
        """Return if this is an externally sequenced sample."""
        return self._is_external
```
So we should remove this... Will do it in a separate PR. There are no references to the field in cg.

### Changed
- Enable delivery of external samples

### How to test
- [ ] Do deliver external sample without reads

### Expected test outcome
- [ ] Files are uploaded


henrikstranneheim (approved):
👍

Vince-janv (approved):
Nice, fix! I thought it would be harder 😅

seallard:
> Regarding the deprecated field I think we can remove it but we would have to go through the database to make sure every sample marked as external has a correct application tag

Created an issue for it https://github.com/Clinical-Genomics/cg/issues/2738

PR #2758, opened 2023-12-11
Add column to determine if eligible for skip qc

islean:
## Description

As part of the ongoing developments to let some customers bypass our qc, we need to keep track of which applications are allowed. The ones currently allowed are described here https://github.com/Clinical-Genomics/project-planning/issues/485#issuecomment-1845032498. This PR adds a column to the Application table which shows whether the application is eligible for skipping qc.

### Added

- New `is_eligible_for_skip_qc` column to the Application table.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b add-eligible-column -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/alembic/versions/2023_12_11_79454ff320bf_add_eligible_for_skip_qc.py
+++ b/alembic/versions/2023_12_11_79454ff320bf_add_eligible_for_skip_qc.py
@@ -1,0 +1,27 @@
+"""add-eligible-for-skip-qc
+
+Revision ID: 79454ff320bf
+Revises: b105b426af99
+Create Date: 2023-12-11 15:02:40.593762
+
+"""
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "79454ff320bf"
+down_revision = "b105b426af99"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "application",
+        sa.Column("is_eligible_for_skip_qc", sa.Boolean(), nullable=False),
+    )
+
+
+def downgrade():
+    op.drop_column("application", "is_eligible_for_skip_qc")
--- a/cg/server/admin.py
+++ b/cg/server/admin.py
@@ -86,6 +86,7 @@ class ApplicationView(BaseView):
         "sample_concentration",
         "priority_processing",
         "is_archived",
+        "is_eligible_for_skip_qc",
     ]
     column_exclude_list = [
         "minimum_order",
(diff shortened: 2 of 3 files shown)


alembic/versions/2023_12_11_79454ff320bf_add_eligible_for_skip_qc.py
```diff
@@ -0,0 +1,27 @@
...
+revision = "79454ff320bf"
+down_revision = "b105b426af99"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "application",
+        sa.Column("is_eligible_for_skip_qc", sa.Boolean(), nullable=False),
```
seallard, line 22: Might be necessary to provide a default value to the column here as well?

seallard (approved):
:+1:  Maybe a default value needs to be supplied in the revision for the new column

alembic/versions/2023_12_11_79454ff320bf_add_eligible_for_skip_qc.py
seallard: Apparently not

diitaz93: but maybe a good idea to set it to False as default, not sure.

diitaz93 (commented):
Looks good, but when are the values updated? Is it going to be a separate PR?

alembic/versions/2023_12_11_79454ff320bf_add_eligible_for_skip_qc.py
islean: A bit unsure what it would add, given that the script has already been tested and added the column with False as default.

islean:
> Looks good, but when are the values updated? Is it going to be a separate PR?

It won't be an automatic thing, rather that lab or production need to be able to set this manually when adding new Applications. Or at least that is my understanding currently, need to check whether this solution is satisfactory.

henrikstranneheim (approved):
👍

islean:
Upon discussion, this will start out as the customer's responsibility. So for now there is no reason to add this functionality.

Issue #2743, opened 2023-12-07
add class AnalysisSampleTags


fevac:
I have seen these tags also in MIP. Maybe we can start a
`class AnalysisSampleTags(StrEnum):`
?

_Originally posted by @ChrOertlin in https://github.com/Clinical-Genomics/cg/pull/2621#discussion_r1418969121_

PR #2759, opened 2023-12-11
rnafusion HK tags as constants

fevac:
## Description

Closes https://github.com/Clinical-Genomics/cg/issues/2743

### Changed

- Use constants for rnafusion HK tags instead of strings

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b delivery_constants -a
    ```

### How to test

- [ ] `cg upload -c RNAFUSION_CASE`. Make sure it's successful

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/constants/delivery.py
+++ b/cg/constants/delivery.py
@@ -1,7 +1,7 @@
 """Constants for delivery."""
 
 from cg.constants.constants import Pipeline
-from cg.constants.housekeeper_tags import AlignmentFileTag
+from cg.constants.housekeeper_tags import AlignmentFileTag, AnalysisTag, HK_DELIVERY_REPORT_TAG
 
 ONLY_ONE_CASE_PER_TICKET: list[Pipeline] = [
     Pipeline.FASTQ,
--- a/cg/constants/delivery.py
+++ b/cg/constants/delivery.py
@@ -115,12 +115,12 @@ MIP_RNA_ANALYSIS_CASE_TAGS: list[set[str]] = [
 ]
 
 MIP_RNA_ANALYSIS_SAMPLE_TAGS: list[set[str]] = [
-    {"fusion", "star-fusion"},
-    {"fusion", "arriba"},
+    {AnalysisTag.FUSION, AnalysisTag.STARFUSION},
+    {AnalysisTag.FUSION, AnalysisTag.ARRIBA},
     {AlignmentFileTag.CRAM},
     {AlignmentFileTag.CRAM_INDEX},
-    {"fusion", "vcf"},
-    {"fusion", "vcf-index"},
+    {AnalysisTag.FUSION, "vcf"},
+    {AnalysisTag.FUSION, "vcf-index"},
     {"salmon-quant"},
 ]
 
--- a/cg/constants/delivery.py
+++ b/cg/constants/delivery.py
@@ -1,6 +1,7 @@
 """Constants for delivery."""
 
 from cg.constants.constants import Pipeline
+from cg.constants.housekeeper_tags import AlignmentFileTag
 
 ONLY_ONE_CASE_PER_TICKET: list[Pipeline] = [
     Pipeline.FASTQ,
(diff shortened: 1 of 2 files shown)


diitaz93 (approved):
Looks good 🙌

cg/constants/delivery.py
```diff
@@ -116,8 +117,8 @@
 MIP_RNA_ANALYSIS_SAMPLE_TAGS: list[set[str]] = [
     {"fusion", "star-fusion"},
     {"fusion", "arriba"},
-    {"cram"},
-    {"cram-index"},
+    {AlignmentFileTag.CRAM},
+    {AlignmentFileTag.CRAM_INDEX},
     {"fusion", "vcf"},
     {"fusion", "vcf-index"},
     {"salmon-quant"},
```
islean, line 124: Would be nice to update these with the new tags as well.

islean (approved):
Looks good! A bit unsure what the scope is - should all instances be switched or is that for a separate PR in some future?

ivadym (commented):
I'm not particularly keen on adding these constants, mainly because they make it more difficult to visualise which set of tags we are delivering and further expand the problem of having the tag logic in two different places, Hermes and CG.

Perhaps this issue needs to be further refined:
- Are these constants file-specific? I don't think so. Should they be? I don't know. Maybe having constants for file names that are assigned a bunch of tags isn't a bad idea.
- Why are we  having this logic on both sides (Hermes and CG) and not using the `deliver` tag that we add in Hermes?
- If we use the `deliver` tag, how do we differentiate between `deliver-sample` and `deliver-case`? Perhaps that's what we need to actually fix and deliver using only those two tags.

fevac:
> I'm not particularly keen on adding these constants, mainly because they make it more difficult to visualise which set of tags we are delivering and further expand the problem of having the tag logic in two different places, Hermes and CG.
@ivadym This pr only replaces strings by constants and I don't think the use of constants make it more difficult to visualize which set of tags are used. It's the opposite to me and it also makes it easier to pinpoint tags that should be the same. I agree that the duplication between hermes and cg is a problem and should be revisited but this PR doesn't expand the problem further.

> Perhaps this issue needs to be further refined:
> * Are these constants file-specific? I don't think so. Should they be? I don't know. Maybe having constants for file names that are assigned a bunch of tags isn't a bad idea.
Some are some are not. I don't think how this is relevant for this.

> * Why are we  having this logic on both sides (Hermes and CG) and not using the `deliver` tag that we add in Hermes?
I don't know and I fully agree. But that is 100% out of the scope of this PR

> * If we use the `deliver` tag, how do we differentiate between `deliver-sample` and `deliver-case`? Perhaps that's what we need to actually fix and deliver using only those two tags.
Good point


I do agree duplication between cg and hermes should be addressed, but in the meantime. Should I leave this as constant or strings? I don't mind which way. I do see benefit of removing strings that are redundant but I don't really care one way or another.

ivadym:
> Some are some are not. I don't think how this is relevant for this.

For me it would make more sense to define files as constants that contain a set of tags. But as you said, some are file specific, some are not. And it's out of scope of this PR 🙃 

> I do agree duplication between cg and hermes should be addressed, but in the meantime. Should I leave this as constant or strings? I don't mind which way. I do see benefit of removing strings that are redundant but I don't really care one way or another.

I think you should proceed with it. Although it doesn’t add any additional semantic meaning it will help to keep our coding style consistent

cg/constants/delivery.py
fevac: done

fevac:
tests are passing

fevac:
Deployed in stage

image


Deployed in prod

image

Issue #2768, opened 2023-12-14
MIP workflow start fails for specific case


seallard:
## Description
As reported by @peterpru, the following command fails `cg workflow mip-dna start savedcoyote` with this error
![image](https://github.com/Clinical-Genomics/cg/assets/33596072/5b35e3f5-30cc-440a-b9fd-2114c5f5a5e2)

## Suggested solution
We are modifying a set while iterating over it, which is not allowed. Refactor it.

## This can be closed when
The case can be started as intended.

PR #2769, opened 2023-12-14
Fix bug causing mip workflow start to fail

seallard:
## Description
Closes https://github.com/Clinical-Genomics/cg/issues/2768.

### Fixed
- Bug due to modifying a set of panels while iterating over it

### How to test

- [ ] Do `cg workflow mip-dna start savedcoyote`


islean (approved):
Good catch!

henrikstranneheim:
@seallard Good catch 👍

PR #2789, opened 2023-12-28
feat(Nanopore) Re-write create manifest command to support Nanopore structure

Vince-janv:
## Description
Due to some differences in the file structures of ONT and Illumina output directories the command had to be re-written slightly.

### Added

- Check for the ONT "Sequencing finished"-file

### Changed

- The directory iteration now uses glob instead if iterdir, to support wildcards in the input


cg/meta/demultiplex/utils.py
```diff
@@ -206,11 +207,18 @@ def parse_manifest_file(manifest_file: Path) -> list[Path]:
...
-        if relevant_directory in file.parts:
-            return True
-    return False
+    irrelevant_directories = [
+        "Thumbnail_Images",
+        ".tmp.cbcl",
+        ".cbcl.tmp",
+        "MyRun",
+        "tmp.filter",
+        "CopyComplete.txt",
```
Vince-janv, line 216: Not sure if these should be constants, Since we per definition fin them irrelevant... 😅

cg/meta/demultiplex/utils.py
```diff
@@ -206,11 +207,18 @@ def parse_manifest_file(manifest_file: Path) -> list[Path]:
 
 def is_file_relevant_for_demultiplexing(file: Path) -> bool:
```
Vince-janv, line 208: Reversed the logic for this function, make sure that any missing files are indeed irrelevant

cg/constants/nanopore_files.py
```diff
@@ -0,0 +1,19 @@
...
+    def sequencing_summary_file(cls, flow_cell_directory: Path) -> Path | None:
+        try:
+            file = Path(next(flow_cell_directory.glob(cls.sequencing_summary_pattern.value)))
+        except StopIteration:
+            file = None
+        return file
+
+    @classmethod
+    def sequencing_complete(cls, flow_cell_directory: Path) -> bool:
+        return bool(cls.sequencing_summary_file(flow_cell_directory))
```
Vince-janv, line 19: Maybe weird to have this in an Enum class. Any suggestions on how to create a constant for a file containing the flow cell in the name?

cg/cli/demultiplex/demux.py
```diff
@@ -178,6 +179,7 @@ def confirm_flow_cell_sync(context: CGConfig, source_directory: str):
     help="The path from where the syncing is done.",
 )
 def create_manifest_files(source_directory: str):
-    for source_flow_cell in Path(source_directory).iterdir():
-        if is_manifest_file_required(source_flow_cell):
-            create_manifest_file(source_flow_cell)
+    """Creates a file manifest for each flow cell in the source directory."""
+    for source_flow_cell in glob(f"{source_directory}/*"):
```
ChrOertlin, line 183:
```suggestion
    for source_flow_cell in get_directories_in_path(Path: Path):

```
does already exist for this

cg/constants/nanopore_files.py
```diff
@@ -0,0 +1,22 @@
...
+class NanoporeDirsAndFiles(StrEnum):
+    """Holds file and directory names for nanopore sequencing output."""
+
+    data_directory: str = "nanopore"
+    sequencing_summary_pattern: str = r"final_summary_*.txt"
+
+    @classmethod
+    def sequencing_summary_file(cls, flow_cell_directory: Path) -> Path | None:
+        try:
+            file = Path(next(flow_cell_directory.glob(cls.sequencing_summary_pattern.value)))
```
ChrOertlin, line 15: I think we have a utils function for this as well
```suggestion
            file = get_files_matching_pattern(directory: Path, pattern: str)

File: cg/constants/nanopore_files.py [hunk 6]
```diff
@@ -0,0 +1,22 @@
+"""Constants related to Oxford Nanopore sequencing."""
+from enum import StrEnum
+from pathlib import Path
+
+
+class NanoporeDirsAndFiles(StrEnum):
+    """Holds file and directory names for nanopore sequencing output."""
+
+    data_directory: str = "nanopore"
+    sequencing_summary_pattern: str = r"final_summary_*.txt"
```
ChrOertlin, line 10:
```suggestion
    DATA_DIRECTORY: str = "nanopore"
    SEQUENCING_SUMMARY_PATTERN: str = r"final_summary_*.txt"
```

cg/constants/nanopore_files.py
islean: Given that it is not really used as an enum (?), maybe it makes more sense to have it as a BaseModel with the Path and bool as fields and an init method setting them? Haven't read through the whole thing so I might be misinterpreting it.

ChrOertlin: You can have a function that generates an f string with certain input parameters I suppose.

ChrOertlin (approved):
Looks good, some comments about likely already implemented functionality in the utils/files.py

diitaz93 (approved):
Looks great 👍 I am a fan of keeping the logic outside the constant modules but do as you consider more convinient

cg/cli/demultiplex/demux.py
Vince-janv: Can't use wildcards in it unfortunately 😢

cg/constants/nanopore_files.py
Vince-janv: Moved the two class methods to utils for now

cg/constants/nanopore_files.py
Vince-janv: Also does not allow for wildcards in `directory`

cg/constants/nanopore_files.py
Vince-janv: Done 👍

cg/cli/demultiplex/demux.py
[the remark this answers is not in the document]
Vince-janv: For the Nanopore data I want to have a wildcard in `source_directory`. For pathlib.Path this is not supported

cg/constants/nanopore_files.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @classmethod
+    def sequencing_summary_file(cls, flow_cell_directory: Path) -> Path | None:
```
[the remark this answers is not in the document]
Vince-janv: Moved to utils for now. Might move to a better suited file if we restructure

cg/meta/demultiplex/utils.py
Vince-janv: No complaints from peers. Keeping as is

Vince-janv:

PR #2794, opened 2024-01-02
Parallelize tests

seallard:
## Description
Run tests in parallel using as many threads as there are cores available for the runner. The macos-13 runner which has 4 cores finishes the tests in less than 2 min, which is about half of what it usually takes. The ubuntu runner is only a bit slower despite only having 2 cores, if we rather use that one.

### Fixed
- Parallelize test execution in CI pipeline


.github/workflows/tests_and_coverage.yml
```diff
@@ -8,7 +8,7 @@ on:
 
 jobs:
   tests-coverage:
-    runs-on: ubuntu-latest
+    runs-on: macos-13
```
islean, line 11: I think we want this as close to hasta and the vms as possible, i.e. CentOS(?) or similar.

islean (approved):
Nice initiative! I think we want to keep it to run on Linux though.

.github/workflows/tests_and_coverage.yml
seallard: Changed it to ubuntu, there are no runners with centos

PR #2796, opened 2024-01-03
Refactor delivery api

seallard:
## Description
TLDR: the code used to be alphabet soup and is now spaghetti.

The code is still next to illegible, but a step in a better direction. Pretty sure that the refactoring has introduced bugs so will need testing, saving it until after your feedback. Needs more refactoring but I got bored.

In the context of this module, a delivery seems to mean copying files from a bundle to the customers directory on Hasta. Which files that are copied depends on the workflow and data delivery. Sounds simple but the code makes it seem like freaking rocket science :sweat_smile: 

## Current refactoring
- Remove ticket/case/tags etc. from instantiation of the delivery API
- Instantiate the delivery API once in the context
- Remove unnecessary logic
- Extract duplicated logic to methods
- Extract logic to smaller methods
- Sprinkle some doc strings and type hints on it

## Future refactoring
- Retrieve the data delivery and workflow from the case, see https://github.com/Clinical-Genomics/cg/pull/2796#discussion_r1442683856
- Trash/rewrite messy logic in utils and delivery api
- Improve clarity by adding extra methods for both sample and case to extract the final list of destination file paths. It will remove the version logic from here, which takes into account if the delivery is fastq or analysis. And it will also allow re-usability, extracting the complete list of delivered files.

### Fixed
- Refactor delivery api

### Testing on hasta
- Test delivering a single case and a ticket for each workflow


cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        version: Version,
+        sample_id: str,
+        pipeline: str,
+    ) -> Iterable[Path]:
+        """Fetch files for a sample from a version that are tagged with sample tags."""
+        file_obj: File
+        for file_obj in version.files:
+            if not include_file_sample(file=file_obj, sample_id=sample_id, pipeline=pipeline):
+                continue
+            yield Path(file_obj.full_path)
```
Vince-janv, line 207: I think this `yield` is confusing. I see no reason why this should not just return `list[Path]`

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+                out_path = Path(delivery_base, file_name)
+                if create_link(source=file_path, destination=out_path, dry_run=self.dry_run):
+                    number_linked_files += 1
+
+            if not files and number_linked_files == 0:
+                raise MissingFilesError(f"Could not find any files for sample {sample_id}")
+
+    def _create_sample_delivery_directory(self, case: Case, sample: Sample, pipeline: str) -> Path:
+        case_name: str | None = get_delivery_case_name(case=case, pipeline=pipeline)
+        delivery_base = get_delivery_dir_path(
```
Vince-janv, line 154: same here regarding variable name. And let's give it a type hint

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        if not self.dry_run:
+            LOG.debug(f"Creating project path {delivery_base}")
+            delivery_base.mkdir(parents=True, exist_ok=True)
+        return delivery_base
+
+    def _get_deliverable_samples(self, case: Case, pipeline: str) -> Iterable[CaseSample]:
+        return filter(
+            lambda link: self._sample_is_deliverable(link=link, case=case, pipeline=pipeline),
+            self.store.get_case_samples_by_case_id(case.internal_id),
+        )
```
Vince-janv, line 170: I kind of like this 😁

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        if not version.files:
+            LOG.warning(f"No files associated with Housekeeper version {version.id}")
+            return []
+
+        version_file: File
+        for version_file in version.files:
+            if not include_file_case(file=version_file, sample_ids=sample_ids, pipeline=pipeline):
+                LOG.debug(f"Skipping file {version_file.path}")
+                continue
+            yield Path(version_file.full_path)
```
Vince-janv, line 194: Same here regarding the yield

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        last_version = self.hk_api.last_version(case.internal_id)
+        if not last_version:
+            case_tags = get_case_tags_for_pipeline(pipeline)
+            skip_missing = pipeline in constants.SKIP_MISSING or self.ignore_missing_bundles
+            if not case_tags:
+                LOG.info(f"Could not find any version for {case.internal_id}")
+            elif not skip_missing:
+                raise SyntaxError(f"Could not find any version for {case.internal_id}")
+            return False
+        return self._case_has_samples(case)
```
Vince-janv, line 219: There should never be cases with no samples. I went and had a look in the database and any such case found was some sort of manual error. So we can remove `_case_has_samples`. We could then remove one indentation level on the rest. Maybe there are more improvements bu this is already better.
```suggestion
    def _case_is_deliverable(self, case: Case, pipeline: str) -> bool:
        last_version = self.hk_api.last_version(case.internal_id)
        if last_version:
            return True
        case_tags = get_case_tags_for_pipeline(pipeline)
        skip_missing = pipeline in constants.SKIP_MISSING or self.ignore_missing_bundles
        if not case_tags:
            LOG.info(f"Could not find any version for {case.internal_id}")
        elif not skip_missing:
            raise SyntaxError(f"Could not find any version for {case.internal_id}")
        return False
```

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    return file.name.replace(sample.internal_id, sample.name)
+
+
+def include_file_case(file: File, sample_ids: set[str], pipeline: str) -> bool:
+    """Check if file should be included in case bundle.
+
+    At least one tag should match between file and tags.
+    Do not include files with sample tags.
+    """
+    file_tags = {tag.name for tag in file.tags}
```
Vince-janv, line 78: Maybe clarifies a bit. Maybe not
```suggestion
    tags_on_file = {tag.name for tag in file.tags}
```

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    case_tags = get_case_tags_for_pipeline(pipeline)
+    all_case_tags: set[str] = {tag for tags in case_tags for tag in tags}
+    if all_case_tags.isdisjoint(file_tags):
+        LOG.debug("No tags are matching")
+        return False
```
Vince-janv, line 83: Not sure we need this 🤔 If this is the case we return False in the end either way

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    For fastq delivery we know that we want to deliver all files of bundle.
+    """
+    file_tags = {tag.name for tag in file.tags}
+    tags: set[str]
+    # Check if any of the file tags matches the sample tags
+    sample_tags = get_sample_tags_for_pipeline(pipeline)
+    for tags in sample_tags:
+        working_copy = deepcopy(tags)
+        if pipeline != "fastq":
+            working_copy.add(sample_id)
```
Vince-janv, line 118: With the work that @diitaz93 and @islean last autumn we can skip this weird check 🥳

Vince-janv (approved):
Wow! Herculean work 💪 Already such a big improvement 🌟

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+    For fastq delivery we know that we want to deliver all files of bundle.
+    """
+    file_tags = {tag.name for tag in file.tags}
+    tags: set[str]
+    # Check if any of the file tags matches the sample tags
+    sample_tags = get_sample_tags_for_pipeline(pipeline)
+    for tags in sample_tags:
+        working_copy = deepcopy(tags)
+        if pipeline != "fastq":
```
islean, line 117: Use constant

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    for tags in case_tags:
+        LOG.debug(f"check if {tags} is a subset of {file_tags}")
+        if tags.issubset(file_tags):
+            return True
+    LOG.debug(f"Could not find any tags matching file {file.path} with tags {file_tags}")
+
+    return False
+
+
+def include_file_sample(file: File, sample_id: str, pipeline: str) -> bool:
```
islean, line 103:
```suggestion
def should_include_file_sample(file: File, sample_id: str, pipeline: str) -> bool:
```

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+def _get_case_out_file_name(file: Path, case: Case) -> str:
+    return file.name.replace(case.internal_id, case.name)
+
+
+def get_sample_out_file_name(file: Path, sample: Sample) -> str:
+    return file.name.replace(sample.internal_id, sample.name)
+
+
+def include_file_case(file: File, sample_ids: set[str], pipeline: str) -> bool:
```
islean, line 72:
```suggestion
def should_include_file_case(file: File, sample_ids: set[str], pipeline: str) -> bool:
```

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    if dry_run:
+        LOG.info(f"Would hard link file {source} to {destination}")
+        return True
+    try:
+        os.link(source, destination)
+        LOG.info(f"Hard link file {source} to {destination}")
+        return True
+    except FileExistsError:
+        LOG.info(f"Path {destination} exists, skipping")
+        return False
```
islean, line 134: I don't think this should return a boolean

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        deliver_failed_samples: bool = self.deliver_failed_samples
+        sample_passes_qc: bool = link.sample.sequencing_qc
+        sample_is_deliverable: bool = (
+            sample_passes_qc or deliver_failed_samples or sample_is_external
+        )
+        last_version: Version = self.hk_api.last_version(case.internal_id)
+        if pipeline == DataDelivery.FASTQ:
+            last_version = self.hk_api.last_version(link.sample.internal_id)
+        if not last_version:
+            skip_missing = pipeline in constants.SKIP_MISSING or self.ignore_missing_bundles
```
islean, line 238: Finding skip_missing very confusing here as well.

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    """Returns the scope of the delivery, ie whether sample and/or case files were delivered."""
+    case_delivery: bool = False
+    sample_delivery: bool = False
+    for delivery in delivery_arguments:
+        if (
+            constants.PIPELINE_ANALYSIS_TAG_MAP[delivery]["sample_tags"]
+            and delivery in constants.ONLY_ONE_CASE_PER_TICKET
+        ):
+            sample_delivery = True
+        if constants.PIPELINE_ANALYSIS_TAG_MAP[delivery]["case_tags"]:
```
islean, line 23:
```suggestion
            get_sample_tags_for_pipeline(delivery)
            and delivery in constants.ONLY_ONE_CASE_PER_TICKET
        ):
            sample_delivery = True
        if get_case_tags_for_pipeline(delivery):
```

islean (commented):
Looks good! Will come back and look at the tests!

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+    def set_dry_run(self, dry_run: bool):
+        self.dry_run = dry_run
+
+    def set_deliver_failed_samples(self, deliver_failed_samples: bool):
+        self.deliver_failed_samples = deliver_failed_samples
+
+    def set_ignore_missing_bundles(self, ignore_missing_bundles: bool):
+        self.ignore_missing_bundles = ignore_missing_bundles
+
+    def deliver(self, ticket: str | None, case_id: str | None, pipeline: str) -> None:
```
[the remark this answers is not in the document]
seallard: This sounds like a design problem. I think that this method should only accept a case id or ticket id by default. Then we retrieve the pipeline and data analysis from the case.

We can expose a separate method for the edge case where production wants to override the data analysis set for the case/deliver some specific analysis for the case.

Does that make sense?

cg/meta/deliver/delivery_api.py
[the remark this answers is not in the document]
islean: Although I agree with @Vince-janv that it is a very neatly written method 💪

cg/meta/deliver/delivery_api.py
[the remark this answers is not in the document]
islean: I like what you suggested! Since data delivery is set on case level, it shouldn't be required by default. I guess there might be reuploads and stuff done manually which can benefit from an override though.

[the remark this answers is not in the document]
Vince-janv: It very much does make sense. As you say we have all the information in the case and it would be nice to just aggregate the tags and call deliver once instead of looping over delivery_types.

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+def get_sample_tags_for_pipeline(pipeline: Pipeline) -> list[set[str]]:
+    return constants.PIPELINE_ANALYSIS_TAG_MAP[pipeline]["sample_tags"]
+
+
+def get_delivery_case_name(case: Case, pipeline: str) -> str | None:
+    return None if pipeline in constants.ONLY_ONE_CASE_PER_TICKET else case.name
+
+
+def get_out_path(out_dir: Path, file: Path, case: Case) -> Path:
```
ChrOertlin, line 59: is this on hasta or on ceasar?

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    out_file_name: str = _get_case_out_file_name(file=file, case=case)
+    return Path(out_dir, out_file_name)
+
+
+def _get_case_out_file_name(file: Path, case: Case) -> str:
```
ChrOertlin, line 64:
```suggestion
def _rename_case_out_file_name(file: Path, case: Case) -> str:
```

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    return file.name.replace(case.internal_id, case.name)
+
+
+def get_sample_out_file_name(file: Path, sample: Sample) -> str:
```
ChrOertlin, line 68:
```suggestion
def rename_sample_out_file_name(file: Path, sample: Sample) -> str:
```

cg/meta/deliver/utils.py
ChrOertlin: a bit more clear what the function does

cg/meta/deliver/utils.py
ChrOertlin, line 134: I think this would better reside within the cg/utils/files.py module as it can be a more general function which we can re-use elsewhere I believe at some point we even worked on a hard/soft link function elsewhere in cg.

cg/models/cg_config.py
```diff
@@ -479,3 +481,16 @@ def trailblazer_api(self) -> TrailblazerAPI:
             api = TrailblazerAPI(config=self.dict())
             self.trailblazer_api_ = api
         return api
+
+    @property
+    def delivery_api(self) -> DeliveryAPI:
```
ChrOertlin, line 486: nice!!!

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        if not case:
+            LOG.warning(f"Could not find case {case_id}")
+            return
+        self.deliver_files(case=case, pipeline=pipeline)
+
+    def deliver_files(self, case: Case, pipeline: str):
+        if not self._case_is_deliverable(case=case, pipeline=pipeline):
+            return
+        self._deliver_case_files(case=case, pipeline=pipeline)
+        self._deliver_sample_files(case=case, pipeline=pipeline)
```
ChrOertlin, line 83:
```suggestion
        if self._case_is_deliverable(case=case, pipeline=pipeline):
            self._deliver_case_files(case=case, pipeline=pipeline)
            self._deliver_sample_files(case=case, pipeline=pipeline)
```

ChrOertlin (approved):
Massive work it is a super clean CLI now :) 💯

cg/cli/deliver/base.py
```diff
@@ -67,48 +67,21 @@ def deliver_analysis(
...
-            cases.append(case_obj)
-        else:
-            cases: list[Case] = status_db.get_cases_by_ticket_id(ticket_id=ticket)
-            if not cases:
-                LOG.warning(f"Could not find cases for ticket {ticket}")
-                return
-
-        for case_obj in cases:
-            deliver_api.deliver_files(case_obj=case_obj)
+        delivery_api.deliver(case_id=case_id, ticket=ticket, pipeline=delivery)
```
diitaz93, line 84: Another thought: Would you consider including the for loop inside the delivery API, similar to what is done in the demux post-processing API?

diitaz93 (commented):
This is already a huge improvement, well done 💯 I will give another review later 👍

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+    case_name: str = None,
+    sample_name: str = None,
+) -> Path:
+    """Get a path for delivering files.
+
+    Note that case name and sample name needs to be the identifiers sent from customer.
+    """
+    delivery_path = Path(base_path, customer_id, constants.INBOX_NAME, ticket)
+    if case_name:
+        delivery_path = delivery_path / case_name
```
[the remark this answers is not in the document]
islean: Very true!

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        self.deliver_failed_samples = deliver_failed_samples
+
+    def set_ignore_missing_bundles(self, ignore_missing_bundles: bool):
+        self.ignore_missing_bundles = ignore_missing_bundles
+
+    def deliver(self, ticket: str | None, case_id: str | None, pipeline: str) -> None:
+        if ticket:
+            self._deliver_files_by_ticket(ticket=ticket, pipeline=pipeline)
+        elif case_id:
+            self._deliver_files_by_case(case_id=case_id, pipeline=pipeline)
```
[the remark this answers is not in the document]
islean: Well in this case I think the risk of the docstring being misleading or outdated outweighs the benefit. But I guess it's a matter of taste.

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        store: Store,
+        hk_api: HousekeeperAPI,
+        project_base_path: Path,
+        force_all: bool = False,
+        ignore_missing_bundles: bool = False,
+        dry_run: bool = False,
+    ):
+        self.store = store
+        self.hk_api = hk_api
+        self.project_base_path = project_base_path
```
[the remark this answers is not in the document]
seallard: Yes, also - it should not be passed to the constructor.

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        self.dry_run = dry_run
+
+    def set_dry_run(self, dry_run: bool):
+        self.dry_run = dry_run
+
+    def set_deliver_failed_samples(self, deliver_failed_samples: bool):
+        self.deliver_failed_samples = deliver_failed_samples
+
+    def set_ignore_missing_bundles(self, ignore_missing_bundles: bool):
+        self.ignore_missing_bundles = ignore_missing_bundles
```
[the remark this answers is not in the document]
seallard: Yes, we should not have getters/setters. That is not how Python works or should be used

cg/meta/deliver/delivery_api.py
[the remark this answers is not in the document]
seallard: I don't want a docstring here

cg/cli/deliver/base.py
```diff
@@ -67,48 +67,21 @@ def deliver_analysis(
...
 
-    inbox: str = context.delivery_path
-    if not inbox:
-        LOG.info("Please specify the root path for where files should be delivered")
-        return
+    delivery_api: DeliveryAPI = context.delivery_api
+
+    delivery_api.set_dry_run(dry_run)
+    delivery_api.set_deliver_failed_samples(force_all)
+    delivery_api.set_ignore_missing_bundles(ignore_missing_bundles)
```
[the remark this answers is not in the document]
seallard: Removed the setters

cg/cli/deliver/base.py
seallard: I'll leave this to limit the scope of the refactoring

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+
+    def deliver(self, ticket: str | None, case_id: str | None, pipeline: str) -> None:
+        if ticket:
+            self._deliver_files_by_ticket(ticket=ticket, pipeline=pipeline)
+        elif case_id:
```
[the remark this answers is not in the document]
seallard: No idea, I've tried to not change the existing logic - just extract it to functions

cg/meta/deliver/delivery_api.py
seallard: I prefer returning early to not introduce unnecessary indentation

cg/meta/deliver/delivery_api.py
seallard: You and me both, the logic is extremely unclear. I think I would just end up lying if I tried to sort out what this mess is supposed to be doing

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        )
+        last_version: Version = self.hk_api.last_version(case.internal_id)
+        if pipeline == DataDelivery.FASTQ:
+            last_version = self.hk_api.last_version(link.sample.internal_id)
+        if not last_version:
+            skip_missing = pipeline in constants.SKIP_MISSING or self.ignore_missing_bundles
+            if skip_missing:
+                LOG.info(f"Could not find any version for {link.sample.internal_id}")
+                return False
+            raise SyntaxError(f"Could not find any version for {link.sample.internal_id}")
```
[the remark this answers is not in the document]
seallard: I did not chose anything, I'm refactoring existing code. I'll leave it to limit the scope of the changes introduced

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+def get_case_tags_for_pipeline(pipeline: Pipeline) -> list[set[str]]:
+    return constants.PIPELINE_ANALYSIS_TAG_MAP[pipeline]["case_tags"]
+
+
+def get_sample_tags_for_pipeline(pipeline: Pipeline) -> list[set[str]]:
+    return constants.PIPELINE_ANALYSIS_TAG_MAP[pipeline]["sample_tags"]
+
+
+def get_delivery_case_name(case: Case, pipeline: str) -> str | None:
+    return None if pipeline in constants.ONLY_ONE_CASE_PER_TICKET else case.name
```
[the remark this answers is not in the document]
seallard: You and me both, but that is the old logic.

cg/meta/deliver/utils.py
seallard: Pretty sure it is hasta

cg/meta/deliver/delivery_api.py
seallard: I think it is cleaner to indent it like this since it is only a few lines 👍

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        if not self._case_is_deliverable(case=case, pipeline=pipeline):
+            return
+        self._deliver_case_files(case=case, pipeline=pipeline)
+        self._deliver_sample_files(case=case, pipeline=pipeline)
+
+    def _deliver_case_files(self, case: Case, pipeline: str) -> None:
+        LOG.debug(f"Deliver case files for {case.internal_id}")
+
+        if not get_case_tags_for_pipeline(pipeline):
+            return
```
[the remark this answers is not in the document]
seallard: Maybe, depends on whether a version can exist for a case without tags

cg/cli/upload/clinical_delivery.py
```diff
@@ -30,31 +29,22 @@ def upload_clinical_delivery(context: click.Context, case_id: str, dry_run: bool
...
+    case: Case = context.obj.status_db.get_case_by_internal_id(case_id)
     delivery_types: set[str] = case.get_delivery_arguments()
     is_sample_delivery: bool
     is_case_delivery: bool
     is_complete_delivery: bool
     job_id: int
-    is_sample_delivery, is_case_delivery = DeliverAPI.get_delivery_scope(
-        delivery_arguments=delivery_types
-    )
+    is_sample_delivery, is_case_delivery = get_delivery_scope(delivery_types)
```
seallard, line 38: Why do we distinguish a sample delivery and a case delivery? Are not samples always in a case and do we not always deliver the files for all samples in a case?

cg/meta/deliver/delivery_api.py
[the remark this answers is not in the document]
seallard: Follow up: can we simplify the logic in this module even further?

A delivery in the context of this module seems to mean: copy the files of a bundle version (the analysis) into the customers folder on Hasta. 

Do we really need to care about the analysis or data delivery at all to do this? Or is it because we need to filter the files stored in the bundle version?

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+
+        deliverable_samples = self._get_deliverable_samples(case=case, pipeline=pipeline)
+        case_name: str | None = get_delivery_case_name(case=case, pipeline=pipeline)
+        for link in deliverable_samples:
+            delivery_base: Path = self._create_sample_delivery_directory(
+                case=case, sample=link.sample, pipeline=pipeline
+            )
+
+            number_linked_files: int = 0
```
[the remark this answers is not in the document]
seallard: That would change the behavior of the code. 
Instead of processing other samples first, it will raise an exception as soon as a sample without files is found.
I'll try to clean it up

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+            create_link(source=file_path, destination=out_path, dry_run=self.dry_run)
+
+    def _get_sample_ids_for_case(self, case: Case) -> set[str]:
+        links = self.store.get_case_samples_by_case_id(case.internal_id)
+        samples: list[Sample] = [link.sample for link in links]
+        sample_ids: set[str] = {sample.internal_id for sample in samples}
+        return sample_ids
+
+    def _create_delivery_directory(self, case: Case) -> Path:
+        delivery_base = get_delivery_dir_path(
```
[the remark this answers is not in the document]
seallard: That is not true, the function is called get_delivery_dir_path.
But we have a standard of type hinting everything, added it

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+        LOG.debug(f"Deliver case files for {case.internal_id}")
+
+        if not get_case_tags_for_pipeline(pipeline):
+            return
+
+        version: Version = self.hk_api.last_version(case.internal_id)
+        out_dir: Path = self._create_delivery_directory(case)
+        self._link_case_files(case=case, version=version, pipeline=pipeline, out_dir=out_dir)
+
+    def _link_case_files(self, case: Case, version: Version, pipeline: str, out_dir: Path):
```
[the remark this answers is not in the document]
seallard: I'll add one once the code is refactored and more clear

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+            out_path: Path = get_out_path(out_dir=out_dir, file=file_path, case=case)
+            create_link(source=file_path, destination=out_path, dry_run=self.dry_run)
+
+    def _get_sample_ids_for_case(self, case: Case) -> set[str]:
+        links = self.store.get_case_samples_by_case_id(case.internal_id)
+        samples: list[Sample] = [link.sample for link in links]
+        sample_ids: set[str] = {sample.internal_id for sample in samples}
+        return sample_ids
+
+    def _create_delivery_directory(self, case: Case) -> Path:
```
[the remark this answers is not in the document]
seallard: I don't agree, the name of function is `create_delivery_directory` and the type hint is `Path`

cg/meta/deliver/utils.py
seallard: Leaving it here for now, if it is needed elsewhere it can be moved at that tpoint

seallard: Other logic depends on it returning a bool, but I agree. Leaving it for now

cg/meta/deliver/utils.py
seallard: Which check? The fastq pipeline if statement? Or the entire sample tag copying shenanigans?

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def get_sample_out_file_name(file: Path, sample: Sample) -> str:
+    return file.name.replace(sample.internal_id, sample.name)
+
+
+def include_file_case(file: File, sample_ids: set[str], pipeline: str) -> bool:
+    """Check if file should be included in case bundle.
+
+    At least one tag should match between file and tags.
```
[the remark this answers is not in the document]
seallard: Unclear to me too, leaving it until refactoring is done. Might be clearer later

cg/meta/deliver/utils.py
seallard: Dict object being constructed, type hint not necessary

cg/meta/deliver/delivery_api.py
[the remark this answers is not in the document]
seallard: Fixed 👍

cg/meta/deliver/utils.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def get_delivery_scope(delivery_arguments: set[str]) -> tuple[bool, bool]:
+    """Returns the scope of the delivery, ie whether sample and/or case files were delivered."""
+    case_delivery: bool = False
+    sample_delivery: bool = False
+    for delivery in delivery_arguments:
+        sample_delivery: bool = is_sample_delivery(delivery)
+        case_delivery: bool = is_case_delivery(delivery)
+    return sample_delivery, case_delivery
```
seallard, line 19: Is it not really weird that we iterate over all "delivery arguments", but we set the scope to that of the last delivery argument? Does this mean that we actually only ever provide a single delivery argument?

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,245 @@
...
+                file_name: str = get_sample_out_file_name(file=file_path, sample=link.sample)
+                if case_name:
+                    file_name: str = file_name.replace(case.internal_id, case.name)
+                out_path = Path(delivery_base, file_name)
+                if create_link(source=file_path, destination=out_path, dry_run=self.dry_run):
+                    number_linked_files += 1
+
+            if not files and number_linked_files == 0:
+                raise MissingFilesError(f"Could not find any files for sample {sample_id}")
+
```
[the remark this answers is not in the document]
seallard: This would change the codes behaviour

cg/meta/deliver/delivery_api.py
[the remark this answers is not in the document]
seallard: Leaving this for a separate PR

cg/meta/deliver/utils.py
seallard: "Delivery argument" means workflow (usually mip-dna + fastq) or something. Can be removed once we stop passing the workflow as an argument and retrieve it from the case instead

cg/cli/deliver/base.py
```diff
@@ -68,48 +64,21 @@ def deliver_analysis(
...
-            cases.append(case_obj)
-        else:
-            cases: list[Case] = status_db.get_cases_by_ticket_id(ticket_id=ticket)
-            if not cases:
-                LOG.warning(f"Could not find cases for ticket {ticket}")
-                return
-
-        for case_obj in cases:
-            deliver_api.deliver_files(case_obj=case_obj)
+        delivery_api.deliver(case_id=case_id, ticket=ticket, workflow=delivery)
```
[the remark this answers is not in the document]
seallard: I agree but will leave that refactoring for a separate PR 👍

cg/meta/deliver/delivery_api.py
```diff
@@ -0,0 +1,205 @@
...
+            customer_id=case.customer.internal_id,
+            ticket=case.latest_ticket,
+            base_path=self.customers_folder,
+        )
+        if not self.dry_run:
+            LOG.debug(f"Creating delivery directory for case {case_target_directory}")
+            case_target_directory.mkdir(parents=True, exist_ok=True)
+        return case_target_directory
+
+    def _deliver_sample_files(self, case: Case, workflow: str) -> None:
```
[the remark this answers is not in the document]
seallard: I agree, but in the interest of merging this PR I'll leave it for subsequent PR:s if that is OK? I added your suggestions to the list of stuff left to do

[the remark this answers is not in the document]
ivadym: Sounds good, I can tackle it in my PR when this one is merged 👍

ivadym:
### Tests

Delivery:
- [x] `cg deliver analysis -c <case-id> -d balsamic -d fastq` (FASTQ + ANALYSIS delivery)
  - [x] Fastqs `sample 1` linked
  - [x] Fastqs `sample 2` linked
  - [x] Case files linked
  - [x] Case `sample 1` files linked
  - [x] Case `sample 2` files linked

- [x] `cg deliver analysis -t 1234567 -d balsamic -d fastq`
  - [x] `case 1` delivered
  - [x] `case 2` delivered

Clinical delivery:
- `cg upload -c <case-id>` (FASTQ + ANALYSIS delivery)
  - [x] Case, sample, and fastq files linked
  - [x] Rsync submitted

seallard:
Tested deliver and clinical-delivery with a mutant case as well 👍

PR #2809, opened 2024-01-08
Improve logging for file reading module

seallard:
## Description
Include the missing file path by default in error when file is missing.

### Fixed
- Improve logging for file reading module


ChrOertlin (approved):
💯

henrikstranneheim (approved):
👍

seallard:
Apparently, this is unnecessary. I was thrown by the use of these kind of logs which removes all context
        except Exception as exception:
            LOG.error(
                f"Error updating analysis for: case - {analysis.case_id} : {exception.__class__.__name__}"
            )

PR #2817, opened 2024-01-10
Remove call to trailblazer endpoint without effect

seallard:
## Description
The `mark-analyses-deleted` did not do anything (it just set a boolean on the cases which is not used anywhere). The endpoint is being removed from trailblazer and we should not call it in cg.

### Fixed
- Remove call to endpoint in trailblazer which does not do anything


diitaz93 (approved):
👍

Issue #2820, opened 2024-01-11
Virus samples run through microsalt fails QC


seallard:
## Description
Other application tags than 'MWRNXTR003', 'MWXNXTR003' or 'mic' are possible for samples run through the microsalt. Specifically, virus samples with application tags like `VWGNXTR001` are also run. This causes the quality control in microsalt to fail.

## Suggested solution
Disable the qc for virus samples run through the microsalt pipeline

seallard:
Example
```
20d [hiseq.clinical@hasta:~] [P_base] 21s $ cg workflow microsalt store-available
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Found 7 cases to perform QC on!
Checking QC for case famousmullet in /home/proj/production/microbial/results/ACC13033_2023.10.31_14.12.6
QC required for case famousmullet
Traceback (most recent call last):
  File "/home/proj/production/bin/miniconda3/envs/P_cg/bin/cg", line 8, in <module>
    sys.exit(base())
             ^^^^^^
  File "/home/proj/production/bin/miniconda3/envs/P_cg/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/proj/production/bin/miniconda3/envs/P_cg/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/home/proj/production/bin/miniconda3/envs/P_cg/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/proj/production/bin/miniconda3/envs/P_cg/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #2825, opened 2024-01-11
Auto pass qc for non microsalt apptags

seallard:
## Description
Closes https://github.com/Clinical-Genomics/cg/issues/2820. 
According to @Karl-Svard, we should disable the microsalt qc for samples with apptags other than MWX/MWR/mic.

New requirement: as to ensure the virus samples have reads, a qc based on the percent guaranteed reads of the target reads will be applied according to @beatrizsavinhas and @karlnyr.

### Fixed
- Disable microsalt qc for non microsalt sample apptags


cg/meta/workflow/microsalt/quality_controller/quality_controller.py
```diff
@@ -83,6 +84,9 @@ def quality_control_sample(self, sample_id: str, metrics: SampleMetrics) -> Samp
             and valid_10x_coverage
         )
 
+        if is_non_microsalt_apptag(sample):
```
seallard, line 87: Is there a better name for such apptags?

diitaz93: one solution is to create the opposite function, `has_sample_microsalt_apptag` and just check if that is False 😅

diitaz93 (approved):
👍

cg/meta/workflow/microsalt/quality_controller/quality_controller.py
Karl-Svard: you could name them "non_microbial" instead of "non_microsalt"

Karl-Svard (approved):
I haven't really been part of the microSALT QC project, but it looks good to me! 

FYI @beatrizsavinhas @karlnyr

Issue #2832, opened 2024-01-12
Ensure analysis are created for fastq cases


seallard:
## Description
Part of https://github.com/Clinical-Genomics/cigrid-ui/issues/478.
To easily filter cases which only consist of a fastq data delivery, we need to create a corresponding analysis for it in trailblazer.

In store_fastq_analysis add a completed FASTQ analysis to trailblazer.
    - add_pending_analysis
    - set_status -> completed

PR #2833, opened 2024-01-12
Export Fastq analyses to Trailblazer

islean:
## Description

Addresses #2832. We add a "dummy" analysis for Fastq cases with status "Completed" to trailblazer. Part of the user story https://github.com/Clinical-Genomics/cigrid-ui/issues/478.

### Changed

- Fastq cases now get an analysis set in Trailblazer with status Completed.

### DIFF
--- a/cg/cli/workflow/fastq/base.py
+++ b/cg/cli/workflow/fastq/base.py
@@ -3,8 +3,11 @@ import logging
 
 import click
 
+from cg.apps.tb import TrailblazerAPI
 from cg.cli.workflow.commands import ARGUMENT_CASE_ID
-from cg.constants.constants import DRY_RUN, Pipeline
+from cg.constants.constants import DRY_RUN, AnalysisType, Pipeline
+from cg.constants.priority import Priority
+from cg.constants.tb import AnalysisStatus
 from cg.meta.workflow.analysis import AnalysisAPI
 from cg.store import Store
 from cg.store.models import Analysis, Case
@@ -39,6 +42,17 @@ def store_fastq_analysis(context: click.Context, case_id: str, dry_run: bool = F
         return
     status_db.session.add(new_analysis)
     status_db.session.commit()
+    trailblazer_api: TrailblazerAPI = context.obj.trailblazer_api
+    trailblazer_api.add_pending_analysis(
+        case_id=case_id,
+        analysis_type=AnalysisType.OTHER,
+        data_analysis=Pipeline.FASTQ,
+        config_path="",
+        out_dir="",
+        slurm_quality_of_service=Priority.priority_to_slurm_qos().get(case_obj.priority),
+        ticket=case_obj.latest_ticket,
+    )
+    trailblazer_api.set_analysis_status(case_id=case_id, status=AnalysisStatus.COMPLETED)
 
 
 @fastq.command("store-available")
--- a/cg/cli/workflow/fastq/base.py
+++ b/cg/cli/workflow/fastq/base.py
@@ -1,16 +1,12 @@
-import datetime as dt
 import logging
 
 import click
 
-from cg.apps.tb import TrailblazerAPI
 from cg.cli.workflow.commands import ARGUMENT_CASE_ID
-from cg.cli.workflow.fastq.utils import get_slurm_priority
-from cg.constants.constants import DRY_RUN, AnalysisType, Pipeline
-from cg.constants.tb import AnalysisStatus
+from cg.cli.workflow.fastq.fastq_service import FastqService
+from cg.constants.constants import DRY_RUN, Pipeline
 from cg.meta.workflow.analysis import AnalysisAPI
 from cg.store import Store
-from cg.store.models import Analysis, Case
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


cg/cli/workflow/fastq/base.py
```diff
@@ -39,6 +42,17 @@ def store_fastq_analysis(context: click.Context, case_id: str, dry_run: bool = F
...
     status_db.session.add(new_analysis)
     status_db.session.commit()
+    trailblazer_api: TrailblazerAPI = context.obj.trailblazer_api
+    trailblazer_api.add_pending_analysis(
+        case_id=case_id,
+        analysis_type=AnalysisType.OTHER,
+        data_analysis=Pipeline.FASTQ,
+        config_path="",
+        out_dir="",
+        slurm_quality_of_service=Priority.priority_to_slurm_qos().get(case_obj.priority),
```
islean, line 52: If these should be changed, please let me know. As far as we could tell, there was no need to set them, given that no actual analysis is run for a Fastq case.

cg/cli/workflow/fastq/base.py
```diff
@@ -27,18 +30,29 @@ def store_fastq_analysis(context: click.Context, case_id: str, dry_run: bool = F
     """Creates an analysis object in status-db for the given fast case"""
     LOG.info(f"Creating an analysis for case {case_id}")
     status_db: Store = context.obj.status_db
-    case_obj: Case = status_db.get_case_by_internal_id(internal_id=case_id)
+    case: Case = status_db.get_case_by_internal_id(case_id)
```
ChrOertlin: ❤️

cg/cli/workflow/fastq/base.py
```diff
@@ -27,18 +30,29 @@ def store_fastq_analysis(context: click.Context, case_id: str, dry_run: bool = F
...
         return
     status_db.session.add(new_analysis)
     status_db.session.commit()
+    trailblazer_api: TrailblazerAPI = context.obj.trailblazer_api
+    trailblazer_api.add_pending_analysis(
+        case_id=case_id,
+        analysis_type=AnalysisType.OTHER,
+        data_analysis=Pipeline.FASTQ,
+        config_path="",
+        out_dir="",
```
islean, line 51: These should be set to fetch the jobs

cg/cli/workflow/fastq/base.py
```diff
@@ -27,18 +30,29 @@ def store_fastq_analysis(context: click.Context, case_id: str, dry_run: bool = F
     """Creates an analysis object in status-db for the given fast case"""
```
diitaz93, line 30:
```suggestion
    """Creates an analysis object in status-db for the given fast case."""
```

cg/cli/workflow/fastq/fastq_service.py
seallard, line 1: Maybe this file belongs in a different module and should be named differently

ChrOertlin (approved):
💯

diitaz93 (approved):
Great 💯

cg/cli/workflow/fastq/base.py
```diff
@@ -23,22 +22,17 @@ def fastq(context: click.Context):
...
-        started_at=dt.datetime.now(),
-        case_id=case_obj.id,
-    )
+
     if dry_run:
         return
-    status_db.session.add(new_analysis)
-    status_db.session.commit()
+
+    fastq_service = FastqService(
```
seallard, line 31: This should not be instantiated here but rather in the app config

cg/cli/workflow/fastq/utils.py
```diff
@@ -0,0 +1,7 @@
+from pathlib import Path
+
+import click
+
+
+def get_config_path(context: click.Context) -> str:
+    return Path(context.obj.data_delivery.base_path, "slurm_job_ids.yaml").as_posix()
```
Vince-janv, line 7: What do we need this path for? Nothing will be created there right?

islean: That path is wrong, but the config path is what we use to find the submitted jobs, right? Or is that via some other parameter? @Vince-janv

seallard: Yes, we need a path to the the `slurm_job_ids.yaml` to track the rsync job. @beatrizsavinhas wants to be able to track the upload job for the fastq case in the new tab.

seallard: The rsync dirs are named inconsistently, some use the database id and others the internal id. For example `271228_240108_10_53_05_784914` vs `betterweevil_240108_23_18_09_159755` which seems super annoying

seallard: We will skip the upload job and fix separately for all pipelines and remove the rsync tab in trailblazer.

cg/cli/workflow/fastq/base.py
seallard: Skipping for now, see below.

cg/cli/workflow/fastq/base.py
islean: Out of scope.

PR #2834, opened 2024-01-12
Fix non microbial qc

seallard:
## Description
Missed to update the actual qc in https://github.com/Clinical-Genomics/cg/pull/2825 for non microbial samples.

### Fixed
- Pass non microbial samples based on total reads qc in microsalt

### DIFF
--- a/cg/meta/workflow/microsalt/quality_controller/quality_controller.py
+++ b/cg/meta/workflow/microsalt/quality_controller/quality_controller.py
@@ -54,6 +54,7 @@ class QualityController:
         return sample_results
 
     def quality_control_sample(self, sample_id: str, metrics: SampleMetrics) -> SampleQualityResult:
+        """Perform a quality control of a sample given its metrics."""
         valid_read_count: bool = self.has_valid_total_reads(sample_id)
         valid_mapping: bool = has_valid_mapping_rate(metrics)
         valid_duplication: bool = has_valid_duplication_rate(metrics)
--- a/cg/meta/workflow/microsalt/quality_controller/quality_controller.py
+++ b/cg/meta/workflow/microsalt/quality_controller/quality_controller.py
@@ -85,7 +85,7 @@ class QualityController:
         )
 
         if has_non_microbial_apptag(sample):
-            sample_passes_qc = True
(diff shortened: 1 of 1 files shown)


cg/meta/workflow/microsalt/quality_controller/quality_controller.py
```diff
@@ -85,7 +85,7 @@ def quality_control_sample(self, sample_id: str, metrics: SampleMetrics) -> Samp
         )
```
diitaz93, line 85: I know this may seem out of scope but from the name of the function is unclear what it does. It also lacks a docstring, it would be great to have one and to rename it as `get_quality_control_sample` if it applies.

cg/meta/workflow/microsalt/quality_controller/quality_controller.py
```diff
@@ -85,7 +85,7 @@ def quality_control_sample(self, sample_id: str, metrics: SampleMetrics) -> Samp
...
 
         if has_non_microbial_apptag(sample):
-            sample_passes_qc = True
+            sample_passes_qc = valid_read_count
```
diitaz93, line 88: Shouldn't this be detected in the tests? does this function have a test?

diitaz93 (approved):
Take into account my comments but feel free to resolve them if you need to merge this sson

cg/meta/workflow/microsalt/quality_controller/quality_controller.py
seallard: There is no test for samples with non microbial apptags, I'll leave it as is for now

cg/meta/workflow/microsalt/quality_controller/quality_controller.py
seallard: Do you mean the `quality_control_sample` function?

diitaz93: add `perform` at the beginning

seallard: Added a docstring to clarify what it does 👍

Issue #2838, opened 2024-01-15
Microsalt cases missing


seallard:
Some microsalt cases are not showing in tb. The oldest missing case is from 2023-12-18. The most recent missing case is from 2024-01-09. There are microsalt analyses in trailblazer after, during and before this period. So it seems to be a bug only affecting certain microsalt cases.

Missing cases:
- smashingwhippet
- wiredbarnacle
- trustedmidge

seallard:
Replaced by https://github.com/Clinical-Genomics/microSALT/issues/167

PR #2840, opened 2024-01-15
Temp fix missing microsalt cases

seallard:
## Description
See https://github.com/Clinical-Genomics/cg/issues/2838.
This temp fix means the cases will be tracked in trailblazer, but jobs for it will not show up.

The issue is that microsalt uses a time stamp in the name for the output directory. This means we cannot pass the correct path when creating the pending analysis in trailblazer, which in turn means that jobs cannot be tracked correctly.

The actual solution consists of renaming the output directory created by microsalt to just use the case name. Making changes in https://github.com/Clinical-Genomics/microSALT seems to run the risk of introducing 1000 regression bugs, but maybe I feel lucky.

### Fixed
- Temp fix for missing microsalt cases


diitaz93 (approved):
👍

PR #2845, opened 2024-01-16
Add delivery message endpoint

seallard:
## Description
Part of https://github.com/Clinical-Genomics/cigrid-ui/issues/482.

### Added
- Endpoint for retrieving the delivery message for a case

- [ ] **MINOR**


islean:
Seems to work. The CiGRID PR needs a fix though.

cg/services/delivery_message/messages/statina_message.py
```diff
@@ -0,0 +1,7 @@
+from cg.services.delivery_message.messages.delivery_message import DeliveryMessage
+from cg.store.models import Case
+
+
+class StatinaMessage(DeliveryMessage):
+    def create_message(self, case: Case) -> str:
+        return "Hello,\n\nBatch BATCH_ID is now available in Statina."
```
seallard, line 7: Missing batch id/url here

cg/services/delivery_message/messages/analysis_scout_message.py
```diff
@@ -0,0 +1,16 @@
+from cg.services.delivery_message.messages.delivery_message import DeliveryMessage
+from cg.services.delivery_message.messages.utils import get_fastq_delivery_path, get_scout_link
+from cg.store.models import Case
+
+
+class AnalysisScoutMessage(DeliveryMessage):
+    def create_message(self, case: Case) -> str:
```
ChrOertlin, line 7: should this be able to get mutlple scout links that can be pooled in one message. Imagine a ticket with 30 cases, that would be 30 emails or?

cg/services/delivery_message/messages/utils.py
```diff
@@ -0,0 +1,18 @@
...
+    ticket_id: str = case.latest_ticket
+    customer_id: str = case.customer.internal_id
+    return f"/home/{customer_id}/inbox/{ticket_id}"
+
+
+def get_scout_link(case: Case) -> str:
+    customer_id: str = case.customer.internal_id
+    case_name: str = case.name
+    return f"https://scout.scilifelab.se/{customer_id}/{case_name}"
+
```
ChrOertlin, line 14: consider adding:

```suggestion
def get_scout_links(cases: list[Case]) -> str:
    scout_links: list[str] = []
    for case in cases:
        customer_id: str = case.customer.internal_id
        case_name: str = case.name
        scout_links.append( f"https://scout.scilifelab.se/{customer_id}/{case_name}")
    return scout_links
```

ChrOertlin (approved):
Looks good! Just a question regarding pooling of multiple scout links into a single message.

cg/services/delivery_message/messages/analysis_scout_message.py
islean: This is not implemented for tickets currently, although that is up for refinement.

ChrOertlin: Ok, but how will the current implementation work in practice? Will there be 30 messages in a ticket with 30 cases like this? Is that what we want in the end? Or maybe I am missing a step here.

seallard: Yes, the implementation follows the acceptance criteria defined by production. They want it on a case basis

tests/server/conftest.py
```diff
@@ -1,20 +1,53 @@
...
+    from cg.server.auto import app
+
+    app.config.update({"TESTING": True})
+    create_all_tables()
+    yield app
+    drop_all_tables()
+
+
+@pytest.fixture
+def case(helpers: StoreHelpers) -> Case:
```
diitaz93, line 35:
```suggestion
def mip_dna_scout_delivery_case(helpers: StoreHelpers) -> Case:
```

cg/services/delivery_message/messages/utils.py
seallard: Not needed currently

tests/server/conftest.py
```diff
@@ -1,20 +1,53 @@
...
+    from cg.server.ext import db as store
```
diitaz93, line 36: Is importing inside a fixture a common practice? Why not import the module outside?

diitaz93 (approved):
Looks amazing 🤩

tests/server/conftest.py
seallard: It was a necessary workaround since we create the flask app which instantiates the store. I agree it can be improved, but we did not find a better way to do it

tests/server/conftest.py
seallard: We do not care what kind of case it is

tests/server/conftest.py
seallard: Moving it outside works fine, good catch 👍

cg/services/delivery_message/messages/utils.py
```diff
@@ -16,3 +16,7 @@ def get_scout_link(case: Case) -> str:
 def get_pangolin_delivery_path(case: Case) -> str:
     customer_id: str = case.customer.internal_id
     return f"/home/{customer_id}/inbox/wwLab_automatisk_hamtning"
+
+
+def get_statina_batch_id(case: Case) -> str:
+    return case.name.split("-")[1]
```
ChrOertlin, line 22: Just a heads up, I don't know if that makes a difference for anything on the user end:

From the case.name for fluffy/nipt/statina cases you get this: `181213-2400953_NIPT`

This function would return: 2400953_NIPT

In the mongodb for statina, batch ids look like this: 2400953

seallard: Good catch, we will check with prod and see what they want in the delivery message

PR #2866, opened 2024-01-24
Add order table

islean:
## Description

Addresses the first point in the user story https://github.com/Clinical-Genomics/cigrid-ui/issues/488, and adds an Order table which will be used to connect cases related to the same ticket in SuSy.

### Added

- Order table
- Column in case table with a foreign key constraint to the order id.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
Deploying to stage (Hasta) works: 

image
 
And new order_id column in case:

image

alembic/versions/2024_01_24_a6befebf1231_add_order_table.py
```diff
@@ -0,0 +1,48 @@
...
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("customer_id", sa.Integer(), nullable=True),
+        sa.Column("order_date", sa.DateTime(), nullable=True),
+        sa.Column("ticket_id", sa.Integer(), nullable=False),
+        sa.ForeignKeyConstraint(
+            ["customer_id"],
+            ["customer.id"],
+        ),
+        sa.PrimaryKeyConstraint("id"),
+        sa.UniqueConstraint("id"),
```
islean, line 32: Would we want this to reference the customer internal_id since it otherwise won't show in statusdb?

alembic/versions/2024_01_24_a6befebf1231_add_order_table.py
```diff
@@ -0,0 +1,48 @@
...
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "order",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("customer_id", sa.Integer(), nullable=True),
+        sa.Column("order_date", sa.DateTime(), nullable=True),
```
ChrOertlin, line 25: also here, why can it be nullable, if an order is placed it should have a date

cg/server/admin.py
```diff
@@ -640,3 +640,12 @@ class SampleLaneSequencingMetricsView(BaseView):
         "flowcell": FlowcellView.view_flow_cell_link,
         "sample": SampleView.view_sample_link,
     }
+
+
+class OrderView(BaseView):
```
seallard, line 645: I'm not sure we should add a view for it yet since production use them and might be annoyed. But maybe it is worth it for dev purposes? Might be better than dbeaver

islean: Oh, I thought we wanted it visible for production. Otherwise I am in favour of removing the view.

cg/store/models.py
```diff
@@ -393,6 +393,7 @@ class Case(Model, PriorityMixin):
     internal_id = Column(types.String(32), unique=True, nullable=False)
     is_compressible = Column(types.Boolean, nullable=False, default=True)
     name = Column(types.String(128), nullable=False)
+    order_id = Column(ForeignKey("order.id", ondelete="CASCADE"))
     ordered_at = Column(types.DateTime, default=dt.datetime.now)
```
ChrOertlin, line 397: is the case.ordered_at redundant when we have an order table that has the order date?

ChrOertlin (approved):
Looks good!

cg/server/admin.py
ChrOertlin: could be a handy resource also to see the number of cases in a ticket easily. I'd check with production if they have use for something like this already.

cg/store/models.py
```diff
@@ -393,6 +393,7 @@ class Case(Model, PriorityMixin):
     internal_id = Column(types.String(32), unique=True, nullable=False)
     is_compressible = Column(types.Boolean, nullable=False, default=True)
     name = Column(types.String(128), nullable=False)
+    order_id = Column(ForeignKey("order.id", ondelete="CASCADE"))
```
seallard, line 396: Will this delete all cases associated with an order? Maybe we want to skip it for now if we want to delete stuff in the order table, not sure! You know best

islean:
Downgrade worked:

```
Logging deploy ...
Getting deployer... done.
done.
Log deploy... done.
cg, version 55.4.11
θ64° [isak.ohlsson@hasta:~] [S_base] 2m59s $ D

```

seallard (approved):
👍 Great start! 🥳

cg/server/admin.py
seallard: We don't have any features implemented yet related to an order, so I vote for hiding it for now

cg/store/models.py
islean: Most likely, but all logic which reroutes functionality will be in other PRs.

alembic/versions/2024_01_24_a6befebf1231_add_order_table.py
```diff
@@ -0,0 +1,48 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "order",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("customer_id", sa.Integer(), nullable=True),
```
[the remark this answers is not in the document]
islean: I do agree that it should not be nullable here, although I don't think it is on the basis of it being a foreign key.

alembic/versions/2024_01_24_a6befebf1231_add_order_table.py
islean: Yeah, that is also true.

cg/store/models.py
islean: We skip it for now!

islean: I will need to test so that we can remove cases still, even if they are in an order though. Or maybe we need to refine that condition? In a separate PR though.

cg/server/admin.py
Vince-janv: I think hiding for now is good

islean: This does not help with that @ChrOertlin, at least not with the current implementation. The case table already has a ticket_id column. And in the Order table the cases are not visible so the only thing they would gain is that they can go from a case to an order and see customer_id and ordered_at, although an order and a case need not be ordered at the same time I guess.

alembic/versions/2024_01_24_a6befebf1231_add_order_table.py
islean: Changed it.

cg/server/admin.py
islean: View is hidden now.

islean:
Deployed to production (Hasta and cg-prod-services):

```
INFO  [alembic.runtime.migration] Running upgrade 584840c706a0 -> a6befebf1231, Add order table
fatal: Not a git repository (or any parent up to mount point /home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
repository is clean
Logging deploy ...
Getting deployer... done.
done.
Log deploy... done.
cg, version 55.5.0
θ61° [isak.ohlsson@hasta:~] [P_base] 6m48s $ 
```

PR #2868, opened 2024-01-24
Add orders endpoint

islean:
## Description

This PR adds an endpoint for fetching orders. Addresses point 4 in https://github.com/Clinical-Genomics/cigrid-ui/issues/488.

### Added

- Endpoint for fetching database entries from the new orders table.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b add-order-endpoint -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -469,7 +469,7 @@ def get_application_pipeline_limitations(tag: str):
 
 
 @BLUEPRINT.route("/orders")
-def get_orders(limit: int):
+def get_orders():
     """Return the latest orders."""
     if request.args.get("limit"):
         orders: list[Order] = db.get_orders(request.args.get(key="limit", type=int))
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -27,7 +27,7 @@ from cg.meta.orders import OrdersAPI
 from cg.models.orders.order import OrderIn, OrderType
 from cg.models.orders.orderform_schema import Orderform
 from cg.server.dto.delivery_message_response import DeliveryMessageResponse
-from cg.server.dto.orders_request import OrdersRequest
+from cg.server.dto.orders_request import OrdersRequest, OrdersResponse
 from cg.server.ext import db, lims, osticket
 from cg.services.delivery_message.delivery_message_service import DeliveryMessageService
 from cg.services.orders.order_service import OrderService
@@ -38,7 +38,6 @@ from cg.store.models import (
     Case,
     Customer,
     Flowcell,
-    Order,
     Pool,
     Sample,
     SampleLaneSequencingMetrics,
@@ -475,10 +474,8 @@ def get_orders():
     """Return the latest orders."""
     orders_request: OrdersRequest = OrdersRequest.model_validate(request.args)
     order_service = OrderService(db)
-    orders: list[Order] = order_service.get_orders(orders_request)
-    if not orders:
-        return jsonify(message="No orders found"), HTTPStatus.NOT_FOUND
-    return jsonify([order.to_dict() for order in orders])
+    response: OrdersResponse = order_service.get_orders(orders_request)
+    return jsonify(response.model_dump())
 
 
 @BLUEPRINT.route("/orderform", methods=["POST"])
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -27,8 +27,10 @@ from cg.meta.orders import OrdersAPI
 from cg.models.orders.order import OrderIn, OrderType
 from cg.models.orders.orderform_schema import Orderform
 from cg.server.dto.delivery_message_response import DeliveryMessageResponse
+from cg.server.dto.orders_request import OrdersRequest
 from cg.server.ext import db, lims, osticket
 from cg.services.delivery_message.delivery_message_service import DeliveryMessageService
+from cg.services.orders.order_service import OrderService
 from cg.store.models import (
     Analysis,
     Application,
@@ -471,10 +473,9 @@ def get_application_pipeline_limitations(tag: str):
 @BLUEPRINT.route("/orders")
 def get_orders():
     """Return the latest orders."""
-    if request.args.get("limit"):
-        orders: list[Order] = db.get_orders(request.args.get(key="limit", type=int))
-    else:
-        orders: list[Order] = db.get_orders()
+    orders_request: OrdersRequest = OrdersRequest.model_validate(request.args)
+    order_service = OrderService(db)
+    orders: list[Order] = order_service.get_orders(orders_request)
     if not orders:
         return jsonify(message="No orders found"), HTTPStatus.NOT_FOUND
     return jsonify([order.to_dict() for order in orders])
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -473,7 +473,7 @@ def get_application_pipeline_limitations(tag: str):
 @BLUEPRINT.route("/orders")
 def get_orders():
     """Return the latest orders."""
-    orders_request: OrdersRequest = OrdersRequest.model_validate(request.args)
+    orders_request: OrdersRequest = OrdersRequest.model_validate(request.args.to_dict())
     order_service = OrderService(db)
     response: OrdersResponse = order_service.get_orders(orders_request)
     return jsonify(response.model_dump())
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -36,6 +36,7 @@ from cg.store.models import (
     Case,
     Customer,
     Flowcell,
+    Order,
     Pool,
     Sample,
     SampleLaneSequencingMetrics,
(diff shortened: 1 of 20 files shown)


cg/server/api.py
```diff
@@ -467,6 +468,18 @@ def get_application_pipeline_limitations(tag: str):
     return jsonify([limitation.to_dict() for limitation in application_limitations])
 
 
+@BLUEPRINT.route("/orders")
+def get_orders():
+    """Return the latest orders."""
+    if request.args.get("limit"):
```
seallard, line 474: I think we should wait with implementing pagination and filtering

islean:
Tested both with and without a limit and with a limit. Without a limit:

```
[
  {
    "customer_id": 1,
    "id": 1,
    "order_date": "Tue, 19 Dec 2023 00:00:00 GMT",
    "ticket_id": 1
  },
  {
    "customer_id": 2,
    "id": 2,
    "order_date": "Tue, 19 Dec 2023 00:00:00 GMT",
    "ticket_id": 2
  }
]
```

With a limit set to 1:

```
[
  {
    "customer_id": 1,
    "id": 1,
    "order_date": "Tue, 19 Dec 2023 00:00:00 GMT",
    "ticket_id": 1
  }
]
```

cg/server/api.py
islean: So you think it is preferable to remove it?

cg/server/api.py
```diff
@@ -467,6 +468,18 @@ def get_application_pipeline_limitations(tag: str):
...
+        orders: list[Order] = db.get_orders(request.args.get(key="limit", type=int))
+    else:
```
seallard, line 476: Don't call database logic directly in the endpoint and don't parse request arguments like this. Follow the [pattern in trailblazer](https://github.com/Clinical-Genomics/trailblazer/blob/0607d626f8e85e07398113db0fd50c4845db1d52/trailblazer/server/api.py#L62).

The pattern is
- Parse request body into pydantic model.
- Send pydantic model into service layer.
- Send data to database layer, which sends up database model to service layer.
- Service layer converts database model to pydantic response model.
- API layer dumps pydantic model into json response.

This can be refactored later, but if you want to practice writing clean endpoints it is a much better pattern.

In this case, you would just create a service called `OrderService` which has a single method `get_orders` which does not accept any arguments. Unless you want to implement pagination, then you can look at how it is done in trailblazer (I don't think it is worth it at this stage though).

cg/store/crud/read.py
```diff
@@ -1735,6 +1736,12 @@ def get_all_pools_to_deliver(self) -> list[Pool]:
         )
         return records.all()
 
+    def get_orders(self, limit: int | None = None) -> list[Order]:
+        records = self._get_query(table=Order)
```
henrikstranneheim, line 1740: Add type hint

seallard (approved):
Nice! If you feel like it, you can try to create a cleaner structure for the endpoint. Maybe `limit` is useful for dev purposes for now. Then, if you feel like doing it cleaner, create a `OrdersRequest` pydantic model which includes the `limit`. That way we get clearer separation between the request validation, business logic, database logic and serialization logic.

henrikstranneheim (approved):
Exciting! Would be nice with unit test for the end point and for untested sql functions. 👍 .

cg/server/api.py
henrikstranneheim: I like this pattern, do you think we can discuss/add it to the coding guidelines?

cg/server/api.py
seallard: Actually, maybe it is nice for dev purposes! You know best

cg/server/api.py
islean: Yeah, I will refactor this endpoint as well! Don't want to add to the technical debt!

islean: Refactored it a bit, although we have a slightly different method of parsing the request - do you see a reason for not using BaseModel.model_validate(request.args)?

cg/server/api.py
```diff
@@ -467,6 +470,17 @@ def get_application_pipeline_limitations(tag: str):
     return jsonify([limitation.to_dict() for limitation in application_limitations])
 
 
+@BLUEPRINT.route("/orders")
+def get_orders():
+    """Return the latest orders."""
+    orders_request: OrdersRequest = OrdersRequest.model_validate(request.args)
```
seallard, line 476: I'd move this logic to a separate function, maybe in `server/utils.py´ called `def parse_orders_request(request: Request) -> OrdersRequest`

cg/server/api.py
```diff
@@ -467,6 +470,17 @@ def get_application_pipeline_limitations(tag: str):
...
+    order_service = OrderService(db)
```
seallard, line 477: If possible, I'd initialize this service in the app context and retrieve it like
```python
order_service: OrderService = current_app.extensions.get("order_service")
``` 
That would require setting it up like this in `app.py`:
```python
# Services
order_service = OrderService(store)

# Register services
app.extensions["order_service"] = analysis_service
```

Or in some other similar way, I have not found a great way for injecting services yet in Python but I think this works for now. Here is how it is [done in Trailblazer](https://github.com/Clinical-Genomics/trailblazer/blob/0607d626f8e85e07398113db0fd50c4845db1d52/trailblazer/server/app.py#L29).

cg/server/dto/orders_request.py
```diff
@@ -0,0 +1,5 @@
+from pydantic import BaseModel
+
+
+class OrdersRequest(BaseModel):
+    limit: int | None = None
```
seallard, line 5: I'd create an `OrderResponse` as well (I think we want to remove the `to_dict` logic from our database models anyway) and then a utils module in the `services/orders` dir with a method.

cg/server/api.py
islean: I think for me that does not add value since it would signal that it needs more manipulation than to be read into an object? Like I saw this in TB but I felt that this was clearer to me.

cg/server/api.py
```diff
@@ -467,6 +470,17 @@ def get_application_pipeline_limitations(tag: str):
...
+    orders: list[Order] = order_service.get_orders(orders_request)
+    if not orders:
+        return jsonify(message="No orders found"), HTTPStatus.NOT_FOUND
```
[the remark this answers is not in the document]
islean: So I should just remove this check?

cg/services/orders/order_service.py
```diff
@@ -0,0 +1,10 @@
+from cg.server.dto.orders_request import OrdersRequest
+from cg.store.store import Store
+
+
+class OrderService:
+    def __init__(self, store: Store) -> None:
+        self.store = store
+
+    def get_orders(self, orders_request: OrdersRequest):
+        return self.store.get_orders(orders_request.limit)
```
seallard, line 10: I'd make this return a `OrdersResponse` instead.

seallard (approved):
Looks great! Made some suggestions if you want to further

- Use dependency injection for the `OrderService`
- Separate the database layer from the api layer
- Separate de-serialization logic from the api layer
- Separate serialization logic from the database model

cg/server/api.py
seallard: Yeah, keep it 👍 Might make more sense later when we add pagination

cg/server/api.py
seallard: No, I think it is fine 👍 Might make more sense to do down the line

cg/server/api.py
```diff
@@ -467,6 +470,17 @@ def get_application_pipeline_limitations(tag: str):
     return jsonify([limitation.to_dict() for limitation in application_limitations])
 
 
+@BLUEPRINT.route("/orders")
```
seallard, line 473: Forgot to say, but I would a test for the endpoint like [this](https://github.com/Clinical-Genomics/cg/blob/master/tests/server/endpoints/test_delivery_message_endpoint.py), if you have the energy. It should be pretty straightforward since we have one test doing exactly this already.

cg/server/api.py
islean: @seallard There is no such pattern in cg though, and the store is never even mentioned in app.py. If anything, I could initialize it in ext.py where the "db" variable is instantiated. That does not sound terribly intuitive to me though.

seallard: Let skip it for now!

cg/server/api.py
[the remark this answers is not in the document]
islean: removed

cg/server/dto/orders_request.py
islean: OrderResponse created!

cg/server/api.py
islean: Added tests

cg/store/crud/read.py
islean: Added

seallard (approved):
Nice 🥳

cg/server/api.py
islean: I'll keep it for now.

islean:
Deployed to production (hasta and cg-prod-services).

PR #2870, opened 2024-01-25
Modify Order table

islean:
## Description

This is a revision on the order table in which we would like to group the orders in CiGRID based on the pipeline, hence why the pipeline column is now being added. I also added a relationship from order to customer to be able to access more customer data.

### Added

- Pipeline column to the order table.
- Order-Customer relationship

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b modify-order -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
Alembic revisions work locally.

cg/server/dto/orders/orders_response.py
```diff
@@ -4,7 +4,7 @@
 
 
 class Order(BaseModel):
-    customer_id: int
+    customer_id: str
```
seallard, line 7: I'd just call this field `customer_name`

islean: Reasonable, but might be confusing when we have a Customer.name field as well?

alembic/versions/2024_01_25_d241d8c493fb_add_pipeline_to_order.py
```diff
@@ -0,0 +1,34 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        table_name="order",
+        column=sa.Column(
+            "pipeline",
+            sa.Enum(*tuple(Pipeline)),
```
islean, line 25: Probably should not be nullable?

tests/store/crud/read/test_read.py
```diff
@@ -1539,8 +1539,17 @@ def test_get_orders_populated_store(store: Store, order: Order, order_another: O
...
     assert len(store.get_orders(limit=1)) == 1
+
+
+def test_order_customer_relationship(store_with_multiple_pools_for_customer: Store, order: Order):
+    # GIVEN a store with an order and a customer
+
+    # WHEN accessing the order
+
+    # THEN the customer should be set via the relationship
+    assert order.customer
```
seallard, line 1555: Is this test really necessary? I think we end up testing the sqlalchemy library here rather than our logic

cg/server/dto/orders/orders_response.py
seallard, line 7: You probably want to add the pipeline here as well, right? 😄

seallard (approved):
Nice 👍

cg/server/dto/orders/orders_response.py
seallard: Ah, we do?! Maybe we should display that instead in the frontend? Your call!

cg/services/orders/utils.py
```diff
@@ -4,5 +4,7 @@
 
 def parse_order(order: DatabaseOrder) -> PydanticOrder:
```
seallard, line 5: Oof, I'd not name this `PydanticOrder` 😢  But stick to it if you want. If it is necessary to have a different name than `Order`, I'd add a `Dto` or something to it

cg/server/dto/orders/orders_response.py
henrikstranneheim: Whenever guarantied to be unique I think `_id` should be used.

henrikstranneheim (approved):
Nice, just want to discuss using workflow instead of pipeline.

Vince-janv:
Quick question on Database normalisation. In this case not duplicating data. Do we think pipeline should be removed from case in the future?

alembic/versions/2024_01_25_d241d8c493fb_add_pipeline_to_order.py
```diff
@@ -0,0 +1,34 @@
...
+down_revision = "a6befebf1231"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        table_name="order",
+        column=sa.Column(
+            "pipeline",
```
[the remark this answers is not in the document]
islean: I will change it!

islean:
> Quick question on Database normalisation. In this case not duplicating data. Do we think pipeline should be removed from case in the future?

I would love that, after all cases have been connected to an order.

cg/server/dto/orders/orders_response.py
islean: Yes! That would be ideal 😅

alembic/versions/2024_01_25_d241d8c493fb_add_pipeline_to_order.py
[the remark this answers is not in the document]
seallard: Maybe order type is a better name for it?

tests/store/crud/read/test_read.py
islean: Probably more for my own dev purposes, like a log message which should be removed before merging 😅

islean: removed

cg/services/orders/utils.py
islean: Changed it to Order, and we'll see if anyone complains in the future.

alembic/versions/2024_01_25_d241d8c493fb_add_pipeline_to_order.py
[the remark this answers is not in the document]
islean: Have we not started out with workflow in other places though?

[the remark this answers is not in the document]
islean: For example /cg/meta/workflow which contains pipelines and fastq etc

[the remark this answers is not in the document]
ChrOertlin: indeed workflow is best here

islean:
Seems to work for the new CiGRID page purposes:

image

cg/server/dto/orders/orders_response.py
islean: I will keep it as the internal id for now - I think production more commonly refers to customers with custXXX than the name chosen by the customers themselves.

alembic/versions/2024_01_25_d241d8c493fb_add_pipeline_to_order.py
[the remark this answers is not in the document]
islean: I will go with workflow as OrderType is a slightly different field in the order forms.

islean:
Deployed to production (Hasta and cg-prod-services):

```
Logging deploy ...
Getting deployer... done.
done.
Log deploy... done.
cg, version 56.0.1
θ69° [isak.ohlsson@hasta:~] [P_base] 7m31s $
```

PR #2875, opened 2024-01-29
Fix parameter name

seallard:
Fix misnamed parameter in function call.

### DIFF
--- a/cg/apps/tb/api.py
+++ b/cg/apps/tb/api.py
@@ -150,8 +150,8 @@ class TrailblazerAPI:
             command="add-comment", request_body=request_body, method=APIMethods.PUT
         )
 
-    def add_upload_job_to_analysis(self, analyis_id: int, slurm_job_id: int) -> None:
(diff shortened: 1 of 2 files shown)


cg/apps/tb/api.py
```diff
@@ -150,8 +150,8 @@ def add_comment(self, case_id: str, comment: str):
             command="add-comment", request_body=request_body, method=APIMethods.PUT
         )
 
-    def add_upload_job_to_analysis(self, analyis_id: int, slurm_job_id: int) -> None:
-        create_request = CreateJobRequest(slurm_id=slurm_job_id, job_type=JobType.UPLOAD)
+    def add_upload_job_to_analysis(self, analyis_id: int, slurm_id: int) -> None:
```
islean, line 153:
```suggestion
    def add_upload_job_to_analysis(self, analysis_id: int, slurm_id: int) -> None:
```

islean (approved):
Another typo as well but good catch!

diitaz93 (approved):
👍

PR #2876, opened 2024-01-29
Create order when an order is placed

seallard:
## Description
This PR adds logic which creates an order when an order is placed. Part of the [implementation plan](https://github.com/Clinical-Genomics/cigrid-ui/issues/488#issuecomment-1905465692) of the [order list view story](https://github.com/Clinical-Genomics/cigrid-ui/issues/488).

### Changed
- Persist an order when it is placed

Test placing an order in stage, I imagine creating an integration test for placing an order is a nightmare. If you think it should be done anyway, let me know and I'll do it in this PR.


cg/server/api.py
```diff
@@ -498,6 +498,8 @@ def parse_orderform():
             order_parser = JsonOrderformParser()
             order_parser.parse_orderform(order_data=json_data)
         parsed_order: Orderform = order_parser.generate_orderform()
+        order_service = OrderService(db)
+        order_service.create_order(parsed_order)
```
islean, line 502: I am uncertain if this is the correct place to put this - this is only used to parse the orderform and return the value to the FE. I don't think an order has actually been submitted at this point, so nothing should be persisted to the database. But please correct me if I'm wrong.

cg/services/orders/utils.py
```diff
@@ -1,8 +1,16 @@
...
     return PydanticOrder(
         customer_id=order.customer_id, ticket_id=order.ticket_id, order_date=order.order_date
     )
+
+
+def create_orders_response(orders: list[DatabaseOrder]) -> OrdersResponse:
+    return OrdersResponse(orders=[_parse_order(order) for order in orders])
+
+
+def create_order_response(order: DatabaseOrder) -> PydanticOrder:
```
islean, line 15: I changed so I did not use "PydanticOrder" so maybe we should refrain from that here as well?

cg/store/models.py
```diff
@@ -883,7 +883,7 @@ class Order(Model):
 
     id = Column(types.Integer, primary_key=True, unique=True)
     customer_id = Column(types.String(64), ForeignKey("customer.id"), nullable=False)
-    order_date = Column(types.DateTime, nullable=False)
+    order_date = Column(types.DateTime, nullable=False, default=dt.datetime.now())
```
islean, line 886: Nice!

islean (commented):
I think we potentially modify the wrong part of the flow. Also - do we not need to add the order id to the connected cases?

cg/store/crud/create.py
```diff
@@ -403,3 +406,14 @@ def add_sample_lane_sequencing_metrics(
             sample_internal_id=sample_internal_id,
             **kwargs,
         )
+
+    def add_order(self, order_data: Orderform):
```
islean, line 410: Feels a bit weird that we would have in the CRUD something which only works with orderforms? I would imagine we potentially want to expand this to a CLI method at some point where only the actual fields are provided? Could we change the parameters to data_analysis, customer_id and ticket_id or something similar?

cg/server/api.py
seallard: Oh, there is a `submit_order` endpoint 😅

cg/store/crud/create.py
seallard: I hope we never create a cli command for placing an order 😭 But yes, the actual data is a `OrderIn` type which is reasonable to pass.

cg/server/api.py
seallard: Moved it 👍

cg/store/crud/create.py
islean: I was thinking also for testing purposes, it is much easier to not mock a whole OrderForm. And still feels odd to me that the CRUD would "know" of anything about the indata. But up to you if you overrule me in this.

seallard: I'll leave it as for now 👍

cg/server/api.py
```diff
@@ -119,15 +119,17 @@ def submit_order(order_type):
...
+        order_in = OrderIn.parse_obj(request_json, project=project)
 
         result = api.submit(
             project=project,
             order_in=order_in,
             user_name=g.current_user.name,
             user_mail=g.current_user.email,
         )
+        order_service = OrderService(db)
+        order_service.create_order(order_in)
```
Vince-janv, line 132: At some point we would like to replace the ticket fields on Sample and Case with a foreign key right? For that the order creation would need to be moved further into `api.submit` where Samples and Cases are created. Maybe placing it somewhere there would save time in the next PR. Up to you though

Vince-janv (approved):
Nice 👍

cg/server/api.py
islean: Also  should not cases be connected to the orders created in this PR?

seallard: Ideally the endpoint will just make a single call `order_service.create_order(order_request)`. I don't think we should move this call further down the chain, but rather a lot of logic should be reshuffled and probably removed.

Vince-janv: Oooh I like the sound of that but I fear you might drown in refactoring 😅

henrikstranneheim: I think it is worth doing now. It will save us time and headache in the future.

henrikstranneheim: Also the entire ordering process is to coupled to LIMS and Susy, so it would be good to rethink this.

seallard: I'm sensing scope creep! In the interest of getting on with the project I'll leave refactoring to subsequent PR:s.

We probably should implement integration tests for the order endpoint before refactoring it. This logic is too integral and messy for me to just refactor and will be a month long project in itself. I'll set up some refinements to stake out what the order endpoint is supposed to do and how we want to implement it.

henrikstranneheim: Sounds like a good plan1

seallard: https://github.com/Clinical-Genomics/cg/issues/2879

seallard:
Deployed to stage and tested submitting an order.

PR #2881, opened 2024-01-30
Add workflow filtering to Orders endpoint

islean:
## Description

To populate different order tabs in CiGRID, there is a need to filter orders on the workflow. This PR adds such support.

### Added

- Support for filtering orders on workflow.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/store/crud/read.py
```diff
@@ -1736,11 +1737,16 @@ def get_all_pools_to_deliver(self) -> list[Pool]:
...
         records: Query = self._get_query(table=Order)
-        if limit is not None:
-            records = records.limit(limit)
+        if orders_request:
+            limit: int | None = orders_request.limit
+            workflow: str | None = orders_request.workflow
+            if limit is not None:
+                records = records.limit(limit)
+            if workflow is not None:
+                records = records.filter(Order.workflow == workflow)
```
ChrOertlin: Have a look at the store filters `cg/store/filters/` when writing queries and filter functions that access the store we have a certain schema that we follow.

We define filters with a single responsiblity that are then called in the crud to construct a more complex query.

ChrOertlin: This also looks like it would be two separate queries.

1) `get_orders`
2) `get_orders_for_workflow`

tests/store/crud/read/test_read.py
```diff
@@ -1539,8 +1540,19 @@ def test_get_orders_populated_store(store: Store, order: Order, order_another: O
...
+
+def test_get_orders_workflow_filter(
+    store: Store, order: Order, order_another: Order, order_balsamic: Order
+):
+    # GIVEN a store with three orders, one of which is a Balsamic order
+
+    # WHEN fetching only balsamic orders
+    orders: list[Order] = store.get_orders(OrdersRequest(workflow=Pipeline.BALSAMIC))
+    # THEN only one should be returned
+    assert len(orders) == 1 and orders[0].workflow == Pipeline.BALSAMIC
```
ChrOertlin, line 1558: this does not test when workflow is none or the limit if supplied.

islean:
Filtering works:

image

image

image

image

cg/store/crud/read.py
```diff
@@ -1736,11 +1737,16 @@ def get_all_pools_to_deliver(self) -> list[Pool]:
         )
         return records.all()
 
-    def get_orders(self, limit: int | None = None) -> list[Order]:
+    def get_orders(self, orders_request: OrdersRequest | None = None) -> list[Order]:
         """Returns a list of entries in the table Order."""
         records: Query = self._get_query(table=Order)
-        if limit is not None:
-            records = records.limit(limit)
+        if orders_request:
```
seallard, line 1743: Extract this to a separate utils method called `parse_orders_request` let it return a `OrdersRequest` pydantic model containing the limit and workflow

cg/store/crud/read.py
```diff
@@ -1736,11 +1737,16 @@ def get_all_pools_to_deliver(self) -> list[Pool]:
         )
         return records.all()
 
-    def get_orders(self, limit: int | None = None) -> list[Order]:
+    def get_orders(self, orders_request: OrdersRequest | None = None) -> list[Order]:
```
seallard, line 1740: `orders_request` cannot be none here, right?

cg/store/crud/read.py
seallard, line 1743: If the orders_request cannot be none this check can be removed

seallard (approved):

tests/store/crud/read/test_read.py
seallard: I think integration test should be added which sends a request against the endpoint with the limit and workflow parameter set

cg/store/crud/read.py
```diff
@@ -1736,12 +1738,14 @@ def get_all_pools_to_deliver(self) -> list[Pool]:
...
```
ChrOertlin, line 1741: This query gets orders by workflow and not just all orders. We have function name for `get_samples` or `get_sample` reserved to either get all or one sample, for instance.

```suggestion
    def get_orders_by_workflow(self, workflow: str = None, limit = None) -> list[Order]:
```

To make the fuction more generally usable should pass on workflow and limit instead of an `OrderRequest` here. In case we need to re-use this functionality elsewhere we don't require the `OrderRequest` class,

cg/store/filters/status_order_filters.py
ChrOertlin, line 1: excellent!

ChrOertlin (approved):
Nice job! One more thing regarding passing `OrderRequest` as parameter to a store function. Maybe we do not want that for re-usability in other places.

cg/store/crud/read.py
islean: Yeah, it is sort of up for debate but I think I agree with you on this

islean: I will rewrite, and maybe have a separate function which just unpacks the request

tests/store/crud/read/test_read.py
islean: Added additional tests

cg/store/crud/read.py
islean: Fixed

cg/store/crud/read.py
islean: @ChrOertlin I don't really agree with the naming convention, though. Or would it not be unclearer if it was called `get_orders_by_workflow` but you can get all orders if you don't provide the workflow?

islean: Also, might not be more flexible to pass all parameters separately here, if we establish a standard where we have a request object containing all data that can be filtered upon. I will stick to the convention for now, but ideally, I would probably want to leave this pattern since it becomes very hard to read when you can filter on many fields.

islean:
Deployed to the VMs by IO.

Issue #2884, opened 2024-01-31
Saving Subject ID for all orders


annagellerbring:
## Description

Subject ID is mandatory to provide in the orderform (1508) for clinical samples but for orders with Data Delivery "fastq" or "no delivery" the Subject ID is not saved in statusdb. The customer invest time to provide it and it's useful to have for future re-runs of the sample.

## Suggested solution

Start saving it in the database at the field "Subject ID".

## This can be closed when

Subject ID is saved for all clinical samples.

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

henrikstranneheim:
@islean Do you have time to fix this?

islean:
> @islean Do you have time to fix this?

Looking into it!

henrikstranneheim:
Seems done!

PR #2888, opened 2024-01-31
Add subject id for fastq orders

islean:
## Description

Addresses #2884.

### Fixed

- Fastq orders get subject_id set on sample level.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/orders/fastq_submitter.py
+++ b/cg/meta/orders/fastq_submitter.py
@@ -46,6 +46,7 @@ class FastqSubmitter(Submitter):
                     "name": sample.name,
                     "priority": sample.priority,
                     "sex": sample.sex,
+                    "subject_id": sample.subject_id,
                     "tumour": sample.tumour,
                     "volume": sample.volume,
                 }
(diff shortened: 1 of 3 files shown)


seallard (approved):
👍 It would be nice to have a better name than `subject_id` (what is that?) but out of scope. Also, do we want to store it on the order model?

islean:
> 👍 It would be nice to have a better name than `subject_id` (what is that?) but out of scope. Also, do we want to store it on the order model?

I don't think there is a need to set it on order_id. This seems to be used to link e.g. RNA samples to corresponding DNA samples. If it should be added, it would need to be refined a bit. Currently I think subject_id can be set per sample, meaning that it is not one subject_id per order.

islean:
Tested on stage and the subject_id is now set for fastq orders: 

image

henrikstranneheim:
Subjekt_id represents the (organism) that a sample has been isolated from. It is a unique identifier used to link multiple samples to an organism.

PR #2892, opened 2024-02-01
refactor external data API

diitaz93:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/cli/add.py
```diff
@@ -359,6 +359,8 @@ def download_external_delivery_data_to_hpc(context: CGConfig, ticket: str, dry_r
 )
 @click.pass_obj
 def add_external_data_to_hk(context: CGConfig, ticket: str, dry_run: bool, force):
-    """Adds external data to Housekeeper"""
-    external_data_api = ExternalDataAPI(config=context)
-    external_data_api.add_transfer_to_housekeeper(dry_run=dry_run, ticket=ticket, force=force)
+    """Adds external data to Housekeeper and starts their analyses."""
+    external_data_api = AddExternalDataAPI(
```
seallard, line 363: Pass via context

cg/cli/add.py
```diff
@@ -359,6 +359,8 @@ def download_external_delivery_data_to_hpc(context: CGConfig, ticket: str, dry_r
...
+        config=context, ticket=ticket, dry_run=dry_run, force=force
```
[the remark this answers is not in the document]
diitaz93: we discussed this with @Vince-janv and we agreed that in this case is convinient

cg/meta/transfer/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def get_all_fastq_from_folder(sample_folder: Path) -> list[Path]:
+    """Returns a list of all fastq.gz files in given folder."""
+    all_fastqs: list[Path] = []
+    for leaf in sample_folder.glob("*fastq.gz"):
+        abs_path: Path = sample_folder.joinpath(leaf)
+        LOG.info(f"Found file {str(abs_path)} inside folder {sample_folder}")
+        all_fastqs.append(abs_path)
+    return all_fastqs
```
ChrOertlin, line 16: We have a `files.py` in `cg/utils` we should add file handling utility functions in there I think and re-use them as much as possible. `get_files_matching_pattern` It is called I believe

cg/cli/add.py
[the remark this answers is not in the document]
diitaz93: Maybe I should elaborate: Each of these "APIs" have a single endpoint, which is either adding or transferring. All the other functions are support functions and all need either the customer id or the ticket. The instantiation of this API does not make sense without a ticket id in the way that it is built right now. Of course, I am open to more suggestions @seallard

cg/meta/transfer/external_data.py
```diff
@@ -8,64 +8,69 @@
...
-        self.source_path: str = config.external.caesar
-        self.base_path: str = config.data_delivery.base_path
+        self.customer_id: str = self.status_db.get_customer_id_from_ticket(ticket=ticket)
+        self._destination_path: str = config.external.hasta
+        self.dry_run: bool = dry_run
+        self.ticket: str = ticket
+
+    def get_destination_path(self, sample_id: str | None = "") -> Path:
+        """Returns the path to where the files are to be transferred."""
+        return Path(self._destination_path % self.customer_id, sample_id)
```
ChrOertlin, line 31:
```suggestion
        return Path(self._destination_path % self.customer_id, sample_id)
```
f-string

ChrOertlin (commented):
To me logically it makes sense to split the functionalities into `Transfer` and `Adding`.

cg/cli/add.py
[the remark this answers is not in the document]
seallard: Why is it more convenient than passing it via the method like it used to?
Stick with this if you like it though 👍 Just be aware that it has downsides. You cannot pass it via the context, the code becomes more difficult to test and usually more difficult to read

cg/meta/transfer/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+def are_all_fastq_valid(fastq_paths: list[Path]) -> bool:
+    """Return True if all fastq files of a given sample pass md5 checksum."""
+    are_all_valid: bool = True
+    for path in fastq_paths:
+        if is_md5sum_correct(path):
+            are_all_valid = False
```
seallard, line 24: This seems incorrect, if the md5sum is valid the function should not return false

cg/cli/add.py
```diff
@@ -341,8 +341,8 @@ def link_sample_to_case(
 def download_external_delivery_data_to_hpc(context: CGConfig, ticket: str, dry_run: bool):
     """Downloads external data from the delivery server and places it in appropriate folder on
     the HPC"""
-    external_data_api = ExternalDataAPI(config=context)
-    external_data_api.transfer_sample_files_from_source(ticket=ticket, dry_run=dry_run)
+    external_data_api = TransferExternalDataAPI(config=context, ticket=ticket, dry_run=dry_run)
+    external_data_api.transfer_sample_files_from_source()
```
seallard, line 345: I'd prefer it that the ticket was passed here like it used to but stick to whatever you prefer 👍

ChrOertlin: I think this is indeed more in line with what we discussed before. You would be able to instantiate the API once and re-use it with any given ticket. 

If we would require this, i'd say it would be worth sticking to the older way.

seallard (approved):
Looks good! Just fix the hash check bug

cg/meta/transfer/utils.py
diitaz93: good catch

cg/cli/add.py
diitaz93: I checked and MetaAPIs are not passed by context but instantiated on site when the command that requires them is called, so for now I will leave this as it is

cg/cli/add.py
islean: I also prefer having it not part of the instantiation. Also because I do not find it clear from the name "external_data_api" what ticket is being handled

cg/cli/add.py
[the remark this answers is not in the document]
islean: I also find adding this data as part of the instantiation of an API confusing, as I am not expecting an API to be tied to one specific file or ticket.

cg/meta/transfer/utils.py
diitaz93: fixed

cg/cli/add.py
diitaz93: fixed

cg/cli/add.py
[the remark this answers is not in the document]
islean: Is it not part of the init still? 🤔

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,194 @@
...
+        self.dry_run: bool | None = None
+        self.force: bool | None = None
         self.mail_user: str = config.data_delivery.mail_user
-        self.slurm_api: SlurmAPI = SlurmAPI()
         self.RSYNC_FILE_POSTFIX: str = "_rsync_external_data"
+        self.slurm_api: SlurmAPI = SlurmAPI()
+        self.source_path: str = config.external.caesar
+        self.ticket: str | None = None
+
+    def _set_parameters(self, ticket: str, dry_run: bool, force: bool | None = False):
```
diitaz93, line 36: Although we agreed that we should not use setters, I think this is more of a wrapper for assigning values to a lot of attributes that is more DRY

diitaz93:
> To me logically it makes sense to split the functionalities into `Transfer` and `Adding`.

I realised this change requires proper test infrastructure (contexts, fixtures, etc) that goes beyond the scope of this PR, which is to include fastq of external samples to Housekeeper bundles.

cg/meta/transfer/utils.py
```diff
@@ -0,0 +1,16 @@
...
+from pathlib import Path
+
+from cg.utils.checksum.checksum import is_md5sum_correct
+
+LOG = logging.getLogger(__name__)
+
+
+def are_all_fastq_valid(fastq_paths: list[Path]) -> bool:
+    """Return True if all fastq files of a given sample pass md5 checksum."""
+    are_all_valid: bool = True
```
ChrOertlin, line 11:
```suggestion
    are_valid: bool = True
```

ChrOertlin (approved):
looks good

cg/meta/transfer/utils.py
diitaz93: I think `all` is more meaningful

cg/meta/transfer/external_data.py
henrikstranneheim: Can't we have these as default when initializing?

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
+        self.force = force
+        self.customer_id = self.status_db.get_customer_id_from_ticket(ticket=self.ticket)
 
-    def create_log_dir(self, dry_run: bool, ticket: str) -> Path:
+    def _create_log_dir(self) -> Path:
         """Creates a directory for log file to be stored"""
         timestamp: dt.datetime = dt.datetime.now()
         timestamp_str: str = timestamp.strftime("%y%m%d_%H_%M_%S_%f")
-        folder_name: Path = Path("_".join([ticket, timestamp_str]))
+        folder_name: Path = Path("_".join([self.ticket, timestamp_str]))
```
henrikstranneheim, line 47:
```suggestion
        folder_name = Path("_".join([self.ticket, timestamp_str]))
```

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
+        self.ticket = ticket
+        self.dry_run = dry_run
+        self.force = force
+        self.customer_id = self.status_db.get_customer_id_from_ticket(ticket=self.ticket)
 
-    def create_log_dir(self, dry_run: bool, ticket: str) -> Path:
+    def _create_log_dir(self) -> Path:
         """Creates a directory for log file to be stored"""
         timestamp: dt.datetime = dt.datetime.now()
         timestamp_str: str = timestamp.strftime("%y%m%d_%H_%M_%S_%f")
```
henrikstranneheim, line 46:
```suggestion
        timestamp: str = timestamp.strftime("%y%m%d_%H_%M_%S_%f")
```

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
-        folder_name: Path = Path("_".join([ticket, timestamp_str]))
+        folder_name: Path = Path("_".join([self.ticket, timestamp_str]))
         log_dir: Path = Path(self.base_path, folder_name)
         LOG.info(f"Creating folder: {log_dir}")
```
henrikstranneheim, line 49: debug?

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
         return log_dir
 
-    def get_source_path(
-        self,
-        customer: str,
-        ticket: str,
-        cust_sample_id: str | None = "",
-    ) -> Path:
+    def _get_source_path(self) -> Path:
         """Returns the path to where the sample files are fetched from"""
```
henrikstranneheim, line 57:
```suggestion
        """Returns the path to where the sample files are fetched from."""
```

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
-        return Path(self.source_path % customer, ticket, cust_sample_id)
+        return Path(self.source_path % self.customer_id, self.ticket)
 
-    def get_destination_path(self, customer: str, lims_sample_id: str | None = "") -> Path:
+    def _get_destination_path(self, lims_sample_id: str | None = "") -> Path:
         """Returns the path to where the files are to be transferred"""
```
henrikstranneheim, line 61:
```suggestion
        """Returns the path to where the files are to be transferred."""
```

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
-        return Path(self.destination_path % customer, lims_sample_id)
+        return Path(self.destination_path % self.customer_id, lims_sample_id)
 
-    def transfer_sample_files_from_source(self, dry_run: bool, ticket: str) -> None:
+    def transfer_sample_files_from_source(self, ticket: str, dry_run: bool = False) -> None:
         """Transfers all sample files, related to given ticket, from source to destination"""
```
henrikstranneheim, line 65:
```suggestion
        """Transfers all sample files on given ticket, from source to destination."""
```

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
-            error=error_function,
+            error=ERROR_RSYNC_FUNCTION.format(),
         )
         self.slurm_api.set_dry_run(dry_run=dry_run)
-        sbatch_content: str = self.slurm_api.generate_sbatch_content(
-            sbatch_parameters=sbatch_parameters
-        )
-        sbatch_path: Path = Path(log_dir, ticket + self.RSYNC_FILE_POSTFIX + ".sh")
+        sbatch_content: str = self.slurm_api.generate_sbatch_content(sbatch_parameters)
+        sbatch_path = Path(log_dir, self.ticket + self.RSYNC_FILE_POSTFIX + ".sh")
```
henrikstranneheim, line 88: Use FileExtension constant.

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
         self.slurm_api.submit_sbatch(sbatch_content=sbatch_content, sbatch_path=sbatch_path)
-        LOG.info(
-            "The folder {src_path} is now being rsynced to hasta".format(
-                src_path=self.get_source_path(customer=cust, ticket=ticket)
-            )
+        LOG.info(f"The folder {self._get_source_path().as_posix()} is now being rsynced to hasta")
```
henrikstranneheim, line 90:
```suggestion
        LOG.info(f"The folder {self._get_source_path().as_posix()} is now being Rsynced to destination")
```

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
+
+    def _curate_sample_folder(self, sample_folder: Path) -> None:
+        """
+        Changes the name of the folder to the sample internal_id. If force is set to True,
+        replaces any previous folder.
+        Raises:
+            Exception if the sample from the folder is not present in statusdb.
```
henrikstranneheim, line 97:
```suggestion
            Exception if the sample from the folder is not present in StatusDB.
```

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
+        """
+        # TODO: Implement dry-run
```
henrikstranneheim, line 99:
```suggestion
```
Or do it in this PR or open issue.

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
+        sample: Sample = self.status_db.get_sample_by_customer_and_name(
+            customer_entry_id=[customer.id], sample_name=sample_folder.name
         )
+        if (sample and not customer_folder.joinpath(sample.internal_id).exists()) or (
+            sample and self.force
+        ):
+            sample_folder.rename(customer_folder.joinpath(sample.internal_id))
+        elif not sample and not self.status_db.get_sample_by_internal_id(sample_folder.name):
+            raise Exception(
+                f"{sample_folder} is not a sample present in statusdb. "
```
henrikstranneheim, line 111:
```suggestion
                f"{sample_folder} is not a sample present in StatusDB. "
```

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
-                f"{sample_folder} is not a sample present in statusdb. Move or remove it to continue"
+    def _add_and_include_files_to_bundles(
+        self, fastq_paths: list[Path], lims_sample_id: str
+    ) -> None:
+        """Add the given fastq files to the the Housekeeper bundle."""
+        if self.dry_run:
+            LOG.info(f"No files will be added for sample {lims_sample_id} since this is a dry-run")
+            return
+        for path in fastq_paths:
+            LOG.info(f"Adding path {path} to bundle {lims_sample_id} in housekeeper")
```
henrikstranneheim, line 161:
```suggestion
            LOG.info(f"Adding path {path} to bundle {lims_sample_id} in Housekeeper")
```

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
-        bundle"""
-        failed_paths: list[Path] = []
-        cust: str = self.status_db.get_customer_id_from_ticket(ticket=ticket)
-        destination_folder_path: Path = self.get_destination_path(customer=cust)
-        for sample_folder in destination_folder_path.iterdir():
-            self.curate_sample_folder(cust_name=cust, sample_folder=sample_folder, force=force)
-        available_samples: list[Sample] = self.get_available_samples(
-            folder=destination_folder_path, ticket=ticket
+        """
+        Add and include available ticket fastq files to a Housekeeper bundle if they are not
```
henrikstranneheim, line 181:
```suggestion
        Add and include available ticket fastq files to a Housekeeper bundles if they are not
```

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
+                )
+                self._start_cases(
+                    cases=self.status_db.get_not_analysed_cases_by_sample_internal_id(
+                        sample_internal_id=sample_id
+                    )
+                )
+            else:
+                LOG.warning(
+                    f"Some files in {sample_id} did not match the given md5sum."
+                    " Changes in housekeeper will not be committed and no cases will be started"
```
henrikstranneheim, line 207:
```suggestion
                    " Changes in Housekeeper will not be committed and no cases will be started"
```

tests/meta/transfer/test_external_data.py
```diff
@@ -97,55 +98,131 @@ def test_transfer_sample_files_from_source(
...
+):
+    """Test getting the sample ids from a folder."""
+    # GIVEN a sample folder
+    tmp_dir_path = Path(tmp_path, sample.internal_id)
+    tmp_dir_path.mkdir()
+    # GIVEN an External API with a ticket number
+    external_data_api.ticket = ticket_id
+    # WHEN getting the sample ids from the folder
+    available_samples = external_data_api._get_sample_ids_from_folder(folder=tmp_dir_path.parent)
+    # THEN the function should return a list containing the sample object
```
henrikstranneheim, line 142:
```suggestion
    
    # WHEN getting the sample ids from the folder
    available_samples = external_data_api._get_sample_ids_from_folder(folder=tmp_dir_path.parent)

     # THEN the function should return a list containing the sample object
```

tests/meta/transfer/test_external_data.py
```diff
@@ -97,55 +98,131 @@ def test_transfer_sample_files_from_source(
...
-        lims_sample_id=sample_id,
+    """Test that getting fastq files to add finds the files in the origin folder."""
+
+    # GIVEN an External API with a sample destination path that has fastq files
+    mocker.patch.object(ExternalDataAPI, "_get_destination_path")
+    ExternalDataAPI._get_destination_path.return_value = fastq_dir
+
+    # GIVEN a sample id
+    sample_id: str = "fastq"
+    # WHEN getting the fastq paths to add
```
henrikstranneheim, line 175:
```suggestion

    # WHEN getting the fastq paths to add
```

tests/meta/transfer/test_external_data.py
```diff
@@ -97,55 +98,131 @@ def test_transfer_sample_files_from_source(
...
+    hk_api: HousekeeperAPI = external_data_api.housekeeper_api
+    assert not hk_api.files(tags=set("fastq"), bundle=sample.internal_id).all()
+
+    # WHEN the fastq file is added and included to Housekeeper
+    external_data_api._add_and_include_files_to_bundles(
+        fastq_paths=[fastq_file], lims_sample_id=sample.internal_id
+    )
+
+    # THEN the files should be in Housekeeper
+    file: File = hk_api.files(tags=set("fastq"), bundle=sample.internal_id).first()
```
henrikstranneheim, line 199: Use constant

tests/utils/test_checksum.py
```diff
@@ -0,0 +1,49 @@
...
+    assert not check_md5sum(
+        file_path=bad_md5sum_file_path, md5sum="c690b0124173772ec4cbbc43709d84ee"
+    )
+
+
+def test_extract_checksum(fastq_file: Path):
+    """Tests if the function successfully extract the correct md5sum."""
+
+    # GIVEN a file containing a md5sum
+    md5sum_file = Path(f"{fastq_file.as_posix()}.md5")
```
henrikstranneheim, line 26: Use FileExtension

henrikstranneheim (approved):
Well done, 👍 !

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
+        hk_version: Version = self.housekeeper_api.get_or_create_version(bundle_name=sample_id)
         fastq_paths_to_add: list[Path] = self.housekeeper_api.check_bundle_files(
-            file_paths=paths,
-            bundle_name=lims_sample_id,
+            file_paths=file_paths,
+            bundle_name=sample_id,
             last_version=hk_version,
             tags=HK_FASTQ_TAGS,
         )
+        LOG.debug(f"{len(fastq_paths_to_add)} samples are not yet in Housekeeper for {sample_id}")
```
islean, line 150: This can't be samples right? Or can you have "5 samples are not yet in housekeeper for <sample_id>"?

islean: Is it supposed to be "fastq files are not yet in housekeeper..."?

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
+        if self.dry_run:
+            LOG.info("No cases will be started since this is a dry-run")
+            return
+        for case in cases:
+            self.status_db.set_case_action(
+                case_internal_id=case.internal_id, action=CaseActions.ANALYZE
+            )
+            LOG.info(f"Case {case.internal_id} has been set to '{CaseActions.ANALYZE}'")
+
     def add_transfer_to_housekeeper(
```
islean, line 177: I am not a fan of this method name. I do not understand what a "transfer" is in this case, and think the added docstring indicates a violation of the SRP. If it both adds and includes fastq files and starts cases, maybe we should consider having two methods. One for adding/including files and one for setting the case actions.

cg/meta/transfer/external_data.py
diitaz93: I will do it in this PR after I solve some issues

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
         log_dir.mkdir(parents=True, exist_ok=False)
         return log_dir
 
-    def get_source_path(
-        self,
-        customer: str,
-        ticket: str,
-        cust_sample_id: str | None = "",
-    ) -> Path:
+    def _get_source_path(self) -> Path:
```
islean, line 56: It will be very confusing to me having both a property called `source_path` and a method called `_get_source_path` and these do not give the same value.

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
         """Returns the path to where the sample files are fetched from"""
-        return Path(self.source_path % customer, ticket, cust_sample_id)
+        return Path(self.source_path % self.customer_id, self.ticket)
 
-    def get_destination_path(self, customer: str, lims_sample_id: str | None = "") -> Path:
+    def _get_destination_path(self, lims_sample_id: str | None = "") -> Path:
```
islean, line 60: Same problem here with name clashing.

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
+        customer: Customer = self.status_db.get_customer_by_internal_id(self.customer_id)
+        customer_folder: Path = sample_folder.parent
+        sample: Sample = self.status_db.get_sample_by_customer_and_name(
+            customer_entry_id=[customer.id], sample_name=sample_folder.name
         )
+        if (sample and not customer_folder.joinpath(sample.internal_id).exists()) or (
+            sample and self.force
+        ):
+            sample_folder.rename(customer_folder.joinpath(sample.internal_id))
+        elif not sample and not self.status_db.get_sample_by_internal_id(sample_folder.name):
```
islean, line 109: Could we consider breaking out these if statements to make it a bit easier to read? Also, seems weird that we have both internal ids and sample names that need to be handled the same way...

islean (approved):
Looks good! I am not too thrilled about initializing an API with data specific to the functionality at hand but do what you think is best.

cg/meta/transfer/external_data.py
diitaz93: Yes, I can do that for `dry_run` and`force`

cg/meta/transfer/external_data.py
```diff
@@ -5,210 +5,204 @@
...
+        self.customer_id: str | None = None
+        self.destination_path: str = config.external.hasta
+        self.dry_run: bool | None = None
+        self.force: bool | None = None
         self.mail_user: str = config.data_delivery.mail_user
-        self.slurm_api: SlurmAPI = SlurmAPI()
         self.RSYNC_FILE_POSTFIX: str = "_rsync_external_data"
+        self.slurm_api: SlurmAPI = SlurmAPI()
+        self.source_path: str = config.external.caesar
+        self.ticket: str | None = None
```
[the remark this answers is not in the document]
diitaz93: We agreed that they would not be parameters for the init function. I though of defining them here so that we only define attributes here. Do you mean defining them inside the endpoint methods? @islean

cg/meta/transfer/external_data.py
diitaz93: it is only one bundle

cg/meta/transfer/external_data.py
diitaz93: fixed

cg/meta/transfer/external_data.py
diitaz93: done

cg/meta/transfer/external_data.py
diitaz93: This renaming is a bit more problematic as it may or may not involve a sample id, any suggestions?

cg/meta/transfer/external_data.py
diitaz93: fixed

cg/meta/transfer/external_data.py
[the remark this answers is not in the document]
islean: Oh yeah right! To me an API does not have hold any data apart from configuration needed for it to "boot". Any data specific for a method call such as customer_id or ticket I would pass between methods. But I do not know if that is the agreed universal way to do things.

cg/meta/transfer/external_data.py
diitaz93: fixed

cg/meta/transfer/external_data.py
islean: I guess you could set the class variable to "destination_path_stub" or something?

cg/meta/transfer/external_data.py
diitaz93: Good that I am not the only one (This is how the code was written originally, I just moved it). I will refactor this.

cg/meta/transfer/utils.py
diitaz93: I will leave it as it is, resolving

cg/meta/transfer/external_data.py
diitaz93: I renamed it to `add_external_data_to_housekeeper`. However modifying the flow for having two functions goes out of the scope of this PR I think, and should be discussed more carefully.

tests/meta/transfer/test_external_data.py
diitaz93: fixed

tests/utils/test_checksum.py
diitaz93: fixed

cg/meta/transfer/external_data.py
diitaz93: done

cg/meta/transfer/external_data.py
diitaz93: fixed @islean! Could you give it a look to check if I missed any logic?

islean: Much cleaner! Only concern would be if the Path.rename method breaks if there is no change in naming (as would happen now if it was already named with the internal id). Other than that, much easier on the eyes!

diitaz93: I guess that if it is working right now there is no issue to it, the behaviour is the same as before

PR #2915, opened 2024-02-07
Rnafusion/taxprofiler workflow: create and use nextflow config by default

fevac:
### Added

- Nextflow config creation for rnafusion and taxprofiler when config-case

### Changed

- Use nextflow config by default in rnafusion/taxprofiler start

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b multiqc_json -a
    ```

### How to test

- [x] `cg workflow rnafusion config-case tendergoose` runs an writes tendergoose_nextflow_config.json
- [x] `cg workflow rnafusion start tendergoose` uses config and succesfully completes
- [x] `cg workflow taxprofiler config-case case_id` runs an writes *_nextflow_config.json
- [x] `cg workflow taxprofiler start -d case_id` prints the command with the --config parameter

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/io/txt.py
+++ b/cg/io/txt.py
@@ -1,6 +1,7 @@
 """Module to read or write txt files."""
 
 from pathlib import Path
+from typing import Any
 
 
 def read_txt(file_path: Path, read_to_string: bool = False) -> list[str] | str:
@@ -11,7 +12,10 @@ def read_txt(file_path: Path, read_to_string: bool = False) -> list[str] | str:
         return list(file)
 
 
-def write_txt(content: list[str], file_path: Path) -> None:
+def write_txt(content: list[str] | Any, file_path: Path) -> None:
     """Write content to a text file."""
     with open(file_path, "w") as file:
-        file.writelines(content)
+        if isinstance(content, list):
+            file.writelines(content)
+        else:
+            file.write(content)
--- a/cg/io/txt.py
+++ b/cg/io/txt.py
@@ -12,7 +12,7 @@ def read_txt(file_path: Path, read_to_string: bool = False) -> list[str] | str:
         return list(file)
 
 
-def write_txt(content: list[str] | Any, file_path: Path) -> None:
+def write_txt(content: list[str] | str, file_path: Path) -> None:
     """Write content to a text file."""
     with open(file_path, "w") as file:
         if isinstance(content, list):
--- a/cg/meta/workflow/nf_analysis.py
+++ b/cg/meta/workflow/nf_analysis.py
@@ -9,14 +9,12 @@ from cg.constants.nextflow import NFX_WORK_DIR
 from cg.constants.tb import AnalysisStatus
 from cg.exc import CgError, MetricsQCError
 from cg.io.controller import ReadFile, WriteFile
+from cg.io.txt import write_txt
 from cg.io.yaml import write_yaml_nextflow_style
 from cg.meta.workflow.analysis import AnalysisAPI
 from cg.meta.workflow.nf_handlers import NextflowHandler, NfTowerHandler
 from cg.models.cg_config import CGConfig
-from cg.models.deliverables.metric_deliverables import (
-    MetricsBase,
-    MetricsDeliverablesCondition,
-)
+from cg.models.deliverables.metric_deliverables import MetricsBase, MetricsDeliverablesCondition
 from cg.models.fastq import FastqFileMeta
 from cg.models.nf_analysis import FileDeliverable, PipelineDeliverables
 from cg.models.rnafusion.rnafusion import CommandArgs
@@ -87,11 +85,13 @@ class NfAnalysisAPI(AnalysisAPI):
         """Get the compute environment for the head job based on the case priority."""
         return f"{self.compute_env_base}-{self.get_slurm_qos_for_case(case_id=case_id)}"
 
-    @staticmethod
-    def get_nextflow_config_path(nextflow_config: str | None = None) -> Path | None:
-        """Path to Nextflow config file."""
+    def get_nextflow_config_path(self, case_id: str, nextflow_config: Path | None = None) -> Path:
+        """Path to nextflow config file."""
         if nextflow_config:
             return Path(nextflow_config).absolute()
+        return Path((self.get_case_path(case_id)), f"{case_id}_nextflow_config").with_suffix(
+            FileExtensions.JSON
+        )
 
     def get_job_ids_path(self, case_id: str) -> Path:
         """Return the path to a Trailblazer config file containing Tower IDs."""
@@ -173,6 +173,15 @@ class NfAnalysisAPI(AnalysisAPI):
             content=workflow_parameters,
             file_path=self.get_params_file_path(case_id=case_id),
         )
+
+    def write_nextflow_config(self, case_id: str) -> None:
+        """Write nextflow config in json format."""
+        if self.get_nextflow_config_content:
+            LOG.debug("Writing nextflow config file")
+            write_txt(
+                content=self.get_nextflow_config_content,
+                file_path=self.get_nextflow_config_path(case_id=case_id),
+            )
 
     @staticmethod
     def write_sample_sheet(
--- a/cg/meta/workflow/nf_analysis.py
+++ b/cg/meta/workflow/nf_analysis.py
@@ -71,6 +71,10 @@ class NfAnalysisAPI(AnalysisAPI):
         """Get pipeline version from config."""
         return self.revision
 
+    def get_nextflow_config_content(self) -> str:
+        """Return nextflow config content."""
+        return None
+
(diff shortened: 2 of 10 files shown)


cg/meta/workflow/nf_analysis.py
```diff
@@ -174,6 +174,15 @@ def write_params_file(self, case_id: str, workflow_parameters: dict) -> None:
...
         )
 
+    def write_nextflow_config(self, case_id: str) -> None:
+        """Write nextflow config in json format."""
+        if self.get_nextflow_config_content:
+            LOG.debug("Writing nextflow config file")
+            write_txt(
+                content=self.get_nextflow_config_content,
+                file_path=self.get_nextflow_config_path(case_id=case_id),
+            )
```
ivadym, line 184:
```suggestion
        nextflow_config_path = self.get_nextflow_config_path(case_id=case_id)
        if self.get_nextflow_config_content:
            LOG.debug(f"Writing nextflow config file {nextflow_config_path}")
            write_txt(
                content=self.get_nextflow_config_content,
                file_path=nextflow_config_path
            )
```

cg/meta/workflow/rnafusion.py
```diff
@@ -64,6 +64,19 @@ def get_deliverables_template_content() -> list[dict]:
             file_path=resources.RNAFUSION_BUNDLE_FILENAMES_PATH,
         )
 
+    @property
```
ivadym, line 67: Properties are associated with attributes. Did you mean it to be a static method? Or perhaps a constant?

cg/meta/workflow/nf_analysis.py
```diff
@@ -87,11 +85,13 @@ def get_compute_env(self, case_id: str) -> str:
         """Get the compute environment for the head job based on the case priority."""
         return f"{self.compute_env_base}-{self.get_slurm_qos_for_case(case_id=case_id)}"
 
-    @staticmethod
-    def get_nextflow_config_path(nextflow_config: str | None = None) -> Path | None:
-        """Path to Nextflow config file."""
+    def get_nextflow_config_path(self, case_id: str, nextflow_config: Path | None = None) -> Path:
+        """Path to nextflow config file."""
         if nextflow_config:
             return Path(nextflow_config).absolute()
```
diitaz93, line 91: it seems that `nextflow_config` is a Path already, so I guess the cast is not necessary, unless the type hint is wrong

diitaz93 (approved):
Looks good 👍 I wonder about not using the JSON controllers for this config file Also @rannick was working on something similar in [this PR](https://github.com/Clinical-Genomics/cg/pull/2725/files#diff-9474c185a688ebf735009d85af462a731e9f462e4c124a4eda2e9c90c1b510e6)

cg/meta/workflow/nf_analysis.py
fevac: type hints are wrong. When getting the config path from the CLI is a string. I'll update it. Thanks

cg/meta/workflow/nf_analysis.py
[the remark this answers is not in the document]
fevac: love it! thanks

fevac:
> I have reviewed this PR, but I'm wondering why are we not using servers for the config? You have a draft PR: [Clinical-Genomics/servers#1081](https://github.com/Clinical-Genomics/servers/pull/1081) and there is also [something in place for RD](https://github.com/Clinical-Genomics/servers/blob/bf821bea85825e2acc8e52f7266070813242f7df/config/hasta.scilifelab.se/nextflow/raredisease/raredisease_v1.0_prod_params.config)

I talked to Annick and concluded that it was better to wait until the config for rarediseases is implemented and tested. Since rnafusion is currently in production and we are not doing a new release, we want to keep changes functional and minimal.  Once RD is tested AND works, we will replace this config to a full config (similar to RD) and move away from using nf-core profiles in favour of configs in server. This is an intermediate step to not hold the generation of json files.

cg/io/txt.py
```diff
@@ -11,7 +12,10 @@ def read_txt(file_path: Path, read_to_string: bool = False) -> list[str] | str:
         return list(file)
 
 
-def write_txt(content: list[str], file_path: Path) -> None:
+def write_txt(content: list[str] | Any, file_path: Path) -> None:
```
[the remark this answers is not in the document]
fevac: I know this is redundant but IMO is also clearer.  But I could update to `list[str] | str` as I haven't tested other types

cg/meta/workflow/nf_analysis.py
fevac: will go with @diitaz93 suggestion instead

cg/meta/workflow/rnafusion.py
```diff
@@ -64,6 +64,19 @@ def get_deliverables_template_content() -> list[dict]:
...
+    def get_nextflow_config_content(self) -> str:
+        """Return nextflow config content."""
+        return """process {
+    withName:'MULTIQC' {
+        memory = { 1.GB * task.attempt }
+        time   = { 4.h  * task.attempt }
+        cpus = 2
+        ext.args = ' --data-format json '
+    }
+}
```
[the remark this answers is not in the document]
fevac: yes, at least taxprofiler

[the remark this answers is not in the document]
ivadym: Maybe it's worth making it general and moving it to nf_analysis.py so we don't have to change this twice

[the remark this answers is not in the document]
fevac: it's not tested yet for taxprofiler, so I'm not sure if the config will be exactly the same. It will use a config, but if the config is the same, I'm not 100% sure yet. Until then I would rather not add it to a common place since this is quite specific.

cg/meta/workflow/rnafusion.py
ivadym: I'm not sure if this would be a correct use of the [property decorator](https://docs.python.org/3/library/functions.html#property). 

In any case you're not invoking it correctly in `write_nextflow_config` with `content=self.get_nextflow_config_content`. Has to be `NfAnalysisAPI.get_nextflow_config_content` since it's a class getter at the end

cg/meta/workflow/rnafusion.py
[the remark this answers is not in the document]
ivadym: Sounds good 👍

cg/meta/workflow/rnafusion.py
fevac: I don't think so, RnafusionAnalysisAPI inherits from NfAnalysisAPI and I'm accessing an attribute from within so self.attribute is ok

fevac: in any case, I can make it into a method if you prefer that. that's ok with me

cg/meta/workflow/nf_analysis.py
```diff
@@ -174,6 +174,15 @@ def write_params_file(self, case_id: str, workflow_parameters: dict) -> None:
             file_path=self.get_params_file_path(case_id=case_id),
         )
 
+    def write_nextflow_config(self, case_id: str) -> None:
+        """Write nextflow config in json format."""
+        if self.get_nextflow_config_content:
+            LOG.debug("Writing nextflow config file")
+            write_txt(
```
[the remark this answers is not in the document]
diitaz93: fair enough

cg/meta/workflow/rnafusion.py
ivadym: That's what I tried to point out in my first message—that you haven't defined `nextflow_config_content` as an attribute

fevac: made into a method

cg/meta/workflow/nf_analysis.py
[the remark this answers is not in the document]
fevac: the txt parser is modified in a general way so it can write from lists or strings. It's has an expanded functionality but it's still very general, nothing config-specific, so this should remain and doesn't have to be cleaned up. 
What this PR adds that will have to be updated is the `get_nextflow_config_content` method, which instead of returning a simple string, will have to parse and concat the servers configs

ivadym (approved):
👍

cg/meta/workflow/nf_analysis.py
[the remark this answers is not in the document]
fevac: resolving this conversation

fevac:
- Deployed stage
image

- Deployed prod
image

PR #2917, opened 2024-02-07
Use workflow instead of order_type

islean:
## Description

Currently order entries are created using order type as workflow. This is not quite correct since for example sars-cov-2 orders should have mutant as workflow but sars-cov-2 as order type.

### Fixed

- Use workflow instead of order type when creating orders.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/store/crud/create.py
+++ b/cg/store/crud/create.py
@@ -409,10 +409,11 @@ class CreateHandler(BaseHandler):
 
     def add_order(self, order_data: OrderIn):
         customer: Customer = self.get_customer_by_internal_id(order_data.customer)
+        workflow: str = order_data.samples[0].data_analysis
         order = Order(
             customer_id=customer.id,
             ticket_id=order_data.ticket,
-            workflow=order_data.order_type,
+            workflow=workflow,
         )
         session = get_session()
         session.add(order)


cg/store/crud/create.py
```diff
@@ -409,10 +409,11 @@ def add_sample_lane_sequencing_metrics(
 
     def add_order(self, order_data: OrderIn):
         customer: Customer = self.get_customer_by_internal_id(order_data.customer)
+        workflow: str = order_data.samples[0].data_analysis
```
seallard, line 412: Can we be sure that samples exist and that the data analysis is set on them?

seallard (approved):
Should work 👍 But not 100 % we can always assume the data analysis is available this way

cg/store/crud/create.py
islean: I interpret this as us having minimum one sample: 
image

and data_analysis is not nullable in the basemodel 
image


Maybe @henrikstranneheim can confirm that we can not have an order without samples, and that we can not have samples without data_analysis set?

karlnyr: We can always assume that all orders should have 1 sample, and that sample should always have a workflow. What I do not know is whether or not we validate this. But we probably do 👍

islean: Yeah, I have seen this validation at least in the new order portal. And if not, I would guess we would want to throw an error.

islean:
Tested on stage and now sars-covid orders are submitted with mutant as workflow.

islean:
Deployed to the VMs.

PR #2927, opened 2024-02-09
Ensure errors during token validation are logged

seallard:
## Description
We are not catching all possible exceptions when validating the oauth token, nice to have these errors in our logs.

### DIFF
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -11,6 +11,7 @@ import requests
 from flask import Blueprint, abort, current_app, g, jsonify, make_response, request
 from google.auth.transport import requests as google_requests
 from google.oauth2 import id_token
+from google.auth import exceptions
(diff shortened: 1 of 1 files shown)


cg/server/api.py
```diff
@@ -97,7 +98,7 @@ def before_request():
     jwt_token = auth_header.split("Bearer ")[-1]
     try:
         user_data = verify_google_token(jwt_token)
-    except ValueError as e:
+    except (exceptions.OAuthError, ValueError) as e:
```
islean, line 101: Nice catch! Won't raised exceptions be logged though?

islean:
Slightly confused - I thought there was an agreement that we wait for a review before merging?

henrikstranneheim:
Yes, we definitively should 👍

Issue #2937, opened 2024-02-13
Concatenate FASTQ files when delivery Microbial samples


Vince-janv:
## Description
The `cg deliver ticket` CLI command concatenates FASTQ files if the samples are micro-sample (see this [part](https://github.com/Clinical-Genomics/cg/blob/4e8a6363fc06fbf42436a4f2aab69d5f386886fd/cg/cli/deliver/base.py#L186-L190)).

The command run in the automation (`cg upload clinical-delivery`) does not perform the same action.

The customers are complaining and we need to introduce the same functionality.

## Suggested solution

If we can move the same logic into cg.meta.deliver.DeliverAPI.deliver_files it will be performed by both commands

## This can be closed when

The automatic delivery flow concatenates FASTQ files if the samples are microbial

PR #2941, opened 2024-02-13
Concatenate fastqs for microsalt

seallard:
## Description
Closes https://github.com/Clinical-Genomics/cg/issues/2937.

The delivery logic is so messy 😭 I probably need to bite the bullet and continue with https://github.com/Clinical-Genomics/cg/pull/2796 but I'm very tempted to just write it from scratch instead

### Fixed
- Concatenate fastqs for microsalt deliveries


seallard:
Will be addressed in a separate PR

PR #2951, opened 2024-02-14
Concatenate fastq files microsalt

seallard:
## Description
Closes https://github.com/Clinical-Genomics/cg/issues/2937.

The concatenation should not really be done in the delivery module since that module only is responsible for copying files from a bundle to the customers folder on hasta. The correct place would probably be in the microsalt pipeline or in another layer of logic. 

Since the delivery module is being refactored and microsalt is being replaced, I think it is ok to add it here for now.

### Added
- Concatenation of fastq files in the delivery module


cg/meta/deliver.py
```diff
@@ -210,6 +213,21 @@ def deliver_sample_files(
             f"There were {number_previously_linked_files} previously linked files and {number_linked_files_now} were linked for sample {sample_id}, case {case_id}"
         )
 
+        if self.delivery_type == Workflow.MICROSALT:
```
seallard, line 216: Will this check work as expected or will the `delivery_type` just be fastq?

cg/meta/deliver.py
```diff
@@ -210,6 +213,21 @@ def deliver_sample_files(
...
+            self.concatenate_fastqs(sample_directory=delivery_base, sample_name=sample_name)
+
+    def concatenate_fastqs(self, sample_directory: Path, sample_name: str):
+        if self.dry_run:
+            return
+        forward_out_path = Path(sample_directory, f"{sample_name}_R1.fastq.gz")
+        reverse_out_path = Path(sample_directory, f"{sample_name}_R2.fastq.gz")
```
seallard, line 223: The old logic added the order date to the concatenated file name. I did it at first, but it looks so ugly. I suggest we skip it unless the customers actually want it (which we do not know)

cg/services/fastq_file_service/fastq_file_service.py
```diff
@@ -0,0 +1,27 @@
...
+class FastqFileService:
+
+    def concatenate(
+        self,
+        fastq_directory: Path,
+        forward_output: Path,
+        reverse_output: Path,
+        remove_raw: bool = False,
+    ):
+        concatenate_forward_reads(directory=fastq_directory, output_file=forward_output)
```
seallard, line 19: Consider generating a unique file name in these calls, storing it and renaming them to the correct names after removing the raw fastqs

Vince-janv (approved):
Nice! Hopefully this code will not stay for too long 😉

cg/meta/deliver.py
Vince-janv: hmm, actually I think you're right. This will not work. The type will be FASTQ

seallard: I sent down the case and validate that the `data_analysis` set on it is `microsalt`.

Vince-janv: Yeah, if I don't misremember thats the same logic we have in the other CLI command that does it 👍

cg/services/fastq_file_service/utils.py
```diff
@@ -0,0 +1,82 @@
+from pathlib import Path
+import re
+import shutil
+import uuid
+
+from cg.services.fastq_file_service.exceptions import ConcatenationError
+
+
+def concatenate_forward_reads(directory: Path) -> Path:
+    fastqs = get_forward_read_fastqs(directory)
```
islean, line 10: Might be nice with a type hint here

cg/services/fastq_file_service/utils.py
```diff
@@ -0,0 +1,82 @@
...
+    if not fastqs:
+        return
+    file: Path = get_new_unique_file(directory)
+    concatenate(input_files=fastqs, output_file=file)
+    validate_concatenation(input_files=fastqs, output_file=file)
+    return file
+
+
+def get_new_unique_file(directory: Path) -> Path:
+    unique_id  = uuid.uuid4()
```
islean, line 30:
```suggestion
    unique_id = uuid.uuid4()
```

cg/services/fastq_file_service/utils.py
```diff
@@ -0,0 +1,82 @@
...
+    total_size: int = get_total_size(input_files)
+    concatenated_size: int = get_total_size([output_file])
+    if total_size != concatenated_size:
+        raise ConcatenationError
+
+
+def sort_files_by_name(files: list[Path]) -> list[Path]:
+    files_map = {file.name: file for file in files}
+    sorted_names = sorted(files_map.keys())
+    return [files_map[name] for name in sorted_names]
```
islean, line 72: I think this can be a one-liner:

`return sorted(files, lambda file: file.name)`

cg/meta/deliver.py
```diff
@@ -29,6 +30,7 @@ def __init__(
         sample_tags: list[set[str]],
         project_base_path: Path,
         delivery_type: str,
+        fastq_file_service: FastqFileService,
```
islean, line 33: Is there a reason for requiring this as input? Might have missed something but can't we just set it as FastqFileService() by default?

islean (commented):
Don't feel confident enough on MicroSalt to do more than comment 🙈

cg/meta/deliver.py
seallard: No, it is a poor design to couple dependencies like that. Even if it seems stupid now, the `FastqFileService` might grow in complexity, requiring more dependencies to be instantiated. As such, we want to keep the logic for instantiating it separate from this class to decouple them.

cg/services/fastq_file_service/fastq_file_service.py
Vince-janv: Which files are we afraid of overwriting? The ones we are concatenating?

PR #2956, opened 2024-02-16
Refactor sql filter functions to use get in name

henrikstranneheim:
### Added

-

### Changed

- Refactor sql filter functions to use get in their name. As they return a query and do not filter anything when called. Also, the end purpose is to get something from the database. The filtering is just a step in the query process.

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


henrikstranneheim:
> I think the reasoning behind having `filter` instead of get was to have a distiction between functions that pass on `Query` objects from functions that pass on, for example, `Sample` objects. I don't mind either way.

Or perhaps even better just omit `_ffilter` and it would read like: `TableFilter.BY_FIELD` @Clinical-Genomics/sysdev ?

ChrOertlin:
> > I think the reasoning behind having `filter` instead of get was to have a distiction between functions that pass on `Query` objects from functions that pass on, for example, `Sample` objects. I don't mind either way.
> 
> Or perhaps even better just omit `filter_` and it would read like: `TableFilter.BY_FIELD` @Clinical-Genomics/sysdev ?

Reads very easy to me, `TableFilter` implied it being a filter already anyway. I think this could work.

diitaz93:
> > I think the reasoning behind having `filter` instead of get was to have a distiction between functions that pass on `Query` objects from functions that pass on, for example, `Sample` objects. I don't mind either way.
> 
> Or perhaps even better just omit `filter_` and it would read like: `TableFilter.BY_FIELD` @Clinical-Genomics/sysdev ?

I like the prefix filter in those functions. I think they perform filter functions, as they are wrappers of the `Query.filter()` function. I know we have the `get` terminology but as these filter functions are only called inside the CRUD, the rest of cg will still use the `get` functions from the CRUD and not use the filter functions directly. I think it is convenient to name them as `filter` so you kind of know you can stack one after the other. 

Regarding the complete omission of prefix, that could work for the capital letters name, as Chris says it reads quite easily, but I think the original function name should still keep the `filter_` prefix.

henrikstranneheim:
Continued here: https://github.com/Clinical-Genomics/cg/pull/2979

PR #2960, opened 2024-02-19
(Streamline delivery) Add OrderCase table

islean:
## Description

We need to track what cases are included in an order, and since a case can be included in several orders, this will be a many-to-many relationship, similar to the case-sample connection. This PR adds an OrderCase table, which should allow us to track the cases in an order.

### Added

- OrderCase table

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b add-order-case-table -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/alembic/versions/2024_02_19_43eb680d6181_add_order_case_table.py
+++ b/alembic/versions/2024_02_19_43eb680d6181_add_order_case_table.py
@@ -1,0 +1,31 @@
+"""Add order case table
+
+Revision ID: 43eb680d6181
+Revises: d241d8c493fb
+Create Date: 2024-02-19 10:13:21.075891
+
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "43eb680d6181"
+down_revision = "d241d8c493fb"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "order_case",
+        sa.Column("id", sa.Integer, nullable=False),
+        sa.Column("order_id", sa.ForeignKey("order.id", ondelete="CASCADE"), nullable=False),
+        sa.Column("case_id", sa.ForeignKey("case.id", ondelete="CASCADE"), nullable=False),
+        sa.UniqueConstraint("order_id", "case_id", name="_order_case_uc"),
+    )
+
+
+def downgrade():
+    op.drop_table(table_name="order_case")
--- a/alembic/versions/2024_02_19_43eb680d6181_add_order_case_table.py
+++ b/alembic/versions/2024_02_19_43eb680d6181_add_order_case_table.py
@@ -21,8 +21,8 @@ def upgrade():
     op.create_table(
         "order_case",
         sa.Column("id", sa.Integer, nullable=False),
-        sa.Column("order_id", sa.Integer, nullable=False),
-        sa.Column("case_id", sa.Integer, nullable=False),
+        sa.Column("order_id", sa.Integer, nullable=False, index=True),
+        sa.Column("case_id", sa.Integer, nullable=False, index=True),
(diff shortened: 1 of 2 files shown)


islean:
Deployed on stage with the following table created:

image

image

image

image

cg/store/models.py
```diff
@@ -884,9 +895,31 @@ class Order(Model):
...
+
+class OrderCase(Model):
+    __tablename__ = "order_case"
+    __table_args__ = (UniqueConstraint("order_id", "case_id", name="_order_case_uc"),)
+
+    id = Column(types.Integer, primary_key=True, unique=True, nullable=False)
+    order_id = Column(ForeignKey("order.id", ondelete="CASCADE"), nullable=False)
+    case_id = Column(ForeignKey("case.id", ondelete="CASCADE"), nullable=False)
+    case = orm.relationship(Case, back_populates="order_links")
+    order = orm.relationship(Order, back_populates="links")
```
Vince-janv, line 925: Not sure what the preferred way is. But for pure link-tables you can define them as line 34-51. Then you can navigate directly between the two models without going via the link model

alembic/versions/2024_02_19_43eb680d6181_add_order_case_table.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def upgrade():
+    op.create_table(
+        "order_case",
+        sa.Column("id", sa.Integer, nullable=False),
+        sa.Column("order_id", sa.Integer, nullable=False, index=True),
+        sa.Column("case_id", sa.Integer, nullable=False, index=True),
+        sa.ForeignKeyConstraint(["order_id"], ["order.id"], ondelete="CASCADE"),
+        sa.ForeignKeyConstraint(["case_id"], ["case.id"], ondelete="CASCADE"),
```
Vince-janv, line 27: Eh, I'm not super familiar with the CASCADE option but I would not like an order to be deleted if one of the cases are 😅 Are you sure "CASCADE" should be set on both these lines?

Vince-janv (commented):
I don't think we need to create an OrderCase class for this 🤔

alembic/versions/2024_02_19_43eb680d6181_add_order_case_table.py
islean: As I understand it, this would delete the link entry, and we would be blocked from deleting both orders and cases with entries in this table without it. I just modeled it after the CaseSample table though, should probably be checked.

cg/store/models.py
islean: Nice, I will change it!

islean:
With rework:

image

islean:
Current behaviour with cascading: Deleting a case or order deletes entries in the case_order-table, but does not reach the other table, i.e. deleting a case does not delete an order, deleting an order does not delete a case and deleting an entry in order_case does not cascade anywhere

cg/store/models.py
islean: Changed!

seallard (approved):
LGTM 👍

alembic/versions/2024_02_19_43eb680d6181_add_order_case_table.py
islean: Checked!

islean:
Deployed to hasta and the vms:

```
repository is clean
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/proj/production
done.
Log deploy... done.
cg, version 59.2.0
[isak.ohlsson@hasta:/home/proj/production] [P_base] 3m6s $ 
```

Issue #2966, opened 2024-02-20
Newly started microsalt analysis displayed as completed before finishing


seallard:
## Description
Re-started a case with this command cg workflow `microsalt start <case_id>`, and it showed up in TB as pending and then completed before the command finished.

![image](https://github.com/Clinical-Genomics/cg/assets/33596072/0023cdcf-c6b4-4e86-a860-948d5a649432)

seallard:
When looking at the data, it seems like the jobs from the previous failed analysis somehow show up as the jobs for the restarted analysis.

islean:
The screenshot's resolution is a bit grainy. Can't really see what it says...

islean:
But also, it needs to be anonymized.

seallard:
> But also, it needs to be anonymized.

We really need to make this repo private, reporting bugs in a meaningful way is impossible now

islean:
> > But also, it needs to be anonymized.
> 
> We really need to make this repo private, reporting bugs in a meaningful way is impossible now

That should indeed be done, but as is, we are stuck with anonymizing data.

Vince-janv:
### Hypothesis:
1. MicroSALT writes the slurm_ids file when it is completed. In this case 30 min after start. 
2. The config_path column for the two analyses in trailblazer for shiningvervet point to the same path. 
3. The slurm_ids file from the first microSALT analysis is still there.
4. Trailblazer scan is run before the second analysis overwrites the slurm_ids file of the first.
5.  Trailblazer will then parse all the old jobs and set the status to completed since those are done.

seallard:
Yes, that is correct. Looking at the analysis entries for the case - the paths are identical and they point to the same `/home/proj/production/microbial/results/reports/trailblazer/<project_id>_slurm_ids.yaml`.

As for point one, microsalt writes the file before.

PR #2967, opened 2024-02-20
Remove existing slurm job ids

seallard:
## Description
Closes https://github.com/Clinical-Genomics/cg/issues/2966.
When a microsalt analysis is started, this fix ensures the previous job id:s are removed.

Fixing this properly would involve restructuring the microsalt output, but I don't think the effort of doing that is justified since it is supposed to be replaced.

### Fixed
- Old jobs being tracked for restarted microsalt analyses

### DIFF
--- a/cg/meta/workflow/microsalt/microsalt.py
+++ b/cg/meta/workflow/microsalt/microsalt.py
@@ -89,13 +89,16 @@ class MicrosaltAnalysisAPI(AnalysisAPI):
 
     def get_job_ids_path(self, case_id: str) -> Path:
         project_id: str = self.get_project_id(case_id)
-        return Path(
+        job_ids_path = Path(
             self.root_dir,
             "results",
             "reports",
             "trailblazer",
             f"{project_id}_slurm_ids{FileExtensions.YAML}",
         )
+        if job_ids_path.exists():
+            job_ids_path.unlink()
+        return job_ids_path
 
     def get_deliverables_file_path(self, case_id: str) -> Path:
         """Returns a path where the microSALT deliverables file for the order_id should be
--- a/cg/meta/workflow/microsalt/microsalt.py
+++ b/cg/meta/workflow/microsalt/microsalt.py
@@ -96,9 +96,12 @@ class MicrosaltAnalysisAPI(AnalysisAPI):
             "trailblazer",
             f"{project_id}_slurm_ids{FileExtensions.YAML}",
         )
+        self._ensure_old_job_ids_file_removed(job_ids_path)
+        return job_ids_path
+
+    def _ensure_old_job_ids_file_removed(self, job_ids_path: Path) -> None:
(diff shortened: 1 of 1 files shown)


cg/meta/workflow/microsalt/microsalt.py
```diff
@@ -89,13 +89,16 @@ def get_config_path(self, filename: str) -> Path:
 
     def get_job_ids_path(self, case_id: str) -> Path:
         project_id: str = self.get_project_id(case_id)
```
ChrOertlin, line 91:
```suggestion
        """ Get job id path and unlink if path exists. """
        project_id: str = self.get_project_id(case_id)
```

cg/meta/workflow/microsalt/microsalt.py
```diff
@@ -89,13 +89,16 @@ def get_config_path(self, filename: str) -> Path:
...
-        return Path(
+        job_ids_path = Path(
             self.root_dir,
             "results",
             "reports",
             "trailblazer",
             f"{project_id}_slurm_ids{FileExtensions.YAML}",
         )
+        if job_ids_path.exists():
+            job_ids_path.unlink()
```
ChrOertlin, line 100: I think this adds more functionality to the function than what can be understood from its name. 

Also, can we add a test case that catches the bug behavior?

ChrOertlin (commented):
Can we add a test that catches this bug?

cg/meta/workflow/microsalt/microsalt.py
islean: Also, slightly unclear why it is there without more context? Maybe we could break it out into a mini-method `_ensure_old_job_ids_file_removed` or something?

seallard:
> Can we add a test that catches this bug?

I don't think that is meaningful given the circumstances and the type of bug this is

cg/meta/workflow/microsalt/microsalt.py
seallard: Extracted a method with a more easily understandable name 👍  Won't add a test for this bug I don't given the circumstances and the type of bug this is. Won't change the name because the caller should not care and this methods is part of the workflow interface

cg/meta/workflow/microsalt/microsalt.py
seallard: Extracted a helper method making the code self documenting

cg/meta/workflow/microsalt/microsalt.py
ChrOertlin: `get_job_ids_path` is a function commonly used in other `workflows`. The `get_job_ids_path` for microsalt now does not behave the same as the other 6 functions. I think this would warrant an added test so that in future if this code where to be changed the bug would be caught.

ChrOertlin (approved):

PR #2952, opened 2024-02-15
(Streamline delivery) Link new cases to orders

islean:
## Description

We have added a table called `order_case`, intended to link cases to the orders they are included in and vice versa. This PR creates entries in the order_case table upon order submission.

Addresses https://github.com/Clinical-Genomics/streamline-delivery/issues/23

### Added

- order_case are created upon order submission.

### How to prepare for test

- [ ] Ssh to relevant server (depending on type of change)
- [ ] Use stage: `us`
- [ ] Paxa the environment: `paxa`
- [ ] Install on stage (example for Hasta):
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b add_order_to_new_cases -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...
- [ ] Take a screenshot and attach or copy/paste the output.

## Review

- [ ] Tests executed by
- [ ] "Merge and deploy" approved by
Thanks for filling in who performed the code review and the test!

### This [version](https://semver.org/) is a

- [ ] **MAJOR** - when you make incompatible API changes
- [ ] **MINOR** - when you add functionality in a backwards compatible manner
- [ ] **PATCH** - when you make backwards compatible bug fixes or documentation/instructions

## Implementation Plan

- [ ] Document in ...
- [ ] Deploy this branch on ...
- [ ] Inform to ...

PR #2976, opened 2024-02-22
Use correct cases in the order_case table

islean:
## Description

In #2952 I wrote code on an assumption that a variable would always contain cases, when in reality it can contain samples or pools as well. This PR ensures that we only connect orders to cases included in new orders.

Order_case is not used anywhere yet, so should not cause any issues, but this should be patched before any functionality is bestowed upon the table.

Part of resolving https://github.com/Clinical-Genomics/streamline-delivery/issues/43.

### Fixed

- Use cases with the specified ticket_id for matching with a new order.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -136,7 +136,7 @@ def submit_order(order_type):
             user_mail=g.current_user.email,
         )
         order_service = OrderService(db)
-        order_service.create_order(order_data=order_in)
+        order_service.create_order(order_in)
 
     except (  # user misbehaviour
         OrderError,
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -136,7 +136,7 @@ def submit_order(order_type):
             user_mail=g.current_user.email,
         )
         order_service = OrderService(db)
-        order_service.create_order(order_data=order_in, cases=result["records"])
+        order_service.create_order(order_data=order_in)
 
     except (  # user misbehaviour
         OrderError,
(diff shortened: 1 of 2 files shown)


cg/server/api.py
```diff
@@ -136,7 +136,7 @@ def submit_order(order_type):
             user_mail=g.current_user.email,
         )
         order_service = OrderService(db)
-        order_service.create_order(order_data=order_in, cases=result["records"])
+        order_service.create_order(order_data=order_in)
```
seallard, line 139: Skip named argument

seallard (approved):
Nice 👍 The ticket will be set. I think this is the only way to do it given the current structure of the order flow. We need to refactor it but that is a separate project

Vince-janv (approved):
Good solution, but do we plan to keep the ticket field on cases?

islean:
> Good solution, but do we plan to keep the ticket field on cases?

No, but that is a later refactoring

islean:
Test with NIPT order worked on stage

islean:
Deployed to cg-services-stage and cg-prod-services.

PR #2977, opened 2024-02-23
add(janusapi)

ChrOertlin:
## Description

Adds JanusAPI client and collect qc metrics endpoint to cg.
requires: https://github.com/Clinical-Genomics/servers/pull/1281

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/apps/janus/api.py
```diff
@@ -0,0 +1,28 @@
...
+from cg.apps.janus.dto.create_collect_qc_request import CreateCollectQCRequest
+from cg.exc import JanusAPIError
+
+
+class JanusAPIClient:
+    def __init__(self, config: dict[str]):
+        self.host = config["janus"]["host"]
+
+    def get_qc_metrics(self, collect_qc_request: CreateCollectQCRequest):
+        endpoint: str = f"{self.host}/collect_qc_metrics"
```
seallard, line 17: Consider renaming this endpoint to follow [RESTful design](https://learn.microsoft.com/en-us/azure/architecture/best-practices/api-design#define-api-operations-in-terms-of-http-methods). 

Following REST standards, the endpoint would be a noun (like `/qc_metrics`). The action, which is to create qc metrics, is indicated by the type of request you send (`POST` here) and should not be specified in the name of the endpoint

cg/models/cg_config.py
```diff
@@ -421,6 +429,15 @@ def housekeeper_api(self) -> HousekeeperAPI:
             self.housekeeper_api_ = housekeeper_api
         return housekeeper_api
 
+    @property
+    def janus_api(self) -> JanusAPIClient:
+        janus_api = self.__dict__.get("janus_api_")
+        if janus_api is None:
+            LOG.debug("Instantiating janus api")
+            janus_api = JanusAPIClient(config=self.dict())
```
seallard, line 437: Consider only passing the specific data the janus client needs instead of everything. It is better to be explicit when composing an object

cg/apps/janus/api.py
```diff
@@ -0,0 +1,28 @@
...
+import requests
+from requests import Response
+
+from cg.apps.janus.dto.create_collect_qc_request import CreateCollectQCRequest
+from cg.exc import JanusAPIError
+
+
+class JanusAPIClient:
+    def __init__(self, config: dict[str]):
+        self.host = config["janus"]["host"]
```
seallard, line 14: Only pass the host as a string

cg/apps/janus/api.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def get_qc_metrics(self, collect_qc_request: CreateCollectQCRequest):
+        endpoint: str = f"{self.host}/collect_qc_metrics"
+        response = requests.post(endpoint, data=collect_qc_request)
+        if response.status_code == HTTPStatus.OK:
+            return response.json()
```
seallard, line 20: Parse this response json into a pydantic model and return that instead

cg/apps/janus/api.py
```diff
@@ -0,0 +1,28 @@
+"""The JanusAPIClient."""
```
seallard, line 1: Consider creating a new directory called clients and placing this one in there. This is a new specific thing, I think we should separate it from the existing apps (even though some of the existing apps basically are web api clients).

seallard (approved):
Nice! Looks great! Some additional design suggestions

cg/apps/janus/api.py
```diff
@@ -0,0 +1,28 @@
...
+from cg.exc import JanusAPIError
+
+
+class JanusAPIClient:
+    def __init__(self, config: dict[str]):
+        self.host = config["janus"]["host"]
+
+    def get_qc_metrics(self, collect_qc_request: CreateCollectQCRequest):
+        endpoint: str = f"{self.host}/collect_qc_metrics"
+        response = requests.post(endpoint, data=collect_qc_request)
```
seallard, line 18: The request data needs to be serialized, you need to do
```python
data = collect_qc_request.model_dump_json()
```

islean (approved):
Looks nice!

cg/apps/janus/api.py
seallard: Yes, we do not want the clients to return a json string. It really needs to be parsed into a model

cg/apps/janus/api.py
```diff
@@ -0,0 +1,28 @@
...
+        if response.status_code == HTTPStatus.OK:
+            return response.json()
+        self._handle_errors(response)
+
+    @staticmethod
+    def _handle_errors(response: Response):
+        if 400 <= response.status_code < 500:
+            raise JanusAPIError(f"Client error: {response.status_code}")
+        elif 500 <= response.status_code:
+            raise JanusAPIError(f"Server error: {response.status_code}")
```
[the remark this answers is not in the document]
seallard: I've added this to the pattern, basically it is better to abstract away the `HTTP` specific errors so that the user of the client does not need to know about them. A single base error for the client should be defined and then specific exceptions inheriting from that exception can be raised. Like `JanusClientInvalidAuthentication` or something

cg/apps/janus/dto/create_collect_qc_request.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class FilePathAndTag(BaseModel):
+    """Model for the file path and its associated tag."""
+
+    file_path: str
+    tag: str
+
+
+class CreateCollectQCRequest(BaseModel):
```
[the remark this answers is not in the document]
seallard: I think it should just be called `CreateQCMetrics` since it is a request used to create QC metrics

cg/apps/janus/api.py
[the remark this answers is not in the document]
islean: Hmm, I don't really see the value in this. Could you give an illuminating example of when it would make a user unhappy? Like in both of these cases you could just catch the parent class HTTPError?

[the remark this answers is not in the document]
seallard: Yes, it is true that we could just catch the HTTP error instead. But I think it is good design to map the errors like this because
- The user of the client does not need to know that the client uses HTTP
- If an exception bubbles up, it is better that its origin is clear rather than being just an HTTPexception

cg/models/cg_config.py
ChrOertlin: I think we have this on the cg-ep for the next meeting. Right now this follows convention how other apis are instantiated into the CGConfig.

cg/apps/janus/api.py
ChrOertlin: See above

cg/apps/janus/api.py
ChrOertlin: fixed

cg/apps/janus/dto/create_collect_qc_request.py
```diff
@@ -0,0 +1,20 @@
...
+    """Model for the qc collection request."""
+
+    case_id: str
+    sample_ids: list[str]
+    files: list[FilePathAndTag]
+    workflow: str
+    prep_category: str
```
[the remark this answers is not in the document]
ChrOertlin: Right I like the idea to have some more strict in model validations, I will try it

cg/apps/janus/api.py
ChrOertlin: fixed that

cg/apps/janus/api.py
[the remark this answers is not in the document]
ChrOertlin: Will use the `_handle_errors` for now as it adheres closest to the cg-ep. We can revert back if discussions alter the proposal.

tests/clients/janus/test_janus_api_client.py
```diff
@@ -0,0 +1,56 @@
...
+import requests
+from pytest_mock import MockFixture
+
+from cg.clients.janus.api import JanusAPIClient
+from cg.clients.janus.dto.create_qc_metrics_request import CreateQCMetricsRequest
+from cg.clients.janus.exceptions import JanusClientError
+
+
+def test_qc_metrics_successful(
+    mocker: MockFixture,
```
ChrOertlin, line 15: I resolved to MockFixture here since this package is already installed. "requests_mock" would introduce a new dependency

seallard: I think it is ok to introduce since it just is a dev depdency

cg/clients/janus/api.py
```diff
@@ -0,0 +1,28 @@
...
+from cg.clients.janus.exceptions import JanusClientError, JanusServerError
+
+
+class JanusAPIClient:
+    def __init__(self, config: dict[str]):
+        self.host = config["janus"]["host"]
+
+    def qc_metrics(self, collect_qc_request: CreateQCMetricsRequest) -> dict:
+        endpoint: str = f"{self.host}/collect_qc_metrics"
+        response = requests.post(endpoint, data=collect_qc_request.model_dump_json())
```
seallard, line 18: Function calls are easier to read when they do not contain other nested function calls. Consider extracting the `collect_qc_request.model_dump_json()` to a separate variable

cg/apps/janus/api.py
seallard: So is the flow like 
- cg collects raw metrics from pipelines
- cg sends raw metrics to janus
- arnold retrieves metrics from janus
Or did I miss something?

ChrOertlin: Arnold does not know when and what metrics to retrieve, that is handled by cg in conjunction with trailblazer and housekeeper. I thought about that as well, but I think we want to let cg handle communications between these things.

tests/clients/janus/test_janus_api_client.py
```diff
@@ -0,0 +1,58 @@
...
+    client: JanusAPIClient,
+    collect_qc_request_balsamic_wgs: CreateQCMetricsRequest,
+    janus_response: dict,
+):
+    # GIVEN a mocked response from the requests.post method that is not successfull
+    mocked_post = mocker.patch.object(requests, "post")
+    mocked_response = mocker.Mock()
+    mocked_response.status_code = HTTPStatus.NOT_FOUND
+    mocked_response.json.return_value = janus_response
+    mocked_post.return_value = mocked_response
```
seallard, line 52: You probably want to extract the mocked response to a fixture

tests/clients/janus/test_janus_api_client.py
ChrOertlin: Let's discuss on the standup monday then I can change it if needed after.

diitaz93 (approved):
Looks good 👍

tests/clients/janus/test_janus_api_client.py
ChrOertlin: fixed

cg/clients/janus/api.py
```diff
@@ -0,0 +1,29 @@
...
+from cg.clients.janus.exceptions import JanusClientError, JanusServerError
+
+
+class JanusAPIClient:
+    def __init__(self, config: dict[str]):
+        self.host = config["janus"]["host"]
+
+    def qc_metrics(self, collect_qc_request: CreateQCMetricsRequest) -> dict:
+        endpoint: str = f"{self.host}/collect_qc_metrics"
+        post_request_data: str = collect_qc_request.model_dump_json()
```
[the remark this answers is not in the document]
ChrOertlin: it is actaully considered a str

PR #2982, opened 2024-02-26
fix: Delivery of SKIP_MISSING bundle analysis

ivadym:
### Added

-

### Changed

-

### Fixed
- Delivery of FASTQ data delivery which does not have a case bundle

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/deliver/delivery_api.py
+++ b/cg/meta/deliver/delivery_api.py
@@ -200,10 +200,8 @@ class DeliveryAPI:
         )
 
     def _bundle_exists(self, bundle_name: str, workflow: str) -> bool:
-        if self.hk_api.last_version(bundle_name):
+        if self.hk_api.last_version(bundle_name) or self.ignore_missing_bundle(workflow):
             return True
-        if self.ignore_missing_bundle(workflow):
-            return False
         raise SyntaxError(f"Could not find any version for {bundle_name}")
 
     def ignore_missing_bundle(self, workflow: str) -> bool:


cg/meta/deliver/delivery_api.py
```diff
@@ -200,10 +200,8 @@ def _is_sample_deliverable(self, sample: Sample) -> bool:
         )
 
     def _bundle_exists(self, bundle_name: str, workflow: str) -> bool:
-        if self.hk_api.last_version(bundle_name):
+        if self.hk_api.last_version(bundle_name) or self.ignore_missing_bundle(workflow):
             return True
-        if self.ignore_missing_bundle(workflow):
-            return False
         raise SyntaxError(f"Could not find any version for {bundle_name}")
```
seallard, line 205: Should we stop raising this weird exception, log a warning and just return false instead?

islean: Maybe better to return False? At least I would expect a method with return type boolean to be able to return False somehow.

islean: Take it with a grain of salt though, I am not so well-versed in this flow.

cg/meta/deliver/delivery_api.py
```diff
@@ -200,10 +200,8 @@ def _is_sample_deliverable(self, sample: Sample) -> bool:
         )
 
     def _bundle_exists(self, bundle_name: str, workflow: str) -> bool:
-        if self.hk_api.last_version(bundle_name):
+        if self.hk_api.last_version(bundle_name) or self.ignore_missing_bundle(workflow):
```
seallard, line 203: Consider extracting the `ignore_missing_bundle` if you think it makes more sense to let the caller check it

seallard (approved):
Great! Double check the other spot where the `_bundle_exists` function is called and whether it expects the same behavior this refactoring introduces

cg/meta/deliver/delivery_api.py
seallard: This change means we will try to retrieve files from a bundle that does not exist (which causes the tests to fail). Not sure if that is what we want

ivadym:
https://github.com/Clinical-Genomics/cg/pull/2984

cg/meta/deliver/delivery_api.py
ivadym: The logic is a mess but we need to keep marking the delivery as failed for case files.  For sample files it's already in place.

PR #2984, opened 2024-02-26
Patch delivery bug

seallard:
## Description
Based on the old code, the intended logic seems to be
- If a bundle is missing for case but missing bundles should be ignored: just continue the flow.
- If a bundle is missing for a sample but missing bundles should be ignored: skip copying files for that sample.
- Only do the sample bundle check above if it is a fastq delivery - otherwise, just  redo the case bundle check 
- Otherwise, raise an exception if a bundle is missing.

### DIFF
--- a/cg/meta/deliver/delivery_api.py
+++ b/cg/meta/deliver/delivery_api.py
@@ -197,11 +197,11 @@ class DeliveryAPI:
         return sample_files
 
     def _is_case_deliverable(self, case: Case, workflow: str) -> bool:
-        bundle_exist: bool = self._bundle_exists(case.internal_id)
+        bundle_exists: bool = self._bundle_exists(case.internal_id)
         ignore_missing_bundle: bool = self.ignore_missing_bundle(workflow)
-        if not bundle_exist and not ignore_missing_bundle:
+        if not bundle_exists and not ignore_missing_bundle:
             raise SyntaxError(f"Could not find bundle {case.internal_id}")
-        return bundle_exist or ignore_missing_bundle
+        return bundle_exists or ignore_missing_bundle
 
     def _sample_passes_quality_check(self, sample: Sample) -> bool:
         return (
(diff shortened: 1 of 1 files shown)


cg/meta/deliver/delivery_api.py
```diff
@@ -147,13 +147,21 @@ def _get_deliverable_samples(self, case: Case, workflow: str) -> list[Sample]:
...
                 deliverable_samples.append(sample)
         return deliverable_samples
 
+    def is_sample_deliverable(self, case: Case, sample: Sample, workflow: str) -> bool:
+        bundle_name: str = get_bundle_name(case=case, sample=sample, workflow=workflow)
+        bundle_exists: bool = self._bundle_exists(bundle_name)
+        ignore_missing_bundle: bool = self.ignore_missing_bundle(workflow)
+
+        if not bundle_exists and not ignore_missing_bundle:
+            raise SyntaxError(f"Could not find bundle {bundle_name}")
```
ivadym, line 160: If we don't get any files it will raise an error in any case for the sample delivery (`MissingFilesError(f"Could not find any files for sample {sample.internal_id}")`). Not for the case though. That's why the other check is important to keep.

We can look into it later with a more intensive refactoring that this API needs

ivadym (commented):
👍

ivadym (approved):

cg/meta/deliver/delivery_api.py
```diff
@@ -147,13 +147,21 @@ def _get_deliverable_samples(self, case: Case, workflow: str) -> list[Sample]:
...
+
+        sample_passes_qc: bool = self._sample_passes_quality_check(sample)
+        return bundle_exists and sample_passes_qc
```
seallard, line 163: Is this return correct?

seallard: I think we should skip it since we cannot copy any files from the bundle if it does not exist

ivadym: We should raise an error, as you did, if the sample doesn't exist

PR #2986, opened 2024-02-26
Fix(Alembic) Remove imports from cg in alembic scripts

Vince-janv:
## Description
Alembic scripts are meant to be static, they should never change since they track the states of a database. Importing code from a repository makes the upgrade/downgrade dependant on the rest of the repository.

This PR removes any imports from cg into alembic scripts.

### Test:
I could downgrade to 2fdb42ba801a and back up again. Due to some poorly written scripts going further would require more changes...

### Fixed

- Removed all import from cg in the alembic scripts


islean:
Do we not want the tight coupling though? I thought that so long as we deploy alembic with cg, the imports will be in sync? Might be good to exclude the imports and so on from the linters though, and make sure that they are not included in the refactoring.

Vince-janv:
> Do we not want the tight coupling though? I thought that so long as we deploy alembic with cg, the imports will be in sync? Might be good to exclude the imports and so on from the linters though, and make sure that they are not included in the refactoring.
> 
> Or what errors are we seeing currently?

I'm not sure we want database changes tightly coupled to code changes. For example the StrEnum Pipeline being renamed to Workflow breaks about 10 revisions. Changes like that should not affect the possibility of rolling back database changes

islean:
> > Do we not want the tight coupling though? I thought that so long as we deploy alembic with cg, the imports will be in sync? Might be good to exclude the imports and so on from the linters though, and make sure that they are not included in the refactoring.
> > Or what errors are we seeing currently?
> 
> I'm not sure we want database changes tightly coupled to code changes. For example the StrEnum Pipeline being renamed to Workflow breaks about 10 revisions. Changes like that should not affect the possibility of rolling back database changes

You have convinced me 💪

islean (approved):
Nice 🔥

seallard (approved):
Great fix!

PR #2989, opened 2024-02-27
fix(ORM) Correct default for Flow cell status

Vince-janv:
## Description
A parenthesis mismatch led to the default value for Flowcell.status not being set

### Fixed

- Flowcell.status now has the default "ondisk" again


seallard (approved):
Sneaky syntax error 😅

PR #2990, opened 2024-02-27
add(collect qc metrics meta api)

ChrOertlin:
## Description

Adds collectqcmetrics meta api to cg.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/clients/janus/dto/create_qc_metrics_request.py
```diff
@@ -1,20 +1,7 @@
...
-from cg.constants import Workflow
-
-
-class SupportedWorkflow(StrEnum):
-    BALSAMIC: str = Workflow.BALSAMIC
-
-
-class SupportedPrepCategory(StrEnum):
-    TGA: str = "tga"
-    WGS: str = "wgs"
```
ChrOertlin, line 16: will move this validation to Janus, cg should not need to care about this just receive errors

cg/clients/janus/dto/create_qc_metrics_request.py
```diff
@@ -29,5 +16,5 @@ class CreateQCMetricsRequest(BaseModel):
     case_id: str
     sample_ids: list[str]
     files: list[FilePathAndTag]
-    workflow: SupportedWorkflow
-    prep_category: SupportedPrepCategory | None
+    workflow: str
+    prep_category: str
```
islean, line 20: So these should be allowed any value? Not only Workflow or PrepCategory or whatever the constants are called?

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+
+class CollectQCMetricsAPI(MetaAPI):
+    """Collect qc metrics API."""
+
+    def __init__(self, config: CGConfig):
+        super().__init__(config)
+        self.janus_api: JanusAPIClient = config.janus_api
+        self.arnold_api: ArnoldAPIClient = config.arnold_api
+
+    def get_qc_metrics_file_paths_for_case(self, case_id: str) -> list[File]:
```
islean, line 31: Should this return the paths or the whole File objects? Method name seems to suggest only paths.

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+        """Get the multiqc metrics files for a case."""
+        files: list[File] = self.housekeeper_api.get_files(
+            bundle=case_id, tags=["qc-metrics", "janus"]
+        ).all()
+        if not files:
+            raise HousekeeperFileMissingError(f"No qc metrics files found for {case_id}.")
+        return files
+
+    @staticmethod
+    def get_tag_to_include(tags: list[Tag]) -> str:
```
islean, line 41: Slight preference for "get_tag_name_to_include" but feel free to keep as is.

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+        """Get the file path and tag for the multiqc files."""
+        file_paths_and_tags: list[FilePathAndTag] = []
+        for file_path_and_tags in files:
+            file_path: str = file_path_and_tags.full_path
+            tag: str = self.get_tag_to_include(file_path_and_tags.tags)
+            file_paths_and_tags.append(FilePathAndTag(file_path=file_path, tag=tag))
+        return file_paths_and_tags
+
+    def get_prep_category(self, case_id: str):
+        """Get the prep category for a sample."""
```
islean, line 58: Should the doc string by "Get the prep category for a case"?

islean: Also might be nice with a type hint

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+        samples: list[Sample] = self.status_db.get_samples_by_case_id(case_id=case_id)
```
islean, line 59:
```suggestion
        samples: list[Sample] = self.status_db.get_samples_by_case_id(case_id)
```
Feel free to disregard

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+        return samples[0].prep_category
+
+    def create_qc_metrics_request(self, case_id: str) -> CreateQCMetricsRequest:
+        """Create a qc metrics request."""
+        files: list[File] = self.get_qc_metrics_file_paths_for_case(case_id)
+        file_paths_and_tags: list[FilePathAndTag] = self.get_file_paths_and_tags(files)
+        case: Case = self.status_db.get_case_by_internal_id(case_id)
+        workflow: str = case.data_analysis
+        sample_ids: Iterator[str] = self.status_db.get_sample_ids_by_case_id(case_id)
+        prep_category: str = self.get_prep_category(case.internal_id)
```
islean, line 69:
```suggestion
        prep_category: str = self.get_prep_category(case_id)
```

tests/meta/qc_metrics/conftest.py
```diff
@@ -0,0 +1,96 @@
...
+def file_with_right_tags() -> str:
+    return Path("tests", "fixtures", "meta", "qc_metrics", "file_with_right_tags.json").__str__()
+
+
+@pytest.fixture
+def file_without_right_tags() -> str:
+    return Path("tests", "fixtures", "meta", "qc_metrics", "file_without_right_tags.json").__str__()
+
+
+@pytest.fixture(scope="function")
```
islean, line 32: I think the scope is function by default but maybe someone else can confirm that

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+            raise HousekeeperFileMissingError(f"No qc metrics files found for {case_id}.")
+        return files
+
+    @staticmethod
+    def get_tag_to_include(tags: list[Tag]) -> str:
+        """Get the multiqc module file tag."""
+        tags_to_exclude: list[str] = ["qc-metrics", "janus", "multiqc"]
+        for tag in tags:
+            if tag.name not in tags_to_exclude:
+                return tag.name
```
islean, line 46: A bit unsure how the files are tagged here but do we know for certain that any tag apart from qc-metrics, janus and multiqc will suffice? Or why do we not choose a specific tag rather than take the first which is not one of them?

tests/meta/qc_metrics/test_collect_qc_metrics.py
```diff
@@ -0,0 +1,58 @@
...
+    for sample_id in qc_metrics_request.sample_ids:
+        assert sample_id == sample.internal_id
+    assert len(qc_metrics_request.files) == 1
+    file_path_and_tag: FilePathAndTag = qc_metrics_request.files[0]
+    assert file_path_and_tag.tag == "hsmetrics"
+    assert qc_metrics_request.prep_category == sample.prep_category
+    case: Case = collect_qc_metrics_api.status_db.get_case_by_internal_id(
+        case_id_with_single_sample
+    )
+    assert qc_metrics_request.workflow == case.data_analysis
```
islean, line 58: Consider creating a fixture for the expected qc_metrics_request and then assering that the qc_metrics_request created is the same as the expected.

islean (approved):
Looks good!

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+            bundle=case_id, tags=["qc-metrics", "janus"]
+        ).all()
+        if not files:
+            raise HousekeeperFileMissingError(f"No qc metrics files found for {case_id}.")
+        return files
+
+    @staticmethod
+    def get_tag_to_include(tags: list[Tag]) -> str:
+        """Get the multiqc module file tag."""
+        tags_to_exclude: list[str] = ["qc-metrics", "janus", "multiqc"]
```
ivadym, line 43:
```suggestion
        tags_to_exclude: list[str] = ["qc-metrics", "janus", "multiqc", "storage"]
```

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+        for tag in tags:
+            if tag.name not in tags_to_exclude:
+                return tag.name
+
+    def get_file_paths_and_tags(self, files: list[File]) -> list[FilePathAndTag]:
```
ivadym, line 48: "Why do we want to use tags and send them to Janus? Don't we only care about the paths there? It seems to me that this is just adding another layer of complexity and would become annoying to maintain if, for some reason, we have to change some tags. Wouldn't it be easier to just send a dictionary of files with descriptive keys? Maybe I'm missing something, but it seems to me we're adding some unnecessary logic.

cg/meta/qc_metrics/collect_qc_metrics.py
ivadym: And I assume you only want to send just one tag. What will happen if we need to add a pipeline specific tag to these multiqc files? Then, would we need to add it to this list? I'm a bit unsure why we don't decouple the tags from these files and leave this logic only in CG/hermes.

cg/meta/qc_metrics/collect_qc_metrics.py
ChrOertlin: These files are tagged with "qc-metrics", "multiqc", "janus" and a special file specific tag. Here I just want to give janus the file specific tag. It does not need to know anything else.

cg/meta/qc_metrics/collect_qc_metrics.py
ChrOertlin: The tags are used in janus to retrieve the proper file model and parsing function.

For example somalier, fastp require a special parsing function. Using the tag I can make link the functions to the files. Alternative would be to rely on file names.

cg/meta/qc_metrics/collect_qc_metrics.py
ChrOertlin: fair consideration, did not think about that. Uff tough one to solve. Will think about it

ivadym: It's unlikely that we'll want to use these files for something other than sending them to Janus, but I guess it's something to take into account 🤔

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+        """Get the file path and tag for the multiqc files."""
+        file_paths_and_tags: list[FilePathAndTag] = []
+        for file_path_and_tags in files:
```
Vince-janv, line 51: `file_paths_and_tags` and `file_path_and_tags` are almost indistinguishable from each other. Please rename one of them

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+            file_path: str = file_path_and_tags.full_path
+            tag: str = self.get_tag_to_include(file_path_and_tags.tags)
+            file_paths_and_tags.append(FilePathAndTag(file_path=file_path, tag=tag))
+        return file_paths_and_tags
```
Vince-janv, line 55: So here we basically parse through `Files` that have tags, and then ignore some tags and create a new object with the path and only one tag?

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+    def get_prep_category(self, case_id: str):
+        """Get the prep category for a sample."""
+        samples: list[Sample] = self.status_db.get_samples_by_case_id(case_id=case_id)
+        return samples[0].prep_category
+
+    def create_qc_metrics_request(self, case_id: str) -> CreateQCMetricsRequest:
+        """Create a qc metrics request."""
+        files: list[File] = self.get_qc_metrics_file_paths_for_case(case_id)
+        file_paths_and_tags: list[FilePathAndTag] = self.get_file_paths_and_tags(files)
+        case: Case = self.status_db.get_case_by_internal_id(case_id)
```
Vince-janv, line 66: Do we need case anywhere?

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+        )
+
+    def get_qc_metrics(self, case_id) -> dict:
+        """Get the qc metrics for a case."""
+        qc_metrics_request: CreateQCMetricsRequest = self.create_qc_metrics_request(case_id)
+        try:
+            qc_metrics: dict = self.janus_api.qc_metrics(qc_metrics_request)
+            return qc_metrics
+        except (JanusClientError, JanusServerError) as error:
+            LOG.info(f"Cannot collect qc metrics from Janus: {error}")
```
Vince-janv, line 85: Clean ⭐

Vince-janv (commented):
Cool stuff! Some comments about the juggling of files and tags

cg/meta/qc_metrics/collect_qc_metrics.py
ChrOertlin: below to get the workflow/data analysis

cg/meta/qc_metrics/collect_qc_metrics.py
ChrOertlin: yes

cg/meta/qc_metrics/collect_qc_metrics.py
ChrOertlin: fixed

cg/meta/qc_metrics/collect_qc_metrics.py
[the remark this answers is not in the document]
ChrOertlin: I changed it to be better

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,85 @@
...
+from cg.clients.janus.exceptions import JanusClientError, JanusServerError
+from cg.exc import HousekeeperFileMissingError
+from cg.meta.meta import MetaAPI
+from cg.models.cg_config import CGConfig
+from cg.store.models import Case, Sample
+
+LOG = logging.getLogger(__name__)
+
+
+class CollectQCMetricsAPI(MetaAPI):
```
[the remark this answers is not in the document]
ChrOertlin: Fixed that. Did not like it either so much.

cg/meta/qc_metrics/collect_qc_metrics.py
ChrOertlin: The tag is the tag name in itself, just the way how it is retrieved from HK it comes with a bunch of extra info. Will keep it as tag. since that is what we want ultimately

cg/constants/housekeeper_tags.py
```diff
@@ -208,3 +208,17 @@ class BalsamicProtectedTags:
...
+    tags_to_retrieve: list[str] = ["qc-metrics", "janus"]
+    multi_qc_file_tags: list[str] = [
+        "hsmetrics",
+        "dups",
+        "stats",
+        "alignmentsummarymetrics",
+        "fastp",
+        "somalier",
+        "insertsize",
+        "wgsmetrics",
```
ivadym, line 223: These are the tags we have in Hermes, right?
```
picard-alignment
picard-duplicates
picard-hs
picard-insert-size
picard-wgs
samtools-stats
somalier
picard-rnaseq
fastp
star
general-stats
```

cg/meta/qc_metrics/collect_qc_metrics.py
```diff
@@ -0,0 +1,92 @@
...
+    def get_tag_to_include(tags: list[Tag]) -> str:
+        """Get the multiqc module file tag."""
+        for tag in tags:
+            if tag.name in JanusTags.multi_qc_file_tags:
+                return tag.name
+
+    def get_file_paths_and_tags(self, files: list[File]) -> list[FilePathAndTag]:
+        """Get the file path and tag for the multiqc files."""
+        file_paths_and_tags: list[FilePathAndTag] = []
+        for file_path_and_tags in files:
```
ivadym, line 58:
```suggestion
        for file in files:
```

ivadym (approved):
Looks good! ⭐  This will make the tags lists less stringent

cg/meta/qc_metrics/collect_qc_metrics.py
Vince-janv: I'm not sure I like the logic. Can't we just give all our tags and janus can decide how to parse it?

cg/meta/qc_metrics/collect_qc_metrics.py
ChrOertlin: opted for hk_file

cg/meta/qc_metrics/collect_qc_metrics.py
ChrOertlin: It will boil down to the exact same logic just on the other end. I'd rather sent a clean request that does not need to be processed further on the other end.

PR #2991, opened 2024-02-27
Add downsampled case to same order as original

islean:
## Description

We now expect every order to have at least one entry in the order_case table, tying it to an order. This does not happen when samples are downsampled which eventually leads to crashes, for example when trying to add analyses to Trailblazer. This PR ties the downsampled case to the latest order the original case was included in.

### Fixed

- Downsampled cases are added to the same order as the latest order for the original case.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b patch-order-case-downsample -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
Do we know that the latest order of the original case is the correct one?

islean:
> Do we know that the latest order of the original case is the correct one?

That is the current behaviour so I say yes. See how ticket is set below:

image

ChrOertlin (approved):
Looks good!

islean:
Tested in stage:

`justhusky` was downsampled and new case `luckytermite` was created. It was coupled to `justhusky`'s latest ticket via order_case:

image

PR #3005, opened 2024-02-29
fix(LIMS) Remove hard-coded user

Vince-janv:
## Description
The LIMS interacting module authenticates itself with the credentials in the .env file. However it creates new projects as the user with ID=3 (apiuser). This hard-coded logic is nonsensical.

In this PR the module fetches its own user using the username from the .env file and uses that when creating new projects.

### Changed
- Researcher ID changed from 3, to whatever is provided in the .env file


islean (approved):
Nice catch!

islean (approved):
🔥

PR #3006, opened 2024-02-29
Fix handle pre-existing ticket

islean:
## Description

Orders with cases should not be able to have new cases added to them. Current behaviour makes it so that the sample and cases are persisted in the database though. Now we will raise an error instead.

### Fixed

- If there is an order with the given ticket_id, an error is raised upon order submission.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


seallard (changes requested):
Is this really what we want? Let me know if this is an urgent bug fix and needs to be patched. It seems very unclear to me though and maybe it would benefit to refine it business wise?

- Why do we allow customers to specify existing tickets in their orders?
- Are the customers billed separately again when they specify an existing ticket? Or not at all?
- If it is necessary to allow customers to specify existing tickets in their orders, does it not make more sense to create a separate order with the same ticket? I.e. drop the ticket uniqueness constraint in the order table

seallard (approved):
I think we need to change this soon. It seems strange to reuse old orders like this

cg/services/orders/order_service/order_service.py
```diff
@@ -36,7 +38,7 @@ def get_orders(self, orders_request: OrdersRequest) -> OrdersResponse:
 
     def create_order(self, order_data: OrderIn) -> OrderResponse:
         """Creates an order and links it to the given cases."""
-        order: Order = self.store.add_order(order_data)
+        order = self.store.add_order(order_data)
```
Vince-janv, line 41: Why do we remove the type hint?

Vince-janv (approved):
Love it 🙌

cg/services/orders/order_service/order_service.py
islean: Oh - remnant! Will bring back!

islean:
Tested and successfully blocked submission:

image

islean:
Deployed to cg-prod-services.

PR #3012, opened 2024-03-01
Support searching orders

seallard:
## Description
Closes https://github.com/Clinical-Genomics/streamline-delivery/issues/50.

### Added
- Support for searching orders

### DIFF
--- a/cg/store/filters/status_order_filters.py
+++ b/cg/store/filters/status_order_filters.py
@@ -72,8 +72,9 @@ def apply_order_filters(
     workflow: SortOrder = None,
     page: int = None,
     page_size: int = None,
-    sort_field: str = None,
-    sort_order: str = None,
+    sort_field: OrderSortField = None,
+    sort_order: SortOrder = None,
+    search: str = None,
 ) -> Query:
     for filter in filters:
         orders: Query = filter(
@@ -86,5 +87,6 @@ def apply_order_filters(
             ticket_id=ticket_id,
             sort_field=sort_field,
             sort_order=sort_order,
+            search=search,
         )
     return orders
--- a/cg/store/filters/status_order_filters.py
+++ b/cg/store/filters/status_order_filters.py
@@ -1,6 +1,6 @@
 from enum import Enum
 from typing import Callable
-from sqlalchemy import asc, desc
+from sqlalchemy import asc, desc, or_
 
 from sqlalchemy.orm import Query
 from cg.server.dto.orders.orders_request import OrderSortField, SortOrder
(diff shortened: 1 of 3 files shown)


cg/store/filters/status_order_filters.py
```diff
@@ -28,6 +28,20 @@ def filter_orders_by_ticket_id(orders: Query, ticket_id: int | None, **kwargs) -
     return orders.filter(Order.ticket_id == ticket_id) if ticket_id else orders
 
 
+def filter_orders_by_search(orders: Query, search: str | None, **kwargs) -> Query:
+    return (
+        orders.filter(
+            or_(
+                Order.id.startswith(search),
+                Order.customer.name.ilike(f"%{search}%"),
+                Order.ticket_id.ilike(f"%{search}%"),
```
seallard, line 37: I'm not 100 % sure how we want to do this (in terms of which operators to use,`begin`/`ilike` or something else)

cg/store/filters/status_order_filters.py
```diff
@@ -28,6 +28,19 @@ def filter_orders_by_ticket_id(orders: Query, ticket_id: int | None, **kwargs) -
     return orders.filter(Order.ticket_id == ticket_id) if ticket_id else orders
 
 
+def filter_orders_by_search(orders: Query, search: str | None, **kwargs) -> Query:
+    if search:
+        orders = orders.join(Order.customer)
+        orders.filter(
+            or_(
+                Order.id.startswith(search),
```
islean, line 36: Should this not be equals? Or would we want to search for 967 and get all 9670, 9671 etc back?

cg/store/filters/status_order_filters.py
```diff
@@ -28,6 +28,19 @@ def filter_orders_by_ticket_id(orders: Query, ticket_id: int | None, **kwargs) -
...
+                Customer.name.ilike(f"%{search}%"),
```
islean, line 37: Probably more relevant with internal id? Maybe both would be nice? cust000 etc are the internal_ids though

cg/store/filters/status_order_filters.py
Vince-janv: Personally I prefer `.contains()`. It feels the most pythonic and I don't think we need anything more advanced. Maybe case insensitivity is good though. Also `ilike` was giving me some issues a while ago

Vince-janv (approved):
Nice 👍

cg/store/filters/status_order_filters.py
islean: I don't really see the use in not setting == all around but on the other hand someone from production is probably better suited to answer this

islean (approved):
Looks good! I would probably to a more strict search but I don't know what would be most useful for production

cg/store/filters/status_order_filters.py
Vince-janv: ooh, I found the `icontains()` function which does what you expect 😁

islean: > Personally I prefer `.contains()`. It feels the most pythonic and I don't think we need anything more advanced. Maybe case insensitivity is good though. Also `ilike` was giving me some issues a while ago

So you don't like I like? 🤪

cg/store/filters/status_order_filters.py
Vince-janv: This is true

cg/store/filters/status_order_filters.py
seallard: Settled on exact matching for order id and ticket id. Using `icontains` for the customer internal id.
Will check with production if they want to adjust the search.

PR #3013, opened 2024-03-01
Remove unnecessary query param

seallard:
## Description
Closes https://github.com/Clinical-Genomics/streamline-delivery/issues/47.

Currently we have no use case where we want to exclude summaries, so I think it is worth to remove it for now to minimize confusion. Will need to remove in cigrid as well.

### Fixed
- Remove redundant query param


cg/services/orders/order_service/utils.py
```diff
@@ -24,6 +24,9 @@ def create_orders_response(
 def _add_summaries(orders: list[Order], summaries: list[OrderSummary]) -> list[Order]:
     order_map = {order.id: order for order in orders}
     for summary in summaries:
-        order = order_map[summary.order_id]
-        order.summary = summary
+        try:
+            order = order_map[summary.order_id]
+            order.summary = summary
+        except KeyError:
+            continue
```
Vince-janv, line 31: Which situation do the exceptions solve? Is there a missmatch between the number of orders and summaries?

cg/services/orders/order_service/utils.py
```diff
@@ -24,6 +24,9 @@ def create_orders_response(
 def _add_summaries(orders: list[Order], summaries: list[OrderSummary]) -> list[Order]:
     order_map = {order.id: order for order in orders}
     for summary in summaries:
-        order = order_map[summary.order_id]
-        order.summary = summary
+        try:
+            order = order_map[summary.order_id]
+            order.summary = summary
```
islean, line 29: When would a key error occur here?

Vince-janv (approved):
Yeah, with how quick the request to trailblazer was I think we can drop `include_summary`

islean (approved):
I like it! Why do we need to catch key error exceptions though? Does Trailblazer leave out fields with no entries?

cg/services/orders/order_service/utils.py
seallard: We can leave them out, but this one occurs if an analysis is tagged with an order that has been removed.

seallard: Good point, this one can be skipped. There will always be a matching summary and order

cg/services/orders/order_service/utils.py
seallard: It won't, good point

seallard:
> I like it! Why do we need to catch key error exceptions though? Does Trailblazer leave out fields with no entries?

Removed one check. The other check ensures the endpoint does not crash when trailblazer returns analyses which the corresponding order has been removed for. That should not occur in general, but I'd rather not cigrid breaks if it would happen at some point.

cg/services/orders/order_service/utils.py
seallard: This exception was hiding a bug from a function (`filter_orders_by_ids`) :+1:

seallard:
> I like it! Why do we need to catch key error exceptions though? Does Trailblazer leave out fields with no entries?

True, removed both of them :+1: We always create summaries for existing orders and retrieve the corresponding analyses summaries, so there should never be any such exception

PR #3014, opened 2024-03-04
Fix order search filter

seallard:
## Description
The search filter was missing a return.

### Fixed
- Order search filter


islean (approved):
LGTM!

PR #3023, opened 2024-03-06
Connect manually created cases to an order

islean:
## Description

Cases created via the CLI do not get an entry in the `order_case` table, which blocks analyses from being uploaded. This PR connects the case to the order specified via the ticket_id.

### Fixed

- Cases added via the CLI gets connected to the order related to the ticket_id.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b fix-add-case -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/cli/add.py
+++ b/cg/cli/add.py
@@ -280,6 +280,14 @@ def add_case(
         ticket=ticket,
     )
     order: Order = status_db.get_order_by_ticket_id(int(ticket))
+    if not order:
+        LOG.warning(f"No order found with ticket_id {ticket}")
+        click.confirm("No order found for the given ticket, proceed?", default=False, abort=True)
+        order = Order(
+            customer_id=customer.id,
+            ticket_id=int(ticket),
+            workflow=data_analysis,
+        )
     new_case.orders.append(order)
 
     new_case.customer: Customer = customer
--- a/cg/cli/add.py
+++ b/cg/cli/add.py
@@ -15,6 +15,7 @@ from cg.store.models import (
     CaseSample,
     Collaboration,
     Customer,
+    Order,
     Panel,
     Sample,
     User,
@@ -278,6 +279,8 @@ def add_case(
         priority=priority,
         ticket=ticket,
     )
+    order: Order = status_db.get_order_by_ticket_id(int(ticket))
+    new_case.orders.append(order)
 
     new_case.customer: Customer = customer
     status_db.session.add(new_case)
--- a/tests/cli/add/test_cli_add_family.py
+++ b/tests/cli/add/test_cli_add_family.py
@@ -1,7 +1,5 @@
 """Tests the CLI methods to add cases to the status database."""
 
-import datetime
-
 from click.testing import CliRunner
 
 from cg.cli.add import add
@@ -19,19 +17,12 @@ def test_add_case_required(
     cli_runner: CliRunner, base_context: CGConfig, helpers: StoreHelpers, ticket_id: str
 ):
     """Test to add a case using only the required arguments"""
-    # GIVEN a database with a customer and an order and a panel
+    # GIVEN a database with a customer and a panel
     disk_store: Store = base_context.status_db
 
     customer: Customer = helpers.ensure_customer(store=disk_store)
     customer_id = customer.internal_id
     panel: Panel = helpers.ensure_panel(store=disk_store)
-    helpers.add_order(
-        store=disk_store,
-        customer_id=customer_id,
-        ticket_id=int(ticket_id),
-        order_date=datetime.datetime.now(),
-        workflow=CLI_OPTION_ANALYSIS,
-    )
     panel_id = panel.name
     name = "case_name"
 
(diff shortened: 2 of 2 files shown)


cg/cli/add.py
```diff
@@ -278,6 +279,8 @@ def add_case(
         priority=priority,
         ticket=ticket,
     )
+    order: Order = status_db.get_order_by_ticket_id(int(ticket))
+    new_case.orders.append(order)
```
islean, line 283: Do we want support for creating an order if it is not existing already? Or is it better to get an error if that happens?

seallard: I think it is better if an error is thrown. Seems the most intuitive to me if orders should only are created when a customer places an order via the order portal

seallard (approved):
LGTM! Might need some hacking in the test suite to get them to pass

Side track: does it make sense to allow adding cases to an order after it has been placed?

islean:
> LGTM! Might need some hacking in the test suite to get them to pass
> 
> Side track: does it make sense to allow adding cases to an order after it has been placed? Maybe these cases are just for dev purposes

Probably not but obviously there are use cases for it currently. Maybe something to refine for the future?

cg/cli/add.py
ChrOertlin: I don't exactly know the extent of the new order flow to be honest. If a new order is created does that trigger something else somewhere, for example invoicing etc? Or are we planning to do so?

How is this used mostly? I guess in the context of the bug it was likely some development thing, should this be considered?

cg/cli/add.py
```diff
@@ -278,6 +279,8 @@ def add_case(
         priority=priority,
         ticket=ticket,
     )
+    order: Order = status_db.get_order_by_ticket_id(int(ticket))
```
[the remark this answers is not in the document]
ChrOertlin: I think we should take a look at this though. We might be representing the same thing in multiple tables as different types now.

cg/cli/add.py
islean: The Order table was introduced to track analyses in trailblazer which belonged to the same order/ticket. Given that we soon will switch ticketing system and that cases can belong to many tickets either way, it seemed reasonable to have a many-to-many solution. 

There are quite a lot of areas where we could use the Order entries, but as of yet, it is only used for new views in CiGRID/Trailblazer. I would like to have a project in the future where we consider other areas which would benefit from using the table though.

In short, a new order does not trigger anything, but hopefully we could find some use cases in the future.

islean: Regarding test cases, it is a good point. Currently there is no available option to create an order via the CLI, so the tester would need to either connect to the database or submit an orderform in stage. Maybe that speaks for the option that a missing order should be created? Maybe with a click.confirm and warning message?

cg/cli/add.py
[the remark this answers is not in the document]
ChrOertlin: I recored this here as an issue https://github.com/Clinical-Genomics/cg/issues/3024

cg/cli/add.py
ChrOertlin: Yes, I think so. 

> Regarding test cases, it is a good point. Currently there is no available option to create an order via the CLI, so the tester would need to either connect to the database or submit an orderform in stage. Maybe that speaks for the option that a missing order should be created? Maybe with a click.confirm and warning message?

Yes, I think so. But if orders will trigger something in future this would require all users to be aware of this and knowledge on other parts of the system. We likely need to document this somehow then.

islean: I guess? I mean it is slightly weird already that you can set a non-existent ticket on your case. Currently it is just a database model intended to contain data related to an order/ticket. Oh well, for a future discussion!

islean:
Tested adding to existing order and order_case entry was created:

image
image

Tested adding to non-existent order and got the option to proceed. Aborted if no confirmation was given:

```
θ65° 8d [hiseq.clinical@hasta:~] [S_base] 22s $ cg add case -a mip-dna -dd analysis -t 167890 cust000 justguppy
No order found with ticket_id 167890
No order found for the given ticket, proceed? [y/N]: N
Aborted!
θ67° 8d [hiseq.clinical@hasta:~] [S_base] 31s 1 $ 
```

When confirmed, it works and an order was created tied to the case:

```
θ68° 8d [hiseq.clinical@hasta:~] [S_base] 37s 1 $ cg add case -a mip-dna -dd analysis -t 167890 cust000 just-guppy
No order found with ticket_id 167890
No order found for the given ticket, proceed? [y/N]: y
polishedeagle: new case added
θ66° 8d [hiseq.clinical@hasta:~] [S_base] 25s $ 
```

islean:
Deployed to production:

```
repository is clean
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/hiseq.clinical
done.
Log deploy... done.
cg, version 59.10.8
θ65° 7d [hiseq.clinical@hasta:~] [P_base] 2m30s $
```

PR #3028, opened 2024-03-07
Unify the fastq concatenation and remove reduntant code

karlnyr:
## Description

The microbial customer reached out to me and described that it is important that they do not have the "R" in the read name. It would leave the clinicians renaming a lot of files manually which was a risk. I checked the cause and saw that in the new flow we added an "R". I created a new file name generator, and static methods to generate forward and reverse. I also removed a redundant check of whether or not data should be concatenated. I also slimmed down the DeliverTicketAPI to the bare minimum and removed tests that are not used any longer.

### Added

- Unified output names for concatenated files.

### Changed

- The DliverTicketAPI to have less features and use the same logic from the DeliverAPI

### Fixed

- The incorrect file name in the concatenation

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/deliver_ticket.py
```diff
@@ -135,48 +88,15 @@ def concatenate(self, ticket: str, dry_run: bool) -> None:
...
-                LOG.info(
-                    "QC PASSED: Total size for files used in concatenation match the size of the concatenated file"
-                )
-                self.remove_files(reads=same_direction)
-
-    def get_app_tag(self, samples: list) -> str:
-        app_tag = samples[0].application_version.application.tag
-        return app_tag
-
-    def check_if_concatenation_is_needed(self, ticket: str) -> bool:
```
ivadym, line 167: Why are we removing this check? Can't we just move it to the FastqFileService?

And I think it will be also a good idea to replace the one we have in clinical-delivery with this one based on tags, what do you think?
https://github.com/Clinical-Genomics/cg/blob/20b14ca20c459c038645d6999e42c7bb19f6bcde/cg/meta/deliver.py#L217-L218

cg/meta/deliver_ticket.py
```diff
@@ -135,48 +88,15 @@ def concatenate(self, ticket: str, dry_run: bool) -> None:
...
+            )
+            reverse_out_path = self.fastq_file_service.get_concatenated_reverse_read_output_path(
+                fastq_directory=dir_path, sample_name=dir_path.name
+            )
+            self.fastq_file_service.concatenate(
+                fastq_directory=dir_path,
+                forward_output=forward_out_path,
+                reverse_output=reverse_out_path,
+                remove_raw=True,
+            )
```
ivadym, line 102: I can't see a reason why `forward_output` and `reverse_output` paths should be passed as parameters for `concatenate`.

Would it be possible to integrate the logic for path extraction into the concatenate function? We already supply the fastq directory, and we could additionally provide the sample name. This will ensure the output file name consistency we are looking for between the clinical delivery and deliver ticket implementations

cg/services/fastq_file_service/fastq_file_service.py
```diff
@@ -31,3 +32,19 @@ def concatenate(
...
+            direction=ReadDirection.FORWARD,
+        )
+
+    @staticmethod
+    def get_concatenated_reverse_read_output_path(fastq_directory: Path, sample_name: str) -> Path:
+        return get_concatenated_read_output_path(
+            fastq_directory=fastq_directory,
+            sample_name=sample_name,
+            direction=ReadDirection.REVERSE,
+        )
```
ivadym, line 50: And then if we move the path logic to this class, this could be also moved to utils 🎈

ivadym (commented):
Great cleanup! 🧼 This will definitely simplify further refactoring ⭐

cg/constants/constants.py
```diff
@@ -267,6 +267,13 @@ class Strandedness(StrEnum):
     UNSTRANDED: str = "unstranded"
 
 
+class ReadDirection:
```
henrikstranneheim, line 270: Make it an IntEnum

cg/meta/deliver.py
```diff
@@ -220,8 +220,12 @@ def deliver_sample_files(
     def concatenate_fastqs(self, sample_directory: Path, sample_name: str):
         if self.dry_run:
             return
-        forward_out_path = Path(sample_directory, f"{sample_name}_R1.fastq.gz")
-        reverse_out_path = Path(sample_directory, f"{sample_name}_R2.fastq.gz")
+        forward_out_path = self.fastq_file_service.get_concatenated_forward_read_output_path(
```
henrikstranneheim, line 223:
```suggestion
        forward_target_path = self.fastq_file_service.get_concatenated_forward_read_target_path(
```
Or perhaps deliver?

cg/services/fastq_file_service/utils.py
```diff
@@ -32,11 +33,17 @@ def get_new_unique_file(directory: Path) -> Path:
...
 
 def get_forward_read_fastqs(fastq_directory: Path) -> list[Path]:
-    return get_fastqs_by_direction(fastq_directory=fastq_directory, direction=1)
+    return get_fastqs_by_direction(fastq_directory=fastq_directory, direction=ReadDirection.FORWARD)
+
+
+def get_concatenated_read_output_path(
+    fastq_directory: Path, sample_name: str, direction: int
+) -> Path:
+    return Path(fastq_directory, f"{sample_name}_{direction}.fastq.gz")
```
henrikstranneheim, line 42: Use FileExtension constant

henrikstranneheim (commented):
Nice 👍

cg/meta/deliver_ticket.py
```diff
@@ -135,48 +88,15 @@ def concatenate(self, ticket: str, dry_run: bool) -> None:
...
-            if app_tag.startswith(prefix):
-                LOG.info(
-                    f"Identified {app_tag} as application tag, i.e. the fastqs should be concatenated",
-                )
-                return True
-        LOG.info(
-            f"The following application tag was identified: {app_tag}, concatenation will be skipped",
-        )
-        return False
+            forward_out_path = self.fastq_file_service.get_concatenated_forward_read_output_path(
```
seallard, line 91: I think this design can be improved. The `get_concatenated_forward_read_output_path` and `get_concatenated_reverse_read_output_path` should not be a part of the `FastqFileService. 

They are internal implementation details specific to this workflow and makes the `FastqFileService` interface less clean.

seallard (changes requested):
Looks good :partying_face:  I think the design can be improved a little.

The logic for retrieving the specific naming convention used in the microbial pipelines for concatenated fastq files should not be in the `FastqFileService`. It makes the service tightly coupled to the microbial pipelines.

Move that logic elsewhere, preferably in the module for the microbial pipelines in cg or a utils function somewhere.

cg/meta/deliver_ticket.py
seallard: It makes the interface of the `FastqFileService` more general and usable. We should not make assumptions about what the names of the output concatenated files should be, that is the callers business. It is better design to provide the output names as parameters.

seallard: Imagine if we need to concatenate files elsewhere but we have other requirements for what the files should be named.

cg/meta/deliver_ticket.py
seallard: I don't  think we should move that logic into the `FastqFileService` concatenation method. It is a violation of the single responsible principle and creates a tight coupling between the service and the microbial workflows

ivadym: I see. If we end up using this module for FASTQ concatenation before starting the analysis (and replacing FastqHandler), then we won't be able to reuse it.

ivadym: But the question that remains is: why are we removing this check? Is it because this is a manual step that's only performed for specific pipelines that will always require concatenation?

cg/meta/deliver_ticket.py
karlnyr: I think I agree with Vadym here. I did not wish to change the logic of the FastqFileService to begin with, but I do not think we should have options for the concatenated filenames, they should be the same for all samples.

cg/services/fastq_file_service/fastq_file_service.py
karlnyr: Yes!

cg/meta/deliver_ticket.py
karlnyr: I think I followed what you and Vadym envisioned in the next push, resolving this for now!

seallard: If you want, I can refactor it later? It is jpoor design to couple the microbial naming requirements to the implementation of a general `FastqFileService`

cg/meta/deliver_ticket.py
seallard: If you want, I can refactor it later? It is poor design to couple the microbial naming requirements to the implementation of a general FastqFileService

islean: In the refactoring I guess we could do a solution with interfaces. Not really to familiar with this though.

cg/meta/deliver_ticket.py
karlnyr: I have an idea for it, no need 👍

cg/meta/deliver_ticket.py
karlnyr: Me and Vadym discussed this and it can be removed since we are currently doing double concatenation when executing the `cg deliver ticket` command

ivadym: We've clarified that this check is duplicated since it calls the delivery API instead of the delivery ticket, as I initially thought. Resolving!

cg/meta/deliver_ticket.py
ivadym: We use `FastqHandler` for `Mutant`, which concatenates files and gives them a specific name like `concatenated_<lane>_<flowcell>_....` If the pipeline parses this name and extracts information from it, and if we intend to deprecate `FastqHandler` in favor of `FastqService` in the future, then we probably need to define the output paths before calling `fastq_file_service.concatenate`

karlnyr:
> Looks good 🥳 I think the design can be improved a little.
> 
> The logic for retrieving the specific naming convention used in the microbial pipelines for concatenated fastq files should not be in the `FastqFileService`. It makes the service tightly coupled to the microbial pipelines.
> 
> Move that logic elsewhere, preferably in the module for the microbial pipelines in cg or a utils function somewhere.

I agree that we can move the naming from the service and into the utils. But the fact remains that we need to be consistent with our fastq naming. That means that EACH time we concatenate, the customer is supposed to expect one output format. This PR achieves that by harmonizing the concatenation in delivery. @ivadym also mentioned the FastqHandler, however, this API only handles fastq that is to be sent into the pipeline, which is vital to have separate from whatever we do in the delivery as each pipeline may or may not expect the fastq format to be in a certain format. Arguably, one could make a "Delivery" workflow, which then could have this concatenation, thoughts?

This PR now removes the double concatenation that was being done in the deliver ticket API. It also locks the delivery format to the one that we offer to customers today, thus making our flow consistent with what we offer our customers.

cg/services/fastq_file_service/utils.py
```diff
@@ -28,19 +30,43 @@ def concatenate_reverse_reads(directory: Path) -> Path | None:
...
-    return Path(directory, f"{unique_id}.fastq.gz")
+    return Path(directory, f"{unique_id}{FileExtensions.FASTQ}{FileExtensions.GZIP}")
 
 
 def get_forward_read_fastqs(fastq_directory: Path) -> list[Path]:
-    return get_fastqs_by_direction(fastq_directory=fastq_directory, direction=1)
+    return get_fastqs_by_direction(fastq_directory=fastq_directory, direction=ReadDirection.FORWARD)
+
+
+def get_concatenated_read_output_path(
```
seallard, line 40: Consider moving this logic to another module - we don't want to spread microbial pipeline logic into this module.

karlnyr: This is not microbial pipeline logic

cg/constants/constants.py
karlnyr: yes!

seallard (approved):
Looks good 👍 It would be good if the microbial file naming logic was moved from the `FastqFileService` module to another one related to the microbial pipelines

cg/services/fastq_file_service/utils.py
karlnyr: It is how we want to deliver all concatenated files to be named

seallard: Alright, lets keep it here then 👍 
From a design perspective, it might better to separate it into another service to keep the concerns separate. We can extract it if other naming standards are requested in the future

karlnyr: I agree! We can make the name better though as this is meant solely for delivery. @henrikstranneheim commented on that earlier as well!

seallard (approved):
Looks great! Nice to get rid of the duplication

cg/meta/deliver/deliver_ticket.py
```diff
@@ -0,0 +1,106 @@
+"""Module for deliver and rsync customer inbox on the HPC to customer inbox on the delivery
+server """
```
ivadym, line 2:
```suggestion
"""
Module for deliver and rsync customer inbox on the HPC to customer inbox on the delivery server.
"""
```

ivadym (approved):
💯 💯

cg/meta/deliver/deliver_ticket.py
```diff
@@ -0,0 +1,106 @@
...
+    def get_all_cases_from_ticket(self, ticket: str) -> list[Case]:
+        return self.status_db.get_cases_by_ticket_id(ticket_id=ticket)
+
+    def get_inbox_path(self, ticket: str) -> Path:
+        cases: list[Case] = self.get_all_cases_from_ticket(ticket=ticket)
+        if not cases:
+            raise CgError(
+                f"The customer id was not identified since no cases for ticket {ticket} was found"
+            )
+        customer_id: str = cases[0].customer.internal_id
```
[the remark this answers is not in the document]
karlnyr: I am not aiming to refactor the API in this ticker, just to harmonize how we concatenate files. I hope that is okay :D

karlnyr:
Tested in stage for both `cg deliver analysis -d fastq -t <ticket_number>` and `cg deliver ticket -d fastq -t <ticket_number>` and both generate the same concatenated file in delivery.

PR #3029, opened 2024-03-07
Add file-centered delivery API & delivered files to delivery report

ivadym:
## Description

Working with the current delivery API is not feasible. We need to add a new file-centered API to handle file uploads and track them through the delivery report. There is also a clinical need for reporting delivered files: https://github.com/Clinical-Genomics/BALSAMIC/issues/1247.

This new API is intended to gradually replace the existing delivery API, and this PR will focus solely on extracting delivered files and integrating them into the delivery report.

For reference, here's the previous implementation using the old delivery API: https://github.com/Clinical-Genomics/cg/pull/2957.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b add-get-delivered-files -a
    ```

### How to test

- [x] generate balsamic report

- [x] generate mip-dna report

- [x] generate rnafusion report

### Expected test outcome

- [x] Expected files to deliver are included in the report
- [x] Report takes into account different delivery types, displaying the files to deliver accordingly

Example of rendered report:
rna-analysis-scout
## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/delivery/delivery.py
```diff
@@ -0,0 +1,170 @@
...
+            analysis_sample_files=True,
+        )
+        return delivery_files
+
+    def convert_files_to_delivery_files(
+        self,
+        files: list[File],
+        case: Case,
+        source_id: str,
+        destination_id: str,
```
ChrOertlin, line 135: I think this would be a bit clearer without the "_id". I see why you opted for the "_id" suffix above since it comes from the sample_id and sample_name. I think just "source" and "destination" would put it into a clearer space though. This is flavour, up to you.

cg/meta/delivery/delivery.py
```diff
@@ -0,0 +1,170 @@
...
+    def get_analysis_case_tags_for_workflow(workflow: Workflow) -> list[set[str]]:
+        return PIPELINE_ANALYSIS_TAG_MAP[workflow]["case_tags"]
+
+    @staticmethod
+    def get_analysis_sample_tags_for_workflow(workflow: Workflow) -> list[set[str]]:
+        return PIPELINE_ANALYSIS_TAG_MAP[workflow]["sample_tags"]
+
+    @staticmethod
+    def is_sample_deliverable(sample: Sample, force: bool = False) -> bool:
+        """Return whether the sample is deliverable or not."""
```
ChrOertlin, line 167:
```suggestion
        """Return whether the sample is deliverable or not. A sample is deliverable if it passes qc or is external. The force parameter can be used to override checks."""
```

cg/meta/delivery/delivery.py
```diff
@@ -0,0 +1,170 @@
...
+        is_external: bool = sample.application_version.application.is_external
+        qc_pass: bool = sample.sequencing_qc
+        return is_external or qc_pass or force
```
ChrOertlin, line 170:
```suggestion
        return any(is_external or qc_pass or force)
```
Is this the same?

cg/models/cg_config.py
```diff
@@ -560,3 +564,21 @@ def trailblazer_api(self) -> TrailblazerAPI:
             api = TrailblazerAPI(config=self.dict())
             self.trailblazer_api_ = api
         return api
+
+    @property
+    def fastq_file_service(self) -> FastqFileService:
+        return FastqFileService()
+
+    @property
+    def delivery_api(self) -> DeliveryAPI:
```
ChrOertlin, line 573: 💯

cg/models/delivery/delivery.py
```diff
@@ -0,0 +1,12 @@
...
+
+from pathlib import Path
+
+from pydantic import BaseModel, FilePath
+
+
+class DeliveryFile(BaseModel):
+    """Model representing a file to be delivered."""
+
+    source_path: FilePath
```
ChrOertlin, line 11: Is there a difference between FilePath and Path?

cg/store/models.py
```diff
@@ -542,6 +542,11 @@ def samples(self) -> list["Sample"]:
         """Return case samples."""
         return self._get_samples
 
+    @property
+    def sample_ids(self) -> list[str]:
+        """Return case sample IDs."""
```
ChrOertlin, line 547:
```suggestion
        """Return sample internal ids for samples in a case."""
```

tests/apps/hk/test_file.py
```diff
@@ -664,3 +664,42 @@ def test_get_file_insensitive_path(bed_file: Path, populated_housekeeper_api: Ho
...
+
+def test_get_files_from_latest_version_by_list_of_tags(
+    populated_housekeeper_api: HousekeeperAPI, sample_id: str, fastq_file: Path, spring_file: Path
+):
+    """Test that a list of files are retrieved given a list of tags."""
+
+    # GIVEN a populated Housekeeper API and a list of sample tags
+    sample_tags: list[set[str]] = [{SequencingFileTag.FASTQ}, {SequencingFileTag.SPRING}]
+
+    # WHEN getting a list of files by list of tags
```
ChrOertlin, line 677:
```suggestion
    # WHEN getting a list of files by tags

```

tests/apps/hk/test_file.py
```diff
@@ -664,3 +664,42 @@ def test_get_file_insensitive_path(bed_file: Path, populated_housekeeper_api: Ho
...
+    assert fastq_file.name in file_names
+    assert spring_file.name in file_names
+
+
+def test_get_files_from_latest_version_by_list_of_tags_excluding_tags(
+    populated_housekeeper_api: HousekeeperAPI, sample_id: str, fastq_file: Path, spring_file: Path
+):
+    """Test that a list of files is retrieved from the latest version, excluding specific tags."""
+
+    # GIVEN a populated Housekeeper API and a list of tags to exclude
```
ChrOertlin, line 693:
```suggestion
    # GIVEN a populated Housekeeper API and tags to exclude

```

tests/apps/hk/test_file.py
```diff
@@ -664,3 +664,42 @@ def test_get_file_insensitive_path(bed_file: Path, populated_housekeeper_api: Ho
...
+        bundle_name=sample_id, tags=sample_tags
+    )
+
+    # THEN the expected sample files should be retrieved
+    file_names: list[str] = [Path(file.full_path).name for file in files]
+    assert fastq_file.name in file_names
+    assert spring_file.name in file_names
+
+
+def test_get_files_from_latest_version_by_list_of_tags_excluding_tags(
```
ChrOertlin, line 688:
```suggestion
def test_get_files_from_latest_version_by_tags_excluding_tags(

```

tests/apps/hk/test_file.py
```diff
@@ -664,3 +664,42 @@ def test_get_file_insensitive_path(bed_file: Path, populated_housekeeper_api: Ho
 
     # THEN the file is fetched correctly
     assert file
+
+
+def test_get_files_from_latest_version_by_list_of_tags(
```
ChrOertlin, line 669:
```suggestion
def test_get_files_from_latest_version_of_tags(

```

cg/apps/housekeeper/hk.py
```diff
@@ -399,7 +401,28 @@ def get_files_from_latest_version(self, bundle_name: str, tags: list[str]) -> Qu
         if not version:
             LOG.warning(f"Bundle: {bundle_name} not found in Housekeeper")
             raise HousekeeperBundleVersionMissingError
-        return self.files(version=version.id, tags=tags)
+        return self.files(version=version.id, tags=tags).all()
```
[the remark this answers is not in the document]
ivadym: surprisingly, there were not many uses of this function, and those that were using it were either expecting a list of files already, or I refactored the query. but it's good to keep it in mind

cg/models/delivery/delivery.py
ivadym: Yes! A `FilePath` must point to an existing file

cg/meta/delivery/delivery.py
ivadym: Any expects an iterable, so it would be the same if we define it as `any([is_external, qc_pass, ...])`. But it just adds more complexity

cg/meta/delivery/delivery.py
ivadym: Yes, these parameters are confusing... Perhaps it would be clearer with `internal_id` & `external_id`, as we want to use the name provided by the customer for reporting the files we deliver. I will rename them

cg/meta/delivery/delivery.py
```diff
@@ -0,0 +1,170 @@
...
+            )
+            delivery_files.extend(sample_delivery_files)
+        return delivery_files
+
+    def get_analysis_sample_delivery_files_by_sample(
+        self, case: Case, sample: Sample
+    ) -> list[DeliveryFile]:
+        """Return a list of analysis files to be delivered for a specific sample."""
+        sample_tags: list[set[str]] = self.get_analysis_sample_tags_for_workflow(case.data_analysis)
+        analysis_sample_tags = [tag | {sample.internal_id} for tag in sample_tags]
```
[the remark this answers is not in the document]
ivadym: I've renamed it and added a type hint. I don't think we need an additional function since it simply performs a union between sets (sample tags and the sample ID).

cg/meta/delivery/delivery.py
```diff
@@ -0,0 +1,170 @@
...
+            delivery_files.extend(fastq_delivery_files)
+        return delivery_files
+
+    def get_fastq_delivery_files_by_sample(
+        self, case: Case, sample: Sample, force: bool = False
+    ) -> list[DeliveryFile]:
+        """Return a list of fastq files to be delivered for a specific sample."""
+        delivery_files: list[DeliveryFile] = []
+        fastq_tags: list[set[str]] = self.get_analysis_sample_tags_for_workflow(Workflow.FASTQ)
+        if self.is_sample_deliverable(sample=sample, force=force):
```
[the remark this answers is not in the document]
ivadym: In the current implementation we're just skipping that fastq sample delivery if the sample is not deliverable. We'll end up doing the same here. But maybe what's it's missing is some logging. I've added it together with a pytest 👍

cg/meta/delivery/delivery.py
```diff
@@ -0,0 +1,170 @@
...
+    ):
+        self.delivery_path: Path = delivery_path
+        self.fastq_file_service: FastqFileService = fastq_file_service
+        self.housekeeper_api: HousekeeperAPI = housekeeper_api
+        self.store: Store = store
+
+    def link_delivery_files_to_inbox(self) -> None:
+        """Link files from Housekeeper bundle to the customer's inbox."""
+        raise NotImplementedError
+
```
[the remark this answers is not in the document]
ivadym: Good to know!

cg/apps/housekeeper/hk.py
```diff
@@ -399,7 +401,28 @@ def get_files_from_latest_version(self, bundle_name: str, tags: list[str]) -> Qu
...
+        filtered_files: list[File] = []
+        if not tags:
+            return filtered_files
+        files: list[File] = self.get_files_from_latest_version(bundle_name=bundle_name)
+        for file in files:
+            file_tags: set[str] = {tag.name for tag in file.tags}
+            if any(tag.issubset(file_tags) for tag in tags):
+                if excluded_tags and any(
+                    exclude_tag.issubset(file_tags) for exclude_tag in excluded_tags
+                ):
```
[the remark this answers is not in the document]
ivadym: I'll rethink this function. It might need a new name and some restructuring. The other queries aim to match the provided tag with the HK tag. But here I'm allowing it to be a subset, which could confuse someone trying to reuse it 🤔

henrikstranneheim (approved):
Massive work, beautiful ⭐

cg/apps/housekeeper/hk.py
```diff
@@ -164,7 +164,7 @@ def get_files_from_version(version: Version, tags: set[str]) -> list[File] | Non
     def get_file_from_version(version: Version, tags: set[str]) -> File | None:
         """Return the first file matching the given tags."""
         files: list[File] = HousekeeperAPI.get_files_from_version(version=version, tags=tags)
-        return files[0] if files else None
+        return next((file for file in files), None)
```
islean, line 167: What does the "(file for file in files)" do here? Turn the list into something more suitable for next()?

cg/apps/housekeeper/hk.py
```diff
@@ -662,3 +664,40 @@ def reset_retrieved_archive_data(self, files_to_reset: list[File]):
...
+        return filtered_files
+
+    def get_files_from_latest_version_with_tags(
+        self, bundle_name: str, tags: list[set[str]], excluded_tags: list[set[str]] | None = None
+    ) -> list[File]:
+        """
+        Return files from the latest version of a bundle matching provided tags. Files containing
+        any tag sets specified in the excluded_tags list will be excluded from the output.
+        """
+        files: list[File] = self.get_files_from_latest_version(bundle_name=bundle_name)
```
islean, line 697: Can you not just pass the tags in this call and not need the middle step with "self.get_files_with_tags"?

cg/apps/housekeeper/hk.py
```diff
@@ -662,3 +664,40 @@ def reset_retrieved_archive_data(self, files_to_reset: list[File]):
...
+        filtered_files: list[File] = []
+        for file in files:
+            file_tags: set[str] = {tag.name for tag in file.tags}
+            if any(tag.issubset(file_tags) for tag in tags):
+                filtered_files.append(file)
+        return filtered_files
+
+    @staticmethod
+    def get_files_without_excluded_tags(
+        files: list[File], excluded_tags: list[set[str]]
```
islean, line 680:
```suggestion
    def get_files_without_tags(
        files: list[File], tags: list[set[str]]
```
I get confused by the "excluded" here 😅

cg/apps/housekeeper/hk.py
ivadym: Not much... I had to change it because SonarCloud was complaining: https://github.com/Clinical-Genomics/cg/runs/22621038550.

Suggestions are welcomed because I didn't quite get why since we were already implementing the suggested solution 😓

cg/apps/housekeeper/hk.py
ivadym: No, I cannot. The `get_files_from_latest_version ` function calls `FILTER_FILES_BY_TAGS`, which returns the files that match the exact tag list. Here, we want all the files that contain the provided tags (i.e. provided tags are subset of file tags)

ivadym: Maybe it would it be clearer if we rename `get_files_with_tags` to `get_files_containing_tags`... 🤔

cg/apps/housekeeper/hk.py
ivadym: I think using `excluded_tags` makes it clearer and easier to distinguish from `get_files_with_tags`. 

And defining the tags as list of sets offers more flexibility for excluding files, like those matching two specific tags

ivadym: We'll hopefully merge the sample and case tags soon, and after that, I believe we could further simplify these functions

cg/meta/delivery/delivery.py
```diff
@@ -0,0 +1,181 @@
...
+        store: Store,
+    ):
+        self.delivery_path: Path = delivery_path
+        self.fastq_file_service: FastqFileService = fastq_file_service
+        self.housekeeper_api: HousekeeperAPI = housekeeper_api
+        self.store: Store = store
+
+    @staticmethod
+    def get_analysis_case_tags_for_workflow(workflow: Workflow) -> list[set[str]]:
+        return PIPELINE_ANALYSIS_TAG_MAP[workflow]["case_tags"]
```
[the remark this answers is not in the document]
ivadym: Yes, I've made the changes, but I'll stash them for the next PRs migrating the delivery API. I would prefer not to touch the old delivery API here

seallard:
What is the the overall intended design here?

From what I gather, we will create a module containing logic for identifying all files in a bundle that should be copied to the customers directory on Hasta based on the case workflow and data delivery?

cg/meta/delivery/delivery.py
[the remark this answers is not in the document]
henrikstranneheim: Great 👍

cg/apps/housekeeper/hk.py
islean: huh, weird. can't really see a reason to complain but it is what it is.

cg/apps/housekeeper/hk.py
islean: I think the unclearest part of this is that filtering on tags requires an exact match of the sets 😅 But oh well, then that is reasonable.

islean: But maybe `get_files_containing_tags` might make the distinction clearer

ivadym: Yes, I think it's clearer, I've changed it. Thanks for the review @islean!

ivadym:
> What is the the overall intended design here?
> 
> From what I gather, we will create a module containing logic for identifying all files in a bundle that should be copied to the customers directory on Hasta based on the case workflow and data delivery?
> 
> In the end, I think a design where the top call only consists of a ticket id or case id is desirable. Then we would retrieve the workflow and data delivery info from the case and simply copy the matching files from the bundle to the customers dir on Hasta. Alternatively just initiate the upload directly with the set of file paths.
> 
> I think it is a good idea to extract the logic for identifying which files to deliver to a separate layer 👍

Exactly, that's the idea: modularising the API so that ticket delivery ends up being processed at the case level

cg/apps/housekeeper/hk.py
```diff
@@ -662,3 +664,40 @@ def reset_retrieved_archive_data(self, files_to_reset: list[File]):
...
+    ) -> list[File]:
+        """Return files without specified tags."""
+        filtered_files: list[File] = []
+        for file in files:
+            file_tags: set[str] = {tag.name for tag in file.tags}
+            if all(not excluded_tag.issubset(file_tags) for excluded_tag in excluded_tags):
```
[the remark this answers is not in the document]
ivadym: I thought about updating `tags` and `excluded_tags` to `list[str]` together when we merge case and sample tags, but I can also do it here probably it's the best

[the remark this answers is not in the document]
ivadym: [44ee6fd](https://github.com/Clinical-Genomics/cg/pull/3029/commits/44ee6fd320ad39d042e7aa856c902ad1eb140d40)

ivadym:
deployed: 
Screenshot 2024-03-19 at 16 53 56

PR #3030, opened 2024-03-08
(Streamline delivery) Add delivery message endpoint for order

islean:
## Description

Closes https://github.com/Clinical-Genomics/streamline-delivery/issues/16. This PR adds an endpoint for fetching the delivery message for an order. It queries trailblazer for each case in the order and only generates the message for those with a "Completed" status.

### Added

- Endpoint for getting the delivery message for an order

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/delivery_message/utils.py
```diff
@@ -87,3 +89,11 @@ def validate_cases(cases: list[Case], case_ids: list[str]) -> None:
 
 def is_matching_order(cases: list[Case]) -> bool:
     return all([case.latest_order == cases[0].latest_order for case in cases])
+
+
+def get_cases_ready_for_delivery(trailblazer_api: TrailblazerAPI, order: Order) -> list[Case]:
+    return [
+        case
+        for case in order.cases
+        if trailblazer_api.get_latest_analysis_status(case.id) == AnalysisStatus.COMPLETED
```
islean, line 98: Do we want to query tb with order_id and status instead? Would require a new method in the TrailblazerAPI

seallard: I like that idea, sounds cleaner and more efficient! But stick to this for now if you want

cg/server/api.py
```diff
@@ -520,6 +520,21 @@ def get_order(order_id: int):
...
+@BLUEPRINT.route("/orders/<order_id>/delivery_message")
+def get_delivery_message_for_order(order_id: int):
+    """Return the delivery message for an order."""
+    try:
+        response: DeliveryMessageResponse = delivery_message_service.get_delivery_message_for_order(
+            order_id
+        )
+        response_dict: dict = response.model_dump()
+        return make_response(response_dict)
+    except IndexError as error:
```
seallard, line 532: This `IndexError` seems a little mystical to me, if you think it makes sense maybe a more specific exception can be raised

cg/services/delivery_message/utils.py
```diff
@@ -87,3 +89,11 @@ def validate_cases(cases: list[Case], case_ids: list[str]) -> None:
 
 def is_matching_order(cases: list[Case]) -> bool:
     return all([case.latest_order == cases[0].latest_order for case in cases])
+
+
+def get_cases_ready_for_delivery(trailblazer_api: TrailblazerAPI, order: Order) -> list[Case]:
```
seallard, line 94: Design: this logic could be placed in a separate service wrapping around the trailblazer API client. Consider refactoring if you think it is worth it

seallard (approved):
Looks good! Some design considerations, but not sure if they are worth the effort

cg/services/delivery_message/utils.py
ChrOertlin: Can a case be part of different orders? If so, not specifying the order could perhaps fetch something unwanted. That being said, I am not up to speed with the order logic.

tests/server/endpoints/test_delivery_message_endpoint.py
```diff
@@ -33,3 +37,30 @@ def test_get_delivery_message_matching_order(
...
+    "analysis_status, expected_status_code",
+    [
+        (AnalysisStatus.COMPLETED, HTTPStatus.OK),
+        (AnalysisStatus.RUNNING, HTTPStatus.PRECONDITION_FAILED),
+    ],
+)
+def test_get_delivery_message_order(
+    client: FlaskClient,
+    server_case: Case,
+    server_case_in_same_order: Case,
```
ChrOertlin, line 52: are these used at all below? 👴

ChrOertlin (approved):

cg/server/api.py
islean: Yeah, I will wrap it in something more descriptive!

cg/services/delivery_message/utils.py
islean: Similar to the housekeeper storage functions? @seallard

cg/services/delivery_message/utils.py
islean: Yeah, I guess we could have that unlucky scenario @ChrOertlin. Seems as if everything points to some refactoring!

tests/server/endpoints/test_delivery_message_endpoint.py
islean: Yes, but testing Flask is very weird 🙃 There must be better ways to fill the database, but specifying these imports do the trick as well. So basically this ensures that the order and the cases are in the database.

cg/services/delivery_message/utils.py
islean: Refactored

cg/services/delivery_message/delivery_message_service.py
```diff
@@ -21,3 +28,13 @@ def get_delivery_message(
         validate_cases(cases=cases, case_ids=case_ids)
         message: str = get_message(cases)
         return DeliveryMessageResponse(message=message)
+
+    def get_delivery_message_for_order(self, order_id: int) -> DeliveryMessageResponse:
+        order: Order = self.store.get_order_by_id(order_id)
+        if not order:
+            raise OrderNotFoundError(f"Order with ID {order_id} not found.")
+        cases_ready_for_delivery: list[Case] = get_cases_ready_for_delivery(
```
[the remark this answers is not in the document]
islean: Changed!

cg/apps/tb/api.py
```diff
@@ -178,3 +178,12 @@ def get_summaries(self, order_ids: list[int]) -> list[AnalysisSummary]:
         response = self.query_trailblazer(command=endpoint, request_body={}, method=APIMethods.GET)
         response_data = SummariesResponse.model_validate(response)
         return response_data.summaries
+
+    def get_analyses_to_deliver(self, order_id: int) -> list[TrailblazerAnalysis]:
```
seallard, line 182: Design: this method is not an "atomic" operation and should be in a service instead of the web API client. Out of scope though

cg/services/delivery_message/delivery_message_service.py
```diff
@@ -1,22 +1,41 @@
...
+    def get_cases_message(
         self, delivery_message_request: DeliveryMessageRequest
     ) -> DeliveryMessageResponse:
         case_ids: list[str] = delivery_message_request.case_ids
+        return self._get_delivery_message(case_ids)
+
+    def get_order_message(self, order_id: int) -> DeliveryMessageResponse:
+        order: Order = self.store.get_order_by_id(order_id)
+        if not order:
+            raise OrderNotFoundError(f"Order with ID {order_id} not found.")
```
seallard, line 29: Design as we discussed: we can consider remapping exceptions in the top layer and enriching them with information there instead of spreading it through the code base. Out of scope

seallard (approved):
👍

cg/services/delivery_message/delivery_message_service.py
islean: For another PR

cg/apps/tb/api.py
islean: Another PR but I like the idea

cg/services/delivery_message/delivery_message_service.py
```diff
@@ -1,22 +1,41 @@
...
+        analyses: list[TrailblazerAnalysis] = self.trailblazer_api.get_analyses_to_deliver(order_id)
+        if not analyses:
+            raise OrderNotDeliverableError(
+                f"No analyses ready to be delivered for order {order_id}"
+            )
+        case_ids: list[str] = [analysis.case_id for analysis in analyses]
```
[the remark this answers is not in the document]
islean: @beatrizsavinhas Any input?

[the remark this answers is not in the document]
beatrizsavinhas: I think only one analysis should be fetched, yes. (The most recent completed analysis, right? Other analyses for the same case should either be due to some fix or manual restart, or they would be ordered by the customer in a different order.)

cg/services/delivery_message/utils.py
seallard: Yes, I don't like those either. They probably belong in an `AnalysisFileService` or something

seallard: But out of scope, if you think it makes sense and feel like doing it - extract it 👍

cg/services/delivery_message/delivery_message_service.py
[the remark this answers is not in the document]
islean: Oh wait, on second thought, the message would probably contain duplicated links. I will fix this

cg/services/delivery_message/utils.py
islean: I will think about it and implement in another PR!

cg/services/delivery_message/delivery_message_service.py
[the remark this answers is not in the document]
islean: Used sets instead

islean:
Deployed to cg-prod-services

PR #3041, opened 2024-03-18
Remove unused sample columns

seallard:
This PR removes the following columns from the sample table
- `sequence_start`
- `invoiced_at`
- `_is_external`

None of these fields are used in cg and they only contain null values in our databases.

EDIT: some of the columns contain a few entries. 

- 6 for `sequence_start` (2015-2023)
- 549 for `is_external` (all are from 2016/17)
- 1 for `invoiced_at` (from 2017)

## Blocked
Need to port the apptags so they are marked as external, production might need that info.


alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
```diff
@@ -0,0 +1,29 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "24c1e9e1476d"
+down_revision = "b2fcb2ada306"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.drop_column("sample", "sequence_start")
```
islean, line 21: image
We do have a few samples with this set

alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
```diff
@@ -0,0 +1,29 @@
...
+    op.drop_column("sample", "is_external")
```
islean, line 22: We do have samples with is_external set although the field is deprecated. Maybe okay to drop but good to double check?:

image

alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
seallard: Yes, I don't think it is significant though since they are so few and we don't use the field. I'm guessing these 5 were set manually.

What do you think?

alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
```diff
@@ -0,0 +1,29 @@
...
+    op.drop_column("sample", "invoiced_at")
```
islean, line 23: We have a single sample with this value set. Probably okay to just drop, but maybe good to check what that sample is about.

seallard: Since we don't use the field for anything, does it matter? I'll check it though

tests/meta/workflow/test_analysis.py
```diff
@@ -89,7 +89,6 @@ def test_is_flow_cell_check_applicable(mip_analysis_api: MipDNAAnalysisAPI, anal
     # GIVEN that no samples are down-sampled nor external
     for sample in case.samples:
         assert not sample.from_sample
-        assert not sample.is_external
```
islean, line 92: sample.is_external is a property though? So why should this be deleted?

alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
islean: I would say it is fine, but on the other hand, I don't consider myself to have the final say in this.

tests/meta/workflow/test_analysis.py
seallard: Because it is not used

alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
islean: I think it is most likely fine. Seems to be from 2017

tests/meta/workflow/test_analysis.py
karlnyr: it is used in the delivery? or is that based on the app tag?

islean: Oh, sorry, I thought it was a property which returned `sample.application_version.application.is_external`. My bad. We would probably want that as a property though 😅 But that is for another PR

seallard: Yes

alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
seallard: Good spot 👍

alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
seallard: Given it is not used anywhere, it is just bloat that pollutes are code and data bases. I guess we just need to make sure that no one is manually setting and reading this field. I'll ask production

alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
seallard: All of them are from 2016/2017 so I don't think it is a problem to remove the column.

alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
seallard: I just saw now that this is a different column. This PR drops the `_is_external` column.
Apparently, status db has a `is_external` column which is not defined in our model.

diitaz93 (approved):
It looks like a good idea 👍 I have checked and as you say, these attributes do not seem to be used anywhere, but better double-check with production

seallard:
Investigated the samples with `is_external` set. They are all associated with an application. No sample with `is_external=1` is associated with an application marked with `is_external=0`. However, 539 samples with `is_external=0` are associated with applications marked with `is_external=1`.

This means that we are good to go and can drop the column. Exported the relevant samples and their `is_external` value.

```sql
SELECT 
    s.id AS sample_id, 
    s.is_external AS sample_is_external, 
    a.is_external AS application_is_external, 
    a.tag AS application_name,
    s.ordered_at
FROM sample s
JOIN application_version av ON s.application_version_id = av.id
JOIN application a ON av.application_id = a.id
WHERE s.is_external IS NOT NULL
AND s.is_external != a.is_external;
```

alembic/versions/2024_03_18_24c1e9e1476d_remove_unused_sample_columns.py
seallard: No ones uses these fields

henrikstranneheim:
Great cleanup! 🧹

PR #2869, opened 2024-01-25
Add upload job to analysis in trailblazer

seallard:
## Description
This PR ensures the rsync jobs are associated with their corresponding analysis in trailblazer.
Part of the [implementation plan](https://github.com/Clinical-Genomics/cigrid-ui/issues/492#issuecomment-1905771598) for https://github.com/Clinical-Genomics/cigrid-ui/issues/492.

### Added
- Associate slurm rsync job with analysis in trailblazer

### This [version](https://semver.org/) is a
- [ ] **MINOR**

PR #3045, opened 2024-03-19
Associate rsync jobs with analysis

seallard:
## Description
Since https://github.com/Clinical-Genomics/cg/pull/2869 we associate the rsync jobs with their rsync analysis in trailblazer. This PR ensures the upload job is associated with the most recent analysis for the case as well.

This is the first step in removing rsync analyses completely since they only serve as a "wrapper" for the upload jobs. We can just track them in the upload job tab in the future instead
image

### Fixed
- Associate upload job with corresponding analysis


cg/services/slurm_upload_service/slurm_upload_service.py
```diff
@@ -36,6 +36,10 @@ def upload(self, upload_command: str, job_name: str, case_id: str):
         slurm_id: int = self.slurm_service.submit_job(job_config)
         self.trailblazer_api.add_upload_job_to_analysis(slurm_id=slurm_id, analysis_id=analysis.id)
 
+    def add_upload_job_to_analysis(self, slurm_id: int, case_id: int) -> None:
+        analysis: TrailblazerAnalysis = self.trailblazer_api.get_latest_completed_analysis(case_id)
+        self.trailblazer_api.add_upload_job_to_analysis(slurm_id=slurm_id, analysis_id=analysis.id)
```
seallard, line 41: Not 100 % convinced this logic belongs here in the `SlurmUploadService`

cg/services/slurm_upload_service/slurm_upload_service.py
```diff
@@ -36,6 +36,10 @@ def upload(self, upload_command: str, job_name: str, case_id: str):
         slurm_id: int = self.slurm_service.submit_job(job_config)
         self.trailblazer_api.add_upload_job_to_analysis(slurm_id=slurm_id, analysis_id=analysis.id)
 
+    def add_upload_job_to_analysis(self, slurm_id: int, case_id: int) -> None:
+        analysis: TrailblazerAnalysis = self.trailblazer_api.get_latest_completed_analysis(case_id)
```
seallard, line 40: Does this assumption hold? That the most recent analysis for a case is the analysis we want to associate the upload job with?

cg/services/slurm_upload_service/slurm_upload_service.py
seallard, line 41: Is there a unique constraint on the `slurm_id` column for jobs in trailblazer?

seallard: No, so there will not be any issues with temporarily creating two upload jobs with the same slurm id before we remove the rsync analyses

islean: Maybe a good start for a wrapper around the TrailblazerAPI?

islean (approved):
Looks good! I would assume that the analysis is correct (seemed that way on KS as well) but maybe get a second opinion.

cg/cli/upload/clinical_delivery.py
```diff
@@ -81,6 +82,10 @@ def upload_clinical_delivery(context: click.Context, case_id: str, dry_run: bool
             ticket=case.latest_ticket,
         )
         trailblazer_api.add_upload_job_to_analysis(analysis_id=analysis.id, slurm_id=job_id)
+
+        slurm_upload_service: SlurmUploadService = context.obj.slurm_upload_service
+        slurm_upload_service.add_upload_job_to_analysis(slurm_id=job_id, case_id=case_id)
```
Vince-janv, line 87: I feel like we're doing the same thing twice here. Should we maybe remove the call to trailblazer_api

cg/services/slurm_upload_service/slurm_upload_service.py
Vince-janv: Yes, I see some edge-cases where that might not be true but it's the best way we have

cg/cli/upload/clinical_delivery.py
Vince-janv: Ah, never mind. So before we added the upload job to the RSYNC-analysis, and now we add it to both the "real" analysis and the rsync one?

Vince-janv (approved):
Nice 💪  I think the hard thing will be to tackle `cg deliver ticket` but that's a beast for another PR

cg/cli/upload/clinical_delivery.py
seallard: Yes

cg/services/slurm_upload_service/slurm_upload_service.py
seallard: Alright 👍

cg/services/slurm_upload_service/slurm_upload_service.py
seallard: Added in https://github.com/Clinical-Genomics/cg/pull/3045/commits/a511368e921f17d2477987f0230c0a5ad22ec484

PR #3049, opened 2024-03-19
Fix sample views

seallard:
Missed removing three dropped columns from the status db views.
Deployed to stage and verified the views work.


islean (approved):
💪

PR #3050, opened 2024-03-19
Speedup order summary service

seallard:
## Description
Convert calculation of order summary into database queries.
Closes https://github.com/Clinical-Genomics/streamline-delivery/issues/52

### Fixed
- Speed up generation of order summaries
- Calculate count of not received cases
- Fix case status logic (earlier sample status takes precedence)


tests/server/endpoints/test_orders_endpoint.py
```diff
@@ -1,62 +1,18 @@
...
-    [
-        (None, Workflow.MIP_DNA, 2),
-        (1, Workflow.MIP_DNA, 1),
-        (2, Workflow.MIP_DNA, 2),
-        (None, Workflow.BALSAMIC, 1),
-        (None, Workflow.FLUFFY, 0),
-        (None, None, 3),
-    ],
-)
-def test_orders_endpoint(
```
seallard, line 29: I think we should drop this test because the exact same logic is covered below in another test. The alternative is to create a messy trailblazer response mock with the summaries.

cg/store/filters/status_case_filters.py
```diff
@@ -200,6 +200,30 @@ def filter_compressible_cases(cases: Query, **kwargs) -> Query:
     return cases.filter(Case.is_compressible)
 
 
+def get_not_received_cases(cases: Query, **kwargs) -> Query:
+    return cases.filter(Sample.received_at != None).distinct()
+
+
+def get_received_cases(cases: Query, **kwargs) -> Query:
+    return cases.filter(Sample.received_at != None).distinct()
```
islean, line 208: Maybe a bit early but these filters are identical. I would guess `get_received_cases` should have a "=="?

tests/server/endpoints/test_orders_endpoint.py
islean: Do we test that the request and URI is parsed correctly in the other tests?

seallard: I decided to go for integration tests on the service level instead, the API layer just calls the service so I don't think it adds value to keep

islean: Yeah, I guess that is fine.

cg/store/filters/status_case_filters.py
```diff
@@ -200,6 +200,30 @@ def filter_compressible_cases(cases: Query, **kwargs) -> Query:
...
+def get_not_prepared_cases(cases: Query, **kwargs) -> Query:
+    return cases.filter(Sample.prepared_at == None).distinct()
+
+
+def get_prepared_cases(cases: Query, **kwargs) -> Query:
+    return cases.filter(Sample.prepared_at != None).distinct()
+
+
+def get_not_sequenced_cases(cases: Query, **kwargs) -> Query:
+    return cases.filter(Sample.last_sequenced_at == None).distinct()
```
seallard, line 220: All of the filters above rely on the `cases` query being passed having been joined with the samples table. Do we place these filters elsewhere usually?

I could also do the join in the filters, but I think that would misbehave when chaining them.

cg/store/filters/status_case_filters.py
```diff
@@ -200,6 +200,30 @@ def filter_compressible_cases(cases: Query, **kwargs) -> Query:
     return cases.filter(Case.is_compressible)
 
 
+def get_not_received_cases(cases: Query, **kwargs) -> Query:
+    return cases.filter(Sample.received_at == None).distinct()
+
+
+def get_received_cases(cases: Query, **kwargs) -> Query:
+    return cases.filter(Sample.received_at != None).distinct()
```
islean, line 208: Will this method not return all cases with at least one received sample, rather than all cases with only received samples?

cg/store/filters/status_case_filters.py
```diff
@@ -200,6 +200,30 @@ def filter_compressible_cases(cases: Query, **kwargs) -> Query:
...
+
+
+def get_not_prepared_cases(cases: Query, **kwargs) -> Query:
+    return cases.filter(Sample.prepared_at == None).distinct()
+
+
+def get_prepared_cases(cases: Query, **kwargs) -> Query:
+    return cases.filter(Sample.prepared_at != None).distinct()
```
islean, line 216: Same concern here

cg/store/filters/status_case_filters.py
islean: I would move the filters on sample to that file probably.

seallard: Yeah, I do use that function and pass it :+1: Do you think these functions belong in that same module (`cg/store/base.py`)?

islean: No, not `base.py` but I would probably let them be in the `store/filters/status_case_sample_filters.py` file.

cg/store/filters/status_case_filters.py
```diff
@@ -200,6 +200,30 @@ def filter_compressible_cases(cases: Query, **kwargs) -> Query:
     return cases.filter(Case.is_compressible)
 
 
+def get_not_received_cases(cases: Query, **kwargs) -> Query:
+    return cases.filter(Sample.received_at == None).distinct()
```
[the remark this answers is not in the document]
islean: Just thinking if we need an extra `select_from(Case)` somewhere.

cg/services/orders/order_status_service/order_summary_service.py
```diff
@@ -0,0 +1,40 @@
...
+from cg.store.store import Store
+
+
+class OrderSummaryService:
+    def __init__(self, analysis_client: TrailblazerAPI, store: Store) -> None:
+        self.analysis_client = analysis_client
+        self.store = store
+
+    def get_summary(self, order_id: int) -> OrderSummary:
+        return self.get_summaries([order_id])[0]
```
ChrOertlin, line 16: I like the re-use of `get_summaries`. Do we need a sanity check to make sure the return of `get_summaries` is length 1?

cg/store/crud/read.py
```diff
@@ -1430,3 +1430,37 @@ def get_order_by_ticket_id(self, ticket_id: int) -> Order | None:
...
         return orders.first()
+
+    def get_not_received_case_count(self, order_id: int) -> int:
+        filters: list[CaseFilter] = [CaseFilter.BY_ORDER, CaseFilter.NOT_RECEIVED]
+        cases: Query = self._get_join_case_and_sample_query()
+        return apply_case_filter(
+            cases=cases,
+            filter_functions=filters,
+            order_id=order_id,
+        ).count()
```
ChrOertlin, line 1441: I think this makes a lot of sense to move the "heavy" lifting to the db as it is designed to do this type of stuff efficiently.

cg/store/filters/status_case_filters.py
seallard: True! I guess I need a subquery

cg/store/crud/read.py
```diff
@@ -1430,3 +1430,37 @@ def get_order_by_ticket_id(self, ticket_id: int) -> Order | None:
             orders=orders, filters=order_filter_functions, ticket_id=ticket_id
         )
         return orders.first()
+
+    def get_not_received_case_count(self, order_id: int) -> int:
```
ChrOertlin, line 1434: I think the name is good. Something to consider is the autocomplete in the IDE is a slight reshuffle. Resolve if you think it is overkill.
```suggestion
    def get_case_not_received_count(self, order_id: int) -> int:
```
Would cut out a lot of functions not related to case queries when writing `store.get_case_` in the IDE and could finding what you need easier. Similar to order code related to tables in specific modules I guess.

cg/store/filters/status_case_filters.py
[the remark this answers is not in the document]
ChrOertlin:
```suggestion
    return Select(cases).filter(Sample.received_at == None).distinct()
```
I think this would be the statement. It would be chained

cg/store/filters/status_case_filters.py
islean: I do genuinely like them

ChrOertlin: I think this is indeed what it does. Also case entries are unique, I don't think the distinct would do anything here. Right?

cg/store/filters/status_case_filters.py
ChrOertlin: I think we have something like this somehwere already.

We would start with a join query on both case and sample and then apply case and sample filters for whatever we need. 

Example:
```Python
    def get_not_analysed_cases_by_sample_internal_id(
        self,
        sample_internal_id: str,
    ) -> list[Case]:
        """Get not analysed cases by sample internal id."""

        query: Query = self._get_join_case_and_sample_query()

        not_analysed_cases: Query = apply_case_filter(
            cases=query,
            filter_functions=[
                CaseFilter.NOT_ANALYSED,
            ],
        )

        return apply_sample_filter(
            samples=not_analysed_cases,
            filter_functions=[SampleFilter.BY_INTERNAL_ID],
            internal_id=sample_internal_id,
        ).all()
```

ChrOertlin (approved):
Looks good, very clean. Making the database do more makes sense to me.

cg/store/filters/status_case_filters.py
ChrOertlin: Although I think this can be done with a subquery within the filter as well. It would make the filter more complex, guess it depends on where we want to go with all of this.

cg/store/filters/status_case_filters.py
islean: I guess when you join with Sample, there will be an entry for each sample within the case?

cg/store/filters/status_case_filters.py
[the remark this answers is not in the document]
islean: Yeah, just thinking how it works, since the join in the SQL query would return a table with much more columns than the Case table alone does.

cg/store/filters/status_case_filters.py
seallard: Ah, that query is perfect :+1: Thanks, I think I'll stick to that syntax for now even if I like the subqueries. I feel like the subquery syntax is a little more expressive, but I'll stick to what I know for now

cg/services/orders/order_status_service/order_summary_service.py
seallard: Yeah, I added it as a sanity check :+1:

cg/store/crud/read.py
seallard: I'll measure just to make sure, but it should be much faster this way!

cg/store/crud/read.py
seallard: Fixed in https://github.com/Clinical-Genomics/cg/pull/3050/commits/189b4fb1393b81cb492f603e058f85ff53a82abf and https://github.com/Clinical-Genomics/cg/pull/3050/commits/bfe3f1183779272c5d79b4d151b7acdb32a91efb

cg/store/filters/status_case_filters.py
seallard: Moved the filters in https://github.com/Clinical-Genomics/cg/pull/3050/commits/4ac0d95aa18f26666310f54174ef24e7f82bfbb6

cg/store/filters/status_case_filters.py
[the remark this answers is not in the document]
islean: Nevermind, I think this is a non-issue. But if you in pure SQL write `SELECT * FROM Case c JOIN CaseSample cs ON c.id = cs.case_id`, you would get rows consisting of all the columns in Case and all columns in CaseSample. I was just wondering how it is done in SQLAlchemy

tests/server/endpoints/test_orders_endpoint.py
```diff
@@ -67,11 +23,8 @@ def test_order_endpoint(
     # THEN the response should be successful
     assert response.status_code == HTTPStatus.OK
 
-    order_summary: OrderSummary = create_summaries(
-        orders=[order], analysis_summaries=[analysis_summary]
-    )[0]
-    # THEN the response should only contain the specified order
-    assert response.json == create_order_response(order=order, summary=order_summary).model_dump()
+    # THEN the response contains the specified order
+    assert response.json["id"] == order_id_to_fetch
```
islean, line 27: Is this not a strictly worse test now? Previously it was tested that the summary was exactly what we want. Now it only tests if it gives any summary for the correct order

tests/services/orders/order_status_service/conftest.py
```diff
@@ -1,16 +1,21 @@
...
 from cg.store.store import Store
 from tests.store_helpers import StoreHelpers
 
 
 @pytest.fixture
-def order_status_service(trailblazer_api: TrailblazerAPI, store: Store):
-    return OrderStatusService(analysis_client=trailblazer_api, store=store)
+def summary_service(trailblazer_api: TrailblazerAPI, store: Store, order: Order):
+    service = OrderSummaryService(analysis_client=trailblazer_api, store=store)
+    service.analysis_client = Mock()
```
islean, line 16: Why do we make this fixture dependent on the trailblazer_api if we override it with a Mock() immediately?

tests/services/orders/order_status_service/conftest.py
```diff
@@ -22,12 +27,63 @@ def order(store: Store, helpers: StoreHelpers) -> Order:
...
+    case_in_sequencing: Case = helpers.ensure_case(
+        store=store,
+        customer=order.customer,
+        order=order,
+        case_name="case_in_sequencing",
+        case_id="case_in_sequencing",
+    )
+    helpers.add_relationship(store=store, sample=sample_not_received, case=case_not_received)
+    helpers.add_relationship(store=store, sample=sample_in_preparation, case=case_in_preparation)
+    helpers.add_relationship(store=store, sample=sample_in_sequencing, case=case_in_sequencing)
```
islean, line 88: Is not the big thing we want to test the interplay that happens when cases have multiple samples with different statuses? Right now I am uncertain if we would catch any such issues.

cg/store/filters/status_case_sample_filters.py
```diff
@@ -18,12 +18,39 @@ def filter_cases_with_sample_by_internal_id(case_samples: Query, sample_internal
...
+    return case_samples.filter(Case.order_id == order_id)
+
+
+def get_not_received_cases(case_samples: Query, **kwargs) -> Query:
+    return case_samples.filter(Sample.received_at == None).distinct()
+
+
+def get_received_cases(case_samples: Query, **kwargs) -> Query:
+    subquery = case_samples.filter(Sample.received_at == None).subquery()
+    return case_samples.filter((CaseSample.case_id != subquery.c.id)).distinct()
```
islean, line 31: I would like a docstring which included the intended SQL query if possible.

tests/services/orders/order_status_service/conftest.py
seallard: Added additional samples to cover these issues https://github.com/Clinical-Genomics/cg/pull/3050/commits/f015e1623c7a117ed082363c3c1d941173a6e1a0

tests/services/orders/order_status_service/conftest.py
seallard: Removed in https://github.com/Clinical-Genomics/cg/pull/3050/commits/769d74ad76ab380540c30de681146519fc29b29b

tests/server/endpoints/test_orders_endpoint.py
seallard: You are right, but I think it is worth it.

The structure of the summaries is bound to change and the cost of patching this test exceeds the value of verifying the structure of the summary. I think it is good enough to validate that A summary is generated for the correct order but I'll re-add it if you think differently.

tests/server/endpoints/test_orders_endpoint.py
seallard: Re-added here: https://github.com/Clinical-Genomics/cg/pull/3050/commits/2c75b690026d7847aca899e152b8acb6fbe33fba

cg/store/base.py
```diff
@@ -59,6 +59,13 @@ def _get_join_case_sample_query(self) -> Query:
         """Return join case sample query."""
         return self._get_query(table=CaseSample).join(CaseSample.case).join(CaseSample.sample)
 
+    def _join_sample_and_case(self):
+        return (
+            self.session.query(Case)
+            .join(CaseSample, Case.id == CaseSample.case_id)
+            .join(Sample, CaseSample.sample_id == Sample.id)
+        )
```
islean, line 67: What is the difference between this method and the _get_join_case_sample_query starting four rows above?

cg/store/filters/status_case_sample_filters.py
```diff
@@ -18,12 +18,51 @@ def filter_cases_with_sample_by_internal_id(case_samples: Query, sample_internal
     return case_samples.filter(Sample.internal_id == sample_internal_id)
 
 
+def filter_by_order(case_samples: Query, order_id: int, **kwargs) -> Query:
+    return case_samples.join(Order, Case.orders).filter(Order.id == order_id)
```
islean, line 22: Is there any point in adding the .distinct here? Or I guess we will never use it without the subsequent filters which already do the .distinct

cg/store/filters/status_case_sample_filters.py
```diff
@@ -18,12 +18,51 @@ def filter_cases_with_sample_by_internal_id(case_samples: Query, sample_internal
...
+    return case_samples.filter(Sample.received_at == None).distinct()
+
+
+def get_received_cases(case_samples: Query, **kwargs) -> Query:
+    not_received = (
+        case_samples.filter(Sample.received_at == None).with_entities(CaseSample.case_id).subquery()
+    )
+    return (
+        case_samples.outerjoin(not_received, CaseSample.case_id == not_received.c.case_id)
+        .filter(not_received.c.case_id == None)
```
islean, line 35: Super cool with an outer join!

tests/server/endpoints/test_orders_endpoint.py
islean: I still think it adds some value for now. But I do see your point. Do as you seem fit, I am fine either way.

tests/services/orders/order_status_service/test_order_summary_service.py
```diff
@@ -0,0 +1,82 @@
...
+
+    # WHEN creating a summary for the order
+    summary: OrderSummary = summary_service.get_summary(order_id)
+
+    # THEN the summary should contain one case in sequencing
+    assert summary.in_sequencing == 1
+
+
+def test_summarize_multiple_samples_not_received(
+    summary_service: OrderSummaryService, order_with_not_received_samples: Order
```
islean, line 51: Just out of curiosity, did we double count this at some point?

islean (approved):
Looks great! Love the outer joins!

tests/services/orders/order_status_service/test_order_summary_service.py
seallard: Yes, there was a bug that introduced double counting for some of the statuses

cg/store/filters/status_case_sample_filters.py
seallard: Added `distinct`

cg/store/base.py
seallard: This one starts from `Case` and uses a more explicit syntax for joining the tables.
The tests fail when using the other join-query

islean: Weird, feels as if one should be sufficient.

islean: image

seallard: I agree, tracked it here https://github.com/Clinical-Genomics/cg/issues/3073.
Will leave it as is for now

PR #3051, opened 2024-03-19
Remove dead code

seallard:
This PR removes a function that was only being called in tests.

`cases` is only called in our test suite. Looking at the code, it seems like it does some sort of TAT calculation.
We should not have code that is not used (and this messy) 🏥 

Removed the following call chain from `cg/store/crud/read.py`
- `cases`
- `_get_empty_case_data`
- `_get_max_tat`
- `_calculate_date_delta`
- `_calculate_estimated_turnaround_time`
- `_is_rerun`
- `_calculate_case_data`
- `_should_be_skipped`
- `_get_case_output`

Removed the tests for `cases` as well.

### Fixed
- Remove dead code


diitaz93 (approved):
🪚

PR #3060, opened 2024-03-21
feat(enum): Refactor

henrikstranneheim:
### Added

-

### Changed

- StrEnums to use the auto function.

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/constants/subject.py
+++ b/cg/constants/subject.py
@@ -1,19 +1,17 @@
-from enum import IntEnum, StrEnum
+from enum import IntEnum, StrEnum, auto
 
 
 class Sex(StrEnum):
-    FEMALE = "female"
-    MALE = "male"
-    UNKNOWN = "unknown"
-    OTHER = "other"
-    MISSING = ""
+    FEMALE = auto()
+    MALE = auto()
+    UNKNOWN = auto()
+    OTHER = auto()
 
 
 class PhenotypeStatus(StrEnum):
-    UNKNOWN = "unknown"
-    UNAFFECTED = "unaffected"
-    AFFECTED = "affected"
-    MISSING = ""
+    UNKNOWN = auto()
+    UNAFFECTED = auto()
+    AFFECTED = auto()
 
 
 class PlinkPhenotypeStatus(IntEnum):


henrikstranneheim:
### Enum translated correctly

```
[henrik.stranneheim@hasta:~] [S_base] 2m47s $ cg workflow mip-dna config-case justhusky -d
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Called undefined __fields__ on HousekeeperAPI, please wrap
Case justhusky exists in Status DB
{'case': 'justhusky', 'default_gene_panels': ['OMIM-AUTO'], 'samples': [{'sample_id': 'earlycasualcaiman', 'sample_display_name': 'ACC5339A2--300M', 'analysis_type': 'wgs', 'sex': 'male', 'phenotype': 'unaffected', 'expected_coverage': 1, 'capture_kit': 'twistexomerefseq_9.1_hg19_design.bed', 'mother': '0', 'father': '0'}, {'sample_id': 'slowlycivilbuck', 'sample_display_name': 'ACC5339A3--300M', 'analysis_type': 'wgs', 'sex': 'female', 'phenotype': 'unaffected', 'expected_coverage': 1, 'capture_kit': 'twistexomerefseq_9.1_hg19_design.bed', 'mother': '0', 'father': '0'}, {'sample_id': 'hugelymodelbat', 'sample_display_name': 'ACC5339A4--300M', 'analysis_type': 'wgs', 'sex': 'male', 'phenotype': 'affected', 'expected_coverage': 1, 'capture_kit': 'twistexomerefseq_9.1_hg19_design.bed', 'mother': 'slowlycivilbuck', 'father': 'earlycasualcaiman'}]}

**Comment by henrikstranneheim** [comment]
Had to use StrEnum and not Enum for correct interpolation.

**Review by diitaz93 — APPROVED** [review]
Should we use Enum instead? That was what I understood from yesterday but Im not 100% sure

**Comment by diitaz93** [comment]
> Should we use Enum instead? That was what I understood from yesterday but Im not 100% sure

Ok nevermind, I just saw the comment

**Comment by henrikstranneheim** [comment]
@Vince-janv Good catch! I removed the `,value` and this seem to work as expected, but I will also verify on stage. I wonder if the use of auto() already bypasses the use of values? The member seem to be interpolated in lower case, see output above.

Also
```
class enum.auto

    auto can be used in place of a value. If used, the Enum machinery will call an Enum’s [_generate_next_value_()](https://docs.python.org/3/library/enum.html#enum.Enum._generate_next_value_) to get an appropriate value. For Enum and IntEnum that appropriate value will be the last value plus one; for Flag and IntFlag it will be the first power-of-two greater than the highest value; for StrEnum it will be the lower-cased version of the member’s name.

Vince-janv:
> @Vince-janv Good catch! I removed the `,value` and this seem to work as expected, but I will also verify on stage. I wonder if the use of auto() already bypasses the use of values? The member seem to be interpolated in lower case, see output above.
> 
> Also
> 
> ```
> class enum.auto
> 
>     auto can be used in place of a value. If used, the Enum machinery will call an Enum’s [_generate_next_value_()](https://docs.python.org/3/library/enum.html#enum.Enum._generate_next_value_) to get an appropriate value. For Enum and IntEnum that appropriate value will be the last value plus one; for Flag and IntFlag it will be the first power-of-two greater than the highest value; for StrEnum it will be the lower-cased version of the member’s name.
> ```

Let's hope it works 🤞 Indeed auto() on StrEnums creates lower case values, so it would have facilitate our current (miss)usage, but what is the inherent problem is that we rely on values as constants instead of the names. If these enums where to be translated into SQLAlcemy Enums the set of options would be {AFFECTED, UNAFFECTED, UNKNOWN}, and not their lower case values.

cg/constants/subject.py
```diff
@@ -1,19 +1,17 @@
-from enum import IntEnum, StrEnum
+from enum import IntEnum, StrEnum, auto
 
 
 class Sex(StrEnum):
```
ChrOertlin, line 4:
```suggestion
class Sex(Enum):
```

ChrOertlin: I think if there is no need to access `str` functions within the enum then there is no need for the `StrEnum` class. Right, @Vince-janv ?

henrikstranneheim:
```
>>> from enum import IntEnum, StrEnum, auto, Enum
>>> class SexEnum(Enum):
...     FEMALE = auto()
...     MALE = auto()
...     UNKNOWN = auto()
...     OTHER = auto()
...
>>> class Sex(StrEnum):
...     FEMALE = auto()
...     MALE = auto()
...     UNKNOWN = auto()
...     OTHER = auto()
...
>>> Sex.MALE
<Sex.MALE: 'male'>
>>> SexEnum.MALE
<SexEnum.MALE: 2>
>>> SexEnum.MALE.name.lower()
<>> if "male" == Sex.MALE:
...     print(True)
...
True
>>> if "male" == SexEnum.MALE:
...     print(True)
...
>>> if "male" == SexEnum.MALE.name.lower():
...     print(True)
...
True
```
I prefer using StrEnum, which has auto(), and hence only uses members and not any values over Enum and `.name.lower():`.

@Vince-janv Is there another way to do this using only the enum class?

Vince-janv:
> I prefer using StrEnum, which has auto(), and hence only uses members and not any values over Enum and `.name.lower():`.
> 
> @Vince-janv Is there another way to do this using only the enum class?

No, you're right for string comparisons we need StrEnum. But I do question whether we need these comparisons 🤔 Do you have an example where we use it?

Vince-janv:
I'm not sure I understand completely 🤔 Within the a repo, I don't see why we would compare an Enum member with a string. Anywhere we would use "male" we should just use `SexEnum.MALE`. But I do see that when interacting with other software (balsamic or scout) there can be problems we might need to solve.

I think both the previous examples could be solved without StrEnum

henrikstranneheim:
I would also like to use Enum over StrEnum if possible, but I currently do not see how (in an easy way). Can you show how you mean to refactor it?

As far as this PR currently goes,I think the usage of the enums (albeit StrEnum) is very close to the intended constant behavior (DRY, use members and easy to understand when it gets interpolated) and is an improvement.

Vince-janv:
No you're right, there is not "easy" way to do it, but to me this PR only solves 1 of the 2 problems we have with Enums. The enum usage here would not match the one expected by SQLAlchemy for example.

If I were to do this I would take one model at a time and follow the steps below:
1. Change the type to Enum and add auto() as the value.
2. If used in models.py change the type of the ORM models (it becomes really clean).
3. Fix all pytests that fail by replacing strings with the Enum member
4. Anywhere the value is dumped into a dict/file or sent in a request use lower() (or similar) to  not break anything in scout, balsamic, MIP, etc. For pydantic BaseModels I think this can be done pretty cleanly by setting some parameters.
5. If the enum was used to set options in the database perform an alembic update to set the new ENUM values to all_caps and parse though the data.

I did most of it for PhenotypeStatus here: https://github.com/Clinical-Genomics/cg/pull/3065/files
It overlapped a bit with StatusOptions but I think you get the point. I got almost all pytests to work but marshmellow (why are we using that?) keeps bugging me.

As you see it's a ton of work for just one model...

islean:
@Vince-janv Although it is not the recommended praxis, would we gain more by using EnumClassName.enum_name_already_in_lower as a middle step? So the refactor at the later stage would instead of removing the .lower() be to switch case?

Vince-janv:
> @Vince-janv Although it is not the recommended praxis, would we gain more by using EnumClassName.enum_name_already_in_lower as a middle step? So the refactor at the later stage would instead of removing the .lower() be to switch case

I have certainly been tempted by that option... I think there is a big risk of not doing the last (hard) step if we decide to have a middle fix. But I started a discussion about it in the CGEP so please chime in if you like 😊

henrikstranneheim:
The current code using StrEnums fulfills point 1-4 above.

Regarding point 5:
`
If the enum was used to set options in the database perform an alembic update to set the new ENUM values to all_caps and parse though the data.
`

Do we really want the data to be upper case? Is that best practice?  In files e.g configs they should be lower case, in code SCREAMING_SNAKE_CASE and for database interactions you want to use the Enum that was used in the schema. This looks clean enough to me (and not that much work 🤞):

```python
class PhenotypeStatus(StrEnum):
    AFFECTED = auto()
    UNAFFECTED = auto()
    UNKNOWN = auto()
```

```python
 phenotype: PhenotypeStatus | None = None
```

```Python
 status: Mapped[PhenotypeStatus] = mapped_column(default=PhenotypeStatus.UNKNOWN)
```

Am I missing something?

henrikstranneheim:
@Clinical-Genomics/sysdev, @ivadym Added to refinement on Wednesday - would be nice to agree on a way forward.

diitaz93:
I am starting to wonder why do we use Enum classes instead of a normal class with `str` members. Is there some functionality from Enums that we need?

ChrOertlin:
# Refinement discussion

1. Go ahead with enum usage as described in this PR.
2. In future refrain from specialised "StrEnum" or "IntEnum" where not needed. "Enum" is the default.
3. Conduct feasibility study before implementation (@Vince-janv, estimated time 5 hours)

Vince-janv:
I checked out the branch and tried som tests. The values seem to be added as all-caps:
Screenshot 2024-04-17 at 16 32 40

henrikstranneheim:
@Vince-janv Thx for filling in teh blanks on the database value.

The I vote for using StrEnum:

- It is designed for this usage case, and it behaves exactly as we want and what you would expect.
- It ensures type safety and conveys the intended usage of the constants.
- You do not have to make additional function calls to manipulate the constant, Python takes care off it based on the class.

@Clinical-Genomics/sysdev @ivadym Please vote 👍 for using StrEnums or 👎 if you want just Enums.

Vince-janv:
I'm not sure the solution will pass the tests on stage. Without `RandomEnum.member.value` I think some things might crash when we connect to an existing database, hence my thumb down. If it works I might rest my case

henrikstranneheim:
> I'm not sure the solution will pass the tests on stage. Without `RandomEnum.member.value` I think some things might crash when we connect to an existing database, hence my thumb down. If it works I might rest my case

Fair point! I will test rigorously in stage before proceeding.

henrikstranneheim:
Stale - won't do!

Issue #3025, opened 2024-03-06
Data duplication linking samples to flow cells


Vince-janv:
### Description
We have a link table called flowcell_sample which links the tables flowcell and sample. We also have the table SampleLaneSequencingMetrics which storest the same information as well as metrics. 

Having both of these creates ambiguity regarding when a sample is linked to a flow cell. There should only be one source of truth for this question and I think it should be SampleLaneSequencingMetrics.

### Suggested solution
Remove the flowcell_sample link table an replace the back_references with properties that use SampleLaneSequencingMetrics instead

### Needs discussion:
- What to do with downsampled samples? They seem to have flow_cell-sample links but no metrics.
- We loose the `flow_cell.samples.append(new_sample)` functionality

Vince-janv:
### Decision:
- Remove the column

#### For downsampled data
- Remove check for flow cells when handling downsampled samples (check if this is done?)
- Maybe add an exception in the .flow_cells property

Vince-janv:
When checking if the data is "ondisk" we already call `cg.meta.workflow.analysis.AnalysisAPI._is_flow_cell_check_applicable` which returns false if any sample in the case is downsampled

PR #3063, opened 2024-03-22
fix(flowcell_sample) Remove flowcell_sample link table

Vince-janv:
## Description
For a description of the issue, see https://github.com/Clinical-Genomics/cg/issues/3025.

This PR removes the use of the flowcell_sample table for connecting samples to flow cells. The backrefs on both models are replaced by properties that look at the SampleLaneSequencingMetrics instead.

### Changed

- Flowcell.samples is now a property instead of a backref
- Sample.flowcells -> Sample.flow_cells
- Sample.flow_cells is now a property instead of a backref
- Removed code linking samples and flow cells via the link table


cg/store/models.py
```diff
@@ -829,6 +826,11 @@ class Sample(Base, PriorityMixin):
     def __str__(self) -> str:
         return f"{self.internal_id} ({self.name})"
 
+    @property
+    def flow_cells(self) -> list[Flowcell]:
```
Vince-janv, line 830: Should this be `flowcells` for consistency or should we keep moving towards flow_cells?

cg/meta/demultiplex/status_db_storage_functions.py
```diff
@@ -39,32 +39,11 @@ def store_flow_cell_data_in_status_db(
...
-    samples: set[Sample] = {
-        store.get_sample_by_internal_id(sample_internal_id)
-        for sample_internal_id in sample_internal_ids
-    }
-    for sample in samples:
-        if isinstance(sample, Sample) and sample not in flow_cell.samples:
-            flow_cell.samples.append(sample)
-    return flow_cell
-
-
```
Vince-janv, line 67: No longer needed

cg/store/base.py
```diff
@@ -109,84 +114,6 @@ def _get_latest_analyses_for_cases_query(self) -> Query:
             ),
         )
 
-    def _get_filtered_case_query(
```
Vince-janv, line 112: This seems unused as far as i can tell

tests/store_helpers.py
```diff
@@ -401,7 +401,14 @@ def add_sample(
...
         for key, value in kwargs.items():
             if key == "flowcell":
-                sample.flowcells.append(kwargs["flowcell"])
+                flow_cell: Flowcell = kwargs["flowcell"]
+                metric: SampleLaneSequencingMetrics = store.add_sample_lane_sequencing_metrics(
+                    sample_internal_id=sample.internal_id,
+                    flow_cell_name=flow_cell.name,
+                    **kwargs,
+                )
+                store.session.add(metric)
```
Vince-janv, line 410: This is messy but I couldn't find a clean way that was within the scope of this PR

cg/store/models.py
```diff
@@ -895,7 +889,7 @@ def to_dict(self, links: bool = False, flowcells: bool = False) -> dict:
         if links:
             data["links"] = [link_obj.to_dict(family=True, parents=True) for link_obj in self.links]
         if flowcells:
-            data["flowcells"] = [flowcell_obj.to_dict() for flowcell_obj in self.flowcells]
+            data["flowcells"] = [flowcell_obj.to_dict() for flowcell_obj in self.flow_cells]
```
ChrOertlin, line 892:
```suggestion
            data["flowcells"] = [flow_cell.to_dict() for flow_cell in self.flow_cells]
```
when you already touch the line :)

cg/store/models.py
ChrOertlin: I think flow_cell is good

ChrOertlin (approved):

cg/store/models.py
diitaz93: I relate a `flow_cell` with a `FlowCellDirectoryData` and a `flowcell` with a `Flowcell`. Maybe we should keep the old way for now and in the future consider a renaming. There are still many instances in the code using `flowcell` so this could be part of the scope of another PR

cg/store/models.py
diitaz93, line 892: for example, here we mix the `flow_cell` terminology with `flowcell`.

cg/store/models.py
```diff
@@ -72,14 +71,6 @@ def to_dict(model_instance):
         }
 
 
-flowcell_sample = Table(
```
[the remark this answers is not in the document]
diitaz93: Does this include an alembic migration or is there no need for that?

diitaz93 (approved):
Looks good 👍

cg/store/base.py
```diff
@@ -69,7 +69,12 @@ def _get_join_sample_and_customer_query(self) -> Query:
 
     def _get_join_flow_cell_sample_links_query(self) -> Query:
         """Return join flow cell samples and relationship query."""
-        return self._get_query(table=Flowcell).join(Flowcell.samples).join(Sample.links)
+        return (
+            self._get_query(table=Flowcell)
+            .join(Flowcell.sequencing_metrics)
+            .join(SampleLaneSequencingMetrics.sample)
+            .join(Sample.links)
```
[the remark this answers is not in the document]
ChrOertlin: Btw, just wanted to point this out, you can resolve this conversation if it prevents merging. I think either way works with our current database.

cg/store/models.py
Vince-janv: @diitaz93 I think your association is logical given our code, but in reality flowcell is just a typo (swedifying an english word). I think I will go for flow_cell

cg/store/base.py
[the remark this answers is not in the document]
Vince-janv: Regarding readability of point one i agree that the joins grow quite long, but thanks to our Query refactoring a year ago I think this can be handled pretty cleanly.

flow_cell -> sample needs to be a many-to-many so a link table is needed, I figured that the metrics table fills this function.

cg/store/models.py
Vince-janv: Since `data["flowcells"]` is exterior-facing (ie the key is probably used in other repos) I don't want to touch that

cg/store/models.py
```diff
@@ -680,15 +671,16 @@ class Flowcell(Base):
...
     sequencing_metrics: Mapped[list["SampleLaneSequencingMetrics"]] = orm.relationship(
         back_populates="flowcell",
         cascade="all, delete, delete-orphan",
     )
 
+    @property
+    def samples(self) -> list["Sample"]:
+        """Return samples sequenced on the flow cell."""
+        return list({metric.sample for metric in self.sequencing_metrics})
+
```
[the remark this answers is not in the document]
Vince-janv: Yes, list is to not change any functionality. Maybe we could do set now that I think about it 🤔 Set is to remove duplicates since the same sample will have one metric per lane.

cg/store/models.py
```diff
@@ -829,6 +818,11 @@ class Sample(Base, PriorityMixin):
     def __str__(self) -> str:
         return f"{self.internal_id} ({self.name})"
 
+    @property
+    def flow_cells(self) -> list[Flowcell]:
+        """Return the flow cells a sample has been sequenced on."""
+        return list({metric.flowcell for metric in self.sequencing_metrics})
```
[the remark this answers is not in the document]
Vince-janv: See answer above

cg/store/models.py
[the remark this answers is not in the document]
islean: maybe out of scope? Feel free to resolve for now though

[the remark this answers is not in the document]
Vince-janv: Keeping as list for now

Vince-janv:
Deployed in production

PR #3080, opened 2024-03-28
fix(Backup) Fix file-touching post decription

Vince-janv:
## Description
Fixes the bug in production where copycomplete and RTAcomplete files cannot be created

### Fixed

- Fixed path to be correct
- Simplified some functions


karlnyr (approved):
Neat 🌞

PR #3095, opened 2024-04-08
Add endpoint for marking order as delivered

seallard:
## Description
Closes https://github.com/Clinical-Genomics/streamline-delivery/issues/98.

### Added
- Endpoint `PATCH /orders/<order_id>/delivered` for marking an order as delivered.

### Changed
- Raise exception in database layer for missing orders


cg/store/crud/read.py
```diff
@@ -1413,14 +1413,17 @@ def get_orders_by_ids(self, order_ids: list[int]) -> list[Order]:
             ids=order_ids,
         ).all()
 
-    def get_order_by_id(self, order_id: int) -> Order | None:
+    def get_order_by_id(self, order_id: int) -> Order:
```
ChrOertlin, line 1416: Nice, we just need to make sure that if this is used in multiple places, these also handle the error.

ChrOertlin (approved):
Looks good!

alembic/versions/2024_04_08_0395f682958e_add_order_delivery.py
```diff
@@ -0,0 +1,24 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "0395f682958e"
+down_revision = "b614a52759c5"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("order", sa.Column("delivered", sa.Boolean(), nullable=False))
```
islean, line 20: Slight preference for `is_delivered` but go with what you like the most.

cg/store/crud/read.py
```diff
@@ -1413,14 +1413,17 @@ def get_orders_by_ids(self, order_ids: list[int]) -> list[Order]:
...
         """Returns the entry in Order matching the given id."""
         orders: Query = self._get_query(table=Order)
         order_filter_functions: list[Callable] = [OrderFilter.BY_ID]
         orders: Query = apply_order_filters(
             orders=orders, filters=order_filter_functions, id=order_id
         )
-        return orders.first()
+        order: Order | None = orders.first()
+        if not order:
```
islean, line 1424:
```suggestion
        if not order := orders.first():
```
Might be nice? If the assignment scope extends past the if-statement?

islean (approved):
Looks great!

alembic/versions/2024_04_08_0395f682958e_add_order_delivery.py
seallard: Renamed in https://github.com/Clinical-Genomics/cg/pull/3095/commits/017bc5dcc43cc1005f405e623c40454b8ae8e587

cg/store/crud/read.py
seallard: I think it looks less readable, prefer this syntax in this context.
```python
  if not (order := orders.first()):
```
The assignment is not as clear.

seallard: Changed in https://github.com/Clinical-Genomics/cg/pull/3095/commits/41ede3a4ff2901f1ea877606b8516cf85e1491a1

cg/store/crud/read.py
seallard: Yeah 👍 All the places which call this function used to raise this exception, so we are golden

PR #3096, opened 2024-04-08
Filter orders by delivered

seallard:
## Description
Closes https://github.com/Clinical-Genomics/streamline-delivery/issues/99.

### Added
- Filter on delivered column in order table


cg/store/filters/status_order_filters.py
```diff
@@ -42,6 +42,10 @@ def filter_orders_by_search(orders: Query, search: str | None, **kwargs) -> Quer
     )
 
 
+def filter_orders_by_delivered(orders: Query, delivered: bool | None, **kwargs) -> Query:
+    return orders.filter(Order.delivered == delivered) if delivered is not None else orders
```
ChrOertlin, line 46: Jus a question, is the None due to the use of a pydantic model when passing parameters? I think if delivered is not supplied to the `apply_delivered_filter` then this function would do nothing, right?

ChrOertlin (approved):
Looks good!

cg/store/filters/status_order_filters.py
islean: The flow goes smoother in this case since we can remove a lot of if-statements depending on what the HTTP request looks like. With this design, we can fetch orders via the /orders endpoint and regardless of what parameters are set (pagination, search, sort order, order_id etc), we can specify "apply all filters" and only the filters with the parameter set in the request will actually get invoked.

ChrOertlin: Yes, makes sense.

alembic/versions/2024_04_08_0395f682958e_add_order_delivery.py
```diff
@@ -0,0 +1,24 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column("order", sa.Column("delivered", sa.Boolean(), nullable=False))
+
+
+def downgrade():
+    op.drop_column("order", "delivered")
```
islean, line 24: This whole alembic script has already been covered in #3095 right?

cg/server/api.py
```diff
@@ -516,6 +517,19 @@ def get_order(order_id: int):
         return make_response(response_dict)
     except OrderNotFoundError as error:
         return make_response(jsonify(error=str(error)), HTTPStatus.NOT_FOUND)
+    except Exception as error:
+        return make_response(jsonify(error=str(error)), HTTPStatus.INTERNAL_SERVER_ERROR)
```
islean, line 521: Is this not the default behaviour already? Or what changes with this?

cg/store/crud/update.py
```diff
@@ -24,3 +24,9 @@ def update_sample_comment(self, sample: Sample, comment: str) -> None:
         else:
             sample.comment = comment
         self.session.commit()
+
+    def update_order_delivery(self, order_id: int, delivered: bool) -> None:
+        """Update the delivery status of an order."""
+        order: Order = self.get_order_by_id(order_id)
+        order.delivered = delivered
```
islean, line 31:
```suggestion
        order.is_delivered = delivered
```
after #3095

islean (approved):
Looks great! Just some conflicts from #3095

PR #3097, opened 2024-04-09
Add endpoint for automatically marking order as delivered

seallard:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/orders/order_service/order_service.py
+++ b/cg/services/orders/order_service/order_service.py
@@ -40,3 +40,9 @@ class OrderService:
 
     def set_delivery(self, order_id: int, delivered: bool) -> None:
         self.store.update_order_delivery(order_id=order_id, delivered=delivered)
+
+    def update_delivered(self, order_id: int, delivered_analyses: int) -> None:
+        order: Order = self.store.get_order_by_id(order_id)
+        case_count: int = len(order.cases)
+        if delivered_analyses >= case_count:
+            self.set_delivery(order_id=order_id, delivered=True)
--- a/cg/server/api.py
+++ b/cg/server/api.py
@@ -43,6 +43,7 @@ from cg.server.dto.delivery_message.delivery_message_request import (
 from cg.server.dto.delivery_message.delivery_message_response import (
     DeliveryMessageResponse,
 )
+from cg.server.dto.orders.order_delivery_update_request import OrderDeliveredUpdateRequest
 from cg.server.dto.orders.order_patch_request import OrderDeliveredPatch
 from cg.server.dto.orders.orders_request import OrdersRequest
 from cg.server.dto.orders.orders_response import Order, OrdersResponse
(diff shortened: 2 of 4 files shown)


cg/services/orders/order_service/order_service.py
```diff
@@ -40,3 +40,9 @@ def create_order(self, order_data: OrderIn) -> OrderResponse:
 
     def set_delivery(self, order_id: int, delivered: bool) -> None:
         self.store.update_order_delivery(order_id=order_id, delivered=delivered)
+
+    def update_delivered(self, order_id: int, delivered_analyses: int) -> None:
+        order: Order = self.store.get_order_by_id(order_id)
+        case_count: int = len(order.cases)
+        if delivered_analyses >= case_count:
```
ChrOertlin, line 47: should the `>` here be able to happen? 
Is it possible to abstract away the check into function. The step of checking the number of delivered analyses to the case_count in the order is a bit difficult to follow.
perhaps:

```suggestion
        if analyses_for_cases_delivered():
```
I think either way works though.

ChrOertlin (approved):
Looks good!

cg/services/orders/order_service/order_service.py
islean: I think for microsalt you can start a partial analysis when not all samples are ready. I guess that could result in a larger than? And also I guess hidden analyses could potentially be marked as delivered..

cg/services/orders/order_service/order_service.py
```diff
@@ -40,3 +40,9 @@ def create_order(self, order_data: OrderIn) -> OrderResponse:
 
     def set_delivery(self, order_id: int, delivered: bool) -> None:
         self.store.update_order_delivery(order_id=order_id, delivered=delivered)
+
+    def update_delivered(self, order_id: int, delivered_analyses: int) -> None:
```
islean, line 44: I think we should add support for unmarking as delivered, say if someone accidentally marks an analysis as delivered and immediately unmarks it?

cg/services/orders/order_service/order_service.py
islean: Also, maybe the uncertainties can be cleared with a short docstring?

islean (approved):
I'd like some support for unmarking the order as delivered but otherwise it looks good to me!

cg/services/orders/order_service/order_service.py
ChrOertlin: Yes, a docstring can also make it clearer!

cg/services/orders/order_service/order_service.py
seallard: That is a separate endpoint implemented here https://github.com/Clinical-Genomics/cg/pull/3095

islean: @seallard no? Is that not for manual markings of the order's delivery status? I would want this to happen when an analysis is unmarked as delivered.

seallard: Ah, nice! Good point 👍

seallard: Added in https://github.com/Clinical-Genomics/cg/pull/3097/commits/f7bc777afbec6e449897c3e7c43ffd2164028004

cg/services/orders/order_service/order_service.py
```diff
@@ -40,3 +40,11 @@ def create_order(self, order_data: OrderIn) -> OrderResponse:
...
+        order: Order = self.store.get_order_by_id(order_id)
+        case_count: int = len(order.cases)
+        if delivered_analyses >= case_count:
+            self.set_delivery(order_id=order_id, delivered=True)
+        else:
+            self.set_delivery(order_id=order_id, delivered=False)
```
islean, line 50: Is it worth to add some check to avoid setting this every time an analysis is marked as delivered? Might just clog the code with no added benefit though.

cg/services/orders/order_service/order_service.py
seallard: Resolved in https://github.com/Clinical-Genomics/cg/pull/3097/commits/4b85a65372a4217ded9f66f2db69ef8de94da352

PR #3100, opened 2024-04-09
No content response codes for delivery messages

islean:
## Description

This PR addresses two instances: 

- When trying to fetch the delivery message for a case with `no-delivery` as delivery method
- When trying to fetch the delivery message for an order in which no analyses are ready.

In both of these instances, we return an error, while it was requested by @moedarrah that they simply return a 204, no content instead.

### Changed

- No-delivery and no analyses ready in an order gives a 204 response for the delivery message.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
Tested in stage for an analyses with no-delivery:

image

Tested in stage for an order with no analyses to deliver:

image

cg/server/api.py
```diff
@@ -540,7 +543,7 @@ def get_delivery_message_for_order(order_id: int):
         response_dict: dict = response.model_dump()
         return make_response(response_dict)
     except OrderNotDeliverableError as error:
-        return make_response(jsonify(error=str(error)), HTTPStatus.PRECONDITION_FAILED)
+        return make_response(jsonify(message=str(error)), HTTPStatus.NO_CONTENT)
```
islean, line 546: Seems as if NO_CONTENT results in no message either way, so maybe just remove it?

cg/exc.py
```diff
@@ -290,5 +290,9 @@ class DeliveryMessageNotSupportedError(CgError):
     """Exception raised when trying to fetch delivery messages for unsupported workflows."""
 
 
+class DeliveryMessageNoDeliveryError(CgError):
```
ChrOertlin, line 293:
```suggestion
class DeliveryMessageNoneError(CgError):
```
Perhaps? Flavour

cg/server/api.py
islean: Like empty string it or something.

ChrOertlin (approved):
Looks good to me

cg/server/api.py
ChrOertlin: Makes sense to have nothing returned in this case except the status

cg/exc.py
islean: Might be better haha I did not like my naming...

seallard: Hm, why do we need an exception? I don't like using exceptions for non-exceptional behaviour

seallard: Hm, might be nice to just raise an exception called `NoAnalysesToDeliver` or something

islean: Yeah, I was thinking that it might be nice to be able to see all errors for the Delivery Message flow by just searching for DeliveryMessage though... Might not be any value in doing so.

islean: And this is for when the data delivery option is "no-delivery"

islean: I will keep as is for now, and we can change it if it is annoying.

diitaz93 (approved):
👍

islean:
Will not do for now after consulting with @moedarrah

PR #3101, opened 2024-04-09
Include is_delivered in orders response

seallard:
The new `is_delivered` field needs to be included in the response from `/orders`.


islean (approved):
💪

PR #3102, opened 2024-04-10
Parent classes for the new Run models

Vince-janv:
## Description
With new instruments being bought and implemented there is a need to store different types of data in the database but still be able to treat the samples the same way. 

This PR implements parent classes which define the relationships between the new models (see image below). This way each child class can hold its own columns but the way they relate to each other is the same across instruments making the flow as a whole less complex (even though this code is more complex...).
image

This flow is designed for the Illumina flow to be integrated into the models as well, but my intention in the near future is to just add the new models and save the refactoring for the future.

### Points to keep in mind:
- An ONT flow cell can be used multiple times
- Saphyr is not a "sequencing" instrument, hence the term digitisation. I'm up for going with sequencing if you like though

Closes: Clinical-Genomics/add-new-tech#1

### DIFF
--- a/cg/store/models.py
+++ b/cg/store/models.py
@@ -815,6 +815,10 @@ class Sample(Base, PriorityMixin):
     )
     invoice: Mapped["Invoice | None"] = orm.relationship(back_populates="samples")
 
+    _non_illumina_metrics: Mapped["SampleDigitisationMetrics"] = orm.relationship(
+        back_populates="sample"
+    )
+
     def __str__(self) -> str:
         return f"{self.internal_id} ({self.name})"
 
@@ -1039,7 +1043,7 @@ class SampleDigitisationMetrics(Base):
     batch_id: Mapped[int] = mapped_column(ForeignKey("batch_digitisation.id"))
 
     batch: Mapped[BatchDigitisation] = orm.relationship(back_populates="sample_metrics")
-    sample: Mapped[Sample] = orm.relationship(back_populates="non_illumina_metrics")
+    sample: Mapped[Sample] = orm.relationship(back_populates="_non_illumina_metrics")
 
     __mapper_args__ = {
         "polymorphic_on": "type",
--- a/cg/store/models.py
+++ b/cg/store/models.py
@@ -990,37 +990,34 @@ class Order(Base):
         return to_dict(model_instance=self)
 
 
-class DataGenerationUnit(Base):
-    """Model for storing data generation unit information."""
+class DigitisationUnit(Base):
+    """Parent model representing the consumable units (cells/chips) on which the sample data is generated."""
 
-    __tablename__ = "data_generation_unit"
+    __tablename__ = "digitisation_unit"
 
     id: Mapped[PrimaryKeyInt]
+    type: Mapped[MachinesTypes]
     model: Mapped[Str32]
     internal_id: Mapped[UniqueStr]
-    type: Mapped[MachinesTypes]
 
-    batch_metrics: Mapped[list["BatchDigitisationMetrics"]] = orm.relationship(
-        back_populates="unit"
-    )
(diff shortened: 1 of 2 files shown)


cg/store/models.py
```diff
@@ -987,3 +988,59 @@ class Order(Base):
 
     def to_dict(self):
         return to_dict(model_instance=self)
+
+
+class DigitisationUnit(Base):
```
Vince-janv, line 993: Describes the actual consumable. Fields such as consumable model (10B, S4, etc), expiration date etc.

cg/store/models.py
```diff
@@ -987,3 +988,59 @@ class Order(Base):
...
+    internal_id: Mapped[UniqueStr]
+
+    batch_metrics: Mapped[list["BatchDigitisation"]] = orm.relationship(back_populates="unit")
+
+    __mapper_args__ = {
+        "polymorphic_on": "type",
+    }
+
+
+class BatchDigitisation(Base):
```
Vince-janv, line 1010: Describes a "run" on the instrument. I see this as anything between starting the instrument, to spitting out BAM/FASTQ files. This can contain time stamps, performance metrics for the consumable used (item to the right). It should be noted that for ONT the flow cells can be run multiple times

seallard:
At a first glance, it seems abstract and I don't know what digitisation means. Are we creating a model that is too abstract?
Will we ever want to write code that interacts with data without knowledge of what kind of machine generated it?

It seems powerful though, I don't fully understand the implications at the moment. It would be nice, if possible, to give the base models more intuitive names.

diitaz93:
Is this limited only to the store? oR will affect models such as `FlowCellDirectoryData`?

ChrOertlin:
> This flow is designed for the Illumina flow to be integrated into the models as well, but my intention in the near future is to just add the new models and save the refactoring for the future.

From my experience with the genotype-api right now I would strongly advice to first get things into shape and then build on top of that.

ChrOertlin:
> At a first glance, it seems abstract and I don't know what digitisation means. Are we creating a model that is too abstract? Will we ever want to write code that interacts with data without knowledge of what kind of machine generated it?
> 
> It seems powerful though, I don't fully understand the implications at the moment. It would be nice, if possible, to give the base models more intuitive names.

I agree with this, digitisation seems like a very abstract thing. 

Something that would be a bit closer to the material could be "Readout". 

ReadoutMachine, ReadoutUnit, ReadoutBatch, ReadoutMetrics...

ChrOertlin:
> In general, I'd like to avoid noSQL since the data is much messier to interact with from my experience.

I think the way we implemented the DTO service layers would handle noSQL (mongodb) data quite nicely. You can basically retrieve jsons from the DB parse and immediatly validate data in the service layer for whatever structure you want.

ChrOertlin:
> Is this limited only to the store? Or will affect models such as `FlowCellDirectoryData`?

This should only be limited to the database. The FlowCellDirectoryData is not a database model in our system. I think the Saphyr, Nanopore etc. will all get their own "DirectoryData" if we choose to follow this pattern.

cg/store/models.py
```diff
@@ -987,3 +992,59 @@ class Order(Base):
...
+class DigitisationUnit(Base):
+    """Parent model representing the consumable units (cells/chips) on which the sample data is generated."""
+
+    __tablename__ = "digitisation_unit"
+
+    id: Mapped[PrimaryKeyInt]
+    type: Mapped[MachinesTypes]
+    model: Mapped[Str32]
+    internal_id: Mapped[UniqueStr]
+
```
ChrOertlin, line 1006:
```suggestion
batch_id: Mapped[int] = mapped_column(ForeignKey("RunBatch.id"))
```
We want to add this here s well?

ChrOertlin:
@Vince-janv can we close this?

Vince-janv:
Performed in other PR:s

PR #3112, opened 2024-04-12
Add conditional parameters for tomte

fevac:
### Added

- Conditional parameters for tomte: gene panel, genome and tissue

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-servers-stage.sh tomte_conditional_config

    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b tomte_update_params -a
    ```

### How to test

- [x] `cg workflow tomte config-case CASE`. 
  - [x] Make sure it creates a gene panel file called `gene_panels.bed` that is not empty and formatted as expected
 
image

  - [x] Make sure that `_nextflow_config.json` contains the following parameters set `params.gene_panel_clinical_filter` `params.tissue`  `params.genome`

image

- [x] `cg workflow tomte start CASE`. Starts as expected without errors. 

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


fevac:
Pending to revise in a different PR:
- get_aggregated_panels: has a weird behaviour. Make it clear and add documentation.
- get source for external and downsampled cases

tests/cli/workflow/nf_analysis/test_cli_start.py
```diff
@@ -7,14 +7,15 @@
...
+from cg.meta.workflow.analysis import AnalysisAPI
 from cg.meta.workflow.nf_analysis import NfAnalysisAPI
 from cg.models.cg_config import CGConfig
 from tests.cli.workflow.conftest import mock_analysis_flow_cell
 
 
 @pytest.mark.parametrize(
     "workflow",
-    [Workflow.RAREDISEASE, Workflow.RNAFUSION, Workflow.TAXPROFILER, Workflow.TOMTE],
+    Workflow.get_nf_workflows(),
```
diitaz93, line 18: I think this is why the tests are failing. Maybe rollback this line to the previous version. See https://github.com/pytest-dev/pytest-xdist/issues/432

diitaz93: or
```suggestion
    sorted(Workflow.get_nf_workflows()),
```
but I would prefer leaving it as it was before, it is more explicit

tests/cli/workflow/nf_analysis/test_cli_start.py
```diff
@@ -49,7 +53,7 @@ def test_start(
...
```
diitaz93, line 56: same here

cg/meta/workflow/nf_analysis.py
```diff
@@ -803,3 +835,68 @@ def get_cases_to_store(self) -> list[Case]:
...
+
+    def get_gene_panel_genome_build(self, case_id: str) -> GenePanelGenomeBuild:
+        """Return build version of the gene panel for a case."""
+        reference_genome = self.get_reference_genome(case_id=case_id)
+        try:
+            return getattr(GenePanelGenomeBuild, reference_genome)
+        except AttributeError as error:
+            LOG.error(
+                f"Reference {reference_genome} has no associated genome build for panels: {error}"
+            )
```
ivadym, line 863: shouldn't we raise a CGError here?

cg/meta/workflow/nf_analysis.py
```diff
@@ -803,3 +835,68 @@ def get_cases_to_store(self) -> list[Case]:
...
+        elif len(reference_genome) > 1:
+            raise CgError(
+                f"Samples linked to case {case_id} have different reference genome versions set"
+            )
+        else:
+            LOG.info(f"No reference genome specified, defaulting to {GenomeVersion.hg38}")
+
+    def get_gene_panel_genome_build(self, case_id: str) -> GenePanelGenomeBuild:
+        """Return build version of the gene panel for a case."""
+        reference_genome = self.get_reference_genome(case_id=case_id)
```
ivadym, line 857: typehint

cg/meta/workflow/nf_analysis.py
```diff
@@ -803,3 +835,68 @@ def get_cases_to_store(self) -> list[Case]:
...
+        try:
+            return getattr(GenePanelGenomeBuild, reference_genome)
+        except AttributeError as error:
+            LOG.error(
+                f"Reference {reference_genome} has no associated genome build for panels: {error}"
+            )
+
+    def get_gene_panel(self, case_id: str, dry_run: bool = False) -> list[str]:
```
ivadym, line 865: I would use directly the `get_gene_panel` function from the analysis API

cg/meta/workflow/nf_analysis.py
```diff
@@ -803,3 +835,68 @@ def get_cases_to_store(self) -> list[Case]:
...
+        return self._get_gene_panel(
+            case_id=case_id,
+            genome_build=self.get_gene_panel_genome_build(case_id=case_id),
+            dry_run=dry_run,
+        )
+
+    def get_source_from_lims_by_sample_id(self, sample_id: str) -> str | None:
+        """Get the source from LIMS for a given sample ID."""
+        lims_sample: dict | None = self.get_lims_sample(sample_id=sample_id, silent_error=True)
+        return lims_sample.get("source", None) if lims_sample else None
```
ivadym, line 876: This should be part of the Lims API

cg/meta/workflow/analysis.py
```diff
@@ -602,25 +598,57 @@ def _write_managed_variants(out_dir: Path, content: list[str]) -> None:
...
+        master_list: list[str] = GenePanelMasterList.get_panel_names()
+        if customer_id in GenePanelMasterList.collaborators() and default_panels.issubset(
+            master_list
+        ):
+            return master_list
+        all_panels: set[str] = add_gene_panel_combo(default_panels=default_panels)
+        all_panels |= {GenePanelMasterList.OMIM_AUTO, GenePanelMasterList.PANELAPP_GREEN}
+        return list(all_panels)
+
+    def get_lims_sample(self, sample_id: str, silent_error: bool = True) -> dict:
```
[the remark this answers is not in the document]
ivadym: And maybe move it to the LIMS API too, that would be great 🙏

cg/meta/workflow/nf_analysis.py
```diff
@@ -803,3 +835,68 @@ def get_cases_to_store(self) -> list[Case]:
...
+
+    def get_source_by_case(self, case_id: str) -> str:
```
ivadym, line 878:
```suggestion
    def get_case_source_type(self, case_id: str) -> str:
```

cg/meta/workflow/nf_analysis.py
```diff
@@ -803,3 +835,68 @@ def get_cases_to_store(self) -> list[Case]:
...
+        LOG.info(f"sources are {sources}")
+        if not sources:
+            return "unknown"
+        if len(sources) == 1:
+            return sources.pop()
+        if len(sources) > 1:
+            raise CgError(
+                f"All samples in a case must be of the same tissue type. "
+                f"Different sources found for {case_id}: {sources}."
+            )
```
ivadym, line 897: This is a bit too verbose, could we copy the implementation Mathias did for `get_case_application_type` here?

cg/meta/workflow/nf_analysis.py
```diff
@@ -803,3 +835,68 @@ def get_cases_to_store(self) -> list[Case]:
...
+
+    @staticmethod
+    def replace_non_alphanumeric(string: str, replace_by="_") -> str:
+        """Replace non-alphanumeric characters from a string."""
+        return re.sub(r"\W+", replace_by, string)
```
ivadym, line 902: This shouldn't be a method but a `TomteParameters` validator.

ivadym: And what do you think about using a more pythonic way instead of the regex?

strings have an `isalnum()` so we could iterate through the string and use a list comprehesion then to join them by the `replace_by` variable

cg/meta/workflow/nf_analysis.py
```diff
@@ -207,6 +221,12 @@ def get_workdir_path(self, case_id: str, work_dir: Path | None = None) -> Path:
             return work_dir.absolute()
         return Path(self.get_case_path(case_id), NFX_WORK_DIR)
 
+    def get_gene_panels_path(self, case_id: str) -> Path:
+        """Path to gene panels bed file exported from Scout."""
+        return Path(self.get_case_path(case_id=case_id), "gene_panels").with_suffix(
+            FileExtensions.BED
+        )
```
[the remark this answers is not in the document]
fevac: I believe it will

cg/apps/scout/scoutapi.py
```diff
@@ -49,7 +49,9 @@ def upload(self, scout_load_config: Path, force: bool = False) -> None:
         self.process.run_command(load_command)
         LOG.debug("Case loaded successfully to Scout")
 
-    def export_panels(self, panels: list[str], build: str = GENOME_BUILD_37) -> list[str]:
+    def export_panels(
+        self, panels: list[str], build: str = GENOME_BUILD_37, dry_run: bool = False
```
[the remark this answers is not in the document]
fevac: that was my intention at first but I chicken out since I saw it had quite a lot of usages and didn't want to make this PR bigger and make things fail. I could include it though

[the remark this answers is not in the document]
ivadym: I see... I would try to change it but it's up to you. When it's parsed as string the conversion is done automatically f"{GenePanelGenomeBuild. GENOME_BUILD_37}". For other cases I think we would need to add `.value` (to this one for example: (`["--build", build])`).

cg/meta/workflow/nf_analysis.py
```diff
@@ -803,3 +835,68 @@ def get_cases_to_store(self) -> list[Case]:
             if self.trailblazer_api.is_latest_analysis_completed(case_id=case.internal_id)
             or self.trailblazer_api.is_latest_analysis_qc(case_id=case.internal_id)
         ]
+
+    def get_reference_genome(self, case_id: str) -> GenomeVersion:
```
ivadym, line 839: Adding this method will create two parallel implementations, as we have `get_genome_build` in the report API (this one should've been added to the analysis API 😓)

cg/meta/workflow/analysis.py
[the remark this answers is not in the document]
ivadym: I'll try to do this in my PR of bypassing LIMS for the delivery reports @fevac

cg/meta/workflow/nf_analysis.py
ivadym: Also this part I'll try to fix in my PR; move all these analysis level methods from the delivery report to the analysis API @fevac

cg/meta/workflow/nf_analysis.py
fevac: there is no defined get_gene_panel in analysis API, are you suggesting that I move it there instead?

cg/meta/workflow/nf_analysis.py
fevac: updated

cg/meta/workflow/nf_analysis.py
fevac: About your second suggestion, I think having a regex is easier to understand that and to follow than iterating and joining over values. So I will keep the regex

cg/meta/workflow/analysis.py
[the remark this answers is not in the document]
fevac: method removed and using lims_api.sample direclty

cg/meta/workflow/nf_analysis.py
```diff
@@ -803,3 +835,68 @@ def get_cases_to_store(self) -> list[Case]:
...
+        """Return reference genome for a case.
+        If this information is not available in StatusDB, it defaults to hg38."""
+        reference_genome: set[str] = {
+            sample.reference_genome
+            for sample in self.status_db.get_samples_by_case_id(case_id=case_id)
+        }
```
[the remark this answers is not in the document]
fevac: I will remove the default value. But we need the reference genome information from the order for tomte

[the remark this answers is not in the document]
ivadym: Please merge master into this branch, we have now `get_genome_build` available in analysis API

cg/meta/workflow/nf_analysis.py
ivadym: 👍

cg/meta/workflow/nf_analysis.py
ivadym: I meant to use directly `_get_gene_panel`, but I see now that we're using it from the base calls.

Could you then just define it in the analysis api and raise there a not implemented error so we fix the references 🙏

fevac: done

cg/meta/workflow/nf_analysis.py
fevac: moved

tests/cli/workflow/nf_analysis/test_cli_start.py
fevac: ok I reverted it for now until we have decided on an approach about how to do this

tests/cli/workflow/nf_analysis/test_cli_start.py
fevac: reverted

cg/apps/lims/api.py
```diff
@@ -9,15 +9,11 @@
...
-    PROP2UDF,
-    DocumentationMethod,
-    LimsArtifactTypes,
-)
+from cg.constants.lims import MASTER_STEPS_UDFS, PROP2UDF, DocumentationMethod, LimsArtifactTypes
 from cg.exc import LimsDataError
 
 from ...constants import Priority
+from ...constants.sample_sources import SourceType
 from .order import OrderHandler
```
diitaz93, line 17: make these absolute imports, i.e. starting with `cg`

cg/apps/lims/api.py
```diff
@@ -68,6 +64,12 @@ def samples_in_pools(self, pool_name, projectname):
         """Fetch all samples from a pool"""
         return self.get_samples(udf={"pool name": str(pool_name)}, projectname=projectname)
 
+    def get_source(self, lims_id: str) -> str:
```
diitaz93, line 67:
```suggestion
    def get_source(self, lims_id: str) -> str | None:
```

cg/meta/workflow/analysis.py
```diff
@@ -165,6 +161,19 @@ def get_case_application_type(self, case_id: str) -> str:
 
         return application_types.pop()
 
+    def get_case_source_type(self, case_id: str) -> str:
+        """Returns the source type for samples in a case.
+        Raises CgError if different sources are set for the samples linked to a case."""
```
diitaz93, line 166:
```suggestion
        """
        Returns the source type for samples in a case.
        Raises: 
            CgError: If different sources are set for the samples linked to a case.
        """
```

cg/meta/workflow/analysis.py
```diff
@@ -165,6 +161,19 @@ def get_case_application_type(self, case_id: str) -> str:
...
+        sample_ids: Iterator[str] = self.status_db.get_sample_ids_by_case_id(case_id=case_id)
+        source_types: set[str] = {
+            self.lims_api.get_source(lims_id=sample_id) for sample_id in sample_ids
+        }
+
+        if len(source_types) > 1:
+            raise CgError(f"Different source types found for case: {case_id} ({source_types})")
```
diitaz93, line 173: The function `self.lims_api.get_source` returns `None` if the sample with the lims id is not found. If this is the case, there will be more than one entry in the set `source_types`. Then the error message would be misleading. This behaviour should be handled explicitly. I'm not sure where, but maybe directly in `self.lims_api.get_source` so that it does not return None if the sample does not exist.

cg/meta/workflow/nf_analysis.py
```diff
@@ -344,6 +363,16 @@ def create_nextflow_config(self, case_id: str, dry_run: bool = False) -> None:
                 file_path=self.get_nextflow_config_path(case_id=case_id),
             )
 
+    def create_gene_panel(self, case_id: str, dry_run: bool) -> None:
+        """Create and write an aggregated gene panel file exported from Scout."""
+        LOG.info("Creating gene panel file")
+        bed_lines: list[str] = self.get_gene_panel(case_id=case_id, dry_run=dry_run)
+        if dry_run:
+            bed_lines: str = "\n".join(bed_lines)
+            LOG.debug(f"{bed_lines}")
```
diitaz93, line 372: unsure if we want debug or info logging here, as `--dry-run` already implies that you are in a debugging more. Feel free to disregard
```suggestion
            LOG.info(f"{bed_lines}")
```

diitaz93 (changes requested):
There is potential unexpected behaviour when the sample is not found in LIMS that is not handled in the function `LimsAPI.get_source()`, could you take a look at that?

cg/meta/workflow/nf_analysis.py
fevac: merged and converged with your PR

cg/meta/workflow/nf_analysis.py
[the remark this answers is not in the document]
fevac: merged and converged with your PR

cg/apps/lims/api.py
fevac: I added a note in the docstring too about returning None

cg/meta/workflow/nf_analysis.py
fevac: I don't think we want this in the Logs as it can be quite long and would clutter it. Also, this is only useful for debugging.
I will leave it as it is

cg/apps/lims/api.py
fevac: Sorry I had missed your comment below. I will modify it so the exception is handled in  `get_source` directly and remove it from the validator

cg/meta/workflow/analysis.py
fevac: updated `get_source` to handle directly the exception and return 'unknown' when no lims information can be fetched retrieved

cg/apps/lims/api.py
ivadym: If there is no sample in LIMS it should return None. Additionally, I don't think we allow `Unknown` anymore from the order portal

cg/constants/constants.py
```diff
@@ -139,6 +139,10 @@ class Workflow(StrEnum):
     TAXPROFILER: str = "taxprofiler"
     TOMTE: str = "tomte"
 
+    @classmethod
+    def get_nf_workflows(cls) -> set:
+        return {cls.RNAFUSION, cls.TAXPROFILER, cls.RAREDISEASE, cls.TOMTE}
```
ivadym, line 144: Should we use the constant that Annick defined? I would copy that constant and add it here would be an easy conflict to resolve

cg/constants/sample_sources.py
```diff
@@ -27,3 +29,7 @@
     "bone marrow",
     "other",
 )
+
+
+class SourceType(StrEnum):
+    UNKNOWN: str = "unknown"
```
ivadym, line 35: I don't think we need this, and it seems from the constants above that the normalised name for unknown is `other`

cg/meta/workflow/analysis.py
ivadym: That's why `source_types` it's a `set`.  We'll have a single `None` there and the method would behave as expected. 

Just make sure to update the typehints of the method and source types

cg/meta/workflow/analysis.py
```diff
@@ -609,26 +619,49 @@ def _write_managed_variants(out_dir: Path, content: list[str]) -> None:
...
+    def _get_gene_panel(self, case_id: str, genome_build: str, dry_run: bool = False) -> list[str]:
         """Create and return the aggregated gene panel file."""
         case: Case = self.status_db.get_case_by_internal_id(internal_id=case_id)
         all_panels: list[str] = self.get_aggregated_panels(
             customer_id=case.customer.internal_id, default_panels=set(case.panels)
         )
-        return self.scout_api.export_panels(build=genome_build, panels=all_panels)
+        return self.scout_api.export_panels(build=genome_build, panels=all_panels, dry_run=dry_run)
+
+    def get_gene_panel(self, case_id: str, dry_run: bool = False) -> list[str]:
```
ivadym, line 638: Could you add the `dry_run` option to the child methods (mip rna & mip dna) 🙏

cg/meta/workflow/nf_analysis.py
```diff
@@ -111,6 +112,11 @@ def is_multiqc_pattern_search_exact(self) -> bool:
         If false, pattern must be present but does not need to be exact."""
         return False
 
+    @property
+    def is_gene_panel_required(self) -> bool:
+        """Return True if a gene panel is needs to be created using the information in StatusDB and exporting it from Scout."""
+        return False
```
ivadym, line 118: I'm not sure if this extra logic is needed. I'd lean towards directly calling `create_gene_panel`. If a workflow doesn't need this functionality, it can override the method in its specific API and just return None.

cg/meta/workflow/nf_analysis.py
```diff
@@ -812,6 +836,41 @@ def get_cases_to_store(self) -> list[Case]:
...
+            for sample in self.status_db.get_samples_by_case_id(case_id=case_id)
+        }
+        if len(reference_genome) == 1:
+            return reference_genome.pop()
+        elif len(reference_genome) > 1:
+            raise CgError(
+                f"Samples linked to case {case_id} have different reference genome versions set"
+            )
+        else:
+            raise CgError("No reference genome specified")
```
ivadym, line 854:
```suggestion
    def get_genome_build(self, case_id: str) -> GenomeVersion:
        """Return reference genome version for a case.
        Raises CgError if this information is missing or inconsistent for the samples linked to a case.
        """
        reference_genome: set[str] = {
            sample.reference_genome
            for sample in self.status_db.get_samples_by_case_id(case_id=case_id)
        }
        if len(reference_genome) == 1:
            return reference_genome.pop()
        if len(reference_genome) > 1:
            raise CgError(
                f"Samples linked to case {case_id} have different reference genome versions set"
            )
        raise CgError("No reference genome specified")
```

ivadym, line 854:
```suggestion
            raise CgError(f"No reference genome specified for case {case_id}")
```

cg/apps/lims/api.py
fevac: That was my original thought. However, unknown is not a value that LIMS can register, so in a way, maybe there is no much confusion. 

If @diitaz93 is ok with this behaviour I will revert it again then and document the return of None for whenever there is either no information available or no sample could be fetch

cg/meta/workflow/nf_analysis.py
```diff
@@ -812,6 +836,41 @@ def get_cases_to_store(self) -> list[Case]:
             or self.trailblazer_api.is_latest_analysis_qc(case_id=case.internal_id)
         ]
 
+    def get_genome_build(self, case_id: str) -> GenomeVersion:
```
ivadym, line 839: This will raise an error for rnafusion as the genome version is not added to the samples (and break the uploads as the delivery report needs this field 😓 )

cg/meta/workflow/nf_analysis.py
diitaz93: leave it, but the production logs won't get clogged because we don't run it in dry-run

cg/constants/constants.py
fevac: I think this can wait. We will have to update all pytests anyways. So we can address it in a different PR to avoid growing changes here

cg/meta/workflow/analysis.py
diitaz93: discussed in huddle. The warning in `self.lims_api.sample()` warns about the sample not being in LIMS. Resolving

cg/models/tomte/tomte.py
```diff
@@ -41,6 +46,25 @@ def list(cls) -> list[str]:
...
+    @validator("tissue", pre=True)
+    def restrict_tissue_values(cls, tissue: str) -> str:
+        return replace_non_alphanumeric(string=tissue)
+
+    @validator("genome", pre=True)
+    def restrict_genome_values(cls, genome: str) -> str:
+        if genome == GenomeVersion.hg38:
+            return "GRCh38"
+        elif genome == GenomeVersion.hg19:
+            return "GRCh37"
```
ivadym, line 65: We should use the genome version directly

```suggestion
    genome: GenomeVersion

    @validator("tissue", pre=True)
    def restrict_tissue_values(cls, tissue: str) -> str:
        return replace_non_alphanumeric(string=tissue)
```

tests/cli/workflow/nf_analysis/test_cli_config_case.py
```diff
@@ -100,13 +103,14 @@ def test_config_case_without_samples(
 
 @pytest.mark.parametrize(
     "workflow",
-    [Workflow.RNAFUSION, Workflow.TAXPROFILER, Workflow.RAREDISEASE, Workflow.TOMTE],
+    [Workflow.RAREDISEASE, Workflow.RNAFUSION, Workflow.TAXPROFILER, Workflow.TOMTE],
```
ivadym, line 106: constant

cg/constants/constants.py
ivadym: 👍

cg/apps/lims/api.py
fevac: discussed and modified

cg/constants/sample_sources.py
fevac: "other" is for other tissues not represented/listed in the order. This is, we have information about the tissue type (source) from the order but non of the accepted fields match the actual tissue type for a sample. This is not the same as not having information about the source at all.

cg/meta/workflow/analysis.py
```diff
@@ -165,6 +161,20 @@ def get_case_application_type(self, case_id: str) -> str:
 
         return application_types.pop()
 
+    def get_case_source_type(self, case_id: str) -> str:
+        """Returns the source type for samples in a case.
+        Raises:
+            CgError: If different sources are set for the samples linked to a case."""
+        sample_ids: Iterator[str] = self.status_db.get_sample_ids_by_case_id(case_id=case_id)
```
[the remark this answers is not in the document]
fevac: added

cg/constants/sample_sources.py
ivadym: Sounds good, I see you're using it as an input parameter for Tomte, that's fine I was unsure about setting it to unknown if it comes from lims since I'm not sure it's supported there 👍

cg/meta/workflow/nf_analysis.py
fevac: I don't really like this suggestion. Overriding a method to return None would imply using a method that doesn't do what's supposed to do. In this case, calling a function named `create_gene_panel` that does not do what it's name says (does not create a gene panel) sounds super confusing to me. In my opinion that would obscure the logic of the code.

I think having an extra property is clearer and I do think this logic is needed if we want to keep a unified `config_case` method.

ivadym: Fair point, but I think that's actually the idea of having nested classes, you try to unify the logic and move it into the workflows. Probably something we should discuss and include in our coding guidelines but let's move on for now 🏁 🏁

cg/models/tomte/tomte.py
fevac: I need the validator for the genome. The values to use  in the pipeline are different than the ones used in cg:
hg19/hg38 vs GRCh37/GRCh38

cg/meta/workflow/nf_analysis.py
fevac: Good catch! I will overwrite for rnafusion then

diitaz93 (approved):
my concerns have been addressed 👍

cg/models/tomte/tomte.py
ivadym: Oh I see 😓

cg/meta/workflow/analysis.py
fevac: and of course if I change this these pytests fail and need to be adapted. I'm giving up today. I will see tomorrow if I decide to address it or not

cg/apps/scout/scoutapi.py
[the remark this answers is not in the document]
fevac: I'm leaving this for now. Too many issues already in this PR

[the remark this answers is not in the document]
fevac: Will address it in a dedicated PR

cg/constants/gene_panel.py
```diff
@@ -8,6 +8,11 @@
 GENOME_BUILD_38: str = "GRCh38"
 
 
+class GenePanelGenomeBuild(StrEnum):
+    hg19: str = GENOME_BUILD_37
+    hg38: str = GENOME_BUILD_38
```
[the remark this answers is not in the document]
fevac: leaving for now - will be updated in the dedicated PR

cg/apps/scout/scoutapi.py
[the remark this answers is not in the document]
ivadym: Yes sounds good 👍

cg/meta/workflow/analysis.py
ivadym: Mmmm they should not, It's adding the argument to the child methods with a default value of false and passing it to the _get_gene_panel method

cg/apps/lims/api.py
```diff
@@ -68,6 +63,13 @@ def samples_in_pools(self, pool_name, projectname):
         """Fetch all samples from a pool"""
         return self.get_samples(udf={"pool name": str(pool_name)}, projectname=projectname)
 
+    def get_source(self, lims_id: str) -> str | None:
+        """Return the source from LIMS for a given sample ID.
+        Return 'None' if no source information is set or
+        if sample is not found or cannot be fetched from LIMS."""
+        if lims_sample := self.sample(lims_id=lims_id):
+            return lims_sample.get("source")
```
ivadym, line 71: No need of an `if` as we discussed:
```suggestion
        lims_sample: dict[str, Any] = self.sample(lims_id=lims_id)
        return lims_sample.get("source")
```

cg/meta/workflow/rnafusion.py
```diff
@@ -57,6 +57,10 @@ def is_multiple_samples_allowed(self) -> bool:
         """Return whether the analysis supports multiple samples to be linked to the case."""
         return False
 
+    def get_genome_build(self, case_id: str) -> GenomeVersion:
+        """Return reference genome for a case. Currently fixed for hg19."""
```
ivadym, line 61:
```suggestion
        """Return reference genome for a case. Currently fixed for hg38."""
```

cg/apps/lims/api.py
fevac: you're right. Banana brain...

cg/meta/workflow/analysis.py
fevac: sorry I misunderstood your comment. Now I get it. Done

fevac:
tests are passing

fevac:
**Note**

This branch as not been deployed in production due to issues with Hasta

PR #3125, opened 2024-04-15
Add admin view for orders

seallard:
Not sure how to setup the visualization for cases in the order since it is not a model. Leaving it for later.


cg/server/admin.py
```diff
@@ -448,6 +448,14 @@ class OrganismView(BaseView):
     column_searchable_list = ["internal_id", "name", "reference_genome"]
 
 
+class OrderView(BaseView):
+    """Admin view for Model.Order"""
+
+    column_default_sort = ("order_date", True)
+    create_modal = True
+    edit_modal = True
```
islean, line 456: Could we add that we can toggle "is_delivered" without editing the whole order?

islean:
FYI, you can view all connected cases by editing an order:

image

cg/server/admin.py
seallard: Not sure how to do that atm, will add later

islean (approved):
Looks good as a first rendition!

PR #3131, opened 2024-04-16
Add analysis ids to the delivery message response

islean:
## Description

In order to mark the analyses included in the delivery message for an order as delivered, the analysis ids used are now included in the response object.

### Changed

- The analysis ids which the order delivery message was generated for are included in the response.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/server/dto/delivery_message/delivery_message_response.py
+++ b/cg/server/dto/delivery_message/delivery_message_response.py
@@ -3,3 +3,7 @@ from pydantic import BaseModel
 
 class DeliveryMessageResponse(BaseModel):
     message: str
+
+
+class DeliveryMessageOrderResponse(DeliveryMessageResponse):
+    analysis_ids: list[int]
--- a/cg/services/delivery_message/delivery_message_service.py
+++ b/cg/services/delivery_message/delivery_message_service.py
@@ -1,14 +1,15 @@
 from cg.apps.tb import TrailblazerAPI
 from cg.apps.tb.models import TrailblazerAnalysis
-from cg.exc import OrderNotDeliverableError, OrderNotFoundError
+from cg.exc import OrderNotDeliverableError
 from cg.server.dto.delivery_message.delivery_message_request import (
     DeliveryMessageRequest,
 )
 from cg.server.dto.delivery_message.delivery_message_response import (
+    DeliveryMessageOrderResponse,
     DeliveryMessageResponse,
 )
 from cg.services.delivery_message.utils import get_message, validate_cases
-from cg.store.models import Case, Order
+from cg.store.models import Case
 from cg.store.store import Store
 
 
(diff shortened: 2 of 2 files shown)


cg/services/delivery_message/delivery_message_service.py
```diff
@@ -21,20 +22,23 @@ def get_cases_message(
         self, delivery_message_request: DeliveryMessageRequest
     ) -> DeliveryMessageResponse:
         case_ids: list[str] = delivery_message_request.case_ids
-        return self._get_delivery_message(set(case_ids))
+        message: str = self._get_delivery_message(set(case_ids))
+        return DeliveryMessageResponse(message=message)
```
seallard, line 26: I like it ⭐

cg/server/dto/delivery_message/delivery_message_response.py
```diff
@@ -3,3 +3,7 @@
 
 class DeliveryMessageResponse(BaseModel):
     message: str
+
+
+class DeliveryMessageOrderResponse(DeliveryMessageResponse):
```
seallard, line 8: I've started to question our naming pattern, what do you think of skipping the `Response` part? If necessary to denote the type of model, maybe we should just add `DTO`.

The data model does not really have anything to do with it ultimately being sent as a payload of a HTTP request. We could reuse the service producing these models in other contexts. I think the `Response` part introduces an implicit coupling to a HTTP response that is potentially misleading and unnecessary

seallard (approved):
Nice 👍 Some thoughts regarding our naming pattern, feel free to start changing it in this PR if you agree

cg/server/dto/delivery_message/delivery_message_response.py
islean: Hmm, I do like the "Response" and the implicit coupling 🤔 Might be enough to look at the typed response type though. How would we discern what the request and response is?

I mean these are designed specifically for the endpoints and I do not like them being used for other things. I was even debating extending the `DeliveryMessageResponse` class for this reason.

islean:
Tested in stage and the analysis_ids are included:

image

cg/server/dto/delivery_message/delivery_message_response.py
islean: I will leave as is for now, and we could consider changing it in the future.

seallard: It is unlikely that we would use the delivery message service in a different context, but not the `OrderService` for instance. The data is just data, it is not necessarily supposed to be sent in a http response.

For instance, consider that the service is called in a CLI command. Or inside another service. I think we should drop the `Response` suffix from models in the service layer (we probably can keep it for HTTP API clients).

Feel free to leave it as is, just want to point out this coupling that is misleading.

islean: I mean, I would always want a DTO specifically for the requests and responses for endpoints. I guess we could discuss where they should be placed though?

I think it adds value that we do not see them as only data. This is data specifically packaged for this purpose in my mind. We could of course discuss it in a meeting or in slack or in person etc.

islean:
Deployed to `cg-services-prod`.

PR #3138, opened 2024-04-16
Remove order date on case

seallard:
## Description
The order date does not belong on a case. It is especially confusing since we can have multiple orders with the same case. We store the correct date in the order table.

We also do not use `case.ordered_at` date for anything, so I say: TRASH IT 🗑️ 

Production is OK with this given https://github.com/Clinical-Genomics/cg/pull/3143

### Fixed
- Drop unused `ordered_at` column from case table


cg/store/filters/status_case_filters.py
```diff
@@ -272,7 +266,6 @@ class CaseFilter(Enum):
     HAS_SEQUENCE: Callable = filter_cases_has_sequence
     IS_RUNNING: Callable = filter_running_cases
     IS_COMPRESSIBLE: Callable = filter_compressible_cases
-    NEW_BY_ORDER_DATE: Callable = filter_newer_cases_by_order_date
```
seallard, line 275: This filter is not used anywhere 🤷

cg/store/crud/read.py
```diff
@@ -1037,7 +1037,7 @@ def cases_to_analyze(
             workflow=workflow,
         )
 
-        families: list[Query] = list(cases.order_by(Case.ordered_at))
```
seallard, line 1040: This sorting is arbitrary 🤷

alembic/versions/2024_04_16_0278168054c8_remove_order_date_from_case.py
```diff
@@ -0,0 +1,25 @@
+"""Remove order date from case
+
+Revision ID: 0278168054c8
+Revises: ac5a804a9f47
```
seallard, line 4: Ensure this is correct with master before merging and deploying.

islean:
Has this been cleared with production? I would guess that they see a value in having these dates even if the code can run without it?

seallard:
> Has this been cleared with production? I would guess that they see a value in having these dates even if the code can run without it?

It has been cleared, once https://github.com/Clinical-Genomics/cg/pull/3143 is merged we can go ahead

seallard:
Blocked because production runs a script weekly using this data

PR #3139, opened 2024-04-16
Remove order date from pool

seallard:
The order date does not belong on a pool. We store it in the order table.

We also do not use `pool.ordered_at` date for anything, so I say: TRASH IT 🗑️ 

Blocked by https://github.com/Clinical-Genomics/cigrid-ui/issues/620

### Fixed
- Drop unused `ordered_at` column from pool table


islean:
Has this been cleared with production?

seallard:
> Has this been cleared with production?

It has been cleared, we can go ahead once https://github.com/Clinical-Genomics/cg/pull/3143 is merged

seallard:
Blocked because production runs a script weekly using this data

PR #3143, opened 2024-04-17
Add delivery date

seallard:
## Description
Add delivery date on orders, will enable turn-around calculations on order level.

### Added
- Delivery date on order


alembic/versions/2024_04_17_752e8f0ce514_add_delivery_date_to_order.py
```diff
@@ -0,0 +1,25 @@
+"""Add delivery date to order
+
+Revision ID: 752e8f0ce514
+Revises: ac5a804a9f47
```
seallard, line 4: Make sure this revision id is correct before merging

diitaz93 (approved):
Looks good 👍

Issue #3145, opened 2024-04-17
Limit number of cases fetched to be started.


Vince-janv:
### Description
Starting mip-dna cases takes so long that start-available sometimes does not finish with 1 hour when the command is run again. The bug is described here: https://github.com/Clinical-Genomics/bug-reports/issues/1

The same scaling issues might exist in the other pipelines.

### Suggested solution:
- Estimate (by looking at the logs) how many cases can be started within an hour for mip and balsamic
- In `cg.meta.workflow.mip.MipAnalysisAPI.get_cases_to_analyze` (and maybe balsamic equivalent) pass the limit parameter to `cg.store.crud.read.ReadHandler.cases_to_analyze`

Vince-janv:
Time to start a case:
- mip-dna ~ 90 seconds
- balsamic ~140 seconds

islean:
What margin would we want? 10 min? @Vince-janv

islean:
And do we see issues with RNAFusion? Like should this be applied for all pipelines?

Vince-janv:
6-10 min sounds good. I added RNAFUSION since it's the only one where I think we might start many cases at once. But since they take 2 sec there's no point in adding a limit IMO

islean:
Fixed in #3147

PR #3147, opened 2024-04-17
Limit analyses started in MIP and Balsamic

islean:
## Description

Addresses #3145. I added a margin of just over 10 minutes.

### Changed

- MIP cases started are limited to 33
- Balsamic cases started are limited to 21

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b limit-mip-balsamic-starts -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/workflow/balsamic.py
+++ b/cg/meta/workflow/balsamic.py
@@ -94,7 +94,7 @@ class BalsamicAnalysisAPI(AnalysisAPI):
 
     def get_cases_to_analyze(self) -> list[Case]:
         cases_query: list[Case] = self.status_db.cases_to_analyze(
-            workflow=self.workflow, threshold=self.use_read_count_threshold
+            workflow=self.workflow, threshold=self.use_read_count_threshold, limit=21
         )
         cases_to_analyze = []
         for case_obj in cases_query:
--- a/cg/meta/workflow/balsamic.py
+++ b/cg/meta/workflow/balsamic.py
@@ -30,6 +30,8 @@ from cg.utils.utils import build_command_from_dict, get_string_from_list_by_patt
 
 LOG = logging.getLogger(__name__)
 
+MAX_CASES_TO_START_IN_50_MINUTES = 21
+
 
 class BalsamicAnalysisAPI(AnalysisAPI):
     """Handles communication between BALSAMIC processes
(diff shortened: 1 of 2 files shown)


cg/meta/workflow/balsamic.py
```diff
@@ -94,7 +94,7 @@ def get_case_path(self, case_id: str) -> Path:
 
     def get_cases_to_analyze(self) -> list[Case]:
         cases_query: list[Case] = self.status_db.cases_to_analyze(
-            workflow=self.workflow, threshold=self.use_read_count_threshold
+            workflow=self.workflow, threshold=self.use_read_count_threshold, limit=21
```
ivadym, line 97: We should define the values as constant, same as we're doing with `MAX_ITEMS_TO_RETRIEVE`

diitaz93 (approved):
👍

ChrOertlin:
the 10 minute margin: is that reducing the number of available cases to start so that there is a buffer of 10 minutes between the first and the second service? 

If so I would likely reduce that margin to 3-4 times the time it takes for an analysis to start given a workflow.

That is:
Balsamic; case start time 140s thus 420second buffer.
Mip; case start time 90s thus 360second buffer.

Just to optimise it a bit.

cg/meta/workflow/balsamic.py
ivadym, line 1: Should we apply this limit to the other pipelines too? I guess we don't have that many cases there, but it might become a problem in the future 🤔

ChrOertlin: seems like mip and balsamic have extreme starting times compared to rnafusion (140s,90s compared to 2s) Perhaps the nf-core pipelines do not need it. Have to see what the rare-disease pipeline also, if that changes things.

islean: So start with this? Or what is the conclusion?

Vince-janv: I think this is only relevant for mip-dna and balsamic. Even though there might be a theoretical bug if we get thousands of RNAFUSION cases.

islean:
> the 10 minute margin: is that reducing the number of available cases to start so that there is a buffer of 10 minutes between the first and the second service?
> 
> If so I would likely reduce that margin to 3-4 times the time it takes for an analysis to start given a workflow.
> 
> That is: Balsamic; case start time 140s thus 420second buffer. Mip; case start time 90s thus 360second buffer.
> 
> Just to optimise it a bit.

Right now it was calculated to run for roughly 50 minutes instead of an hour. I guess we could potentially optimise it further but I have no idea how much it fluctuates. Is the optimisation worth the risk? @ChrOertlin

cg/meta/workflow/balsamic.py
islean: Done

PR #3148, opened 2024-04-17
add(create validation case)

ChrOertlin:
## Description

Add a CLI command that allows users to create a validation case from an existing case in statusdb. 

Creates a new case in statusdb as well as new samples.
Samples will be renamed to use "VAL" (validation) instead of "ACC" sample internal id prefix. 
Case internal id will be generated by cg and cannot be assigned. 
Case sample relationships are taken over.

In housekeeper new sample bundles will be created.
Fastq files from original samples will be copied into the new validation bundle and renamed to contain the new validation identifier.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/create_validation_cases/validation_cases_api.py
```diff
@@ -0,0 +1,175 @@
...
+from cg.constants import SequencingFileTag
+from cg.models.validation_cases.validation_case_data import ValidationCaseData
+from cg.store.models import Case, CaseSample, Sample
+from cg.store.store import Store
+from cg.utils.files import copy_file, get_files_matching_pattern, rename_file
+
+LOG = logging.getLogger(__name__)
+
+
+class ValidationCaseAPI:
```
ivadym, line 18:
```suggestion
class CreateValidationCaseAPI:
```

Or are you leaving the option for this API to also perform the validation in the future?

cg/models/validation_cases/validation_case_data.py
```diff
@@ -0,0 +1,118 @@
...
+from cg.utils.calculations import multiply_by_million
+
+LOG = logging.getLogger(__name__)
+
+
+class ValidationCaseData:
+    def __init__(
+        self,
+        status_db: Store,
+        hk_api: HousekeeperAPI,
```
ivadym, line 19: I don't think we should have the DB APIs here. Models are a way to encapsulate data, and provide methods to work with that data, but we should extract it in the CreateValidationCase API. Then you could also make use of a pydantic models here to perform all the validations


```suggestion
```

cg/meta/create_validation_cases/validation_cases_api.py
```diff
@@ -0,0 +1,175 @@
...
+            fastq_file.absolute().as_posix().replace(validation_sample.from_sample),
+            validation_sample.internal_id,
+        )
+
+    def add_validation_samples_to_hk(
+        self, sample_id: str, file_path: Path, tags: list[str]
+    ) -> None:
+        """Add the new validation sample bundle to housekeeoer."""
+        version: Version = self.hk_api.get_latest_bundle_version(sample_id)
+        self.hk_api.add_file(
```
ivadym, line 158: could we directly use `include_files_to_latest_version` here?

cg/models/validation_cases/validation_case_data.py
```diff
@@ -0,0 +1,118 @@
...
+        """
+        Generate a case for the validation samples.
+        The new case uses existing case data.
+        Customer will be set to cust000.
+        """
+        if self.status_db.case_with_name_exists(case_name=self.validation_case_name):
+            raise ValueError(f"Case with name {self.case_name} already exists.")
+        validation_case: Case = self.status_db.add_case(
+            data_analysis=self.original_case.data_analysis,
+            data_delivery=DataDelivery.NO_DELIVERY,
```
ivadym, line 104: These two fields should be part of the CLI options when generating the new validation cases. If not provided, they could default to using the original cases data analysis and `NO_DELIVERY`

ivadym: Any other options we should be able to select when creating new validation cases that you think we would need? @fevac @mathiasbio

cg/meta/create_validation_cases/validation_cases_api.py
ChrOertlin: include_files would attempt to copy the files into the bundle directory, since these are already there we don't need this. And I think it would also fail since the files it tries to copy already exist.

cg/models/validation_cases/validation_case_data.py
ChrOertlin: I moved it out of the models as the class does not represent a typical model class indeed.

cg/models/validation_cases/validation_case_data.py
mathiasbio: Not that I can think of! There is more information that could be added for future automatic validations, such as what to validate for the specific case we're adding but I don't think we're ready for that yet!

mathiasbio:
This looks so nice..... 😍

cg/models/validation_cases/validation_case_data.py
mathiasbio: I guess the only thing I can think of, but for which I cannot remember a use-case for at the moment, is the option to add the level of downsampling for the new samples created for the validation-case. 

If I understand it correctly. In case someone needed a downsampled sample specifically for a validation, they would with the current solution need to downsample it first using the downsample feature, which would create the downsampled case, and then specify that downsampled case here, which if I understood it correctly will create 2 additional new samples. If we could do the downsampling here already it could save some time and space 🤷

cg/meta/create_validation_cases/validation_case_data.py
```diff
@@ -0,0 +1,124 @@
...
+                if self.input_data.delivery
+                else self.original_case.delivery
+            ),
+            name=self.validation_case_name,
+            panels=self.original_case.panels,
+            priority=self.original_case.priority,
+            ticket=self.original_case.latest_ticket,
+        )
+        validation_case.orders.append(self.original_case.latest_order)
+        validation_case.customer = "cust000"
```
fevac, line 117: use constant

cg/meta/create_validation_cases/validation_case_data.py
```diff
@@ -0,0 +1,124 @@
...
+        validation_case.is_compressible = False
```
fevac, line 118: Add action as `hold`

cg/meta/create_validation_cases/validation_cases_api.py
```diff
@@ -0,0 +1,187 @@
...
+                )
+                rename_file(file_path=fastq_file, renamed_file_path=new_fast_file_path)
+                new_tags: list[str] = self.get_new_tags(validation_sample)
+                self.add_validation_samples_to_hk(
+                    sample_id=validation_sample.internal_id,
+                    file_path=new_fast_file_path,
+                    tags=new_tags,
+                )
+
+    def get_new_tags(self, validation_sample: Sample) -> list[str]:
```
fevac, line 134: what if the case is compressed and we do not have the fastq tag?

cg/meta/create_validation_cases/validation_cases_api.py
```diff
@@ -0,0 +1,187 @@
...
+        self.dry_run = dry_run
+
+    def get_validation_case_data(self, validation_data_input: ValidationDataInput):
+        return ValidationCaseData(
+            status_db=self.status_db,
+            hk_api=self.hk_api,
+            validation_data_input=validation_data_input,
+        )
+
+    def store_validation_samples(self, validation_case_data: ValidationCaseData) -> None:
```
fevac, line 33: should this also store info in LIMS?

fevac (commented):
Looking good! few comments

cg/models/validation_cases/validation_case_data.py
fevac: As cli options I think it's enough with data analysis and data delivery. If any, adding/appending a comment but I don't use it and I don't know if someone does.

Then other options that should be not copied from the original case are: customer (cust000), action (hold), priority (standard?)

cg/meta/create_validation_cases/validation_case_data.py
```diff
@@ -0,0 +1,124 @@
...
+        self.original_case: Case = self.get_case_to_copy()
+        self.original_samples: list[Sample] = self.get_samples_to_copy()
+        self.validation_samples: list[Sample] = self.get_validation_samples()
+        self.validation_case: Case = self._generate_validation_case()
+        LOG.info(f"Validation Data checks completed for {self.case_id}")
+
+    @staticmethod
+    def _validation_sample_name(sample: Sample) -> str:
+        """Return a new validation sample identifier. The identifier removes "ACC" and prepends "VAL"."""
+        return sample.internal_id.replace("ACC", "VAL")
```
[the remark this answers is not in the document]
islean: We do have a stage user in lims so all orders submitted via the stage order portal will have "STG" instead of "ACC". Just adding an example.

cg/meta/create_validation_cases/validation_cases_api.py
ChrOertlin: You will need to make sure the case is available and decompressed at the moment.

cg/meta/create_validation_cases/validation_case_data.py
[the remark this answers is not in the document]
ChrOertlin: It will now take the first three letter and changes them to "VAL"... In cases where a sample has something else then a conventional id (3 letters and some numbers) this would be a bit iffy...

perhaps a "validation" suffix would be better?

cg/meta/create_validation_cases/validation_cases_api.py
ChrOertlin: What information is pulled from LIMS

ChrOertlin:
# housekeeper
image

ChrOertlin:
# statusdb
image
image
image

cg/cli/validation.py
```diff
@@ -0,0 +1,71 @@
+"""cg module for validation ofcases."""
```
ivadym, line 1:
```suggestion
"""CG module for validation of cases."""
```

cg/meta/create_validation_cases/validation_cases_api.py
```diff
@@ -0,0 +1,185 @@
+"""Module that holds the api to create validation cases."""
```
ivadym, line 1:
```suggestion
"""Module that holds the API to create validation cases."""
```

cg/meta/create_validation_cases/validation_cases_api.py
ivadym: I would say mainly the data that we have mocked in the `MockLimsAPI`

cg/cli/validation.py
```diff
@@ -0,0 +1,71 @@
...
+def create_validation_case(
+    context: CGConfig,
+    case_id: str,
+    case_name: str,
+    delivery: str,
+    data_analysis: str,
+    dry_run: bool,
+):
+    """Create a validation case to be used in workflow validation."""
+    validation_case_api = CreateValidationCaseAPI(
```
seallard, line 57: Consider passing the `CreateValidationCaseAPI` via the context instead of instantiating it here

cg/meta/create_validation_cases/validation_case_data.py
```diff
@@ -0,0 +1,121 @@
...
+from cg.store.store import Store
+from cg.utils.calculations import multiply_by_million
+
+LOG = logging.getLogger(__name__)
+
+
+class ValidationCaseData:
+    def __init__(self, status_db: Store, validation_data_input: ValidationDataInput):
+        """Initialize the validation sample data and perform integrity checks."""
+        self.status_db: Store = status_db
```
seallard, line 20: The class name sounds a bit like it is a pure data object, but it contains database access logic. Maybe there is a better way to encapsulate the logic/data? Not sure

cg/meta/create_validation_cases/validation_cases_api.py
```diff
@@ -0,0 +1,185 @@
...
+        for sample in validation_case_data.validation_samples:
+            if self.status_db.sample_with_id_exists(sample_id=sample.internal_id):
+                raise ValueError(f"Sample {sample.internal_id} already exists in StatusDB.")
+            LOG.info(
+                f"New validation sample created: {sample.internal_id} from {sample.from_sample}"
+                f"Application tag set to: {sample.application_version.application.tag}"
+            )
+            if not self.dry_run:
+                self.status_db.session.add(sample)
+                LOG.info(f"Added {sample.name} to StatusDB.")
```
seallard, line 49: It might make sense to move logic like this into the store where other pure data access logic is. In general, maybe we don't want to touch the session outside of the `Store` layer.

cg/meta/create_validation_cases/validation_cases_api.py
```diff
@@ -0,0 +1,185 @@
...
+from cg.meta.create_validation_cases.validation_case_data import ValidationCaseData
+from cg.meta.create_validation_cases.validation_data_input import ValidationDataInput
+from cg.store.models import Case, CaseSample, Sample
+from cg.store.store import Store
+from cg.utils.files import copy_file, get_files_matching_pattern, rename_file
+
+LOG = logging.getLogger(__name__)
+
+
+class CreateValidationCaseAPI:
```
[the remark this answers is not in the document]
diitaz93: What should be called API and what should be called service? @seallard

cg/meta/create_validation_cases/validation_cases_api.py
ChrOertlin: decided not to store anything in LIMS. Information retrieval from lims will be handled during validation call.

cg/meta/create_validation_cases/validation_cases_api.py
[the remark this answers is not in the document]
ChrOertlin: will be a service

cg/meta/create_validation_cases/validation_case_data.py
ChrOertlin: it contains data only but uses access logic to make sure the stored data is correct. It is similar to the downsampledata class.

ChrOertlin:
closing this pull request for now. There are questions regarding the order table and the stored worfklow thereon. I think this should be addressed first before moving on with this

PR #3150, opened 2024-04-18
fix(start analyses) Move case-limit to final step before iteration

Vince-janv:
## Description
Currently the limit is applied before additional filtering is performed. This results in a lot fewer cases being started than intended.

### Changed

- Moved the limit to take place just before iteration


henrikstranneheim (approved):
👍

PR #3155, opened 2024-04-19
Handle all orders being filtered away

seallard:
## Description
Closes https://github.com/Clinical-Genomics/cigrid-ui/issues/625 and https://github.com/Clinical-Genomics/cigrid-ui/issues/619.

We should return early when all orders are filtered away in the request against `/orders`.

### Fixed
- Handle all orders being filtered away correctly in the `/orders` endpoint.


islean (approved):
Looks good!

PR #3157, opened 2024-04-19
Update Dockerfile

seallard:
## Description
Continues the work started in https://github.com/Clinical-Genomics/cg/pull/2610.

This PR changes the base image and adds some binaries (gpg, pigz, tar) to the image so we do not need to mount them from Hasta.

The remaining steps
- Identify necessary directories from Hasta to mount into the container
- Test on Hasta
- Add deploy scripts to stage and prod
- Update aliases in servers
- Remove cg conda environments

### DIFF
--- a/Dockerfile-cli
+++ b/Dockerfile-cli
@@ -1,16 +1,40 @@
-FROM docker.io/library/python:3.11-slim-bullseye
+FROM registry.access.redhat.com/ubi8/ubi:latest
 
-RUN addgroup --system nonroot \
-    && adduser --system --ingroup nonroot nonroot
+ENV PYTHON_VERSION=3.11
+
+RUN yum -y update && \
+    yum -y install gcc openssl-devel bzip2-devel libffi-devel zlib-devel make wget && \
+    yum -y install gnupg2 pigz tar && \  # Install gpg, pigz, and tar
+    yum clean all
+
+
+RUN mkdir -p /home/bin
+
+# CG expects these binaries to be availbable in /home/bin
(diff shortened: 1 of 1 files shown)


Dockerfile-cli
```diff
@@ -1,16 +1,43 @@
...
+    yum -y install gnupg2 pigz tar && \
+    yum clean all
+
+
+RUN mkdir -p /home/bin
+
+# CG expects these binaries to be availbable in /home/bin
+RUN ln -s /usr/bin/gpg /home/bin/gpg && \
+    ln -s /usr/bin/pigz /home/bin/pigz && \
+    ln -s /usr/bin/tar /home/bin/tar
```
Vince-janv, line 16: I'm quite sure these are not needed on Kasta. But maybe this is intended to run on Hasta as well?

Vince-janv (approved):
Nice! ⭐

Dockerfile-cli
```diff
@@ -1,16 +1,43 @@
...
+
+RUN cd /usr/src && \
+    wget https://www.python.org/ftp/python/3.11.9/Python-3.11.9.tgz && \
+    tar xzf Python-3.11.9.tgz && \
+    cd Python-3.11.9 && \
+    ./configure --enable-optimizations && \
+    make altinstall
```
seallard, line 23: I'm not sure if it is a good idea to build Python like this, it is really slow. But it does not seem like 3.11 is available via yum.

Dockerfile-cli
seallard: Yeah, this is for Hasta

PR #2886, opened 2024-01-31
Add pre-analysis quality check service

karlnyr:
## Description

https://github.com/Clinical-Genomics/cg/issues/2770

In this PR I aim to standardize the way that any pre-analysis quality checks are done, currently, it includes only sequencing qc but could also include others in the future by adding them into the run_qc command.

### Changed

- Sequencing Quality Check of samples before analysis start

### How to prepare for test

- [ ] Ssh to relevant server (depending on type of change)
- [ ] Use stage: `us`
- [ ] Paxa the environment: `paxa`
- [ ] Install on stage (example for Hasta):
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b [THIS-BRANCH-NAME] -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...
- [ ] Take a screenshot and attach or copy/paste the output.

## Review

- [ ] Tests executed by
- [ ] "Merge and deploy" approved by
Thanks for filling in who performed the code review and the test!

### This [version](https://semver.org/) is a

- [ ] **MAJOR** - when you make incompatible API changes
- [x] **MINOR** - when you add functionality in a backwards compatible manner
- [ ] **PATCH** - when you make backwards compatible bug fixes or documentation/instructions

## Implementation Plan

- [ ] Document in ...
- [ ] Deploy this branch on ...
- [ ] Inform to ...

PR #3196, opened 2024-05-02
Update use_read_count_threshold for taxprofiler

sofstam:
### Changed

- According to previous communication, cust053 would like to have data and analysis even though they are just below the targeted threshold of 75%. In case the data are below the targeted threshold, they will assess if the quality of the data is good enough for the purposes. In this PR, the function `use_read_count_threshold` is updated to fit this purpose.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/workflow/taxprofiler.py
+++ b/cg/meta/workflow/taxprofiler.py
@@ -104,3 +104,9 @@ class TaxprofilerAnalysisAPI(NfAnalysisAPI):
     def get_genome_build(self, case_id: str) -> str:
         """Return the reference genome build version of a Taxprofiler analysis."""
         return GenomeVersion.T2T_CHM13.value
+
+    @property
+    def use_read_count_threshold(self) -> bool:
+        """Defines whether the threshold for adequate read count should be passed for all samples
+        when determining if the analysis for a case should be automatically started."""
+        return False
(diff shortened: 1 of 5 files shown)


cg/meta/workflow/taxprofiler.py
```diff
@@ -104,3 +104,9 @@ def get_multiqc_search_patterns(self, case_id: str) -> dict:
     def get_genome_build(self, case_id: str) -> str:
         """Return the reference genome build version of a Taxprofiler analysis."""
         return GenomeVersion.T2T_CHM13.value
+
+    @property
+    def use_read_count_threshold(self) -> bool:
```
ivadym, line 109: I'm not sure this method is used somewhere, seems like it's just defined in the nextflow analysis api but it's never called 🤔

fevac: the parent method was removed there https://github.com/Clinical-Genomics/cg/pull/2886/files
so it has to be adapted on the nf-analysis class or has already be done and can just be removed

sofstam: I can remove those methods and test `start-available` for all pipelines (rare disease, tomte, taxprofiler, rnafusion)

fevac: not the test. Only the methods. I'm not sure if the behaviour is correct right now. Can you check?

sofstam: https://github.com/Clinical-Genomics/cg/blob/b6ad74c3a62d6b829475ca37f85bf39fcfe1d0c9/cg/services/quality_controller/quality_checks/checks.py#L25-L53

It is correct.

PR #3198, opened 2024-05-03
Count cases failing sequencing qc

islean:
## Description

Cases failing the sequencing QC are currently left out from the order status summary. This PR counts the cases and tallys them.

### Fixed

- Cases failing sequencing QC are counted in the order summary.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
Comparison in stage:

image

in production:

image

cg/services/orders/order_summary_service/utils.py
```diff
@@ -46,3 +48,14 @@ def _get_analysis_map(analysis_summaries: list[AnalysisSummary]) -> dict:
...
+
+def _is_case_failed_sequencing_qc(case: Case) -> bool:
+    return case.are_all_samples_sequenced and not QualityControllerService.case_pass_sequencing_qc(
+        case
+    )
+
+
+def get_cases_failed_sequencing_qc_count(order: Order) -> int:
+    cases: list[Case] = order.cases
+    return sum(1 for case in cases if _is_case_failed_sequencing_qc(case))
```
seallard, line 61: Would it be possible to move this logic into a db query instead?

seallard: Alright! If this is something production wants? If so, I think we should merge. But lets make it a priority to clean up the logic as well as improve the performance (maybe track it in an issue as technical debt?). I don't think it is acceptable to add so much latency and complexity in the endpoint.

Given that the calculation is expensive, it might make sense to store it. Basically, once sequencing is done and we do the post processing of the flow cell, I think it might make sense to do the QC there at some point and store the result.

We will probably add more logic to the endpoint over time, and if we double the latency each time, the order views will quickly become unusable. The performance will be even worse in production given that the load is higher.

seallard (approved):
I don't think the hit in performance and endpoint complexity is acceptable, but if this is important to production - we should merge and track this as technical debt.

I think we need to do and store the sequencing QC somewhere else since it is expensive to calculate.

cg/services/orders/order_summary_service/utils.py
```diff
@@ -37,6 +38,7 @@ def create_order_summary(
         cancelled=analysis_summary.cancelled,
         failed=analysis_summary.failed,
         completed=analysis_summary.completed,
+        failed_sequencing_qc=case_summary.failed_sequencing_qc,
```
seallard, line 41: Cigrid needs to be updated as to not always render this status (when the count is None/0)

cg/services/orders/order_summary_service/utils.py
beatrizsavinhas: This is definitely something production needs now, in fact I think it solves this issue https://github.com/Clinical-Genomics/cigrid-ui/issues/631.

Regarding the calculation cost and storing of the status, in this upcoming project [Improve filtering functionality in Trailblazer](https://github.com/Clinical-Genomics/project-planning/issues/575) (continuation of streamline delivery) I proposed including these statuses, that are currently calculated for orders only, on analysis and sample level in Trailblazer. If we implement that and store status in the database, we would not need to calculate everything for orders again and could just gather information from the analyses to display in the order view. This would cut the calculation costs!
But in the meantime, even with the performance hit, I think this fix is useful.

seallard: Yeah lets merge this then 👍  But track the technical debt in an issue so we do not forget it

seallard: It does not resolve https://github.com/Clinical-Genomics/cigrid-ui/issues/631

islean: I think it would resolve this though: https://github.com/Clinical-Genomics/cigrid-ui/issues/624

beatrizsavinhas: Ah yes! Indeed it is https://github.com/Clinical-Genomics/cigrid-ui/issues/624 as @islean pointed out!

islean: I created https://github.com/Clinical-Genomics/cg/issues/3223 to track this.

islean:
Deployed to cg-services-prod

seallard:
I should have blocked this PR 😅 So much work to fix now

PR #3200, opened 2024-05-03
Update deplyoment instructions

seallard:
## Description
The publish action needs to finish before we deploy.

I don't like how deploys are done. Might get a bit better once we have containerized cg.

### Fixed
- Missing instruction


ivadym (approved):
👍

PR #3202, opened 2024-05-03
add( tables for new prod tech)

ChrOertlin:
## Description

Closes: https://github.com/Clinical-Genomics/add-new-tech/issues/1

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/store/models.py
```diff
@@ -971,3 +971,46 @@ class Order(Base):
...
         return to_dict(model_instance=self)
+
+
+class RunDevice(Base):
+    """Model for storing run devices."""
+
+    __tablename__ = "run_device"
+
+    id: Mapped[PrimaryKeyInt]
+    type: DeviceType = mapped_column(types.Enum(*(device.value for device in DeviceType)))
```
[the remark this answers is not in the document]
islean: I think we wanted to avoid .value everywhere if I am not mistaken

cg/constants/devices.py
```diff
@@ -0,0 +1,10 @@
+"""Enums for devices."""
+
+from enum import Enum, auto
+
+
+class DeviceType(Enum):
+    ILLUMINA = auto()
+    SEPHYR = auto()
```
Vince-janv, line 8: This should be BIONANO for consistency

cg/store/models.py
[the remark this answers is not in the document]
Vince-janv: This should do it 🤩 
```suggestion
    type: DeviceType
```

alembic/versions/9b188aee9577_add_new_tech_tables.py
```diff
@@ -0,0 +1,47 @@
...
+depends_on = None
+Base = declarative_base()
+
+
+def upgrade():
+    op.create_table(
+        "run_device",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("type", sa.String(128), nullable=False),
+        sa.Column("device_id", sa.String(128), nullable=False),
```
Vince-janv, line 26:
```suggestion
        sa.Column("device_id", sa.Integer(), nullable=False),
```

cg/store/models.py
```diff
@@ -971,3 +972,46 @@ class Order(Base):
...
+    device_id: Mapped[int] = mapped_column(ForeignKey("device.id"))
+
+    __mapper_args__ = {
+        "polymorphic_identity": "run_device",
```
Vince-janv, line 987: I think we can skip the `"polymorphic_identity": "run_device"` part on all these. If I've understood it correctly it does the following:

if the type column for entry 123 is equal to "run_device" return an object of type `RunDevice`. Since we never plan to instantiate these models (only inherit from them), we can skip this line.

cg/store/models.py
```diff
@@ -971,3 +972,46 @@ class Order(Base):
...
+    }
+
+
+class RunDevice(Base):
+    """Model for storing run devices."""
+
+    __tablename__ = "run_device"
+
+    id: Mapped[PrimaryKeyInt]
+    type: DeviceType = mapped_column(types.Enum(*(device.value for device in DeviceType)))
```
Vince-janv, line 998:
```suggestion
    type: DeviceType
```

cg/store/models.py
```diff
@@ -971,3 +972,46 @@ class Order(Base):
...
+        "polymorphic_on": "type",
+    }
+
+
+class SampleRunMetrics(Base):
+    __tablename__ = "sample_run_metrics"
+    id: Mapped[PrimaryKeyInt]
+    sample_id: Mapped[int] = mapped_column(ForeignKey("sample.id"))
+    run_metrics_id: Mapped[int] = mapped_column(ForeignKey("run_metrics.id"))
+    type: DeviceType = mapped_column(types.Enum(*(device.value for device in DeviceType)))
```
Vince-janv, line 1012:
```suggestion
    type: DeviceType
```

alembic/versions/9b188aee9577_add_new_tech_tables.py
```diff
@@ -0,0 +1,47 @@
...
+branch_labels = None
+depends_on = None
+Base = declarative_base()
+
+
+def upgrade():
+    op.create_table(
+        "run_device",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column("type", sa.String(128), nullable=False),
```
Vince-janv, line 25: I don't think we need 128 characters, and 32 is the one that's most used in other columns
```suggestion
        sa.Column("type", sa.String(32), nullable=False),
```

Vince-janv (approved):
Exciting 😁

cg/store/models.py
```diff
@@ -971,3 +972,42 @@ class Order(Base):
...
+class RunDevice(Base):
+    """Model for storing run devices."""
+
+    __tablename__ = "run_device"
+
+    id: Mapped[PrimaryKeyInt]
+    type: Mapped[DeviceType]
+
+    __mapper_args__ = {
+        "polymorphic_on": "type",
```
diitaz93, line 986: do we need a `polimorphic_identity` field here for the parent classes?

ChrOertlin: Not strictly necessary unless we would like to call these tables directly according to what the docs say. Not sure how it will behave exactly. We can see and test.

diitaz93 (approved):
Looks great 💯

ChrOertlin:
# Tables show up
image

PR #3203, opened 2024-05-06
Add RIN thresholds for the Rnafusion delivery report

ivadym:
## Description

Documentation: https://github.com/Clinical-Genomics/atlas/pull/2693

### Added
- RIN max and min thresholds for delivery report validation

### Fixed
- Zero integers treated as `N/A` when parsing metrics for the delivery report

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [x] `cg generate delivery-report <case-id>`

### Expected test outcome

- [ ] Report renders N/A for RIN values outside range, including 0 
  - No available cases in LIMS. Tested only locally.
- [x] Report render N/A for empty RIN values.
- [x] Report generation Taxprofiler.

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/models/report/validators.py
```diff
@@ -33,14 +33,18 @@ def get_boolean_as_string(value: bool | None) -> str:
...
 
 
-def get_float_as_string(value: float | None) -> str:
-    """Return string representation of a float value."""
-    return str(round(float(value), PRECISION)) if value or isinstance(value, float) else NA_FIELD
+def get_number_as_string(value: int | float | None) -> str:
+    """Return string representation of a number."""
+    return (
+        str(round(float(value), PRECISION))
+        if value or isinstance(value, (int, float))
```
diitaz93, line 40: Im wondering what would happen if the input is bool (True) or str, would the behaviour be unexpected?
```suggestion
        if isinstance(value, (int, float))
```

diitaz93 (commented):
will finish later

cg/models/report/validators.py
ivadym: If bool or an invalid string (not coercible to a float) are passed we would like this to fail actually.

ivadym: just edited the comment as I realised some coverage values are parsed as stirngs for mip :( 

>  I will keep the value check as some fields may be parsed as strings

tests/meta/report/test_rnafusion_api.py
```diff
@@ -33,3 +37,35 @@ def test_get_sample_metadata(
 
     # THEN the sample metadata should be correctly retrieved and match the expected validated metrics
     assert sample_metadata.model_dump() == rnafusion_validated_metrics
+
+
+def test_ensure_rin_thresholds(
```
diitaz93, line 42: I think more tests with different values of RIN are important, maybe you can parametrise this test

tests/models/report/test_validators.py
```diff
@@ -63,27 +63,53 @@ def test_get_boolean_as_string():
...
 
     # WHEN performing the validation
-    validated_float_value: str = get_float_as_string(float_value)
+    validated_float_value: str = get_number_as_string(float_value)
+
+    # THEN check if the input value was formatted correctly
+    assert validated_float_value == "0.0"
+
+
+def test_get_int_number_as_string():
```
diitaz93, line 92: O parametrise this one instead

diitaz93 (approved):
Looks great 👍 Maybe you can add more tests scenarios to the number to str function

ivadym:
Deployed! Thanks for the review @diitaz93!

Issue #3072, opened 2024-03-27
bug: dry-run of cg downsample samples creates directory


fevac:
## Description

- Dry-run of `cg downsample samples -c CASE -cn CASENAME -i SAMPLEID 1 -d` creates /home/proj/stage/downsample/DOWNSAMPLENAME

## Suggested solution

- No files or directories should be created since this is a dry-run. 

## This can be closed when

*Describe what needs to be done for this issue to be closed*

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

PR #3210, opened 2024-05-08
fix(downsample api)

ChrOertlin:
## Description

Fixes a bug that the directories get created when running downsample in dry run mode. @fevac closes: #3072
Fixes using an underscore in case name in the downsample api. @karlnyr 
Provide proper accounts to slurm API depending on environment: "development" for stage and "production" for prod. @sofstam , @henrikstranneheim 

merge in concert with: https://github.com/Clinical-Genomics/servers/pull/1354
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/models/cg_config.py
```diff
@@ -85,6 +85,12 @@ class DemultiplexConfig(BaseModel):
     slurm: SlurmConfig
 
 
+class DownsampleConfig(BaseModel):
+    downsample_dir: str
+    downsample_script: str
```
islean, line 90: Do we need to prepend "downsample" given the class name?

cg/models/downsample/downsample_data.py
```diff
@@ -52,7 +52,7 @@ def downsampled_case_name(
         self,
     ) -> str:
         """Return a case name with _downsampled appended."""
-        return f"{self.case_name}_downsampled"
+        return f"{self.case_name}-downsampled"
```
fevac, line 55: is this needed?

tests/models/downsample/test_down_sample_meta_data.py
```diff
@@ -46,4 +46,4 @@ def test_downsample_meta_data_pass_checks(
     # THEN all necessary models to run the down sample command are created
     assert meta_data.downsampled_sample.internal_id == f"{sample.internal_id}DS{number_of_reads}M"
```
fevac, line 47: the command currently adds the sample with a suffix like DS5C0M, why the C? is that for "comma", mixing english (DS) and swedish (C)?

ChrOertlin: @karlnyr changed it to this. I will not address it in this PR.

cg/models/downsample/downsample_data.py
ChrOertlin: I thikn this is what the discussion was about in slack if I remember correctly? @karlnyr wanted this not to be allowed?

karlnyr: I think @fevac means if the "-downsampled" affix is needed. I think that it is, because it is the only thing, except for the connected sample, that tells us if this is a real case or not.

ChrOertlin:
> Great - would it also be possible to override the default from the config by setting the account from the CLI?

yes I can add an account default override

henrikstranneheim (approved):
👍

cg/models/cg_config.py
ChrOertlin: I just took over how it was previously and will leave it at that. We can rework at another time if it is wished.

ChrOertlin:
# tests completed 
- can add account
- default account developtment on stage
- config works

PR #3221, opened 2024-05-13
Add internal_id column to RunDevice model

diitaz93:
## Description
Closes https://github.com/Clinical-Genomics/add-new-tech/issues/19

### Added

- `intenal_id` column to `RunDevice` model
- New alembic migration with the changes

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/store/models.py
+++ b/cg/store/models.py
@@ -45,6 +45,7 @@ VarChar128 = Annotated[str, 128]
 
 PrimaryKeyInt = Annotated[int, mapped_column(primary_key=True)]
 UniqueStr = Annotated[str, mapped_column(String(32), unique=True)]
+UniqueStr64 = Annotated[str, mapped_column(String(64), unique=True)]
 
 
 class Base(DeclarativeBase):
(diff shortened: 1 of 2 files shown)


cg/store/models.py
```diff
@@ -967,6 +968,7 @@ class RunDevice(Base):
 
     id: Mapped[PrimaryKeyInt]
     type: Mapped[DeviceType]
+    internal_id: Mapped[UniqueStr64]
```
seallard, line 971: Would it make more sense to call this `device_name`?

seallard (approved):
Nice 👍  Maybe you want to consider other names? I feel like `internal_id` might be unnecessarily arcane and CG-like 😅

cg/store/models.py
diitaz93: This is the name that was decided in the refinement. We wanted it to be an id because is unique and we wanted to mimic what it is used in the `Sample` model, so referring to it as an "internal_id" in the database models but as just `_id` in cg

Vince-janv: In the current context, name how the customers refer to the objects (like sample.name and case.name). I think internal_id makes sense in the reality we have. The alternative would be to refactor the other ones to sample.external_id or similar which I think is out of scope here

Vince-janv (approved):
Looks great ⭐

Issue #3228, opened 2024-05-15
Reference genome not saved in database


islean:
## Description

There are examples of tomte orders being submitted with reference genome set but the data is not persisted in statusdb.

seallard:
Resolved in https://github.com/Clinical-Genomics/cg/issues/3243, closing.

PR #3231, opened 2024-05-15
Patch setting reference genome for existing samples

islean:
## Description

Existing samples are not updated with the reference genome even though it is sent in the order. This prevents new Tomte orders with existing samples from being run. Addresses #3228.

### Fixed

- Reference genome is updated for existing samples if set in a new order.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
This is an improvement but ideally maybe the reference genome should be placed in the case_sample table so that it can be set per order? Now any new order will override the old value.

islean:
Tested submitting an order in stage setting the reference genome for a sample with it previously not set. After submission it was set:

image

diitaz93 (approved):
Looks good 👍

diitaz93:
> This is an improvement but ideally maybe the reference genome should be placed in the case_sample table so that it can be set per order? Now any new order will override the old value.

Yeah good point, maybe lets consult with someone more familiar to pipelines before merging this

diitaz93:
@islean Talking to @ivadym we decided that it is better to set this for refinement until we have a better understanding of the reference genome usage.

fevac:
We need a quick fix for this. We could then refine a more permanent solution.

My suggestion:
- Add reference genome information in an old sample information if this information is missing (not overwrite if it was added)
- I had previously talked about this with @karlnyr 

We can then refine the reference_genome information with a bit more time. 

We cannot have this bug in prod for too long as it is a pain in the butt and we are still modifying (just manually) the old sample information

My point is, that the workaround for production is still modifying the old sample. There is no other workaround. Setting a default genome version is not an option, because they already provide that information in the order and is not always 19 or always 38. So that cannot be an option.

We could also add a comment also if you want to keep traceability

fevac:
Decision from meeting 20240520 with @karlnyr @Vince-janv @ivadym @fevac :

- Short term:
  -  @Vince-janv confirm with customers about updating the reference genome for their already existing samples. 
  - @Vince-janv update statusDB for customers involved for all their already existing RNA samples:
    - set reference genome (this field was previously not used for any logic, so it should not entail a problem in the data flow. However it does become a bit misleading at the "information level" since the reference genome might differ from what it was previously ran)
    - append a comment informing about the changes done and the date

- Long term:
  - Move reference genome information to a case level. This requires a bit of refactoring and it's a bigger project. @Vince-janv will create or link an issue to track this. The priority will depend on whether other workflows will require setting these fields or not

diitaz93:
Does this still need refinement? @fevac @Vince-janv ?

seallard:
> Decision from meeting 20240520 with @karlnyr @Vince-janv @ivadym @fevac :
> 
>     * Short term:
>       
>       * @Vince-janv confirm with customers about updating the reference genome for their already existing samples.
>       * @Vince-janv update statusDB for customers involved for all their already existing RNA samples:
>         
>         * set reference genome (this field was previously not used for any logic, so it should not entail a problem in the data flow. However it does become a bit misleading at the "information level" since the reference genome might differ from what it was previously ran)
>         * append a comment informing about the changes done and the date
> 
>     * Long term:
>       
>       * Move reference genome information to a case level. This requires a bit of refactoring and it's a bigger project. @Vince-janv will create or link an issue to track this. The priority will depend on whether other workflows will require setting these fields or not

I don't quite understand the role of a reference genome, but it sounds like it is more of a setting for a specific analysis of the case? As such, would it not make more sense to introduce a separate model for it than to store it on the case?

Or is it not likely that we will re-run cases with different reference genomes?
@Vince-janv, @fevac

henrikstranneheim:
Reanalysis using a different reference genome would yield a completely different result and I think it is more than a rerun and warrants a separate case. Or we have to refine a lot of logic.

islean:
Closing due to the work done in https://github.com/Clinical-Genomics/cg/issues/3243 and the outcome of the refinement meetings.

PR #3239, opened 2024-05-16
add IlluminaFlowCellDevice model

diitaz93:
## Description
Closes https://github.com/Clinical-Genomics/add-new-tech/issues/2

### Added

- IlluminaFlowCellDevice model
- Alembic migration

### How to prepare for test

### How to test

- [x] Do `alembic -c $ALEMBIC_CG_STAGE upgrade fe23de4ed528`
- [x] Verify that the new alembic version is `fe23de4ed528`
- [x] Verify that the new table ìllumina_flow_cell` exists
- [x] Create a new entry in the `run_device` table of type `ILLUMINA`
- [x] Create a new entry in the table `illumina_flow_cell` with the same id as the previous entry and mode `10B`.
- [x] Query the new table filtering by flow cells with mode == "10B"
- [x] Do `alembic -c $ALEMBIC_CG_STAGE downgrade 6e6c36d5157b`

### Expected test outcome

- [x] Check that the filtering returns a flow cell and no errors are raised
Screenshot 2024-05-20 at 09 34 09

## Review

- [x] Tests executed by SD

## Implementation Plan

- [ ] Deployed to stage:
```shell

```
- [ ] Deployed to production:
```shell

```


alembic/versions/0768c90806f5_add_illumina_flow_cell.py
```diff
@@ -0,0 +1,28 @@
+"""add_illumina_flow_cell
+
+Revision ID: 0768c90806f5
+Revises: 5c6de08c4aca
```
diitaz93, line 4: fix this with latest migration

alembic/versions/0768c90806f5_add_illumina_flow_cell.py
```diff
@@ -0,0 +1,28 @@
+"""add_illumina_flow_cell
```
diitaz93, line 1: add date to final name

alembic/versions/0768c90806f5_add_illumina_flow_cell.py
```diff
@@ -0,0 +1,28 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_table(
+        "illumina_flow_cell",
+        sa.Column("id", sa.ForeignKey("run_device.id"), nullable=False, primary_key=True),
+    )
+
```
diitaz93, line 25: add all columns

diitaz93:
Apparently what we called "flow cell model" (10B, 25B) is officially called "type", which is the field we use for polymorphism. I changed it to `flow_cell_type`, any other suggestions? @Vince-janv @ChrOertlin

cg/store/models.py
```diff
@@ -975,6 +976,18 @@ class RunDevice(Base):
     }
 
 
+class IlluminaFlowCellDevice(RunDevice):
```
Vince-janv, line 979: Device seems superfluous
```suggestion
class IlluminaFlowCell(RunDevice):
```

cg/store/models.py
```diff
@@ -975,6 +976,18 @@ class RunDevice(Base):
...
+    """Model for storing Illumina flow cells."""
+
+    __tablename__ = "illumina_flow_cell"
+
+    id: Mapped[int] = mapped_column(ForeignKey("run_device.id"), primary_key=True)
+    flow_cell_type: Mapped[IlluminaFlowCellType | None]
+    flow_cell_name: Mapped[UniqueStr64 | None]
```
Vince-janv, line 986: Is this not in the parent class? (RunDevie.internal_id)

cg/store/models.py
```diff
@@ -975,6 +976,18 @@ class RunDevice(Base):
     }
 
 
+class IlluminaFlowCellDevice(RunDevice):
+    """Model for storing Illumina flow cells."""
+
+    __tablename__ = "illumina_flow_cell"
+
+    id: Mapped[int] = mapped_column(ForeignKey("run_device.id"), primary_key=True)
+    flow_cell_type: Mapped[IlluminaFlowCellType | None]
```
Vince-janv, line 985: I don't think we need to care that they call it type. As long as what we name it makes sense I think it works. In that way I think just `model` is better

cg/constants/demultiplexing.py
```diff
@@ -212,3 +212,14 @@ class IndexSettings(BaseModel):
...
+
+class IlluminaFlowCellType(StrEnum):
+    # "10B"
+    # "25B"
+    # "1.5B"
+    # "S4"
+    # "S1"
+    # "SP"
+    # "S2"
+    pass
```
Vince-janv, line 225: Why StrEnum? 👀

Vince-janv (commented):
I think short and descriptive names for the fields are better. We already have a clear name for the Model, so we don't need to duplicate the information in the field name

cg/store/models.py
diitaz93: fixed

cg/store/models.py
diitaz93: I am referring here to the full name, so `20231108_LH00188_0028_B22F52TLT3` for example. Is this something relevant in this class or in the metrics class?

cg/store/models.py
diitaz93: Another option is follow how it is called in the RunParameters.xlm:
```
<Mode>10B</Mode>
```

Vince-janv: I find mode even more confusing. It sounds like you can choose what to use for the specific flow cell when in fact its an unchangeable property of the flow cell

cg/store/models.py
ChrOertlin: We don't want to store this string. It is a concatenation of:

date_sequencername_???_flow_cell_name

alembic/versions/0768c90806f5_add_illumina_flow_cell.py
diitaz93: updated

cg/store/models.py
Vince-janv: Refinement decision: go with `model`

alembic/versions/0768c90806f5_add_illumina_flow_cell.py
diitaz93: updated

alembic/versions/0768c90806f5_add_illumina_flow_cell.py
diitaz93: updated

cg/constants/demultiplexing.py
diitaz93: decision for refinement: Test it with Enum

cg/store/models.py
```diff
@@ -975,6 +976,24 @@ class RunDevice(Base):
...
+class IlluminaFlowCell(RunDevice):
+    """Model for storing Illumina flow cells."""
+
+    __tablename__ = "illumina_flow_cell"
+
+    id: Mapped[int] = mapped_column(ForeignKey("run_device.id"), primary_key=True)
+    model: Mapped[IlluminaFlowCellModels | None] = mapped_column(
+        types.Enum(
+            IlluminaFlowCellModels,
+            values_callable=lambda: [
```
diitaz93, line 988: Im getting `E   TypeError: IlluminaFlowCell.<lambda>() takes 0 positional arguments but 1 was given` with this alternative, will try the other one

diitaz93: switched to the other approach

ChrOertlin (approved):
Looks good

PR #3246, opened 2024-05-20
Feat(Sequencing models) Add relationships

Vince-janv:
## Description
With the models in place relationships can be defined to related them in python. Also two basic properties where added.

### Added

- Relationships between device -> run_metrics -> sample_run_metrics
- property to access devices from samples
- property to access samples from device

### DIFF
--- a/cg/store/models.py
+++ b/cg/store/models.py
@@ -782,6 +782,8 @@ class Sample(Base, PriorityMixin):
     )
     invoice: Mapped["Invoice | None"] = orm.relationship(back_populates="samples")
 
+    _new_run_metrics: Mapped[list["SampleRunMetrics"]] = orm.relationship(back_populates="sample")
+
     def __str__(self) -> str:
         return f"{self.internal_id} ({self.name})"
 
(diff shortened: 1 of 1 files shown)


cg/store/models.py
```diff
@@ -1030,6 +1055,9 @@ class SampleRunMetrics(Base):
     run_metrics_id: Mapped[int] = mapped_column(ForeignKey("run_metrics.id"))
     type: Mapped[DeviceType]
 
+    run_metrics: Mapped[RunMetrics] = orm.relationship(back_populates="sample_metrics")
+    sample: Mapped[Sample] = orm.relationship(back_populates="_new_run_metrics")
```
ChrOertlin, line 1059: we should add a cascade on delete here, if a sample gets deleted, delete the sequencing info.

cg/store/models.py
```diff
@@ -1030,6 +1055,9 @@ class SampleRunMetrics(Base):
     run_metrics_id: Mapped[int] = mapped_column(ForeignKey("run_metrics.id"))
     type: Mapped[DeviceType]
 
+    run_metrics: Mapped[RunMetrics] = orm.relationship(back_populates="sample_metrics")
```
ChrOertlin, line 1058: we should add a cascade on delete. If a run_metrics entry gets deleted, this should also remove associated sample_run_metrics

cg/store/models.py
```diff
@@ -987,6 +1007,11 @@ class RunMetrics(Base):
     type: Mapped[DeviceType]
     device_id: Mapped[int] = mapped_column(ForeignKey("run_device.id"))
 
+    device: Mapped[RunDevice] = orm.relationship(back_populates="run_metrics")
```
ChrOertlin, line 1010: cascade on delete: if a run_device gets deleted we should cascade this.

ChrOertlin (approved):
Looks good, we should add on delete behaviour so that we only need to remove a run_device if we want to delete associated metrics.

PR #3267, opened 2024-05-23
feat(Raredisease) add delivery report

peterpru:
### Added

- Delivery report for raredisease.

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/report/raredisease.py
+++ b/cg/meta/report/raredisease.py
@@ -1,0 +1,127 @@
+from cg.meta.report.field_validators import get_million_read_pairs
+from cg.meta.report.report_api import ReportAPI
+from cg.models.analysis import NextflowAnalysis
+from cg.models.cg_config import CGConfig
+from cg.meta.workflow.raredisease import RarediseaseAnalysisAPI
+from cg.models.mip.mip_metrics_deliverables import get_sample_id_metric
+from cg.models.raredisease.raredisease import RarediseaseQCMetrics
+from cg.models.report.metadata import RarediseaseSampleMetadataModel
+from cg.store.models import Case, Sample
+
+
+class RarediseaseReportAPI(ReportAPI):
+    """API to create Rare disease DNA delivery reports."""
+
+    def __init__(self, config: CGConfig, analysis_api: RarediseaseAnalysisAPI):
+        super().__init__(config=config, analysis_api=analysis_api)
+
+    def get_sample_metadata(
+        self, case: Case, sample: Sample, analysis_metadata: NextflowAnalysis
+    ) -> RarediseaseSampleMetadataModel:
+        """Return sample metadata to include in the report."""
+        sample_metrics: RarediseaseQCMetrics = analysis_metadata.sample_metrics[sample.internal_id]
+
+        sample_coverage: dict = self.get_sample_coverage(sample=sample, case=case)
+        return RarediseaseSampleMetadataModel(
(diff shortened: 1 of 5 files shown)


cg/apps/coverage/api.py
```diff
@@ -87,3 +88,57 @@ def sample_coverage(self, sample_id: str, panel_genes: list) -> dict:
...
+    def __init__(self, base_url, config):
+        self.base_url = base_url
+        self.chanjo2_config = config["chanjo2"]["config_path"]
+        self.chanjo2_binary = config["chanjo2"]["binary_path"]
+        self.process = Process(binary=self.chanjo2_binary, config=self.chanjo2_config)
+
+    def get_coverage_statistics(
+        self, coverage_file_path, intervals_bed_path, completeness_thresholds
+    ):
+        endpoint = f"{self.base_url}/coverage/d4/interval_file/"
```
northwestwitch, line 105: This looks OK I think. Should be called for each sample of the case and it returns a response like this:

image

What is the exact data that you need to include in the delivery report?

peterpru: thanks! It should be median coverage over the omim panel, and the completeness at 10X (medelsekvensdjup & täckningsgrad in the delivery report).
image

cg/meta/report/raredisease.py
```diff
@@ -0,0 +1,131 @@
...
+            mapped_reads=sample_metrics.percentage_mapped_reads,
+            mean_target_coverage=sample_coverage.get("mean_coverage"),
+            pct_10x=sample_coverage.get("mean_completeness"),
+            duplicates=sample_metrics.fraction_duplicates,
+        )
+
+    ##TODO #here it looks for chanjo API, need to add chanjo2 instead
+    def get_sample_coverage(self, sample: Sample, case: Case) -> dict:
+        """Return coverage values for a specific sample."""
+        genes = self.get_genes_from_scout(panels=case.panels)
```
northwestwitch, line 56: The api endpoint in chanjo2 that you are calling is expecting a bed file with the gene coordinates, not gene IDs. So perhaps export the gene panel from scout as a bed file? 

In scout you would do it with this command: `scout export panel --build 37 --bed OMIM-AUTO` but I don't know if there is some shortcut in cg to do the same. Then I would save this bed file as a temporary file and send it as a request to the chanjo2 api (intervals_bed_path param)

northwestwitch (commented):
It's hard to foresee if this would work without test. but looks like you are calling a chanjo2 endpoint that expects a bed file with a list of genes?

I don't see that your code is parsing what you get from chanjo2 (A very long list of dictionaries with stats for each of line of the bed file you provided), so you have to include the code that does it. Alternatively I can also create a new endpoint in chanjo2 that returns exactly the stats you need, if you describe exactly what format they should have

northwestwitch:
If it's easier for you to provide gene IDs I can also write an endpoint in chanjo2 that accepts those and returns the minimal info that you need!

ivadym:
Switching to a more client-oriented approach:

- https://github.com/Clinical-Genomics/cg/pull/3450
- https://github.com/Clinical-Genomics/cg/pull/3459

PR #3270, opened 2024-05-23
Add DV200 and Reception QC metrics to delivery report

ivadym:
## Description

https://github.com/Clinical-Genomics/MTP-RNAFUSION/issues/80

### Added:
- DV200 metric to Tomte and Rnafusion delivery reports
- Sample reception control metric to lab prep step (to all delivery report supported workflow)

### Changed:
- RIN value set as not mandatory
- RIN metric description

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] `cg generate delivery-report <case-id>`

### Expected test outcome

- [ ] DV200 reported for Tomte and Rnafusion
- [ ] RIN reported for Tomte and Rnafusion
- [ ] QC flag reported for all delivery report supported workflows
- [ ] Sex report is still correct
- [ ] Analysis type report is still correct (Balsamic)

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/apps/lims/api.py
```diff
@@ -420,6 +426,26 @@ def get_sample_rin(self, sample_id: str) -> float | None:
...
+            dv200: float = sample_artifact.udf.get(PROP2UDF["dv200"])
+        except HTTPError as error:
+            LOG.warning(f"Sample {sample_id} not found in LIMS: {error}")
+        return dv200
+
+    def get_sample_reception_qc_flag(self, sample_id: str) -> ReceptionQCFLag | None:
+        """Return the sample reception QC flag."""
+        qc_flag = None
+        try:
+            sample_artifact: Artifact = Artifact(self, id=f"{sample_id}PA1")
```
fevac, line 443: is this the same for both DNA and RNA? I can see that in the lims the "step" they belong to have different names

cg/constants/report.py
```diff
@@ -44,17 +46,22 @@
 RIN_MAX_THRESHOLD: int = 10
 RIN_MIN_THRESHOLD: int = 1
 
-REPORT_GENDER: dict[str, str] = {
-    "unknown": "Okänd",
-    "female": "Kvinna",
-    "male": "Man",
+REPORT_SEX: dict[str, str] = {
```
fevac, line 49: 🌟

cg/constants/report.py
```diff
@@ -207,14 +215,14 @@
...
     "bias_5_3",
+    "dv200",
     "input_amount",
     "mrna_bases",
     "pct_adapter",
     "pct_surviving",
     "q20_rate",
     "q30_rate",
     "ribosomal_bases",
-    "rin",
```
fevac, line 217: we should keep the rin value. Both rin and dv200 should be shown

cg/meta/report/templates/macros/data_analysis/qc_metrics/rnafusion_qc_metrics.html
```diff
@@ -3,7 +3,8 @@
 {% macro rnafusion_qc_metrics(samples) %}
   {% set
     metrics = [
-      {"name": "RNA Integrity Number (RIN)",                   "key": "rin",                   "description": "Mått på skala från 1 till 10 för kvalitet och integritet hos RNA-prover."},
+      {"name": "RNA Integrity Number (RIN)",                   "key": "rin",                   "description": "Mått på skala från 1 till 10 för kvalitet och integritet hos RNA-prover. N/A som ett värde är en varning att RIN kanske inte är tillförlitlig för det specifika provet."},
+      {"name": "DV200",                                        "key": "dv200",                 "description": "Andelen RNA-fragment längre än 200 nukleotider. Denna metrik fungerar som en kvalitetsindex och kategoriserar prover som hög kvalitet (> 70%), medelkvalitet (50–70%), låg kvalitet (30–50%) eller för degraderade (< 30%)."},
```
fevac, line 7: Nice! out of curiosity, where did you get this from?

cg/meta/report/templates/macros/data_analysis/qc_metrics/rnafusion_qc_metrics.html
```diff
@@ -3,7 +3,8 @@
 {% macro rnafusion_qc_metrics(samples) %}
   {% set
     metrics = [
-      {"name": "RNA Integrity Number (RIN)",                   "key": "rin",                   "description": "Mått på skala från 1 till 10 för kvalitet och integritet hos RNA-prover."},
+      {"name": "RNA Integrity Number (RIN)",                   "key": "rin",                   "description": "Mått på skala från 1 till 10 för kvalitet och integritet hos RNA-prover. N/A som ett värde är en varning att RIN kanske inte är tillförlitlig för det specifika provet."},
```
fevac, line 6: Not reliable or not available. What about something like this (Swedish to be corrected)?
```suggestion
      {"name": "RNA Integrity Number (RIN)",                   "key": "rin",                   "description": "Mått på skala från 1 till 10 för kvalitet och integritet hos RNA-prover. N/A som värde är en varning om att inget tillförlitligt RIN är tillgängligt för det specifika det provet."},
```

tests/models/report/test_validators.py
```diff
@@ -173,20 +173,20 @@ def test_get_path_as_string(filled_file: Path):
...
-    """Test report gender parsing."""
+def test_get_sex_as_string():
+    """Test report sex parsing."""
 
-    # GIVEN an invalid gender category
-    gender: Sex = Sex.FEMALE
-    invalid_gender: str = "not_a_gender"
+    # GIVEN an invalid sex category
+    sex = Sex.FEMALE
+    invalid_sex = "not_a_sex"
```
fevac, line 181: type hint

tests/mocks/limsmock.py
```diff
@@ -132,3 +133,11 @@ def get_sample_rin(self, sample_id: str) -> float:
     def get_latest_rna_input_amount(self, sample_id: str) -> float:
         """Mock return input amount used in the latest preparation of an RNA sample."""
         return 300.0
+
+    def get_sample_dv200(self, sample_id: str) -> float | None:
```
fevac, line 137:
```suggestion
    def get_sample_dv200(self, sample_id: str) -> float:
```

tests/mocks/limsmock.py
```diff
@@ -132,3 +133,11 @@ def get_sample_rin(self, sample_id: str) -> float:
...
+        """Mock return sample's percentage of RNA fragments greater than 200 nucleotides."""
+        return 75.0
+
+    def get_sample_reception_qc_flag(self, sample_id: str) -> ReceptionQCFLag | None:
```
fevac, line 141:
```suggestion
    def get_sample_reception_qc_flag(self, sample_id: str) -> ReceptionQCFLag:
```

fevac (approved):
Nice! 🌟

cg/constants/report.py
ivadym: Yes, they will. These are just the mandatory fields. If DIV200 is missing it will fail the report, but we're jusr allowing RIN N/A values here by removing it from that list

cg/meta/report/templates/macros/data_analysis/qc_metrics/rnafusion_qc_metrics.html
ivadym: Was about to write you a slack regarding this to confirm the values.

But it's from [illumina's guidelines](https://www.illumina.com/content/dam/illumina-marketing/documents/products/technotes/evaluating-rna-quality-from-ffpe-samples-technical-note-470-2014-001.pdf). The only thing is that the study is only about FFPE samples.

cg/meta/report/templates/macros/data_analysis/qc_metrics/rnafusion_qc_metrics.html
ivadym: Will also double check with KN

cg/apps/lims/api.py
ivadym: At least for MIP WGS samples this value is reported in the Artifact object, but I will make sure it's the same for Balsamic as well

cg/constants/report.py
fevac: oh, I don't think DV200 should be mandatory. Sorry if I wasn't clear. I meant we should show those values if we have them or NA if not. But they should not be mandatory

cg/meta/report/templates/macros/data_analysis/qc_metrics/rnafusion_qc_metrics.html
fevac: Let's double check with the lab if they interpret these values in the same way. Otherwise we can skip the second line.

tests/models/report/test_validators.py
fevac: aha! good to know! Thanks

cg/constants/report.py
ivadym: I've removed it from mandatory fields 👍

cg/apps/lims/api.py
```diff
@@ -420,6 +426,26 @@ def get_sample_rin(self, sample_id: str) -> float | None:
             LOG.warning(f"Sample {sample_id} not found in LIMS: {error}")
         return rin
 
+    def get_sample_dv200(self, sample_id: str) -> float | None:
+        """Return the sample's percentage of RNA fragments greater than 200 nucleotides."""
+        dv200 = None
```
ChrOertlin, line 431: typehint

cg/apps/lims/api.py
```diff
@@ -420,6 +426,26 @@ def get_sample_rin(self, sample_id: str) -> float | None:
...
+        try:
+            sample_artifact: Artifact = Artifact(self, id=f"{sample_id}PA1")
+            dv200: float = sample_artifact.udf.get(PROP2UDF["dv200"])
+        except HTTPError as error:
+            LOG.warning(f"Sample {sample_id} not found in LIMS: {error}")
+        return dv200
+
+    def get_sample_reception_qc_flag(self, sample_id: str) -> ReceptionQCFLag | None:
+        """Return the sample reception QC flag."""
+        qc_flag = None
```
ChrOertlin, line 441:
```suggestion
        qc_flag: ReceptionQCFLag | None = None
```

cg/apps/lims/api.py
```diff
@@ -420,6 +426,26 @@ def get_sample_rin(self, sample_id: str) -> float | None:
...
+        """Return the sample's percentage of RNA fragments greater than 200 nucleotides."""
+        dv200 = None
+        try:
+            sample_artifact: Artifact = Artifact(self, id=f"{sample_id}PA1")
+            dv200: float = sample_artifact.udf.get(PROP2UDF["dv200"])
+        except HTTPError as error:
+            LOG.warning(f"Sample {sample_id} not found in LIMS: {error}")
+        return dv200
+
+    def get_sample_reception_qc_flag(self, sample_id: str) -> ReceptionQCFLag | None:
```
ChrOertlin, line 439: Is this behaviour wanted - i.e. returning None when the sample cannot be found in lims?

cg/meta/report/balsamic.py
```diff
@@ -80,34 +88,38 @@ def get_panel_metadata(
...
         """Return report metadata for Balsamic WGS analysis."""
         return BalsamicWGSSampleMetadataModel(
-            million_read_pairs=million_read_pairs,
+            duplicates=sample_metrics.percent_duplication if sample_metrics else None,
+            fold_80=sample_metrics.fold_80_base_penalty if sample_metrics else None,
+            mean_insert_size=sample_metrics.mean_insert_size if sample_metrics else None,
             median_coverage=sample_metrics.median_coverage if sample_metrics else None,
+            million_read_pairs=million_read_pairs,
             pct_15x=sample_metrics.pct_15x if sample_metrics else None,
             pct_60x=sample_metrics.pct_60x if sample_metrics else None,
```
ChrOertlin, line 118: Just for my understanding, these field are not mandatory at all?

cg/meta/report/balsamic.py
```diff
@@ -80,34 +88,38 @@ def get_panel_metadata(
...
-            duplicates=sample_metrics.percent_duplication if sample_metrics else None,
-            mean_insert_size=sample_metrics.mean_insert_size if sample_metrics else None,
-            fold_80=sample_metrics.fold_80_base_penalty if sample_metrics else None,
             pct_reads_improper_pairs=(
                 sample_metrics.pct_pf_reads_improper_pairs if sample_metrics else None
             ),
+            reception_qc_flag=reception_qc_flag,
```
ChrOertlin, line 122: This can be none as well following the function above, perhaps it would be nice to have this explicitly stated:
```suggestion
            reception_qc_flag=reception_qc_flag if reception_qc_flag else None,
```

cg/meta/report/templates/macros/data_analysis/qc_metrics/rnafusion_qc_metrics.html
ChrOertlin: the DV200, do we obtain this from the sequencing? Just curious

cg/apps/lims/api.py
ivadym: Yes, it is

cg/meta/report/balsamic.py
ivadym: They are not mandatory in terms of  pydantic models because we still want to render the report even if some fields are missing. However, the report is incomplete if any key fields specified in `REQUIRED_REPORT_FIELDS` are missing, which acts as a check for prod-bioinfo to catch any issues with the case

cg/meta/report/balsamic.py
ivadym: I think it's redundant for the `reception_qc_flag`. The check is necessary for the other metrics since we are accessing some attributes from the sample metrics, which could be empty

ivadym:
> the DV200, do we obtain this from the sequencing? Just curious

@ChrOertlin no, it's calculated during the RNA reception QC step with the Tapestation

cg/apps/lims/api.py
ivadym, line 1: Could you have a look at the LIMS part @Karl-Svard? 🙏 

- `DV200` will be used by `Tomte` and `Rnafusion`
- The reception QC flag will be included in all the reports (MIP, Balsamic, Tomte, Rnafusion, and Taxprofiler). Can we extract it from the same artifact, or is this value computed at different steps for different workflows?

cg/meta/report/templates/macros/data_analysis/qc_metrics/rnafusion_qc_metrics.html
ivadym: no, it's calculated during the RNA reception QC step with the Tapestation

cg/meta/report/templates/macros/data_analysis/qc_metrics/rnafusion_qc_metrics.html
```diff
@@ -3,7 +3,8 @@
 {% macro rnafusion_qc_metrics(samples) %}
   {% set
     metrics = [
-      {"name": "RNA Integrity Number (RIN)",                   "key": "rin",                   "description": "Mått på skala från 1 till 10 för kvalitet och integritet hos RNA-prover."},
+      {"name": "RNA Integrity Number (RIN)",                   "key": "rin",                   "description": "Mått på skala från 1 till 10 för kvalitet och integritet hos RNA-prover. N/A som värde är en varning om att inget tillförlitligt RIN är tillgängligt för det specifika det provet."},
+      {"name": "DV200",                                        "key": "dv200",                 "description": "Andelen RNA-fragment längre än 200 nukleotider. Denna metrik fungerar som en kvalitetsindex och kategoriserar prover som hög kvalitet (> 70%), medelkvalitet (50–70%), låg kvalitet (30–50%) eller för degraderade (< 30%)."},
```
ivadym, line 7: Could you have a look at the Swedish translation here @karlnyr? 🙏 🇸🇪

cg/constants/report.py
```diff
@@ -44,17 +46,23 @@
...
+    CancerAnalysisType.TUMOR_NORMAL_PANEL: "Tumör/normal (panelsekvensering)",
+    CancerAnalysisType.TUMOR_NORMAL_WGS: "Tumör/normal (helgenomsekvensering)",
+    CancerAnalysisType.TUMOR_PANEL: "Tumör-endast (panelsekvensering)",
+    CancerAnalysisType.TUMOR_WGS: "Tumör-endast (helgenomsekvensering)",
+}
+
+REPORT_QC_FLAG: dict[ReceptionQCFLag, str] = {
+    ReceptionQCFLag.FAILED: "Underkänd",
+    ReceptionQCFLag.PASSED: "Godkänd",
+    ReceptionQCFLag.UNKNOWN: "Okänd",
```
ivadym, line 65: Here as well @karlnyr 🙏

cg/apps/lims/api.py
Karl-Svard: I'll give it a look on Monday if that's fine :) 

But the sample artifact is used for the RC of all those preps (WGS, TGA, RNA) so the QC flag will behave the same.

ivadym: Nice, then I think we're good but let double check on Monday. Thanks!!

cg/meta/report/templates/macros/data_analysis/qc_metrics/rnafusion_qc_metrics.html
ivadym: > Let's double check with the lab if they interpret these values in the same way. Otherwise we can skip the second line.

[For now we are only recording the DV200-values and they don't need to be evaluated.](https://atlas.scilifelab.se/production/lab/supporting_procedures/rc_rna/#322-tapestation).

Should we keep all of those thresholds or just keep the bottom one (<30%) in the description? @fevac

fevac: I would say we only keep: `"description": "Andelen RNA-fragment längre än 200 nukleotider."`

cg/apps/lims/api.py
```diff
@@ -412,14 +418,34 @@ def get_sample_project(self, sample_id: str) -> str | None:
...
+        """Return the sample's percentage of RNA fragments greater than 200 nucleotides."""
+        dv200: float | None = None
+        try:
+            sample_artifact: Artifact = Artifact(self, id=f"{sample_id}PA1")
+            dv200: float = sample_artifact.udf.get(PROP2UDF["dv200"])
+        except HTTPError as error:
+            LOG.warning(f"Sample {sample_id} not found in LIMS: {error}")
+        return dv200
+
+    def get_sample_reception_qc_flag(self, sample_id: str) -> ReceptionQCFLag | None:
```
Karl-Svard, line 439: This function will work for almost all preps, but we have some (like TGA for example) where the QC flag is set on an artifact further down in the dataflow. We have created a specific EPP in LIMS because of this. It saves the Initial QC outcome to the sample level on the UDF "Passed Initial QC" that has the potential string values "True" or "False". The function could therefore look something like this:

```python
def has_sample_passed_initial_qc(self, sample_id: str) -> bool | None:
    """Return boolean representing the outcome of the Initial QC protocol of the given sample."""
    boolean_outcome: bool | None = None
    try:
        sample: Sample = Sample(self, id=sample_id)
        boolean_outcome: bool = bool(sample.udf["Passed Initial QC"])
    except HTTPError as error:
        LOG.warning(f"Sample {sample_id} not found in LIMS: {error}")
    return boolean_outcome
``` 

Another option is to use `sample.udf.get(<udf name>)`, which will return `None` if the UDF value isn't present on the sample.

ivadym: Nice, thanks for the review and all the info. I've followed this suggestion 💯

ivadym:
deployed!

PR #3273, opened 2024-05-24
Add Illumina sample sequencing metrics admin view table

diitaz93:
## Description

Add table **Illumina Sample Sequencing Metrics** to StatusDB with the same columns as the **Sample Lane Sequencing Metrics** table and the same operations (search and sort).

### Added

- `IlluminaSampleSequencingMetricsView` model

### How to prepare for test

- [x] Ssh to relevant server (depending on type of change) or deploy locally with script

### How to test

- [x] Do deploy in the VM or locally

### Expected test outcome

- [x] Check that the table is rendered correctly, the search and sort operations work
## Review

- [x] Tests executed by SD

## Implementation Plan

To be merged into a dev branch, will be deployed later


diitaz93:
## Tests on local web API
Searchable and sortable works, links for flow cells and sample work
Screenshot 2024-05-24 at 12 19 46

cg/server/admin.py
```diff
@@ -640,6 +708,25 @@ class UserView(BaseView):
...
 
+class IlluminaSampleSequencingMetricsView(BaseView):
+    column_list = [
+        "flow_cell",
+        "sample",
+        "flow_cell_lane",
+        "total_reads_in_lane",
+        "base_passing_q30_percent",
+        "base_mean_quality_score",
+        "_yield",
```
ChrOertlin, line 719:
```suggestion
        "yield",
```
Does this pick names from the DB table directly? In that case the name is just `yield`

cg/server/admin.py
```diff
@@ -441,6 +447,68 @@ class AnalysisView(BaseView):
...
+    column_sortable_list = [
+        "sequenced_at",
+        "sequencer_type",
+        "sequencer_name",
+        "data_availability",
+        "has_backup",
+        "sequencing_started_at",
+        "sequencing_completed_at",
+        "demultiplexing_started_at",
+        "demultiplexing_completed_at",
```
ChrOertlin, line 492: Once I know if we can get accurate dates here we can keep this..

Also, if we have these 4 time stamps. What purpose does the `sequenced_at` serve?

diitaz93: fixed in source branch

cg/server/admin.py
diitaz93: I am not sure, but if it doesn't I have to change this to `yield_` anyway. I will test this when the source branch is merged

ChrOertlin (approved):
With the new changes, looks good!

cg/server/admin.py
diitaz93: fixed to `yield_`

Issue #3223, opened 2024-05-14
Calculation of case sequencing QC slows down /orders


islean:
## Description

The calculations done by the `QualityControllerService` are expensive, meaning that usage of the methods in the web services, e.g. CiGRID, adds quite a lot of latency. As an example found in #3198, calculating the number of cases failing sequencing QC within orders roughly doubles the response time when rendering the orders page. Since this data is useful, we should think of a better way to calculate this.

## Suggested solution

Easiest would probably be to save this data on case and sample level in status db. Would require refinement on when these values should be set. At the end of demultiplexing?

## This can be closed when

Finding cases which fail sequencing qc can be done quickly.

seallard:
I think the calculation is fine as longs as we store the result

seallard:
Given that the sequencing QC is calculated per case (and pass on individual samples but not on the case level) and the analyses are started per case (given that the sequencing QC passes), we should store the overall case sequencing QC on the case. We can also store the sample QC on the samples.

seallard:
### Implementation plan

- Add `sequencing_qc_status` to case model with enum `pending`, `passed`, `failed` (default is `pending`).
- Add new CLI command which picks up cases for which the qc needs to be done and updates the `sequencing_qc_status`.
- Refactor old logic to decouple the sequencing qc from the other logic and just filter based on the stored `sequencing_qc_status`.

seallard:
Leaving this for a future project

seallard:
This is very straightforward to implement, we just need to have a meeting to get everyone on the same page and as to ensure the required PR:s do not get blocked.

seallard:
Rename field on case to `ready_for_analysis`.

seallard:
Pick up the same as `start-available`.
Wait with changing the `start-available` logic to use the cached database value.

seallard:
Replacing the calculations with db queries brings down the response time for 500 ms. I think the overall implementation of the order status summary logic needs to be revisited and a better design for tracking the status of our samples and statuses is merited. See https://github.com/Clinical-Genomics/cg/issues/3452.

PR #3279, opened 2024-05-27
Add sequencing QC field to case model

seallard:
## Description
Part of resolving https://github.com/Clinical-Genomics/cg/issues/3223. We need a field on the case level to store the case sequencing QC result.

### Added
- Sequencing qc field on the case table


cg/constants/constants.py
```diff
@@ -306,3 +306,9 @@ class MultiQC(StrEnum):
 
 
 NG_UL_SUFFIX: str = " ng/uL"
+
+
+class SequencingQCStatus(Enum):
+    FAILED = "failed"
```
diitaz93, line 312: shouldn't this be `auto()`?

alembic/versions/2024_05_27_769f8c6595a4_add_sequencing_qc.py
```diff
@@ -0,0 +1,32 @@
...
+
+def upgrade():
+    op.add_column(
+        "case",
+        sa.Column(
+            "sequencing_qc_status",
+            sa.Enum("FAILED", "PASSED", "PENDING", name="sequencingqcstatus"),
+            nullable=True,
+        ),
+    )
```
islean, line 28: I think we should specify "PENDING" as the default here as well, so that the MYSQL table also has that set.

cg/constants/constants.py
```diff
@@ -306,3 +306,9 @@ class MultiQC(StrEnum):
...
+    PASSED = "passed"
+    PENDING = "pending"
```
islean, line 314: Consider using `auto()` for the values.

diitaz93 (approved):
👍

cg/constants/constants.py
ChrOertlin, line 314:
```suggestion
    FAILED = auto()
    PASSED = auto()
    PENDING = auto()
```

the enum and sql alchemy will handle the rest in this case. We can use the enum fields in the code then for the rest. right?

alembic/versions/2024_05_27_769f8c6595a4_add_sequencing_qc.py
```diff
@@ -0,0 +1,32 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.add_column(
+        "case",
+        sa.Column(
+            "sequencing_qc_status",
+            sa.Enum("FAILED", "PASSED", "PENDING", name="sequencingqcstatus"),
```
islean, line 25: Should this be lower case or upper? The Python models seem to indicate lower case in the values but we might have agreed to use upper case for new enums?

cg/store/models.py
```diff
@@ -446,6 +447,10 @@ class Case(Base, PriorityMixin):
     priority: Mapped[Priority] = mapped_column(
         default=Priority.standard,
     )
+
+    sequencing_qc_status: Mapped[str | None] = mapped_column(
```
ChrOertlin, line 451:
```suggestion
    sequencing_qc_status: Mapped[SequencingQCStatus | None] = mapped_column(
```
I think this should be the mapped enum

cg/constants/constants.py
seallard: Right! Thanks 👍

islean (approved):
Looks good! Mainly concerned if the cases match between the MYSQL table and the python models.

cg/constants/constants.py
seallard: Resolved in https://github.com/Clinical-Genomics/cg/pull/3279/commits/437ad152d55cb26dcf1a0cd5adb940674800c925

cg/store/models.py
seallard: Resolved in https://github.com/Clinical-Genomics/cg/pull/3279/commits/4c83668d722edcb0335eec2a704e331e3b13e860.

cg/constants/constants.py
seallard: Resolved in https://github.com/Clinical-Genomics/cg/pull/3279/commits/437ad152d55cb26dcf1a0cd5adb940674800c925

alembic/versions/2024_05_27_769f8c6595a4_add_sequencing_qc.py
seallard: Resolved in https://github.com/Clinical-Genomics/cg/pull/3279/commits/d90df282770028edb0eeac43b4731ef5245bf221

alembic/versions/2024_05_27_769f8c6595a4_add_sequencing_qc.py
ChrOertlin: UPPER was aggreed upon and should be used for enums regardless

seallard:
Deployed to stage and prod.

PR #3281, opened 2024-05-27
Fix serialization of enum

seallard:
We need to introduce a separate layer for our serialization logic. Tracked technical debt in https://github.com/Clinical-Genomics/cg/issues/3283

This patch ensures the serialization works in the `/cases` endpoint.


cg/server/ext.py
```diff
@@ -1,4 +1,5 @@
 from decimal import Decimal
+from enum import Enum
```
Vince-janv:
```suggestion
```

Vince-janv (approved):
👍

PR #3287, opened 2024-05-27
Add sequencing QC CLI command

seallard:
## Description
Add sequencing QC CLI command which stores the result on the case.

Part of resolving https://github.com/Clinical-Genomics/cg/issues/3223.


seallard:
Tested in stage
```bash
[sebastian.allard@hasta:~] [S_base] $ cg sequencing-qc
Sequencing QC status for case asasdas: SequencingQCStatus.FAILED
No samples in case have reads.
Sequencing QC status for case aasasd: SequencingQCStatus.FAILED
No samples in case have reads.
Sequencing QC status for case asn: SequencingQCStatus.FAILED
Sequencing QC status for case tas: SequencingQCStatus.PASSED
Sequencing QC status for case u: SequencingQCStatus.PASSED
Sequencing QC status for case xi: SequencingQCStatus.PASSED
```

cg/cli/base.py
```diff
@@ -117,3 +118,4 @@ def init(context: CGConfig, reset: bool, force: bool):
 base.add_command(generate_cmd)
 base.add_command(downsample)
 base.add_command(validate)
+base.add_command(sequencing_qc)
```
ChrOertlin, line 121: Do you think this would fit under the "validate" command?

`cg validate sequencing-qc`?

cg/store/crud/read.py
```diff
@@ -1481,3 +1481,10 @@ def get_illumina_flow_cell_by_internal_id(self, internal_id: str) -> IlluminaFlo
             flow_cells=self._get_query(table=IlluminaFlowCell),
             internal_id=internal_id,
         ).first()
+
+    def get_cases_for_sequencing_qc(self) -> list[Case]:
+        """Return all cases that are ready for sequencing QC."""
+        return apply_case_filter(
+            cases=self._get_query(table=Case),
+            filter_functions=[CaseFilter.PENDING_OR_FAILED_SEQUENCING_QC, CaseFilter.HAS_SAMPLES],
+        ).all()
```
islean, line 1490: Should there not be some condition of it having been run? Or does the "HAS_SAMPLES" do that?

islean: Because currently everything that has not finished running would be set to FAILED right?

cg/services/sequencing_qc_service/sequencing_qc_service.py
```diff
@@ -6,10 +11,25 @@
...
+    def __init__(self, store: Store):
+        self.store = store
+
+    def run_sequencing_qc(self) -> None:
+        """Run sequencing QC for all pending or failed cases."""
+        cases: list[Case] = self.store.get_cases_for_sequencing_qc()
+        for case in cases:
+            passes_qc: bool = self.case_pass_sequencing_qc(case)
+            qc_status: SequencingQCStatus = qc_bool_to_status(passes_qc)
+            self.store.update_sequencing_qc_status(case=case, status=qc_status)
```
ChrOertlin, line 30: We wanted to put the sequencingqc status on the sample, not the case if I remember correctly.

seallard: This is the case level QC, so it belongs on the case. It is separate from the sample QC

cg/store/crud/read.py
seallard: Hm, what do you mean by having been run? The sequencing QC is done on cases after sequencing and before analysis. So I don't think it is necessary to do any additional filtering

cg/cli/base.py
seallard: I'd rather keep the sequencing quality control as a separate top level command for now

cg/services/sequencing_qc_service/sequencing_qc_service.py
ChrOertlin: Cases are not sequenced, samples are. We can check whether all samples in the case pass the sequencing qc. 
If this is to be on the case level, which we did not say in the refinement, then it should be named differently. 

Perhaps: ready_to_analyse

ChrOertlin: My understanding from the refinement was -> put sequencing qc on the sample.
Then a case can be analysed when all the samples pass the qc. but maybe I misunderstood.

cg/store/crud/read.py
islean: Yeah, but I mean now we are picking up all cases with pending sequencing qc. Is that not the default that all completely new cases have?

islean: So freshly ordered cases will immediately be picked up by this? Or am I mistaken?

seallard: Ah, true 👍 What do you think is a good approach to avoid that?

cg/services/sequencing_qc_service/sequencing_qc_service.py
ChrOertlin: The app-tag on the sample determines the amount of reads it has to have.

Priority is on both the sample AND the case level, beats me what either of those mean now... will have to check the relationship there... 

Yes, I would not call it "sequencing" on a case level.

I would use terms that are sequencing related on sample and analysis related on case

Then let sequencing related attributes on the sample determine the case attribute, since the samples will be re-used and the number of reads should not make a difference since that is determined by the apptag (sample level).

seallard:
Leaving this for a future project

henrikstranneheim:
Perhaps not enough for an entire project, but clearly needs more refinement. Added to refinement 29/5.

cg/services/sequencing_qc_service/sequencing_qc_service.py
seallard: We renamed it to `aggregated_sequencing_qc` instead of `sequencing_qc`. The `ready_for_analysis` is misleading since other criteria factor in such as whether the data needs to be retrieved from pdc. What do you think?

seallard: We majority support for this naming, closing this to move forward getting rid of the technical debt. We can rename it later

cg/store/crud/read.py
seallard: Lets use the has sequence filter 👍

islean (approved):
Looks good as far as I can tell! I guess the drawbacks of adding additional code identifying only cases with new sequencing data compared to the last seq qc check are bigger than just running the check for more cases?

seallard:
> Looks good as far as I can tell! I guess the drawbacks of adding additional code identifying only cases with new sequencing data compared to the last seq qc check are bigger than just running the check for more cases?

Yes, I think that adds needless complexity. The number of cases which has failed sequencing should never be that big (I guess at most in the hundreds).

PR #3310, opened 2024-06-03
refacto(flow cell dir data pt1)

ChrOertlin:
## Description

First part of the larger refactoring.
Will do more specific refactoring of other modules separately.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/cli/backup.py
```diff
@@ -239,7 +239,7 @@ def _get_samples(status_api: Store, object_type: str, identifier: str) -> list[S
     get_samples = {
         "sample": status_api.sample,
         "case": status_api.get_samples_by_case_id,
-        "flow_cell": status_api.get_samples_from_flow_cell,
+        "devices": status_api.get_samples_from_flow_cell,
```
islean, line 242: Not something that needs fixing now, but I find this dict a bit unclear. Is it to get around doing three ifs?

cg/models/devices/illumina_run_directory_data.py
```diff
@@ -36,53 +36,53 @@
...
-        self.path: Path = flow_cell_path
+    def __init__(self, sequencing_run_path: Path):
+        LOG.debug(f"Instantiating FlowCellDirectoryData with path {sequencing_run_path}")
+        self.path: Path = sequencing_run_path
         self.machine_name: str = EMPTY_STRING
         self._run_parameters: RunParameters | None = None
         self.run_date: datetime.datetime = datetime.datetime.now()
         self.machine_number: int = 0
-        self.base_name: str = EMPTY_STRING  # Base name is flow cell-id + flow cell position
+        self.base_name: str = EMPTY_STRING  # Base name is sequencing run-id + flow cell position
```
islean, line 49: Would it not make sense to call it flow cell id here? Since we are in the IlluminaRunDirectoryData?

cg/models/devices/illumina_run_directory_data.py
```diff
@@ -36,53 +36,53 @@
 }
 
 
-class FlowCellDirectoryData:
-    """Class to collect information about flow cell directories and their particular files."""
+class IlluminaRunDirectoryData:
```
islean, line 39: This is probably due to me being a bit fuzzy on what the intent is with this class. But is this not intended for Illumina flow cells? So would it not make sense to call the sequencing run for a flow cell here? 

For some cases, e.g. method names, I guess it would be to make the exposed methods have the same name? Although I see no inheritance from some interface. In for example comments, though, is there a benefit to calling it a sequencing run?

cg/models/devices/illumina_run_directory_data.py
```diff
@@ -317,17 +319,17 @@ def __str__(self):
...
+        LOG.debug(f"Found directory: {run_dir}")
         try:
-            flow_cell = FlowCellDirectoryData(flow_cell_path=flow_cell_dir)
+            sequencing_run = IlluminaRunDirectoryData(sequencing_run_path=run_dir)
         except FlowCellError:
             continue
-        flow_cells.append(flow_cell)
-    return flow_cells
+        sequencing_runs.append(sequencing_run)
+    return sequencing_runs
```
islean, line 335: Just in general for this class, I do not see the benefit from renaming flow cell to sequencing run. Is sequencing run more specific for Illumina or are we not losing information in logs, comments and method names here?

islean: I guess for the method names the class name might be enough though.

islean (approved):
Looks good! Slightly confused as to why we need to change so much in the actual Illumina flow cell to not mention Illumina flow cells though.

cg/cli/backup.py
ChrOertlin: this is an error will change.

cg/models/devices/illumina_run_directory_data.py
ChrOertlin: sequencing run is more specific to what it actually represents. The use of flow cell as an output directory is a bit misleading since you just get data that was generated by a sequencing run that consumed a flow cell.

cg/models/devices/illumina_run_directory_data.py
ChrOertlin: Indeed, I think the comment should say flow cell-id as well.

ChrOertlin:
# Tests

`cg demultiplex flow-cell all --dry-run` instantiates properly.

Issue #2821, opened 2024-01-11
Increase characters in Click help


islean:
## Description

In our CLI, the descriptions of commands are often not shown in their entirety, see below

<img width="655" alt="image" src="https://github.com/Clinical-Genomics/cg/assets/40887124/72fde5ab-20ca-4f57-bc3f-27fb78ad776b">

I would guess that this is a setting which can be modified in Click, which I suggest we set higher.

## Suggested solution

Increase allowed characters in the CLI command prints.

## This can be closed when

Help descriptions are shown in their entirety.

henrikstranneheim:
@islean https://click.palletsprojects.com/en/8.1.x/documentation/
```
To change the maximum width, pass max_content_width when calling the command.

cli(max_content_width=120)

islean:
@henrikstranneheim Might be this instead https://click.palletsprojects.com/en/8.1.x/documentation/#command-short-help? I interpret the max_content_width as allowing the lines to be wider than the terminal, whereas this is an issue only with the shorter help messages being truncated much earlier than that.

henrikstranneheim:
@eliottBo would you like to look into this once the dry run issue is solved?

eliottBo:
Yes I can look into it after the dry-run 👍

eliottBo:
Solved with https://github.com/Clinical-Genomics/cg/pull/3315
Closing

PR #3315, opened 2024-06-04
Increase click character display

eliottBo:
## Description
This PR addresses issue #2821. The help messages were truncated. I found a way to display a number of character adapted to the terminal size. 
Might not be the most optimal solution, if you have any other suggestions I can look at them.
I tried to only modify the cg command and that all subcommands inherit this context but I could not find a way (@click.pass_context does not seem to work when I tried).

### Added

- Added a constant in cg/cli/utils.py that is a dictionary with the value of the terminal size -10.
- In `@click.group()` decorators, added the parameter `context_settings=CLICK_CONTEXT_SETTINGS`. The command will therefore display a number of character equal to terminal size - 10.

### Fixed

- Truncated characters in `--help` command.

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    ```

### How to test

- [x] Run `--help` option for different cg commands

### Expected test outcome

- [x] Check that the help messages display completely

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


eliottBo:
Also I am unsure if I have to add anything special at other places because I added an import statement

eliottBo:
## Here are some tests after installing the branch to cg-stage hasta:
```bash
## install branch:
[eliott.bosshard@hasta:~] [S_base] $ bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b increase_click_character
(...)
Log deploy... done.
cg, version 60.9.3
```
### Before installing the branch:
Screenshot 2024-06-04 at 15 32 34
Screenshot 2024-06-04 at 16 02 49

### After installing the branch:
Screenshot 2024-06-04 at 15 34 11
Screenshot 2024-06-04 at 15 32 59

cg/cli/utils.py
```diff
@@ -12,3 +14,7 @@ def is_case_name_allowed(name: str) -> bool:
     """Returns true if the given name consists only of letters, numbers and dashes."""
     allowed_pattern: re.Pattern = re.compile("^[A-Za-z0-9-]+$")
     return bool(allowed_pattern.fullmatch(name))
+
+
+def click_context_setting_max_content_width() -> dict[str, int]:
+    return {"max_content_width": shutil.get_terminal_size().columns - 10}
```
ChrOertlin, line 20:
```suggestion
MAX_CONTENT_WIDTH :dict[str, int] = {"max_content_width": shutil.get_terminal_size().columns - 10}
```
I think this could be a constant. Right @islean ?

eliottBo:
@islean the inheritance is from the parent to the child only that is why I also added the context to the sub commands of ```workflow``` because they are also groups with sub-commands (ex: ```balsamic``` is a group that has ```config-case```, ```run``` etc). It would be Ideal to add it, as you said, in the cg base command and that it is inherited through all the sub commands but I could not find a way to do so unfortunately. But if you have an idea lmk!

eliottBo:
Tests after changing the function to a constant:
```bash
[eliott.bosshard@hasta:~] [S_base] $ bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b increase_click_character

(...)

done.
Log deploy... done.
cg, version 61.0.0
```
Before installing the branch:
Screenshot 2024-06-11 at 13 01 11

After installing the branch:
Screenshot 2024-06-11 at 13 08 15

eliottBo:
@ChrOertlin works fine with the constant thanks for the comment!

islean:
> @islean the inheritance is from the parent to the child only that is why I also added the context to the sub commands of `workflow` because they are also groups with sub-commands (ex: `balsamic` is a group that has `config-case`, `run` etc). It would be Ideal to add it, as you said, in the cg base command and that it is inherited through all the sub commands but I could not find a way to do so unfortunately. But if you have an idea lmk!

No, if setting it in the @click.group for the base command is not passed down all levels, I don't have any clue. Go ahead!

This is where I would hope would be enough:
image

ivadym (approved):
Nice work, well done! 💯 

Should we add the context to all the `click.group()` then? I've found some are missing it 🤔 

- `cg/cli/base.py`
- `cg/cli/get.py`
- `cg/cli/transfer.py`
- `cg/cli/demultiplex/finish.py`
- `cg/cli/set/base.py`

ivadym:
> @ivadym I did not add in these on purpose bc the help message displayed fully. However, I can add it if it is more consistent from a programming point-of-view 😛

Yes, I think it's a good idea to add it in case in the future we expand those help messages. And I suppose the time we might add to the CLI commands is insignificant, but out of curiosity, have you quantified it?

eliottBo:
Ok I will add it then. I quantified the time difference but in a code replica not the actual code and it is the same.

eliottBo:
## Deployment to prod:
Screenshot 2024-06-12 at 11 09 59

Screenshot 2024-06-12 at 11 10 12

Screenshot 2024-06-12 at 11 10 26

PR #3317, opened 2024-06-04
Feat(freshdesk) Add Freshdesk Client

Vince-janv:
## Description
Adds a basic client for communicating with the freshdesk REST API. 

Currently only the create ticket method is implemented

### Added

- Freshdesk client
- Create ticket method


cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -0,0 +1,32 @@
...
+    """Client for communicating with the freshdesk REST API."""
+
+    def __init__(self):
+        self.api_key = None
+        self.url = None
+
+    def init_app(self, url, api_key):
+        """Set up the client."""
+        self.url = url
+        self.api_key = api_key
```
Vince-janv, line 21: I really don't like the way of instantiating stuff. But to that seems to be way we do it in ext.py. I think a more intuitive way results in circular imports. Will evaluate in the coming PRs

cg/clients/freshdesk/models.py
```diff
@@ -0,0 +1,44 @@
+from datetime import datetime
+
+from pydantic import BaseModel
+
+
+class Ticket(BaseModel):
+    """Freshdesk ticket."""
```
Vince-janv, line 7: Not sure what the custom is here. Should one have separate models for request and response? Or just have a big one with everything optional so which can be used both as input and output?

The reason for unifying them was because the overlap was so large

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def create_ticket(self, ticket: Ticket) -> Ticket:
+        """Create a ticket."""
+        LOG.debug(ticket.model_dump_json())
+        response = requests.post(
+            url=f"{self.url}{EndPoints.TICKETS}",
+            headers={"Content-Type": "application/json"},
+            auth=(self.api_key, "X"),
+            json=ticket.model_dump(exclude_none=True),
+        )
```
islean, line 31: You could consider hiding this away a bit. I think saving the headers and auth as class variables or something similar might make the more relevant info stand out more.

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -0,0 +1,32 @@
...
+        return Ticket.model_validate(response.json())
```
islean, line 32: Consider catching HTTP errors with response.raise_for_status for example. Now I guess everything would be a pydantic validation error.

cg/clients/freshdesk/models.py
```diff
@@ -0,0 +1,44 @@
...
+    cc_emails: list[str] = []
+    company_id: int | None = None
+    custom_fields: dict | None = None
+    deleted: bool | None = None
+    description: str | None = None
+    description_text: str | None = None
+    due_by: datetime | None = None
+    email: str | None = None
+    email_config_id: int | None = None
+    facebook_id: str | None = None
```
ChrOertlin, line 20: are we ever going to use this?

cg/clients/freshdesk/models.py
```diff
@@ -0,0 +1,44 @@
...
+    requester_id: int | None = None
+    responder_id: int | None = None
+    source: int = 2
+    source_additional_info: str | None = None
+    spam: bool | None = None
+    status: int = 3
+    subject: str = ""
+    tags: list[str] = []
+    to_emails: list[str] = []
+    twitter_id: str | None = None
```
ChrOertlin, line 41: same here, also I reckon this should be 'x_id' nowadays.

cg/clients/freshdesk/models.py
islean: What does the response add? Timestamps for creation? And are they left blank if something goes wrong? You could consider having a parent class with two child classes.

cg/clients/freshdesk/models.py
```diff
@@ -0,0 +1,44 @@
...
+    phone: str | None = None
+    priority: int = 1
+    product_id: int | None = None
+    reply_cc_emails: list[str] = []
+    requester_id: int | None = None
+    responder_id: int | None = None
+    source: int = 2
+    source_additional_info: str | None = None
+    spam: bool | None = None
+    status: int = 3
```
islean, line 37: Is there a reason for the default values 1,2,3 here? Seems very cryptic to me.

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+class FreshdeskClient:
+    """Client for communicating with the freshdesk REST API."""
+
+    def __init__(self):
+        self.api_key = None
+        self.url = None
+
+    def init_app(self, url, api_key):
```
islean, line 18: Would be nice with type hints.

islean: And what is the different use cases for the init and the init_app?

islean (approved):
Looks good! Some minor comments

cg/clients/freshdesk/models.py
ChrOertlin: I would make a model with the required fields for a CreateTicket model that we actually will use. We can then still adjust if we need to add something later on. 

I think having all these fields could potentially be confusing.

cg/clients/freshdesk/freshdesk_client.py
ChrOertlin, line 21: why do you need this here? isn't it enough to assign it in the __init__()?

cg/clients/freshdesk/models.py
Vince-janv: So for the creation only about 5 fields are used, but the objects that are returned contain a lot. But importantly an id and timestamps.

I agree that most fields are unnecessary. Would you both prefer two classes TicketCreate and TicketResponse which only contain the fields we intend to use, and add more fields if we decide to use them?

cg/clients/freshdesk/models.py
Vince-janv: Nope, I just added everything the return-model contains

cg/clients/freshdesk/models.py
Vince-janv: see comment above

cg/clients/freshdesk/freshdesk_client.py
Vince-janv: Makes sense, will do!

cg/clients/freshdesk/freshdesk_client.py
Vince-janv: Will look into this in the next PR combined with retry strategies

cg/clients/freshdesk/models.py
Vince-janv: Yes, I should probably make them constants. Source: 2 is from the API (weird that one can choose that 😅 ). Status 3 is open, and priority 1 is medium i think. But that should be clarified

cg/clients/freshdesk/freshdesk_client.py
Vince-janv: Will add type hints! Regarding the `init_app` thingy there is a weird way we instantiate services for the API. We basically instantiate them in one file, before we read the config with env variables. Hence we can't assign them on instantiation... I might spend some time to fix it in a future PR, but this just follows what has been done for other services

cg/clients/freshdesk/models.py
islean: I agree, sort of similar to the request/response from DDN.

Vince-janv: Addressed in [3dca545](https://github.com/Clinical-Genomics/cg/pull/3317/commits/3dca545ffb56567f9685a269e81d154581de9f88)

cg/clients/freshdesk/models.py
Vince-janv: removed in [3dca545](https://github.com/Clinical-Genomics/cg/pull/3317/commits/3dca545ffb56567f9685a269e81d154581de9f88)

cg/clients/freshdesk/models.py
Vince-janv: removed in [3dca545](https://github.com/Clinical-Genomics/cg/pull/3317/commits/3dca545ffb56567f9685a269e81d154581de9f88)

cg/clients/freshdesk/models.py
Vince-janv: Added constants in [3dca545](https://github.com/Clinical-Genomics/cg/pull/3317/commits/3dca545ffb56567f9685a269e81d154581de9f88)

cg/clients/freshdesk/freshdesk_client.py
Vince-janv: Fixed type hints in [4eb3e02](https://github.com/Clinical-Genomics/cg/pull/3317/commits/4eb3e027287cc3751ae83eed06d269ef3292cc51)

cg/clients/freshdesk/freshdesk_client.py
Vince-janv: Yes, this is not very logical... See this comment: https://github.com/Clinical-Genomics/cg/pull/3317#discussion_r1627195681

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -0,0 +1,32 @@
...
+        self.url = url
+        self.api_key = api_key
+
+    def create_ticket(self, ticket: Ticket) -> Ticket:
+        """Create a ticket."""
+        LOG.debug(ticket.model_dump_json())
+        response = requests.post(
+            url=f"{self.url}{EndPoints.TICKETS}",
+            headers={"Content-Type": "application/json"},
+            auth=(self.api_key, "X"),
```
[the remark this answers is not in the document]
Vince-janv: Addressed in [4eb3e02](https://github.com/Clinical-Genomics/cg/pull/3317/commits/4eb3e027287cc3751ae83eed06d269ef3292cc51)

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -0,0 +1,32 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+class FreshdeskClient:
+    """Client for communicating with the freshdesk REST API."""
+
+    def __init__(self):
+        self.api_key = None
+        self.url = None
```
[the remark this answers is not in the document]
Vince-janv: Changed in https://github.com/Clinical-Genomics/cg/pull/3328

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def create_ticket(self, ticket: TicketCreate) -> TicketResponse:
+        """Create a ticket."""
+        LOG.debug(ticket.model_dump_json())
+        response = requests.post(
+            url=f"{self.url}{EndPoints.TICKETS}",
+            headers=self.headers,
+            auth=self.auth_header,
+            json=ticket.model_dump(exclude_none=True),
+        )
```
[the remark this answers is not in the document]
Vince-janv: Changed in https://github.com/Clinical-Genomics/cg/pull/3328

PR #3320, opened 2024-06-05
fix(allow multiple data analysis type in mip)

ChrOertlin:
## Description

Currently cg upload -c fails for mip-dna cases that have multiple analysis types (WES;WGS) for the same case. however, the workflow supports the use of multiple data analysis types for the same case.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/workflow/analysis.py
+++ b/cg/meta/workflow/analysis.py
@@ -712,7 +712,7 @@ class AnalysisAPI(MetaAPI):
         analysis_types: set[str] = {
             link.sample.application_version.application.analysis_type for link in case.links
         }
-        if len(analysis_types) > 1:
+        if len(analysis_types) > 1 and self.workflow != Workflow.MIP_DNA:
             raise ValueError(
                 f"Case samples have different analysis types {', '.join(analysis_types)}"
             )


cg/meta/workflow/analysis.py
```diff
@@ -712,7 +712,7 @@ def get_data_analysis_type(self, case_id: str) -> str | None:
         analysis_types: set[str] = {
             link.sample.application_version.application.analysis_type for link in case.links
         }
-        if len(analysis_types) > 1:
```
islean, line 715: Won't this give an unclear return statement? It just pops the first analysis type in the list but if we have several we might need some extra logic

cg/meta/workflow/analysis.py
```diff
@@ -712,7 +712,7 @@ def get_data_analysis_type(self, case_id: str) -> str | None:
...
+        if len(analysis_types) > 1 and self.workflow != Workflow.MIP_DNA:
```
islean, line 715: Won't this give an unclear return statement? It just pops the first analysis type in the list but if we have several we might need some extra logic

ChrOertlin: hmm I see... perhaps we should not even go into the function for ip dna but handle the function call based on the workflow?

islean: A bit uncertain but it is set in the RarediseaseParameters. Might need to check with some Raredisease developer to know the implications?

fevac (approved):
Looks good to me as a quick fix. However, this logic would make more sense during the ordering or the start step. If we don't handle multiple analyses types by the workflow, why even run them?

ChrOertlin:
> Is this only to upload one case? If so I don't think this is worth introducing this extra logic layer

I don't know. From the infomration that I have gathered.
1. MIP can be run like this
2. Logic was introduced that now causes an error 27-5-24, which did not happen previously.
3. I don't know if this is an isolated incident or a more common occurence.

henrikstranneheim:
I remember being very hesitant adding this feature in MIP. @jemten Is this allowed in raredisease? Given that MIP is nearing its end-of-life - is this important enough to fix @Clinical-Genomics/prodbioinfo ?

Karl-Svard:
It's currently blocking a case in production from being uploaded, `comicwasp`. So would be great if it could be fixed shortly! 

In the past, these types of cases (WGS + WES) were analyzed as WES. This shouldn't have changed as far as I know

ChrOertlin:
Ok, since this is a no longer supported feature I will close the PR.

PR #3328, opened 2024-06-10
Feat(Freshdesk) Added retry functionality

Vince-janv:
## Description
This PR adds retry functionality and error handling to the freshdesk client which addresses comments made on the previous PR.

### Added

- Retry functionality
- Error handling for the one existing method

### Changed

- Requests are now submitted using a session
- The session is configured upon instantiation


cg/clients/freshdesk/exceptions.py
```diff
@@ -0,0 +1,16 @@
...
+class FreshdeskAPIException(FreshdeskException):
+    """The Freshdesk API returned an error."""
+
+    pass
+
+
+class FreshdeskModelException(FreshdeskException):
+    """Pydantic model validation failed."""
+
+    pass
```
Vince-janv, line 16: Open for renaming these for more clarity

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -11,28 +18,51 @@
...
+
+    @staticmethod
+    def _configure_retries(session: Session) -> None:
+        """Configures retries for the session."""
+        retry_strategy = Retry(
+            total=5,
+            status_forcelist=[TO_MANY_REQUESTS_CODE],
+            allowed_methods=None,
+            backoff_factor=2,
         )
```
Vince-janv, line 66: This strategy means:
1. Retry 5 times
2. On every retry wait twice as long (2,4,8,16,32) seconds.
3. This should only be done on 429 (too many requests) exit status
4. No restrictions on which methods this applies to

Vince-janv: By default retries is only applied to idempotent methods (ie not POST) but we do want it for POST as well. 

That's why I only added 429 for the strategy, but please give your opinion on how this should be configured

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -11,28 +18,51 @@
...
+        self.session = self._configure_session()
+
+    def __enter__(self) -> "FreshdeskClient":
+        return self
 
-    @property
-    def auth_header(self):
-        return self.api_key, "X"
+    def __exit__(self, exc_type, exc_value, traceback) -> None:
+        self.session.close()
```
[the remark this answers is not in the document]
seallard: What is a request session?

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -11,28 +18,51 @@
...
-            headers=self.headers,
-            auth=self.auth_header,
-            json=ticket.model_dump(exclude_none=True),
+        try:
+            response: Response = self.session.post(
+                url=f"{self.base_url}{EndPoints.TICKETS}",
+                json=ticket.model_dump(exclude_none=True),
+            )
+            response.raise_for_status()
+            ticket_response: TicketResponse = TicketResponse.model_validate(response.json())
```
seallard, line 41: Redundant type hint

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -11,28 +18,51 @@
...
+        except RequestException as error:
```
islean, line 42: What errors will be caught here? raise_for_status raises an HTTPException.

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -11,28 +18,51 @@
...
+        return session
+
+    @staticmethod
+    def _configure_retries(session: Session) -> None:
+        """Configures retries for the session."""
+        retry_strategy = Retry(
+            total=5,
+            status_forcelist=[TO_MANY_REQUESTS_CODE],
+            allowed_methods=None,
+            backoff_factor=2,
```
seallard, line 65: Is it possible to explicitly set the delay time between the requests explicitly?

cg/clients/freshdesk/constants.py
```diff
@@ -19,3 +19,6 @@ class Priority:
     MEDIUM = 2
     HIGH = 3
     URGENT = 4
+
+
+TO_MANY_REQUESTS_CODE = 429
```
islean, line 24:
```suggestion
TOO_MANY_REQUESTS_CODE = 429
```

cg/clients/freshdesk/freshdesk_client.py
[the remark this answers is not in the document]
Vince-janv: From https://requests.readthedocs.io/en/latest/user/advanced/#session-objects:

> The Session object allows you to persist certain parameters across requests. It also persists cookies across all requests made from the Session instance, and will use urllib3’s [connection pooling](https://urllib3.readthedocs.io/en/latest/reference/urllib3.connectionpool.html). So if you’re making several requests to the same host, the underlying TCP connection will be reused, which can result in a significant performance increase (see [HTTP persistent connection](https://en.wikipedia.org/wiki/HTTP_persistent_connection)).

Basically you don't have to give parameters to every request, and it also reuses the TCP connection. But the main reason is for configuring retry strategies which seems to be applied on a session rather than individual requests

cg/clients/freshdesk/freshdesk_client.py
islean: A bit unsure of what this means. What does "this" in point 3 refer to? The retries? Because if we only send one request and we get an error, that would not be a "too many requests" error?

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -11,28 +18,51 @@
...
+        session.auth = (self.api_key, "X")
+        session.headers.update({"Content-Type": "application/json"})
+        return session
+
+    @staticmethod
+    def _configure_retries(session: Session) -> None:
+        """Configures retries for the session."""
+        retry_strategy = Retry(
+            total=5,
+            status_forcelist=[TO_MANY_REQUESTS_CODE],
```
seallard, line 63: Spelling `TOO`, but also it is not necessary to redefine these constants. They are part of the HTTP package, which I think is preferable to use.

```python
from http import HTTPStatus

HTTPStatus.TOO_MANY_REQUESTS
```

Vince-janv: Very true. Will import 👍

cg/clients/freshdesk/freshdesk_client.py
Vince-janv: Yes, "this" refers to retries. Basically if we get any other error exit code than 429 we will not retry.

cg/clients/freshdesk/freshdesk_client.py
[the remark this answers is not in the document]
seallard: I'm not 100 % sure why it is beneficial to use a context manager like this. Does it not make more sense to instantiate the client once centrally and keep the client lifetime for the lifetime of the app instead of tying it to a `with FreshdeskClient as client` block?

seallard (approved):
Nice 👍 I don't understand why we want to use a context manager for the client yet

cg/clients/freshdesk/freshdesk_client.py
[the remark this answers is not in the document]
seallard: We should use the session, I'm just not sure whether we want to tie it to the lifetime of the app or a context manager block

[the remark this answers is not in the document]
Vince-janv: After reading up a bit I think I'm convinced. Could you have a look at [f441155](https://github.com/Clinical-Genomics/cg/pull/3328/commits/f441155bc691ac4d12fabbdb15924ac0943a463e)?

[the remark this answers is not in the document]
seallard: I'd instantiate the session in the constructor and keep it there to get more value from the pooled TCP connections.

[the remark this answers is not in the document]
seallard: And it seems like the package should handle all stuff related to keeping the connection alive automatically https://requests.readthedocs.io/en/latest/user/advanced/#keep-alive

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -11,28 +19,47 @@
...
+        session.auth = (self.api_key, "X")
+        session.headers.update({"Content-Type": "application/json"})
+        return session
+
+    @staticmethod
+    def _configure_retries(session: Session) -> None:
+        """Configures retries for the session."""
+        retry_strategy = Retry(
+            total=5,
+            status_forcelist=[HTTPStatus.TOO_MANY_REQUESTS],
```
seallard, line 60: Maybe we can use it for some more transient errors as well? Here is another library that I've used and the defaults they use: [default transient http errors](https://github.com/App-vNext/Polly/wiki/Polly-and-HttpClientFactory#extending-the-convenience-addtransienthttperrorpolicy-definition).

I think it makes sense to also, at least, use the retry for these errors
- 503 Service Unavailable
- 504 Gateway Timeout Error
- 408 Request Timeout Error

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -11,28 +19,47 @@
...
 
     def create_ticket(self, ticket: TicketCreate) -> TicketResponse:
         """Create a ticket."""
         LOG.debug(ticket.model_dump_json())
-        response = requests.post(
-            url=f"{self.url}{EndPoints.TICKETS}",
-            headers=self.headers,
-            auth=self.auth_header,
-            json=ticket.model_dump(exclude_none=True),
+        session: Session = self._get_session()
```
seallard, line 29: I'd instantiate the session in the constructor and keep it there to get more value from the pooled TCP connections.

seallard (approved):
Nice!!! If you have the energy, consider updating our http client docs example with the error handling and retries https://atlas.scilifelab.se/infrastructure/development/cgep/cgep_6_consume_web_apis/

I think this client is cleaner than the example in the docs, so consider replacing it 😄

cg/clients/freshdesk/freshdesk_client.py
Vince-janv: solved in [f441155](https://github.com/Clinical-Genomics/cg/pull/3328/commits/f441155bc691ac4d12fabbdb15924ac0943a463e)

cg/clients/freshdesk/constants.py
Vince-janv: solved in [f441155](https://github.com/Clinical-Genomics/cg/pull/3328/commits/f441155bc691ac4d12fabbdb15924ac0943a463e)

cg/clients/freshdesk/freshdesk_client.py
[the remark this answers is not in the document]
Vince-janv: Decided to keep one session for the lifetime of the app

cg/clients/freshdesk/freshdesk_client.py
Vince-janv: `RequestException` is a parent to `HTTPErrror` so it will catch more stuff as well. Such as `ConnectionError`, `Timeout` and `ProxyError`. I think it makes sense to catch it quite wide

cg/clients/freshdesk/constants.py
Vince-janv: Imported instead

cg/clients/freshdesk/freshdesk_client.py
Vince-janv: Yes! ChatGPT recomended these: HTTPStatus.TOO_MANY_REQUESTS, HTTPStatus.INTERNAL_SERVER_ERROR, HTTPStatus.BAD_GATEWAY, HTTPStatus.SERVICE_UNAVAILABLE, HTTPStatus.GATEWAY_TIMEOUT. Will add 408 as well 👍

cg/clients/freshdesk/freshdesk_client.py
Vince-janv: Done 👍

Issue #3327, opened 2024-06-10
Add optional parameters for cg downsample samples


fevac:
As a bioinformatician,
I want to be able to set the action in StatusDB, the ticket number, customer, pipeline and delivery type when downsampling a case,
So that new cases do not have to be manually modified to avoid starting automatically and having incorrect information.

# Acceptance Criteria

- [x] Optional parameter to set action in StatusDB
- [x] Optional parameter to set the customer in StatusDB
- [x] Optional parameter to set the workflow/pipeline in  StatusDB
- [x] Optional parameter to set delivery type in StatusDB

# Notes

- Additional information.
- Dependencies.
- Related user stories.

# Implementation plan

- [ ] https://github.com/Clinical-Genomics/cg/pull/3329

fevac:
@ChrOertlin @islean thoughts about this? I would really like to get this feature so I don't have to be manually changing information from downsampled cases

ChrOertlin:
Are these all case level parameters?

islean:
> @ChrOertlin @islean thoughts about this? I would really like to get this feature so I don't have to be manually changing information from downsampled cases

Should be quite doable. I guess we should make them all optional though.

fevac:
> Are these all case level parameters?

yes

Vince-janv:
Validation and downsampling are not prioritised at the moment. CG as an organisation needs a solution for this, but it is unclear cg should take care of it

PR #3329, opened 2024-06-10
feature(downsample api parameters)

ChrOertlin:
## Description
resolves https://github.com/Clinical-Genomics/cg/issues/3327
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


ChrOertlin:
# tests
## new options are registered:
image

## testing dry-run

Dry run works

## submitting job

submitting job works

cg/apps/downsample/downsample.py
```diff
@@ -25,9 +26,7 @@ def __init__(
         super().__init__(config)
         self.dry_run: bool = dry_run
 
-    def get_downsample_data(
-        self, sample_id: str, number_of_reads: float, case_id: str, case_name: str
-    ) -> DownsampleData:
+    def get_downsample_data(self, downsample_input: DownsampleInput) -> DownsampleData:
         """Return the DownSampleData.
         Raises:
             DownsampleFailedError
```
henrikstranneheim, line 32: Should we mention on what grounds this exception is raised?

cg/apps/downsample/downsample.py
```diff
@@ -126,22 +122,12 @@ def start_downsample_job(
...
-        case_name: str,
-        case_id: str,
-        number_of_reads: float,
-        account: str | None = None,
-    ) -> int | None:
+    def downsample_sample(self, downsample_input: DownsampleInput) -> int | None:
         """Downsample a sample."""
-        LOG.info(f"Starting Downsampling for sample {sample_id}.")
-        validate_sample_id(sample_id)
+        LOG.info(f"Starting Downsampling for sample {downsample_input.sample_id}.")
```
henrikstranneheim, line 127:
```suggestion
        LOG.info(f"Starting downsampling for sample {downsample_input.sample_id}.")
```

tests/models/downsample/test_down_sample_meta_data.py
```diff
@@ -47,3 +48,56 @@ def test_downsample_meta_data_pass_checks(
...
+        hk_api=downsample_hk_api,
+        downsample_input=downsample_input,
+        out_dir=tmp_path_factory.mktemp("tmp"),
+    )
+    # WHEN the number of reads in millionts is converted to a string
+    number_of_reads: str = meta_data.convert_number_of_reads_to_string
+
+    # THEN all necessary models to run the down sample command are created
+    assert meta_data.downsampled_sample.internal_id == f"{sample.internal_id}DS{number_of_reads}M"
+    assert meta_data.downsampled_sample.reads == number_of_reads_in_millions * 1_000_000
```
henrikstranneheim, line 97: `number_of_reads_in_millions * 1_000_000` is repeated - should we make it a fixture?

tests/models/downsample/test_down_sample_meta_data.py
```diff
@@ -47,3 +48,56 @@ def test_downsample_meta_data_pass_checks(
...
+    # GIVEN that the sample has enough reads to down sample
+    assert sample.reads > number_of_reads_in_millions * 1_000_000
+
+    # GIVEN a bundle exists in housekeeper for the sample
+    assert downsample_hk_api.get_latest_bundle_version(sample.internal_id)
+
+    # GIVEN a DownsampleInput with action, ticket, customer_id, data_analysis and data_delivery
+    downsample_input.action = CaseActions.HOLD
+    downsample_input.ticket = "random_ticket"
+    downsample_input.customer_id = "cust000"
```
henrikstranneheim, line 81: Is there a constant in that we can use here?

henrikstranneheim (approved):
Great, 👍 !

islean:
From the discussion:

We have duplicated data in status_db in that workflow is stored on both the case and the order. The same is true for the ticket. Ideally this data should only be on the order and in the order_case table. The changes in this PR will lead to some weird clashes. Currently the PR does not link the downsampled case to the provided order, although the tickets column is updated. This will have the consequence that started analyses will have the original case's order_id in Trailblazer, i.e. they will not show up for the provided ticket. This can probably be solved by just changing the current order_case entry which is created to instead link to the provided ticket's order.

What is more tricky and where we probably need input from @fevac is with workflow. Will you want to downsample a case with workflow set to one thing but link it to a ticket with a different workflow? I.e. would you want to downsample and set the workflow to tomte while the ticket you provide is a balsamic order or will this always be in agreement? Could we just use the workflow from the ticket/order?

fevac:
So, what I need to do is:
- Downsample existing samples and run them with Tomte. Samples were originally ordered with other pipelines, so I do want to change the workflow. So, no I cannot use the workflow given in the original order. 

I'm only interested in downsampling samples and not really on the orignal case nor the original order. So it's quite inconvenient that I need to propagate so much information from the case and the order. I could create a new case from scratch instead of reusing information if that's better, however if I recall properly this was problematic (at least in the past) as cases created from the terminal would lack an order and would mess-up with cigrid/etc. And to do this, the downsample command should be also modified to only downsample samples and not create new cases.

Another option to avoid propagating wrong information from the order:
1. I can create a new ticket with already existing samples from Tomte. Downside of this: I will create full size cases which I don't need. I will need to set those to hold and then downsample those. And it involves manual steps with the ordering that I was hoping to avoid (but I could live with this if this is the preferred option)

ChrOertlin:
> terminal would lack an order

This is an oversight from implementing the order thing and should like be adjusted @islean

islean:
> > terminal would lack an order
> 
> This is an oversight from implementing the order thing and should likely be adjusted. I think we patched the downsample api at some point regarding this as well, right @islean ?

I can't recall one of those bugs not being patched. But if we did not, that is definitely not something to be proud of and it should be fixed immediately. If so, let me know and I will take care of it.

islean:
> So, what I need to do is:
> 
>     * Downsample existing samples and run them with Tomte. Samples were originally ordered with other pipelines, so I do want to change the workflow. So, no I cannot use the workflow given in the original order.
> 
> 
> I'm only interested in downsampling samples and not really on the orignal case nor the original order. So it's quite inconvenient that I need to propagate so much information from the case and the order. I could create a new case from scratch instead of reusing information if that's better, however if I recall properly this was problematic (at least in the past) as cases created from the terminal would lack an order and would mess-up with cigrid/etc. And to do this, the downsample command should be also modified to only downsample samples and not create new cases.
> 
> Another option to avoid propagating wrong information from the order:
> 
>     1. I can create a new ticket with already existing samples from Tomte. Downside of this: I will create full size cases which I don't need. I will need to set those to hold and then downsample those. And it involves manual steps with the ordering that I was hoping to avoid (but I could live with this if this is the preferred option). **Update**: it seems that adding existing samples to an order can only be done manually by clicking one by one. Since I have to add 40-50 samples that's not ideal.

I was a bit unclear. Given that you want the option to provide a ticket, can we use the workflow from the ticket/order you provide yourself? I.e. not the one from the original case but the CLI parameter?

fevac:
> I was a bit unclear. Given that you want the option to provide a ticket, can we use the workflow from the ticket/order you provide yourself? I.e. not the one from the original case but the CLI parameter?

If it's just a "normal" ticket the ticket won't have any information about workflow. If it's a ticket related to an order then I need to make an order first with the proper order information (in which case changing these options is meaningless). Am I missing something?

islean:
> > I was a bit unclear. Given that you want the option to provide a ticket, can we use the workflow from the ticket/order you provide yourself? I.e. not the one from the original case but the CLI parameter?
> 
> If it's just a "normal" ticket the ticket won't have any information about workflow. If it's a ticket related to an order then I need to make an order first with the proper order information (in which case changing these options is meaningless). Am I missing something?

What is a "normal" ticket here? One that was opened without an order of samples being submitted? Basically we have the order table in statusdb. The entries there correspond to an order, i.e. a collection of samples and cases that some customer wants to run with a specific workflow. Hence, the table stores data of which customer, workflow and ticket the order has, and is then linked to the cases in it. Given that you would want to provide a ticket id in the downsampling command, this raises a number of questions.

1. Will this ticket/order be a new order? Tickets opened in SuSy without an order having been submitted will not have an entry in the order table, so those scenarios would fall into this category so one would need to be created. Then we need the customer and the workflow. Currently the original sample's latest case is used, but I guess the suggestion is to override these if any of these parameters is provided?
2. Will this ticket be for an existing order? Then that order will already have a customer and workflow connected to it. I would not want an order tied to cases with a workflow which does not match. Ideally I think we should not have a workflow on case level and only store it on the order level, since we do not support orders with multiple workflows. Hence the question "What do we do if a ticket and workflow is specified?" Do we take the workflow associated with the provided ticket or the workflow provided as parameter? If the latter, should the order entry be changed to said workflow, which would indicate that all cases in the order is using that workflow? 

What I am trying to avoid is having orders with one workflow configured linked to cases with a different workflow. This stems from our database models needing a cleanup since this data should not exist in multiple places, but as it is this is very important to keep in mind. It is the same thing with the customer. Should we be able to create downsampled samples/cases linked to an order for a different customer (and potentially outside the collaboration)? Although the suggested parameters might seem independent, they are quite tightly connected, and the combinations could give conflicting/malformed data if they are not chosen carefully.

henrikstranneheim:
Would it work to create a "downsample order" per workflow and always reuse it when downsampling? It would be nice to separate production from validation/verification as much as possible.

islean:
> Would it work to create a "downsample order" per workflow and always reuse it when downsampling? It would be nice to separate production from validation/verification as much as possible.

I think that could work. We would just need to store the ids of the downsampling orders in some place. Will all downsampled cases be for cust000 though?

fevac:
>  Ideally I think we should not have a workflow on case level and only store it on the order level, since we do not support orders with multiple workflows.

This seems backwards to me. I would say that ideally you would want to store the workflow only on a case level and not on the order level. The order table should be able to be linked to the cases and the workflow should be taken from the case table. To me it seems that the workflow column should be drop from the order table.

fevac:
> Would it work to create a "downsample order" per workflow and always reuse it when downsampling? It would be nice to separate production from validation/verification as much as possible.

I'm not sure this separation is possible. In the future we would like to downsample all our control samples automatically before running the analyses so trending is possible (we can then compare QC values when having the same amount of reads). This would still be part of production and not really a validation/verification.

That being said, I'm not sure I understand what you envision by "downsample order", do you mean creating a new order entry from the non-downsampled one? I don't think we want to downsample all cases in an order, generally it would only be a few. But maybe I'm not understanding your idea, could you expand?

islean:
From a discussion with Eva:

I think we could support the following scenarios currently:

Providing ticket_id, customer and workflow for instances where there is no entry in the order table for the ticket_id. In this case we would create a new order for the ticket_id with the corresponding cases.

Providing only the ticket_id for instances where there is an entry in the order table for it. We would not want to override the workflow set for that order, and it would be bad to have an order with cases having non-matching workflows.

After the database models have been cleared up, we could probably support a wider range of scenarios.

ChrOertlin:
# Refinement

1. Create a new order every time we downsample, generate a new ticket it 5 digits.

ChrOertlin:
closing this pull request for now. There are questions regarding the order table and the stored worfklow thereon. I think this should be addressed first before moving on with this

PR #3343, opened 2024-06-13
Avoid duplicating TB statuses in the order summary

islean:
## Description

Concerns. https://github.com/Clinical-Genomics/trailblazer/issues/452 in conjunction with https://github.com/Clinical-Genomics/trailblazer/pull/454. It will need a patch in the frontend to handle the new response structure though.

### Fixed

- Cases in trailblazer do not get any additional statuses inferred.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/orders/order_summary_service/order_summary_service.py
+++ b/cg/services/orders/order_summary_service/order_summary_service.py
@@ -1,9 +1,8 @@
 from cg.apps.tb.api import TrailblazerAPI
 from cg.apps.tb.dto.summary_response import AnalysisSummary
-from cg.services.orders.order_summary_service.dto.case_summary import CaseSummary
 from cg.services.orders.order_summary_service.dto.order_summary import OrderSummary
 from cg.services.orders.order_summary_service.utils import (
-    create_summaries,
+    _get_analysis_map,
     get_cases_failed_sequencing_qc_count,
 )
 from cg.store.models import Order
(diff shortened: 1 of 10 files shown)


ChrOertlin (approved):
I understand that the "cases_to_exclude" will filter the output. But how are these "cases_to_Exclude" generated? I cant seem to find it here.

Overall, looks good!

cg/services/orders/order_summary_service/order_summary_service.py
```diff
@@ -20,27 +19,46 @@ def get_summary(self, order_id: int) -> OrderSummary:
...
-                in_preparation=in_preparation,
-                in_sequencing=in_sequencing,
-                failed_sequencing_qc=failed_sequencing_qc,
-            )
-            summaries.append(summary)
-
-        return summaries
+    def create_order_summary(self, order: Order, summary: AnalysisSummary) -> OrderSummary:
+        order_id = order.id
+        counted_cases: list[str] = summary.case_ids if summary else []
```
islean, line 39: These are the cases_to_exclude @ChrOertlin. In the AnalysisSummary object you will find an order id and a summary of all the statuses found in Trailblazer. Since the conclusion of the refinement was that these are not to be duplicated, they are excluded when calculating outside-of-tb-statuses.

ChrOertlin: ah and then there is a priority list and something cannot be counted twice in a different list. I see.

islean:
Filtering does not work currently: 

image

cg/services/orders/order_summary_service/order_summary_service.py
islean: Essentially that is the deal. But the only priority this PR adds is that if we get a status from TB for a case, then that is the only status for the case.

islean:
Filtering works better now: 
image

islean:
Deployed to cg-services-prod.

PR #3350, opened 2024-06-18
Feat(config) Modernize api config

Vince-janv:
## Description
The way the cg app is configured is very old fashioned. This causes instantiation of services used by the app to be very cumbersome (see `ext.py` and `app.py`).

This PR add pydantic-basesettings and utilises it to simplify the configuration. This in preparation for adding the freshdesk client.

### Added

- app_config.py containing the BaseSettings class AppConfig

### Changed

- keys in the app.config are now in snake_case to follow the normal use of BaseSettings 

### Fixed

- Unused config entries are removed


Vince-janv (commented):
Removed some config settings that were not used

cg/server/app_config.py
```diff
@@ -0,0 +1,26 @@
+from pydantic_settings import BaseSettings
+
+
+class AppConfig(BaseSettings):
```
ChrOertlin, line 4: I think you might need to specify the ENV file you get the values from, right? 

i.e. add a class Config in here that parses the env file:
```suggestion
class AppConfig(BaseSettings):
    ...
    ...
    class Config:
        env_file = str(ENV_FILE)

```

cg/apps/osticket.py
```diff
@@ -26,10 +26,10 @@ def __init__(self):
     def init_app(self, app: Flask):
         """Initialize the API in Flask"""
         self.setup(
-            api_key=app.config["OSTICKET_API_KEY"],
-            domain=app.config["OSTICKET_DOMAIN"],
-            osticket_email=app.config["SUPPORT_SYSTEM_EMAIL"],
-            email_uri=app.config["EMAIL_URI"],
+            api_key=app.config["osticket_api_key"],
```
ChrOertlin, line 29:
```suggestion
            api_key=app.config.osticket_api_key,
```
Does this not work as well?

ChrOertlin (approved):
Nice modernisation!

cg/server/app_config.py
```diff
@@ -0,0 +1,26 @@
...
+    gunicorn_workers: int = 4
+    gunicorn_threads: int = 4
+    gunicorn_bind: str
+    gunicorn_timeout: int = 400
+    cg_sql_database_uri: str
+    cg_secret_key: str = "thisIsNotASafeKey"
+    invoice_max_price: int = 750000
```
henrikstranneheim, line 11:
```suggestion
    invoice_max_price: int = 750_000
```

henrikstranneheim (approved):
Great 👍

cg/server/app_config.py
Vince-janv: BaseSettings takes the variables straight from the environment variables. But if you want to skip exporting them to the environment you can use the example you added to point to a file instead. 

Since the cg container has all variables exported to it I didn't add the Config class since it will not currently be used in production. I think it can easily be added if we decide to point to a file instead. 

Basically my reasoning is. Since we're not using it we can leave it out for now

cg/apps/osticket.py
Vince-janv: Unfortunately `app.config` is just a dict that I update with the contents of `app_config`. Hence we can only access with values via square brackets

Vince-janv:
StatusDB and order submission works in stage

PR #3378, opened 2024-07-01
Fix get samples CLI to work with Illumina flow cells

diitaz93:
## Description

Fixed the CLI command `cg get samples` so that it works with the new Illumina models

### Changed

- Renamed CRUD function `get_samples_by_illumina_flow_cell_internal_id` to `get_samples_by_illumina_flow_cell` as suggested by @Vince-janv in #3377


ChrOertlin (approved):
💯

PR #3391, opened 2024-07-03
Improve order flow Tomte models

islean:
## Description

This PR adds parent models and Tomte models for the new validation flow.


seallard (approved):
Good start 👍

Issue #3064, opened 2024-03-22
clean `scout-finished-cases` has a potential issue with old `solved` cases


dnil:
## Description

Freshly uploaded rerun solved cases have alignment files erroneously cleaned, although their apparent analysis date is correct. Compare cases `cleverpup`, `tendergator` and `tidyimpala`. On load they would all have been sitting around with fresh files and scout status `solved` after re-uploading, since we only change the status back to inactive for cases with status active and archived if they are rerun.
 
## Suggested solution

First, trouble-shoot: primary candidate would be date parsing on output from `scout export cases --json`.

## This can be closed when

Freshly uploaded rerun alignments are not deleted nightly for previously `solved` cases.

# Clarification
The scout cleaning command removes the CRAM file based on time (older than 300 days) or if the case is solved. Hence if a solved case is re-run the file will be removed directly. The CRAM file is stored in the housekeeper bundle on hasta.

The current logic looks at the field `analysis_date` which should be correctly set for re-analysed cases.

Vince-janv:
Interesting, that does sound annoying. Do we know why a user would like to re-run a solved case? It sounds like a design-flaw somewhere 🤔

dnil:
Yes, the nomenclature of “solving” a genomics investigation is flawed in general. ☺️Life is seldom  that simple. One possibility is the advent of new unexpected symptoms in an rd case, the end of a remission period in leukemia, or perhaps just technical reasons such as wanting a control sample or make a picture for a new report.

One could also say Scout is  it granular enough in the solve situation. I do’nt know if you looked at the three examples provided? For one of the cases, a strong class 3 was answered out as unclear but possibly significant in 2020, and the case marked solved in part to indicate it had been reported non-negative and in part to have the interesting variant show up in matching causatives to try and catch more cases. Fast forward a few years, it’s time to reevaluate the patient for whatever reason. Now the variant is more common in the public dbs, and still no more affected carriers. It was probably benign after all.

Vince-janv:
Not knowing exactly how users work in scout, would it make sense to just create a new case with the old samples? Would the users loose any functionality compared to re-running the same case again.

We are striving for unifying re-runs in the future to only go via new cases, and if this can solve this problem it would be one more argument for it

dnil:
Yes! There are many reasons we allow a choice between fresh new case for a rerun or saving old data on a case.
[...]

PR #3400, opened 2024-07-05
Patch(Scout cleaning) Decrease logging level

Vince-janv:
## Description
Issue #3064 is very difficult to debug due to the large amount of uninformative logging. 

This PR reduces the log level for the unimportant stuff

### Changed

- Log level decreased for  `get_cases` in `scoutapi.py`
- Log level decreased for `hk_alignment_files` in `clean.py`

### DIFF
--- a/cg/cli/clean.py
+++ b/cg/cli/clean.py
@@ -88,7 +88,7 @@ def hk_alignment_files(
         tag_files = set(housekeeper_api.get_files(bundle=bundle, tags=[tag]))
 
         if not tag_files:
-            LOG.warning(
+            LOG.debug(
                 f"Could not find any files ready for cleaning for bundle {bundle} and tag {tag}"
             )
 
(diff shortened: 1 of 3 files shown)


cg/cli/clean.py
```diff
@@ -88,7 +88,7 @@ def hk_alignment_files(
         tag_files = set(housekeeper_api.get_files(bundle=bundle, tags=[tag]))
 
         if not tag_files:
-            LOG.warning(
+            LOG.debug(
```
islean, line 91: Is this not valuable to have as info?

islean (approved):
LGTM!

cg/cli/clean.py
Vince-janv: Sort of torn here... The way we use it I feel like it adds nothing, it just bloats our logs

islean: Feel free to go ahead with your choice!

PR #3403, opened 2024-07-05
Replace calculation with cached qc value in /orders endpoint

seallard:
Replace the repeated sequencing QC calculations in the `/orders` endpoint with queries instead using the stored result.

Part of closing https://github.com/Clinical-Genomics/cg/issues/3223


cg/store/filters/status_case_sample_filters.py
```diff
@@ -56,6 +57,25 @@ def get_not_sequenced_cases(case_samples: Query, **kwargs) -> Query:
     return case_samples.filter(Sample.last_sequenced_at == None).distinct()
 
 
+def get_sequenced_cases(case_samples: Query, **kwargs) -> Query:
```
diitaz93, line 60: Let's add some unit tests to these too

cg/store/crud/read.py
```diff
@@ -1521,6 +1521,23 @@ def get_case_in_sequencing_count(self, order_id: int, cases_to_exclude: list[str
             cases_to_exclude=cases_to_exclude,
         ).count()
 
+    def get_case_failed_sequencing_count(self, order_id: int, cases_to_exclude: list[str]) -> int:
```
diitaz93, line 1524: It would be great a unit test for this

diitaz93 (approved):
Looks good! just would. be good to add some tests

cg/store/filters/status_case_sample_filters.py
```diff
@@ -56,6 +57,25 @@ def get_not_sequenced_cases(case_samples: Query, **kwargs) -> Query:
...
+    not_sequenced = (
+        case_samples.filter(Sample.last_sequenced_at == None)
+        .with_entities(CaseSample.case_id)
+        .subquery()
+    )
+    return (
+        case_samples.outerjoin(not_sequenced, CaseSample.case_id == not_sequenced.c.case_id)
+        .filter(not_sequenced.c.case_id == None)
+        .distinct()
+    )
```
ChrOertlin, line 70: Could you not just look for last_sequenced_at is not None?

seallard: The query does rely on `last_sequenced_at`. The query is a bit tricky because a case is only defined as sequenced if all samples in it are sequenced. I think I'll keep it as is since it works and follows the structure of the other queries.

I have spent some time in the past trying to simplify these queries, but it is tricky 😅

cg/store/filters/status_case_sample_filters.py
seallard: We will end up refactoring the order summary logic extensively and these queries will not live past that, so I'll skip writing tests for these queries.

PR #3405, opened 2024-07-05
Restructure order validation modules

seallard:
## Description
Restructured module to split the validation logic based on workflows instead of splitting the various components per type (like models, validators etc). It will be much easier to encapsulate the logic cleanly that way, instead of mixing the various logic in the same modules.


Vince-janv (approved):
I agree with this structure 👍

diitaz93 (approved):
👍

PR #3410, opened 2024-07-08
Add validation logic skeleton

seallard:
Add overall skeleton for order validation logic, specifically for `Tomte` orders.

I'll start adding actual validation rules in subsequent PR:s with unit tests etc.


cg/services/order_validation_service/order_validation_service.py
```diff
@@ -6,3 +8,9 @@ class OrderValidationService:
 
     def __init__(self, tomte_service: TomteOrderValidationService):
         self.tomte_service = tomte_service
+
+    def validate(self, order_json: str):
+        order = Order.model_validate_json(order_json)
+
+        if order.workflow == Workflow.TOMTE:
+            self.tomte_service.validate(order_json)
```
Vince-janv, line 16: Are just going to add more ifs here? Is a dict not better suited?
```python
def __init__(self, some_params...)
    self.workflow_validators: dict[Workflow, Any] = {
        Workflow.TOMTE = TomteOrderValidationService(some_params...)
    }

def validate(self, order_json: str):
    self.workflow_validators[order.workflow](order_json)
```

Vince-janv (approved):
It's taking shape! 😁

cg/services/order_validation_service/order_validation_service.py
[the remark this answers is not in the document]
seallard: Sure!

The order service accepts the Tomte service in its constructor as an argument, meaning that to instantiate the `OrderValidationService`, you need to pass an instance of the `TomteOrderValidationService`. In the future, it will accept services for the other workflows as well.

This is a design pattern called dependency injection. 

Dependency Injection involves passing (injecting) dependencies, like the `TomteOrderValidationService`, into a class rather than letting the class create the dependencies itself. It separates the concern of using a class from that of creating it. Wikipedia describes it well https://en.wikipedia.org/wiki/Dependency_injection.

## Design
Why not initialize the Tomte service directly?

1. Separation of Concerns: By injecting dependencies, we separate the configuration and instantiation logic from the business logic. This means the `OrderValidationService` doesn’t need to know how to create a `TomteOrderValidationService`; it just needs to use it.
2. Flexibility: This design allows the `OrderValidationService` to support multiple workflows easily. If a new workflow is added, you only need to inject the appropriate service without modifying the existing validation logic.
3. Testability: With dependency injection, you can easily mock dependencies for unit tests. This makes the `OrderValidationService` easier to test in isolation.
4. Complex Dependencies: In the future, if instantiating `TomteOrderValidationService` requires complex setup (e.g., initializing a store or another service), this complexity is handled outside the `OrderValidationService`, keeping the business logic clean and straightforward.

seallard: I think I prefer explicitly passing each of the services in the constructor. There is probably a better design, but I don't think using a dict is better.

seallard: Yes, I prefer passing the services explicitly in the constructor. We can keep the services in a dict internally, but I'll leave it for now. We don't want to have complex logic for instantiating the validation services here.

There might be a better design than an "orchestrator" validation service, but it is the best I could think of for now.

Vince-janv: I see the point of extracting the instantiation. But passing each service as an argument will grow to be quite ugly with the 14 validation_services we have planned. Would it make sense to pass a dict to this service containing instantiated objects?

seallard: There is something called service locator pattern which might be a better fit for us.

Basically, we would let all validation services implement an `OrderValidationService` interface and instantiate them centrally in some container.

Then we would retrieve them like `validation_service = service_locator.get_service(order.workflow)`.

What do you think? It would look something like this

```python

def validate_order(order_json: str):
    order = Order.model_validate_json(order_json)
    validation_service: OrderValidationService = service_locator.get_service(order.workflow)
    validation_service.validate(order)
```

ChrOertlin: I think what you want to do is along the lines of:

```suggestion
class OrderValidator:
    def __init__(self, validation_service: ValidationService = TomteValidationService()):
        self.validation_service = validation_service

    def validate(self, order_json: str):
            self.tomte_service.validate(order_json)
```

Then the place where you would need to validate something:

```
 # create dependency
validationService = TomteOrderValidationService()
# inject dependency
validator = OrderValidator(validationService)
validator.validate(order_json)
```
Then you circumvent the if statements that will need to grow with each worfklow, right?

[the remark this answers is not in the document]
ChrOertlin: Thanks, for the info. It was not immediately clear that you wanted to setup dependency injection here.

ChrOertlin: I was curious about the "registry" and "service locator" and looked it up a bit. Is this really what we want? It seems like it is akin to what we wanted to get away from by not passing the CgConfig directly, instead being more explicit by passing e.g. the store / housekeeper api. 

Some things I found about using a service locator/ registry / context also:
https://blog.ploeh.dk/2010/02/03/ServiceLocatorisanAnti-Pattern/
https://blog.ploeh.dk/2014/05/15/service-locator-violates-solid/
https://blog.ploeh.dk/2015/10/26/service-locator-violates-encapsulation/

this video was quite ifnoramtive:
https://www.youtube.com/watch?v=RlfLCWKxHJ0

I can see what they are talking about given our test suite where we for example have to pass the whole config or just create a context. It is often quite painful. 

Maybe I am missing something?

Vince-janv: A registry is the first thing ChatGPT suggests when I describe what we want to do and to me it makes sense if we map an order type to its appropriate validator. In his examples the Services seem a bit more diverse than what we have, like roof, house, window. While we have chefs knife, butcher knife, gutting knife etc.

I think what is important for us is that we can extend it easily which is something they don't give a solution for 🤔

Vince-janv: But I feel like all these examples dismantled our CGContext pretty well...

seallard: The problem with `CgConfig` is that we pass it down to classes, that is a bad design. It should be used at a top level, in a CLI command or endpoint, retrieving a service and using it.

I think service locator or a factory pattern will work well for our order validation services

ChrOertlin: I think what we do not want to do is pass a service locator in the constructor of the service. 
But rather get the "right" service first and then inject the service and not the service locator. If you want to do it like that.

seallard: Alternatively, we can setup a dependency injection framework (similar to what we have in trailblazer) and let it automatically resolve the dependencies 😄 I guess our validation logic is an edge case, like Vincent said it is a bit different because we have many flavors of the same thing. I'll think about it some more today

ChrOertlin: This is what you get with my suggestion above if I understand the strategy approach correctly.

Each service (tomte, rare-disease, mip, fluffy and whatnot) will have a OrderValidationService that you inject to the validator that you then call in the needed places. 

I think the question here is "with what strategy do we retrieve the right service". Honeslty a workflow to service map might be the easiest to expand, maintain and understand.

seallard: I agree, implementing a draft now 👍

cg/services/order_validation_service/models/order.py
```diff
@@ -1,13 +1,13 @@
...
 
 from cg.constants import DataDelivery
-from cg.models.orders.orderform_schema import OrderCase
+from cg.constants.constants import Workflow
 
 TICKET_PATTERN = r"^#\d{4,}"
 
 
 class Order(BaseModel):
-    cases: list[OrderCase]
```
seallard, line 10: All orders do not contain cases and as such, it does not make sense to keep cases in our base model for an order.

cg/services/order_validation_service/workflows/tomte/validators/case_validators.py
```diff
@@ -1,13 +0,0 @@
...
-    DataDelivery.ANALYSIS_FILES,
-    DataDelivery.FASTQ,
-    DataDelivery.FASTQ_ANALYSIS,
-    DataDelivery.NO_DELIVERY,
-]
-
-
-def validate_tomte_delivery_type(delivery_type: DataDelivery):
-    if delivery_type not in TOMTE_DELIVERY_TYPES:
-        raise ValueError("Delivery type not allowed.")
```
seallard, line 13: This rule was better applied as a Pydantic type in the `TomteOrder` model.

cg/services/order_validation_service/validators/sample_validators.py
```diff
@@ -1,70 +0,0 @@
-from pydantic import ValidationError
-from pydantic_core import InitErrorDetails, PydanticCustomError
-
-from cg.models.orders.sample_base import ContainerEnum
-from cg.services.order_validation_service.models.order_sample import OrderSample
-
-
-def validate_required_buffer(sample: OrderSample) -> OrderSample:
```
seallard, line 8: I'll refactor and add these validation rules in subsequent PR:s

Vince-janv (commented):
The removal of cases from orders and samples from cases I think it warrants some discussion. It's true that cases are not created by the customers for some workflows, but we create them upon order submission. Would it not make sense to validate whatever would be created?

cg/services/order_validation_service/workflows/tomte/validation/inter_field_validators.py
```diff
@@ -0,0 +1,42 @@
+from cg.services.order_validation_service.models.validation_error import (
+    ValidationError,
+    ValidationErrors,
+)
+from cg.services.order_validation_service.workflows.tomte.models.case import TomteCase
+from cg.services.order_validation_service.workflows.tomte.models.order import TomteOrder
+
+
+def each_case_contains_mother(order: TomteOrder) -> ValidationErrors:
```
ChrOertlin, line 9: I wonder whether these can be implemented as Field validators directly on the Pydantic model so we levarage those features a bit more? Or is there merit to split it up like this?

cg/services/order_validation_service/workflows/tomte/validation_service.py
```diff
@@ -1,5 +1,23 @@
...
+)
 
-class TomteOrderValidationService:
 
-    def __init__(self):
-        pass
+class TomteValidationService(OrderValidationService):
+
+    def __init__(self, field_validator: TomteFieldValidator):
+        self.field_validator = field_validator
```
ChrOertlin, line 15: In case we could use the pydantic field validators we can omit this injection here and rely on just the model_validate_json() functionality if I am not mistaken? Would that be easier to maintain in the long run?

cg/services/order_validation_service/workflows/tomte/validation/inter_field_validators.py
```diff
@@ -0,0 +1,42 @@
...
+    errors: list[ValidationError] = []
+    for case in cases:
+        has_father: bool = case_has_father(case)
+        if not has_father:
+            error = ValidationError(
+                message="All cases must contain a father",
+                field="unique_field_identifier_here",
+            )
+            errors.append(error)
+    return ValidationErrors(errors)
```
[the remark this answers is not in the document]
seallard: I removed these rules, I'll add them in subsequent PR:s

cg/services/order_validation_service/workflows/tomte/validation_service.py
seallard: That was our first approach! But we want to map the pydantic errors to a general error object and it was easier to extract it to a separate class

cg/services/order_validation_service/order_validation_service.py
```diff
@@ -1,8 +1,9 @@
...
 
-class OrderValidationService:
-    """Service to orchestrate the order validation."""
 
-    def __init__(self, tomte_service: TomteOrderValidationService):
-        self.tomte_service = tomte_service
+class OrderValidationService(ABC):
+    @abstractmethod
+    def validate(self, order_json: str) -> ValidationError:
+        pass
```
ChrOertlin, line 9: Just a question in regards to the endpoint the ordervalidationservice will be used.

How would that look like? Is the OrderValidationService called in that place, if so, should we not inject the other service here like we inject the field validator in the TomteValidationService further down? Or is this handled by the ABC?

seallard: This class is actually an abstract class so it cannot be instantiated.
I'm not sure how it will look in the end, but the endpoint will somehow get a hold of a service exposing the validation logic

cg/services/order_validation_service/workflows/tomte/validation/inter_field_validators.py
seallard: We've decided to keep interfield validation outside of the Pydantic models because it makes the models messy

seallard:
> The removal of cases from orders and samples from cases I think it warrants some discussion. It's true that cases are not created by the customers for some workflows, but we create them upon order submission. Would it not make sense to validate whatever would be created?

The models are used to parse the order form the customers are building. If a specific type of order form does not include cases, why would we add that to the data structure?

If there are rules related to cases for orders which are placed without creating cases, we can add them separately without muddying the models used to parse the order form.

PR #3415, opened 2024-07-09
Parse Pac-Bio metrics for Productivity - raw data report

diitaz93:
## Description

Closes https://github.com/Clinical-Genomics/add-new-tech/issues/47. Adds infrastructure for parsing Pac-Bio polymerase metrics

### Added

- Pydantic model for the polymerase metrics
- Constants with the attribute names of the raw-data report file
- Fixtures for raw-data report file and productivity metrics object
- Function in `utils` to divide by 1000 and keep only one decimal for changing bases into kilobases and its unit test

### Changed

- Updated test of metrics parser to work with polymerase model

### Fixed

- Metric attributes needed in kb instead of bases were changed to float and validated before assignment. Their value will only have one decimal figure (as in the SMRTlnk web interface).
- Removed unused report file fixture

### How to test

- No endpoint, no production nor automation code modified

## Review

- [x] "Merge and deploy" approved by SA

## Implementation Plan

- [ ] Deployed to stage:
```shell

```
- [ ] Deployed to production:
```shell

```


cg/constants/pacbio.py
```diff
@@ -31,3 +31,10 @@ class LoadingAttributesIDs:
     P_0: str = "loading_xml_report.productivity_0_n"
     P_1: str = "loading_xml_report.productivity_1_n"
     P_2: str = "loading_xml_report.productivity_2_n"
+
+
+class PolymeraseDataAttributeIDs:
+    MEAN_READ_LENGTH: str = "raw_data_report.read_length"
+    READ_LENGTH_N50: str = "raw_data_report.read_n50"
+    MEAN_LONGEST_SUBREAD_LENGTH: str = "raw_data_report.insert_length"
+    LONGEST_SUBREAD_LENGTH_N50: str = "raw_data_report.insert_n50"
```
seallard, line 40: I have no clue what `N50` means, is it meaningful to include that in the names? Also, maybe this class should be an enum!

cg/utils/calculations.py
```diff
@@ -1,6 +1,11 @@
 """Module to hold functions for calculations."""
 
 
+def divide_by_thousand_with_one_decimal(number: float | int) -> float:
+    """Divide a given number by a thousand."""
```
seallard, line 5: Prime example of an unnecessary doc string according to me 😉

seallard (approved):
👍

cg/utils/calculations.py
diitaz93: true

cg/constants/pacbio.py
diitaz93: https://en.wikipedia.org/wiki/N50,_L50,_and_related_statistics It is important to make the distinction between normal read length and N50 read length.

cg/utils/calculations.py
diitaz93: fixed

cg/constants/pacbio.py
diitaz93: The discussion about StrEnums, Enums and normal classes is still open and no conclusion has been reached https://github.com/Clinical-Genomics/atlas/pull/2485 I will leave this as it is for now.

PR #3436, opened 2024-07-16
Parse metrics from smrtlnk datasets file

diitaz93:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/pacbio/metrics/metrics_parser.py
```diff
@@ -23,43 +25,48 @@ def __init__(self, smrt_cell_path: Path) -> None:
         self.base_calling_report_file: Path = get_file_in_directory(
             directory=self.report_dir, file_name=PacBioDirsAndFiles.BASECALLING_REPORT
         )
-        self.hifi_metrics: HiFiMetrics = self.parse_attributes_to_model(
+        self.hifi_metrics: HiFiMetrics = self.parse_report_to_model(
```
seallard, line 28: This seems to repeat the mistakes we discussed earlier today. The parser should not accept a path in the constructor.

seallard (commented):
The `MetricsParser` class should implement an interface defined in an abstract base class. It should not accept run time data in its constructor. It should be instantiated centrally, if it is a CLI command it should be created in the cg config module.

EDIT: at least, I think the approach above is more flexible once we integrate more instruments. You are more familiar with the project, so make a judgement call.

cg/services/pacbio/metrics/metrics_parser.py
seallard: This is probably out of scope for this PR, but it seems like you are repeating the exact same design mistakes in the code for the new instruments

diitaz93: I was planning to fix that in another PR. Let's first complete all the metrics that we need

cg/services/pacbio/metrics/models.py
```diff
@@ -101,3 +107,31 @@ class PolymeraseMetrics(BaseModel):
...
+    device_internal_id: str = Field(..., alias=SmrtLinkDatabasesIDs.CELL_ID)
+    well_name: str = Field(..., alias=SmrtLinkDatabasesIDs.WELL_NAME)
+    well_sample_name: str = Field(..., alias=SmrtLinkDatabasesIDs.WELL_SAMPLE_NAME)
+    sample_internal_id: str = Field(..., alias=SmrtLinkDatabasesIDs.BIO_SAMPLE_NAME)
+    movie_name: str = Field(..., alias=SmrtLinkDatabasesIDs.MOVIE_NAME)
+    cell_index: int = Field(..., alias=SmrtLinkDatabasesIDs.CELL_INDEX)
+    path: str = Field(..., alias=SmrtLinkDatabasesIDs.PATH)
+
+    @property
+    def plate(self) -> int:
```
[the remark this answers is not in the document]
Vince-janv: Since this class is supposed to only hold data I prefer to not add this kind of logic here, rather set it using a before_validator

cg/services/pacbio/metrics/models.py
```diff
@@ -101,3 +107,31 @@ class PolymeraseMetrics(BaseModel):
...
+        """
+        Extract the plate number from the plate_well name in the path.
+        Example:
+            /1_A01/ -> plate = 1
+            /2_D01/ -> plate = 2
+        """
+        match = re.search(r"/(\d)_", self.path)
+        if not match:
+            raise ValidationError("Path should contain the plate-well name")
+        plate = int(match.group(1))
```
Vince-janv, line 134: This seems quite sketchy. If we search the whole path I think we should have a more exact regexp

tests/fixture_plugins/pacbio_fixtures/metrics_fixtures.py
```diff
@@ -61,3 +63,17 @@ def pac_bio_polymerase_metrics() -> PolymeraseMetrics:
...
     }
     return PolymeraseMetrics.model_validate(data, from_attributes=True)
+
+
+@pytest.fixture
+def pac_bio_smrtlink_databases_metrics() -> SmrtlinkDatasetsMetrics:
+    data: dict[str, Any] = {
+        SmrtLinkDatabasesIDs.BIO_SAMPLE_NAME: "1247014000119",
+        SmrtLinkDatabasesIDs.CELL_ID: "EA094834",
+        SmrtLinkDatabasesIDs.CELL_INDEX: 0,
```
Vince-janv, line 73: What is the cell_index?

diitaz93: The position of the smrt cell in the sequencing tray. It can hold 4 smart cells positioned 0, 1, 2 and 3

cg/services/pacbio/metrics/models.py
diitaz93: I also though about this:
```python
@property
    def plate(self) -> int:
        """
        Extract the plate number from the plate_well name in the path.
        Example:
            /1_A01/ -> plate = 1
            /2_D01/ -> plate = 2
        """
        path_parts: list[str] = self.path.split("/")
        plate_well: str = [part for part in path_parts if self.well_name in part][0]
        plate = int(plate_well[0])
        if plate not in [1, 2]:
            raise ValidationError("Value of 'plate' should be 1 or 2")
        return plate
```

cg/services/pacbio/metrics/models.py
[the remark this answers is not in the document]
diitaz93: updated with model validator

cg/services/pacbio/metrics/models.py
diitaz93: updated to model validator according to documentation

PR #3437, opened 2024-07-16
Add well validation

seallard:
If an order contains a plate, each well on the plate is only allowed to contain a single sample. This PR adds that validation rule for tomte orders.

Also restructured the error classes a bit to be able to differentiate errors originating from order, case, sample and case sample fields.


cg/services/order_validation_service/workflows/tomte/validation/inter_field_validators.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+def _get_colliding_samples(
+    samples_with_cases: list[tuple[TomteSample, str]]
+) -> list[tuple[TomteSample, str]]:
+    colliding_samples = []
+    sample_well_map = _get_sample_well_map(samples_with_cases)
+    for _, well_samples in sample_well_map.items():
+        if len(well_samples) > 1:
+            colliding_samples.extend(well_samples[1:])
```
diitaz93, line 40: It is unclear why you only choose the last samples. Shouldn't you choose all samples that are in one well? I guess the order is arbitrary

cg/services/order_validation_service/workflows/tomte/validation/inter_field_validators.py
```diff
@@ -0,0 +1,52 @@
+from cg.models.orders.sample_base import ContainerEnum
+from cg.services.order_validation_service.models.errors import OccupiedWellError, ValidationError
+from cg.services.order_validation_service.workflows.tomte.models.order import TomteOrder
+from cg.services.order_validation_service.workflows.tomte.models.sample import TomteSample
+from cg.services.order_validation_service.workflows.tomte.models.case import TomteCase
+
+
+def validate_wells_contain_at_most_one_sample(order: TomteOrder) -> list[OccupiedWellError]:
```
diitaz93, line 8: This module should have tests

cg/services/order_validation_service/workflows/tomte/validation/inter_field_validators.py
seallard: The reason is that a well can contain one sample but not more. So one of the samples is allowed to be in the well, but all others need to be moved to other wells.

seallard: Extracted it to a variable with a descriptive name https://github.com/Clinical-Genomics/cg/pull/3437/commits/82c8160c4929ee649237c44d905e37e58f2c2318

PR #3443, opened 2024-07-18
feat - add second SMRTcell fixture and files

diitaz93:
## Description
- Changes the structure of the reports in the fixtures to include the zipped reports folder
- Adds the new ccs report file to the fixture dir
- Fixtures

### How to test

- [No endpoint, no production code modified

## Review

- [x] "Merge and deploy" approved by SA

## Implementation Plan

- [ ] Deployed to stage:
```shell

```
- [ ] Deployed to production:
```shell

```


cg/models/run_devices/pacbio_cell_directory_data.py
```diff
@@ -0,0 +1,20 @@
...
+        self.path = pac_bio_run_directory
+        self.statistics_dir = Path(self.path, "statistics")
+
+    @property
+    def ccs_report(self) -> Path:
+        """Finds the json file finishing with ccs_report.json in the statistics dir."""
+        files: list[Path] = get_files_in_directory_with_pattern(
+            directory=self.statistics_dir, pattern=PacBioDirsAndFiles.CCS_REPORT_SUFFIX
+        )
+        return files[0]
```
ChrOertlin: I think if we start putting the files as property on this model we will end up with the same that happened with the IlluminaRunDirectoryData, which is totally overloaded and lost its single responsibility.

So what should this class achieve? It should collect information but here in this property it returns a file. Should it not just return the expected path? And then getting the file should be done somewhere else?

diitaz93: Fixing in pair programming session

cg/models/run_devices/pacbio_cell_directory_data.py
```diff
@@ -0,0 +1,20 @@
+from pathlib import Path
+
+from cg.constants.pacbio import PacBioDirsAndFiles
+from cg.utils.files import get_files_in_directory_with_pattern
+
+
+class PacBioCellDirectoryData:
```
seallard, line 7: I find this naming confusing, it sounds like the class will encapsulate data but it is just paths

cg/models/run_devices/pacbio_cell_directory_data.py
```diff
@@ -0,0 +1,20 @@
...
+        self.path = pac_bio_run_directory
+        self.statistics_dir = Path(self.path, "statistics")
+
+    @property
+    def ccs_report(self) -> Path:
+        """Returns the CCS report file path."""
+        files: list[Path] = get_files_in_directory_with_pattern(
+            directory=self.statistics_dir, pattern=PacBioDirsAndFiles.CCS_REPORT_SUFFIX
+        )
+        return files[0]
```
seallard, line 20: What happens if no files are found? Consider catching an exception, logging an error message and raising a more descriptive exception.

seallard (approved):
👍

diitaz93:
Changed implementation, closing

cg/models/run_devices/pacbio_cell_directory_data.py
diitaz93: removed

cg/models/run_devices/pacbio_cell_directory_data.py
diitaz93: removed

tests/models/pacbio/test_pac_bio_cell_directory_data.py
```diff
@@ -0,0 +1,33 @@
...
+):
+    """Test that the correct ccs report file is returned."""
+    # GIVEN a PacBioCellDirectoryData object
+
+    # WHEN accessing the ccs_report property
+    ccs_report: Path = pac_bio_cell_directory_data_1_a01.ccs_report
+
+    # THEN the correct ccs report file is returned
+    assert ccs_report.name == ccs_report_1_a01_name
+    assert ccs_report.exists()
```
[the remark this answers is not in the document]
diitaz93: removed

PR #3447, opened 2024-07-18
Fixed percentage calculations in productivity metrics

diitaz93:
## Description
Changed the calculation of percentage values in PacBio Productivity metrics model from a property to a model validator. This to be more "pydantic" and so that these values are exported when dumping into dictionaries

### How to test

- [x] No production code touched

## Review

- [ ] "Merge and deploy" approved by SA

## Implementation Plan

- [ ] Deployed to stage:
```shell

```
- [ ] Deployed to production:
```shell

```

### DIFF
--- a/cg/services/pacbio/metrics/models.py
+++ b/cg/services/pacbio/metrics/models.py
@@ -63,24 +63,22 @@ class ProductivityMetrics(BaseModel):
     p_0: int = Field(..., alias=LoadingAttributesIDs.P_0)
     p_1: int = Field(..., alias=LoadingAttributesIDs.P_1)
     p_2: int = Field(..., alias=LoadingAttributesIDs.P_2)
+    percentage_p_0: float
+    percentage_p_1: float
+    percentage_p_2: float
 
-    @property
-    def percentage_p_0(self) -> float:
-        return self._calculate_percentage(self.p_0)
-
-    @property
-    def percentage_p_1(self) -> float:
-        return self._calculate_percentage(self.p_1)
-
-    @property
-    def percentage_p_2(self) -> float:
-        return self._calculate_percentage(self.p_2)
-
-    def _calculate_percentage(self, value: int) -> float:
-        """Calculates the percentage of a value to productive_zmws."""
-        if self.productive_zmws == 0:
-            return 0.0
-        return round((value / self.productive_zmws) * 100, 0)
+    @model_validator(mode="before")
+    @classmethod
+    def set_percentages(cls, data: Any):
+        if isinstance(data, dict):
+            productive_zmws = data.get(LoadingAttributesIDs.PRODUCTIVE_ZMWS)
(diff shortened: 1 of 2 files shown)


cg/services/pacbio/metrics/models.py
```diff
@@ -63,24 +63,22 @@ class ProductivityMetrics(BaseModel):
     p_0: int = Field(..., alias=LoadingAttributesIDs.P_0)
     p_1: int = Field(..., alias=LoadingAttributesIDs.P_1)
     p_2: int = Field(..., alias=LoadingAttributesIDs.P_2)
+    percentage_p_0: float
```
seallard, line 66: Can these be given more descriptive names?

diitaz93: These are the names of the PacBio attributes that we want to Parse. We agreed not to include unnecessary biological context at this level as we are only parsing data
Screenshot 2024-07-18 at 16 29 30

cg/services/pacbio/metrics/models.py
```diff
@@ -63,24 +63,22 @@ class ProductivityMetrics(BaseModel):
...
+    @classmethod
+    def set_percentages(cls, data: Any):
+        if isinstance(data, dict):
+            productive_zmws = data.get(LoadingAttributesIDs.PRODUCTIVE_ZMWS)
+            p_0 = data.get(LoadingAttributesIDs.P_0)
+            p_1 = data.get(LoadingAttributesIDs.P_1)
+            p_2 = data.get(LoadingAttributesIDs.P_2)
+            data["percentage_p_0"] = round((p_0 / productive_zmws) * 100, 0)
+            data["percentage_p_1"] = round((p_1 / productive_zmws) * 100, 0)
+            data["percentage_p_2"] = round((p_2 / productive_zmws) * 100, 0)
```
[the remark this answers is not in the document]
seallard: I guess it can make sense to normalize values

seallard (approved):
Not a big fan of the `p_0`, `p_1` naming 😄 Otherwise looks good, a general design thing we can think about is how much logic to keep in the Pydantic models for transforming the data. I think what is done here makes sense since we just scale it

cg/services/pacbio/metrics/models.py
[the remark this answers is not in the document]
diitaz93: We are not normalising the original values, just adding new fields for the normalised values. I don't see why it can't be done in the same step. That is why pydantic offers the model validator. This approach is also use below in line 122

cg/services/pacbio/metrics/models.py
seallard: I think the alias decouples us from having to name the field the same way. I think having the domain language in the 
models is a good thing, better than `p_0` etc. Maybe you can discuss it with fellow project members :)

PR #3450, opened 2024-07-19
Add Chanjo2 API

ivadym:
## Description

Context: taking over https://github.com/Clinical-Genomics/cg/pull/3267 

### Added
- Chanjo2 API client
- Chanjo2 get coverage endpoint

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


ChrOertlin (commented):
Looks good minor comments

cg/clients/chanjo2/api.py
```diff
@@ -0,0 +1,40 @@
...
+LOG = logging.getLogger(__name__)
+
+
+class Chanjo2APIClient:
+    """
+    Chanjo2 API to communicate with a d4tools software in order to return coverage
+    over genomic intervals.
+    """
+
+    def __init__(self, config: dict[str, str]):
```
seallard, line 19: Don't pass the config like this in the constructor, provide what is needed explicitly so it is clear what the client needs. Seems like it needs a `base_url` or whatever `host` refers to 😄

cg/clients/chanjo2/api.py
```diff
@@ -0,0 +1,40 @@
...
+            "Content-Type": "application/json",
+            "accept": "application/json",
+        }
+        self.host = config["chanjo2"]["host"]
+
+    def get_coverage(self, coverage_request: CoverageRequest) -> CoverageData | None:
+        """Send a POST request to the coverage endpoint to retrieve gene coverage summary data."""
+        endpoint: str = f"{self.host}/coverage/d4/genes/summary"
+        post_data: dict[str, Any] = coverage_request.model_dump()
+        try:
```
seallard, line 30: We have found a nice pattern for removing these repeated try/catches from code like this 🥳 Basically, you can define a decorator which does the error handling, see this example [Tower API client](https://github.com/Clinical-Genomics/trailblazer/blob/648eeca3683407fd8e1ff75c29f256f0c712d01d/trailblazer/clients/tower/tower_client.py) and the decorator used for error handling [here](https://github.com/Clinical-Genomics/trailblazer/blob/648eeca3683407fd8e1ff75c29f256f0c712d01d/trailblazer/clients/tower/utils.py).

cg/clients/chanjo2/api.py
```diff
@@ -0,0 +1,40 @@
...
+            response = requests.post(url=endpoint, headers=self.headers, json=post_data)
+            response.raise_for_status()
+            response_content: dict[str, Any] = response.json().values()
```
seallard, line 33: It is better to directly parse the response json into a Pydantic model like this [example](https://github.com/Clinical-Genomics/trailblazer/blob/648eeca3683407fd8e1ff75c29f256f0c712d01d/trailblazer/clients/tower/tower_client.py#L28)

cg/clients/chanjo2/api.py
```diff
@@ -0,0 +1,40 @@
...
+            if not response_content:
+                LOG.error("The POST get coverage response is empty")
+                return None
+            return CoverageData(**next(iter(response_content)))
```
seallard, line 37: What is the next and iter stuff doing? Not familiar with it! If you use a Pydantic model and parse the response directly into the model as described above I think this can be skipped

seallard: I'm pretty sure you can just do this instead
```python
json = response.json()
return CoverageData.model_validate(json)
```

tests/clients/chanjo2/test_chanjo2_api_client.py
```diff
@@ -0,0 +1,69 @@
...
+
+    # GIVEN a mocked POST request returning an empty JSON
+    mocker.patch.object(requests, "post", return_value=coverage_post_response_empty)
+
+    # WHEN getting the coverage data
+    sample_coverage: CoverageData | None = chanjo2_api_client.get_coverage(coverage_request)
+
+    # THEN an empty data error should have been thrown
+    assert sample_coverage is None
+    assert "The POST get coverage response is empty" in caplog.text
```
seallard, line 69: Might be better to check that a specific exception is raised

seallard (commented):
I think this can be made a bit cleaner, made some suggestions! But looks great anyway 👍

cg/clients/chanjo2/api.py
```diff
@@ -0,0 +1,40 @@
...
+            "accept": "application/json",
+        }
+        self.host = config["chanjo2"]["host"]
+
+    def get_coverage(self, coverage_request: CoverageRequest) -> CoverageData | None:
+        """Send a POST request to the coverage endpoint to retrieve gene coverage summary data."""
+        endpoint: str = f"{self.host}/coverage/d4/genes/summary"
+        post_data: dict[str, Any] = coverage_request.model_dump()
+        try:
+            response = requests.post(url=endpoint, headers=self.headers, json=post_data)
```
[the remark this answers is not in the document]
seallard: I guess it also depends on the chanjo API, if it is built wrong then it could be the case that it has to be a post request 🤢 If that is the case and we maintain the chanjo API, we should patch it to be a get request.

cg/clients/chanjo2/api.py
```diff
@@ -0,0 +1,40 @@
...
+            response.raise_for_status()
+            response_content: dict[str, Any] = response.json().values()
+            if not response_content:
+                LOG.error("The POST get coverage response is empty")
+                return None
+            return CoverageData(**next(iter(response_content)))
+        except (requests.RequestException, ValueError) as error:
+            LOG.error(f"Error during coverage POST request: {error}")
+            return None
```
seallard, line 40: Better to raise a custom exception here! I think this is something exceptional 😉

seallard: Consider creating a `ChanjoAPIClientException` base exception and other more specific ones from it. That way, any user of the client will easily catch any exceptions originating from the client

cg/clients/chanjo2/api.py
```diff
@@ -0,0 +1,40 @@
...
+        """Send a POST request to the coverage endpoint to retrieve gene coverage summary data."""
+        endpoint: str = f"{self.host}/coverage/d4/genes/summary"
+        post_data: dict[str, Any] = coverage_request.model_dump()
+        try:
+            response = requests.post(url=endpoint, headers=self.headers, json=post_data)
+            response.raise_for_status()
+            response_content: dict[str, Any] = response.json().values()
+            if not response_content:
+                LOG.error("The POST get coverage response is empty")
+                return None
```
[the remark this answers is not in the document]
seallard: Yes, also avoid returning None when something weird happens. This should be an exception and it would be thrown when trying to parse the data into the Pydantic model.

cg/clients/chanjo2/api.py
```diff
@@ -0,0 +1,40 @@
+"""Coverage analysis API for human clinical sequencing data."""
```
seallard, line 1: Consider renaming this file to `client.py` or `chanjo_api_client.py` or something

cg/clients/chanjo2/api.py
ivadym: makes sense, thanks!

cg/clients/chanjo2/api.py
ivadym: I will give it a try, looks clean!

cg/clients/chanjo2/api.py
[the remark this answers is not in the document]
ivadym: The problem is that we can't allow this method to raise an error since we need to be able to force the generation of the delivery report even if the request fails. But I will move this logic to the Analysis API so that it is independent of the client API

cg/clients/chanjo2/api.py
ivadym: The response we get is in the format {sample_id: {coverage: ..., completeness: ...}}. I want to get rid of the sample ID, which is why I extract the first value with the iterator

ivadym: Actually the request seems odd for a specific sample as well as its return, but since the discussion in https://github.com/Clinical-Genomics/cg/pull/3267 have been going on for two months, it might be time to move forward with what we have 🥵

cg/clients/chanjo2/api.py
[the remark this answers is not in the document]
seallard: It sounds inefficient to send such a list. Would it be possible to send a panel id instead? This is definitely out of scope and it sounds like there is a lot to clean up in https://github.com/Clinical-Genomics/chanjo2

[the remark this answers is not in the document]
seallard: I'll start creating issues in this repo, we can have stuff like this going on

cg/clients/chanjo2/api.py
seallard: Assuming that the API actually provides valid JSON data which it does not seem to do?

cg/clients/chanjo2/api.py
seallard: I think it is better to incorporate that in the pydantic model so we have an exact map of the actual response data structure. The consumer of the client will have to discard whatever data is not needed.

cg/clients/chanjo2/api.py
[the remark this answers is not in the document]
seallard: I'll create issues in that repo to track these problems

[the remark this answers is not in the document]
ivadym: That might be complicated since these panels are specific to our use case. But starting a discussion with an issue could be a good approach. Thanks!

cg/clients/chanjo2/client.py
```diff
@@ -0,0 +1,37 @@
...
+
+import logging
+from typing import Any
+
+import requests
+
+from cg.clients.chanjo2.models import CoveragePostRequest, CoveragePostResponse
+from cg.clients.chanjo2.utils import handle_client_errors
+
+LOG = logging.getLogger(__name__)
```
seallard, line 11: I guess all logging is done in the error handler now, so this can be removed? 😄

seallard (approved):
Nice, this looks great 🚀

cg/clients/chanjo2/client.py
ivadym: Good catch! 

And thanks for the review, very useful suggestions 🙏

ivadym:
Deployed with https://github.com/Clinical-Genomics/cg/pull/3459

PR #3451, opened 2024-07-19
Example refactoring

seallard:
Refactoring applied

- Declutter parser class by extracting class methods to utils module
- Declutter parser class by removing setter methods
- Declutter parser class by not storing paths on class

Additionally
- Restructure pydantic model for pac bio metrics

The last point might be more in your domain, but I think it makes sense to keep the different metrics in their respective models.

### DIFF
--- a/cg/services/pacbio/metrics/metrics_parser.py
+++ b/cg/services/pacbio/metrics/metrics_parser.py
@@ -1,11 +1,5 @@
 from pathlib import Path
-from typing import Any, Type
-
-from cg.constants.constants import FileFormat
-from cg.constants.pacbio import PacBioDirsAndFiles
-from cg.io.controller import ReadFile
 from cg.services.pacbio.metrics.models import (
-    BaseMetrics,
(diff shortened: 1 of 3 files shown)


cg/services/pacbio/metrics/metrics_parser.py
```diff
@@ -1,72 +1,37 @@
...
+    parse_control_metrics,
+    parse_hifi_metrics,
+    parse_polymerase_metrics,
+    parse_productivity_metrics,
+    parse_dataset_metrics,
+)
 
 
-class MetricsParser:
+class PacBioMetricsParser:
```
seallard, line 19: Anyone going into this module does not what to be overwhelmed by details and lots of helper methods in the class. It is better to separate them. It is also easier to follow the control flow if you don't store calculated paths on the class.

diitaz93:
## Implementing suggestion in branch

diitaz93 (approved):

PR #3458, opened 2024-07-23
add pacbio run data generator

ChrOertlin:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/utils/string.py
```diff
@@ -0,0 +1,5 @@
+"""Utils related to str manipulation."""
+
+
+def get_element_from_split(value: str, element_position: int, split: str) -> str:
```
diitaz93, line 4:
```suggestion
def get_element_from_split(value: str, element_position: int, split: str) -> str:
    """Split a string on a given separator and return the split element given by the position. """
```

cg/utils/string.py
```diff
@@ -0,0 +1,5 @@
+"""Utils related to str manipulation."""
```
diitaz93, line 1:
```suggestion
"""Utils related to string manipulation."""
```

cg/services/post_processing/exc.py
```diff
@@ -0,0 +1,5 @@
+from cg.exc import CgError
```
diitaz93, line 1: We could refine if we want a parent post-processing error and many errors inheriting from it, but for now it looks good

cg/utils/string.py
diitaz93: lets add a simple test function for this

diitaz93 (approved):
Looks good 👍 Let's add some tests

PR #3460, opened 2024-07-24
add PacBioRunFileManager

diitaz93:
## Description
Closes https://github.com/Clinical-Genomics/add-new-tech/issues/67
Add PacBioRunFileManager, fixture and tests


cg/services/post_processing/pacbio/run_file_manager/run_file_manager.py
```diff
@@ -0,0 +1,67 @@
...
+        statistics_dir: Path = Path(run_path, PacBioDirsAndFiles.STATISTICS_DIR)
+        files: list[Path] = get_files_matching_pattern(
+            directory=statistics_dir, pattern=f"*{PacBioDirsAndFiles.CCS_REPORT_SUFFIX}"
+        )
+        if not files:
+            raise PostProcessingFileNotFoundError(f"No CCS report file found in {statistics_dir}")
+        return files[0]
+
+    @staticmethod
+    def _get_unzipped_report_files(run_path: Path) -> list[Path]:
```
ChrOertlin, line 40:
```suggestion
    def _get_report_files(run_path: Path) -> list[Path]:
```
I don't think we need to care whether they have been zipped or not previosuly

ChrOertlin (approved):
Looks good, and seems very easy to extend

cg/services/post_processing/pacbio/run_file_manager/run_file_manager.py
diitaz93: fixed

cg/services/post_processing/pacbio/run_file_manager/run_file_manager.py
```diff
@@ -0,0 +1,67 @@
...
+from cg.services.post_processing.abstract_classes import RunFileManager
+from cg.services.post_processing.exc import PostProcessingFileNotFoundError
+from cg.services.post_processing.pacbio.run_data_generator.run_data import PacBioRunData
+from cg.services.post_processing.validators import validate_files_exist
+from cg.utils.files import get_files_matching_pattern
+
+
+class PacBioRunFileManager(RunFileManager):
+
+    def get_files_to_parse(self, run_data: PacBioRunData) -> list[Path]:
```
diitaz93, line 14: I'm thinking we could make this methods receive the full path instead of the whole RunData object @ChrOertlin

ChrOertlin: if that is all it needs, sure!

cg/constants/pacbio.py
```diff
@@ -2,13 +2,15 @@
 
 
 class PacBioDirsAndFiles:
-    ADAPTER_REPORT: str = "adapter.report.json"
     BASECALLING_REPORT: str = "ccs.report.json"
```
ChrOertlin, line 5: we don't need this anymore, right?

diitaz93: no, we will get all those metrics from the other CCS report. I will remove it in the parsing refactoring (so next PR)

cg/services/post_processing/pacbio/run_file_manager/run_file_manager.py
diitaz93: Will evaluate in another PR

PR #3467, opened 2024-07-26
refactor pacbio metrics parser

diitaz93:
## Description
Refactor PacBio metrics parser to compile with the post-processing flow

### Changed

- Condensed all individual parsers into one
- Refactored tests

### DIFF
--- a/cg/services/post_processing/abstract_classes.py
+++ b/cg/services/post_processing/abstract_classes.py
@@ -1,12 +1,10 @@
 """Post-processing service abstract classes."""
 
-from abc import abstractmethod, ABC
+from abc import ABC, abstractmethod
 from pathlib import Path
 
-from pydantic import BaseModel
-
 from cg.apps.housekeeper.hk import HousekeeperAPI
-from cg.services.post_processing.abstract_models import PostProcessingDTOs, RunMetrics, RunData
+from cg.services.post_processing.abstract_models import PostProcessingDTOs, RunData, RunMetrics
 from cg.store.store import Store
 
 
@@ -38,8 +36,9 @@ class RunFileManager(ABC):
 
 class PostProcessingMetricsParser(ABC):
 
+    @staticmethod
     @abstractmethod
-    def parse_metrics(self, metrics_paths: list[Path]) -> RunMetrics:
+    def parse_metrics(metrics_files: list[Path]) -> RunMetrics:
         pass
 
 
--- a/cg/services/post_processing/abstract_classes.py
+++ b/cg/services/post_processing/abstract_classes.py
@@ -36,9 +36,10 @@ class RunFileManager(ABC):
 
 class PostProcessingMetricsParser(ABC):
 
-    @staticmethod
-    @abstractmethod
-    def parse_metrics(metrics_files: list[Path]) -> RunMetrics:
+    def __init__(self, file_manager: RunFileManager):
+        self.file_manager: RunFileManager = file_manager
+
+    def parse_metrics(self, run_data: RunData) -> RunMetrics:
         pass
 
 
--- a/cg/services/post_processing/pacbio/metrics_parser/metrics_parser.py
+++ b/cg/services/post_processing/pacbio/metrics_parser/metrics_parser.py
@@ -13,14 +13,21 @@ from cg.services.post_processing.pacbio.metrics_parser.models import (
 from cg.services.post_processing.pacbio.metrics_parser.utils import (
     get_parsed_metrics_from_file_name,
 )
+from cg.services.post_processing.pacbio.run_data_generator.run_data import PacBioRunData
+from cg.services.post_processing.pacbio.run_file_manager.run_file_manager import (
+    PacBioRunFileManager,
+)
 
 
 class PacBioMetricsParser(PostProcessingMetricsParser):
     """Class for parsing PacBio sequencing metrics."""
 
-    @staticmethod
-    def parse_metrics(metrics_files: list[Path]) -> PacBioMetrics:
+    def __init__(self, file_manager: PacBioRunFileManager):
+        super().__init__(file_manager=file_manager)
+
+    def parse_metrics(self, run_data: PacBioRunData) -> PacBioMetrics:
         """Return all the relevant PacBio metrics parsed in a single Pydantic object."""
+        metrics_files: list[Path] = self.file_manager.get_files_to_parse(run_data)
         hifi_metrics: HiFiMetrics = get_parsed_metrics_from_file_name(
             metrics_files=metrics_files, file_name=PacBioDirsAndFiles.BASECALLING_REPORT
         )
--- a/cg/services/post_processing/pacbio/metrics_parser/metrics_parser.py
+++ b/cg/services/post_processing/pacbio/metrics_parser/metrics_parser.py
@@ -1,5 +1,6 @@
 from pathlib import Path
 
+from cg.constants.pacbio import PacBioDirsAndFiles
 from cg.services.post_processing.abstract_classes import PostProcessingMetricsParser
 from cg.services.post_processing.pacbio.metrics_parser.models import (
     ControlMetrics,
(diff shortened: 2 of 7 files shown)


cg/services/post_processing/pacbio/metrics_parser/utils.py
```diff
@@ -42,37 +48,42 @@ def parse_report_to_model(report_file: Path, data_model: Type[BaseMetrics]) -> B
 
 
 @handle_pac_bio_parsing_errors
-def parse_dataset_metrics(report_dir: Path) -> SmrtlinkDatasetsMetrics:
-    file_name = PacBioDirsAndFiles.SMRTLINK_DATASETS_REPORT
-    report: Path = get_file_in_directory(directory=report_dir, file_name=file_name)
-    parsed_json: dict = ReadFile.read_file[FileFormat.JSON](report)
+def parse_dataset_metrics(metrics_files: list[Path]) -> SmrtlinkDatasetsMetrics:
+    report_file: Path = get_report_file_from_pattern(
```
ChrOertlin, line 52: If we add another function:

```Python
get_data_model_for_pattern()
```

Then we can handle all files /models in one function. What do you think?

ChrOertlin (approved):

cg/services/post_processing/pacbio/metrics_parser/utils.py
diitaz93: Good idea, implementing to see how it looks

cg/utils/files.py
```diff
@@ -23,6 +23,18 @@ def get_file_in_directory(directory: Path, file_name: str) -> Path:
     raise FileNotFoundError(f"File {file_name} not found in {directory}")
 
 
+def get_file_with_pattern_from_list(files: list[Path], pattern: str) -> Path | None:
```
diitaz93, line 26: Will add a unit test for this

cg/services/post_processing/pacbio/metrics_parser/utils.py
```diff
@@ -32,47 +32,30 @@ def wrapper(*args, **kwargs):
...
+        PacBioDirsAndFiles.SMRTLINK_DATASETS_REPORT: SmrtlinkDatasetsMetrics,
+    }
+    return pattern_to_model.get(pattern)
+
+
+def _parse_report_to_model(report_file: Path, data_model: Type[BaseMetrics]) -> BaseMetrics:
     """Parse the metrics report to a data model."""
-    parsed_json: dict = ReadFile.read_file[FileFormat.JSON](file_path=report_file)
+    parsed_json: dict = ReadFile.read_file[FileFormat.JSON](report_file)
+    if data_model == SmrtlinkDatasetsMetrics:
```
diitaz93, line 50: @ChrOertlin The only downside of this approach is that there is less flexibility when adding a new file to parse if the file has a different structure, but other than that it encapsulates a lot of logic

cg/services/post_processing/abstract_classes.py
```diff
@@ -38,8 +36,9 @@ def get_files_to_store(self, run_data: RunData) -> list[Path]:
 
 class PostProcessingMetricsParser(ABC):
 
+    @staticmethod
     @abstractmethod
-    def parse_metrics(self, metrics_paths: list[Path]) -> RunMetrics:
+    def parse_metrics(metrics_files: list[Path]) -> RunMetrics:
```
seallard, line 41: Can we assume that all metrics will share at least one meaningful common parent directory? If so, maybe it is better to encapsulate the logic for identifying the individual report files in the concrete implementation of the parser

cg/services/post_processing/pacbio/metrics_parser/utils.py
seallard: I think it is also a bit more difficult to understand what is going on, but sure it removes some duplication 😄

seallard (commented):
Not sure about passing a list of paths, it makes the interface for the metrics parser leaky since suddenly the client needs to have logic for finding the different paths and passing them into the parser.

I think it is better to encapsulate the logic for identifying the individual report files in the metrics parser.

cg/services/post_processing/abstract_classes.py
seallard: It makes the interface for the metrics parser class leaky. Code using the parser will need to have logic for finding the paths and I think it would be better to encapsulate that in the metrics parser class.

cg/services/post_processing/pacbio/metrics_parser/utils.py
ChrOertlin: You can get the function and the model from a map via the file name. This should not need any IF statement

cg/services/post_processing/abstract_classes.py
ChrOertlin: There is a RunFilemanager that handles the path constructions. It is not part of this PR.

cg/services/post_processing/abstract_classes.py
```diff
@@ -38,8 +36,9 @@ def get_files_to_store(self, run_data: RunData) -> list[Path]:
 
 class PostProcessingMetricsParser(ABC):
 
+    @staticmethod
```
ChrOertlin, line 39:
```suggestion
    def __init__(self, file_manager: RunFileManager):
    self.file_manager: RunFileManager = file_manager
```

cg/services/post_processing/pacbio/metrics_parser/metrics_parser.py
```diff
@@ -9,26 +11,31 @@
...
-    parse_hifi_metrics,
-    parse_polymerase_metrics,
-    parse_productivity_metrics,
+    get_parsed_metrics_from_file_name,
 )
 
 
-class PacBioMetricsParser:
+class PacBioMetricsParser(PostProcessingMetricsParser):
     """Class for parsing PacBio sequencing metrics."""
```
ChrOertlin, line 20: def __init__(
        self,
        file_manager: PacBioRunFileManager,
    ):
        super().__init__(file_manager=file_manager)

cg/services/post_processing/pacbio/metrics_parser/metrics_parser.py
```diff
@@ -9,26 +11,31 @@
...
 
     @staticmethod
-    def parse_metrics(smrt_cell_path: Path) -> PacBioMetrics:
+    def parse_metrics(metrics_files: list[Path]) -> PacBioMetrics:
```
ChrOertlin, line 22:
```suggestion
    def parse_metrics() -> PacBioMetrics:
    metrics_files: list[Path] = self.file_manager.get_files_to_parse()
```

cg/services/post_processing/abstract_classes.py
seallard: Ah, yeah maybe the metrics parser should just accept that in the constructor and handle it internally still

cg/services/post_processing/pacbio/metrics_parser/utils.py
diitaz93: @ChrOertlin  All the other reports have the data we want in a section of the JSON called `attributes`. Those JSON files have the data of the plots, so they have some other figure metadata This smrlink dataset is not a file containing data shown in any plot, it seems to hold run metadata, and it does not have this structure

cg/services/post_processing/pacbio/metrics_parser/metrics_parser.py
diitaz93: implemented

cg/services/post_processing/abstract_classes.py
diitaz93: implemented

cg/services/post_processing/abstract_classes.py
diitaz93: implemented

cg/utils/files.py
diitaz93: done

cg/services/post_processing/pacbio/metrics_parser/utils.py
diitaz93: Lets leave this as it is and if we get rouble further down we can change it

cg/services/post_processing/pacbio/metrics_parser/metrics_parser.py
diitaz93: implemented

PR #3469, opened 2024-07-26
PacBio - implement parsing of failed reads metrics

diitaz93:
## Description
Closes https://github.com/Clinical-Genomics/add-new-tech/issues/64
The read metrics we parsed came from a file containing only HiFi data. It is important to parse the failed read metrics too. There is a file containing both HiFi and failed metrics (`m84202_240522_135641_s1.ccs_report.json`). It is, however, generated by **another software** so the values are not exactly the same as the metrics parsed before. @J35P312 assured that the difference was negligible. 

### Added

- New parameters to parse from ccs file:
  - [x] <Q20 Reads
  - [x] <Q20 Yield (bp)
  - [x] <Q20 Read Length (mean, bp)

### Changed

- Renamed `HiFiMetrics` model to `ReadMetrics`
- The path to the ccs file

### Fixed

- Removed old ccs file usage and fixture


ChrOertlin (approved):
Nice!

PR #3477, opened 2024-07-29
Add PacBio Data Transfer service

diitaz93:
## Description
Closes https://github.com/Clinical-Genomics/add-new-tech/issues/69
Add PacBio data transfer service

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/post_processing/pacbio/data_transfer_service/data_transfer_service.py
+++ b/cg/services/post_processing/pacbio/data_transfer_service/data_transfer_service.py
@@ -12,6 +12,7 @@ from cg.services.post_processing.pacbio.data_transfer_service.utils import (
 )
 from cg.services.post_processing.pacbio.metrics_parser.metrics_parser import PacBioMetricsParser
 from cg.services.post_processing.pacbio.metrics_parser.models import PacBioMetrics
+from cg.services.post_processing.pacbio.run_data_generator.run_data import PacBioRunData
 
 
 class PacBioDataTransferService(PostProcessingDataTransferService):
@@ -19,8 +20,8 @@ class PacBioDataTransferService(PostProcessingDataTransferService):
     def __init__(self, metrics_service: PacBioMetricsParser):
         super().__init__(metrics_service=metrics_service)
 
-    def get_post_processing_dtos(self) -> PacBioDTOs:
-        metrics: PacBioMetrics = self.metrics_service.parse_metrics()
+    def get_post_processing_dtos(self, run_data: PacBioRunData) -> PacBioDTOs:
+        metrics: PacBioMetrics = self.metrics_service.parse_metrics(run_data)
         smrt_cell_dto: PacBioSMRTCellDTO = get_smrt_cell_dto(metrics)
         sequencing_run_dto: PacBioSequencingRunDTO = get_sequencing_run_dto(metrics)
         sample_sequencing_metrics_dtos: list[PacBioSampleSequencingMetricsDTO] = (
--- a/cg/services/post_processing/pacbio/data_transfer_service/data_transfer_service.py
+++ b/cg/services/post_processing/pacbio/data_transfer_service/data_transfer_service.py
@@ -1,0 +1,33 @@
+from cg.services.post_processing.abstract_classes import PostProcessingDataTransferService
+from cg.services.post_processing.pacbio.data_transfer_service.dto import (
+    PacBioDTOs,
+    PacBioSampleSequencingMetricsDTO,
+    PacBioSequencingRunDTO,
+    PacBioSMRTCellDTO,
+)
+from cg.services.post_processing.pacbio.data_transfer_service.utils import (
+    get_sample_sequencing_metrics_dtos,
+    get_sequencing_run_dto,
+    get_smrt_cell_dto,
+)
+from cg.services.post_processing.pacbio.metrics_parser.metrics_parser import PacBioMetricsParser
+from cg.services.post_processing.pacbio.metrics_parser.models import PacBioMetrics
+
+
+class PacBioDataTransferService(PostProcessingDataTransferService):
+
+    def __init__(self, metrics_service: PacBioMetricsParser):
+        super().__init__(metrics_service=metrics_service)
+
+    def get_post_processing_dtos(self) -> PacBioDTOs:
+        metrics: PacBioMetrics = self.metrics_service.parse_metrics()
+        smrt_cell_dto: PacBioSMRTCellDTO = get_smrt_cell_dto(metrics)
+        sequencing_run_dto: PacBioSequencingRunDTO = get_sequencing_run_dto(metrics)
+        sample_sequencing_metrics_dtos: list[PacBioSampleSequencingMetricsDTO] = (
+            get_sample_sequencing_metrics_dtos(metrics)
+        )
+        return PacBioDTOs(
+            run_device=smrt_cell_dto,
+            sequencing_run=sequencing_run_dto,
+            sample_sequencing_metrics=sample_sequencing_metrics_dtos,
+        )
--- a/cg/services/post_processing/pacbio/data_transfer_service/utils.py
+++ b/cg/services/post_processing/pacbio/data_transfer_service/utils.py
@@ -1,0 +1,63 @@
+from cg.constants.devices import DeviceType
+from cg.services.post_processing.pacbio.data_transfer_service.dto import (
+    PacBioSampleSequencingMetricsDTO,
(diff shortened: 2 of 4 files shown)


cg/services/post_processing/pacbio/data_transfer_service/utils.py
```diff
@@ -0,0 +1,63 @@
...
+
+def get_smrt_cell_dto(metrics: PacBioMetrics) -> PacBioSMRTCellDTO:
+    internal_id: str = metrics.dataset_metrics.cell_id
+    return PacBioSMRTCellDTO(type=DeviceType.PACBIO, internal_id=internal_id)
+
+
+def get_sequencing_run_dto(metrics: PacBioMetrics) -> PacBioSequencingRunDTO:
+    hifi_mean_read_quality: str = f"Q{metrics.read.hifi_median_read_quality}"
+    return PacBioSequencingRunDTO(
+        type=DeviceType.PACBIO,
```
islean, line 18: Feel free to disregard but feels off that we would need to specify this in a DTO object whose name already contains "PacBio"

cg/services/post_processing/pacbio/data_transfer_service/utils.py
```diff
@@ -0,0 +1,63 @@
...
+        polymerase_mean_longest_subread=metrics.polymerase.mean_longest_subread_length,
+        polymerase_longest_subread_n50=metrics.polymerase.longest_subread_length_n50,
+        control_reads=metrics.control.reads,
+        control_mean_read_length=metrics.control.mean_read_length_kb,
+        control_mean_read_concordance=metrics.control.percent_mean_concordance_reads,
+        control_mode_read_concordance=metrics.control.percent_mode_concordance_reads,
+        failed_reads=metrics.read.failed_reads,
+        failed_yield=metrics.read.failed_yield,
+        failed_mean_read_length=metrics.read.failed_mean_read_length_kb,
+        movie_name=metrics.dataset_metrics.movie_name,
```
islean, line 44: Could we have these in alphabetical order? 🥺

cg/services/post_processing/pacbio/data_transfer_service/utils.py
```diff
@@ -0,0 +1,63 @@
...
+    sample_sequencing_metrics_dto = PacBioSampleSequencingMetricsDTO(
+        hifi_reads=metrics.read.hifi_reads,
+        hifi_yield=metrics.read.hifi_yield,
+        hifi_mean_read_length=metrics.read.hifi_mean_read_length_kb,
+        hifi_median_read_length=metrics.read.hifi_median_read_length,
+        hifi_median_read_quality=hifi_mean_read_quality,
+        percent_reads_passing_q30=metrics.read.percent_q30,
+        failed_reads=metrics.read.failed_reads,
+        failed_yield=metrics.read.failed_yield,
+        failed_mean_read_length=metrics.read.failed_mean_read_length_kb,
```
islean, line 61: same here?

cg/services/post_processing/pacbio/data_transfer_service/utils.py
seallard, line 44: Consider keeping some nestling in this metrics object if the different metrics fall into clearly distinguishable categories

cg/services/post_processing/pacbio/data_transfer_service/data_transfer_service.py
```diff
@@ -0,0 +1,34 @@
...
+)
+from cg.services.post_processing.pacbio.metrics_parser.metrics_parser import PacBioMetricsParser
+from cg.services.post_processing.pacbio.metrics_parser.models import PacBioMetrics
+from cg.services.post_processing.pacbio.run_data_generator.run_data import PacBioRunData
+
+
+class PacBioDataTransferService(PostProcessingDataTransferService):
+
+    def __init__(self, metrics_service: PacBioMetricsParser):
+        super().__init__(metrics_service=metrics_service)
```
seallard, line 21: Why are you calling the parent constructor like this? Seems wonky! 😄 
I'd expect the constructor to look like
```python
def __init__(self, metrics_service: PacBioMetricsParser):
    self.metrics_service = metrics_service
```

cg/services/post_processing/pacbio/data_transfer_service/utils.py
diitaz93: I think the point is to parse this as a dictionary to the database model (`DBModel(**dto.model_dump())`), so the nesting will complicate that

cg/services/post_processing/pacbio/data_transfer_service/data_transfer_service.py
```diff
@@ -0,0 +1,34 @@
...
+
+    def get_post_processing_dtos(self, run_data: PacBioRunData) -> PacBioDTOs:
+        metrics: PacBioMetrics = self.metrics_service.parse_metrics(run_data)
```
seallard, line 24: I'm guessing exceptions can be thrown from the metrics parsing, should they be caught here or just bubble up?

cg/services/post_processing/pacbio/data_transfer_service/utils.py
seallard: I'd say it is not a great idea to couple the DTO to the database model like that. 

What if a new field is added or renamed in the database model? Then that sort of implicit dump relying on the fields being named the same way would break! It is less brittle to convert it explicitly like

```python
def convert_metrics_dto(metrics: PacBioMetrics) -> DbModel:
    return DbModel(
        some_field_x = metrics.category.some_field_x_named_differently
    )
```

ChrOertlin (approved):
Looks good, we can change the constructors for all the other classes in one go later on.

cg/services/post_processing/pacbio/data_transfer_service/utils.py
diitaz93: Ok, then it makes sense the nesting

seallard (approved):
Looks good! Some concerns about how you intend to dump the DTO directly into the database model, creating a tight coupling between the two. I'd keep some nestling in the DTO because it is easier to understand what it contains that way.

Might need some error handling for the metrics parsing, unless that is done elsewhere.

Let me know if you have questions, ping me in slack or here 😄

cg/services/post_processing/pacbio/data_transfer_service/utils.py
diitaz93: This is part of the DTO functionality as this is one of the attributes we want to parse to the db for this instance.

cg/services/post_processing/pacbio/data_transfer_service/data_transfer_service.py
seallard: I read some more and it might make sense to call the parent constructor, I'll leave it to you! I guess you want to enforce a strict design of the components.

I'm more used to interfaces which work a bit differently, only specifying the expected methods to be available on the class implementing the interface. A single class can also implement multiple interfaces

ChrOertlin: We are fixing this later on.

cg/services/post_processing/pacbio/data_transfer_service/data_transfer_service.py
ChrOertlin: we are fixing error handling in a later step

cg/services/post_processing/pacbio/data_transfer_service/utils.py
diitaz93: We will consider this when fixing details in the dev branch

cg/services/post_processing/pacbio/data_transfer_service/utils.py
diitaz93: We will fix in dev branch

cg/services/post_processing/pacbio/data_transfer_service/utils.py
ChrOertlin: We are not transferring the data as a model dump but will write in explicitly. 
The reason why it is flat is because it reads a nested structure (```PacBioMetrics```) into the database structure. The DTOs are then consumed in the CRUD.

PR #3482, opened 2024-07-29
Validate unique case names

seallard:
Validate case names are unique in order.

### DIFF
--- a/cg/services/order_validation_service/models/errors.py
+++ b/cg/services/order_validation_service/models/errors.py
@@ -53,3 +53,8 @@ class OrderNameRequiredError(OrderError):
 class OccupiedWellError(CaseSampleError):
     field: str = "well_position"
     message: str = "Well is already occupied"
+
+
+class ReusedCaseNameError(CaseError):
+    field: str = "name"
+    message: str = "Duplicate case name"
--- a/cg/services/order_validation_service/models/errors.py
+++ b/cg/services/order_validation_service/models/errors.py
@@ -55,7 +55,7 @@ class OccupiedWellError(CaseSampleError):
     message: str = "Well is already occupied"
 
 
-class ReusedCaseNameError(CaseError):
+class RepeatedCaseNameError(CaseError):
     field: str = "name"
     message: str = "Duplicate case name"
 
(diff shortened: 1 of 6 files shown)


cg/services/order_validation_service/models/errors.py
```diff
@@ -55,6 +55,11 @@ class OccupiedWellError(CaseSampleError):
     message: str = "Well is already occupied"
 
 
+class RepeatedCaseNameError(CaseError):
```
ChrOertlin, line 58: nice

cg/services/order_validation_service/workflows/tomte/validation/inter_field/utils.py
```diff
@@ -53,12 +54,23 @@ def _get_sample_well_map(plate_samples_with_cases: list[tuple[TomteSample, Tomte
     return sample_well_map
 
 
-def get_duplicate_sample_names(case: TomteCase) -> list[str]:
+def get_repeated_case_names(order: TomteOrder) -> list[str]:
+    case_names = [case.name for case in order.cases]
+    count = Counter(case_names)
+    return list({name for name, freq in count.items() if freq > 1})
```
ChrOertlin, line 60:
```suggestion
    return [name for name, freq in count.items() if freq > 1]
```
Guess the same as with the sample?

ChrOertlin: I wonder can that function be generalised so we only need 1 for both case and sample?
Even more, can you imagine this function to be required for other workflows as well?

```Python 
get_repeated_names(tome_model: TomteCase | TomteSample, associated_error: RepeatedCaseNameError | RepeatedSampleNameError)
```

ChrOertlin (approved):
Looks good.

cg/services/order_validation_service/workflows/tomte/validation/inter_field/utils.py
seallard: I'll keep them separate for now, that signature looks confusing! Will consider it in the future if there is a lot of duplication

ChrOertlin: The signature was just an example to make clearer what I meant. and sounds good!

PR #3484, opened 2024-07-29
Add application validation

islean:
### Added

- Validation that the provided applications exist in the database

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/models/errors.py
+++ b/cg/services/order_validation_service/models/errors.py
@@ -59,6 +59,7 @@ class ApplicationNotValidError(SampleError):
     field: str = "application"
     message: str = "Application is not valid"
 
+
 class RepeatedCaseNameError(CaseError):
     field: str = "name"
     message: str = "Case name already used"
--- a/cg/services/order_validation_service/models/errors.py
+++ b/cg/services/order_validation_service/models/errors.py
@@ -55,7 +55,7 @@ class OccupiedWellError(CaseSampleError):
     message: str = "Well is already occupied"
 
 
-class ApplicationNotValidError(SampleError):
+class ApplicationNotValidError(CaseSampleError):
     field: str = "application"
     message: str = "Application is not valid"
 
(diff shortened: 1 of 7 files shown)


cg/services/order_validation_service/workflows/tomte/validation_rules.py
```diff
@@ -18,3 +19,5 @@
 ]
 
 TOMTE_CASE_SAMPLE_RULES = [validate_wells_contain_at_most_one_sample]
+
+TOMTE_SAMPLE_RULES = [validate_application_exists]
```
seallard, line 23: This should be case sample rule. The sample rules are for samples which can be uniquely identified by the sample name alone, so the microbial stuff.

All tomte samples require both a case name and a sample name to be uniquely identified in the order.

islean: Okay, that naming is very confusing. I interpreted CASE_SAMPLE_RULES as rules which are applied to the samples within a given case. Could we clear these namings up?

seallard: That is the correct interpretation 👍

islean: Resolved in person

cg/services/order_validation_service/models/errors.py
```diff
@@ -55,6 +55,11 @@ class OccupiedWellError(CaseSampleError):
     message: str = "Well is already occupied"
 
 
+class ApplicationNotValidError(SampleError):
```
seallard, line 58: I think this error should be a `CaseSampleError` since we need both the case and sample name to uniquely identify the sample

seallard (approved):
Nice! I think we should use the `CaseSampleError`

cg/services/order_validation_service/models/errors.py
islean: Fixed!

PR #3486, opened 2024-07-30
Disallow reusing sample names across cases

seallard:
The customer needs to provide a sample internal id if they want to have the same sample in multiple cases. We should not allow them to reuse a sample name in different cases since it is a potential source of confusion for us and them.

Judgement call by @karlnyr

### DIFF
--- a/cg/services/order_validation_service/models/errors.py
+++ b/cg/services/order_validation_service/models/errors.py
@@ -60,6 +60,6 @@ class RepeatedCaseNameError(CaseError):
     message: str = "Case name already used"
 
 
-class RepeatedSampleNameError(CaseSampleError):
+class RepeatedSampleNameError(SampleError):
     field: str = "name"
     message: str = "Sample name already used"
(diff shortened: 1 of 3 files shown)


ChrOertlin:
Just as a note, there should likely be 1. a unique constraint in the DB on that column. 2. We should also make sure that there are no duplicate entires for any given sample name in statusdb.

seallard:
> Just as a note, there should likely be 1. a unique constraint in the DB on that column, if that isn't the case already. 2. We should also make sure that there are no duplicate entires for any given sample name in statusdb.

Might be the case! Not going to fix that in this PR though

islean:
Also - what is the intent with the internal id? The customer knows no such thing so that can't be the way customers are supposed to go about this. To add to this, there is no way of providing internal ids in an order.

ChrOertlin (approved):
Looks good. But I think we need to ensure that sample names are also unique considering existing database entries

ChrOertlin:
> Also - what is the intent with the internal id? The customer knows no such thing so that can't be the way customers are supposed to go about this. To add to this, there is no way of providing internal ids in an order.

Yes indeed, it has to be per customer!

islean:
> Looks good. But I think we need to ensure that sample names are also unique considering existing database entries

That is true! But we thought of that as a separate upcoming validation so we can give a more descriptive error message (sample already taken vs repeated sample name).

seallard:
I'll hold off on this, introduces some confusion at this stage. Lets fix it later

PR #3488, opened 2024-07-30
Validate father sex

seallard:


cg/services/order_validation_service/workflows/tomte/models/case.py
```diff
@@ -10,3 +10,15 @@ class TomteCase(Case):
     panels: list[GenePanelMasterList]
     synopsis: str | None = None
     samples: list[TomteSample]
+
+    def get_sample(self, sample_name: str) -> TomteSample | None:
```
seallard, line 14: We can make these access more efficient later

cg/services/order_validation_service/workflows/tomte/validation/inter_field/utils.py
```diff
@@ -74,3 +76,22 @@ def get_repeated_sample_names(case: TomteCase) -> list[str]:
...
+    for child in children:
+        if is_father_sex_invalid(child=child, case=case):
+            error = create_father_sex_error(case=case, sample=child)
+            errors.append(error)
+    return errors
+
+
+def is_father_sex_invalid(child: TomteSample, case: TomteCase) -> bool:
+    father: TomteSample | None = case.get_sample(child.father)
+    return father and father.sex != Sex.MALE
```
islean, line 93: This makes the checks not atomic. If the father does not exist in the case we would not want a InvalidFatherSexError. That does not seem like what we want

seallard: That won't result in an error here, if the father does not exist the check returns false and no error is created 👍

islean (approved):
Looks good to me!

cg/services/order_validation_service/workflows/tomte/validation/inter_field/utils.py
seallard: But maybe I should add a test for it 🤔

PR #3489, opened 2024-07-30
Add application workflow compatibility check

islean:
## Description

This PR adds validation that the application is allowed for the workflow.

### Added

- Check that the application provided is compatible with the chosen workflow.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/models/errors.py
+++ b/cg/services/order_validation_service/models/errors.py
@@ -57,7 +57,12 @@ class OccupiedWellError(CaseSampleError):
 
 class ApplicationNotValidError(CaseSampleError):
     field: str = "application"
-    message: str = "Application is not valid"
+    message: str = "Chosen application does not exist"
+
+
+class ApplicationNotCompatibleError(CaseSampleError):
+    field: str = "application"
+    message: str = "Application is not compatible with the chosen workflow"
 
 
 class RepeatedCaseNameError(CaseError):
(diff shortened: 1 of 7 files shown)


cg/services/order_validation_service/models/errors.py
```diff
@@ -57,7 +57,12 @@ class OccupiedWellError(CaseSampleError):
 
 class ApplicationNotValidError(CaseSampleError):
     field: str = "application"
-    message: str = "Application is not valid"
+    message: str = "Chosen application does not exist"
+
+
+class ApplicationNotCompatibleError(CaseSampleError):
+    field: str = "application"
+    message: str = "Application is not compatible with the chosen workflow"
```
seallard, line 65: A clearer message might be `Application is not allowed for the chosen workflow`.

cg/services/order_validation_service/validators/inter_field/rules.py
```diff
@@ -24,3 +29,21 @@ def validate_name_required_for_new_order(order: Order, **kwargs) -> list[OrderEr
         error = OrderNameRequiredError()
         errors.append(error)
     return errors
+
+
+def validate_application_compatibility(
+    order: Order, store: Store, **kwargs
```
seallard, line 35: The order model does not contain cases/samples so this type hint is incorrect! I'd rather we defined this rule for the tomte workflow since we do not have an appropriate base class for orders with cases (unless you want to add such a base class in this PR)

PR #3492, opened 2024-07-30
Validate father in same case as child

seallard:


cg/services/order_validation_service/models/errors.py
```diff
@@ -71,5 +71,10 @@ class RepeatedSampleNameError(CaseSampleError):
 
 
 class InvalidFatherSexError(CaseSampleError):
-    field: str = "sex"
+    field: str = "father"
     message: str = "Father must be male"
+
+
+class InvalidFatherCaseError(CaseSampleError):
```
islean, line 78: Slightly unclear naming but probably workable. I would suggest `FatherNotInCaseError`.

islean (approved):
Looks good to me! Sometime soon it might be nice to start writing tests that check multiple validations? I think your tests should have covered for the case when a father is missing - If so, we do not want has failed the sex check as well but we want this PR's validation to give errors. 

Not the best example since I think your unit tests are sufficient for it, but something to consider.

cg/services/order_validation_service/models/errors.py
seallard: Nice, changed it https://github.com/Clinical-Genomics/cg/pull/3492/commits/59c55d28cdd64bd4530f77b20ba88e7aa2f3cfda

seallard:
> Looks good to me! Sometime soon it might be nice to start writing tests that check multiple validations? I think your tests should have covered for the case when a father is missing - If so, we do not want has failed the sex check as well but we want this PR's validation to give errors.
> 
> Not the best example since I think your unit tests are sufficient for it, but something to consider.

Good point! I'll add unit tests for this rule as well. Maybe we can hold off on integration tests until we have the entire tomte flow in place.

PR #3494, opened 2024-07-30
Validate samples are not marked as their own parents

seallard:
Disallow the father/mother of a sample to be set to itself


seallard:
Implement complete cycle check instead

PR #3499, opened 2024-07-31
Add gene panel validation

islean:
### Added

- Added validation making sure that the gene panels exist in status db

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/utils.py
+++ b/cg/services/order_validation_service/utils.py
@@ -3,6 +3,7 @@ from typing import Callable
 from pydantic import ValidationError
 
 from cg.services.order_validation_service.models.errors import (
+    CaseError,
     CaseSampleError,
     OrderError,
     ValidationErrors,
@@ -15,6 +16,14 @@ def apply_order_validation(rules: list[Callable], order: Order, store: Store) ->
     errors: list[OrderError] = []
     for rule in rules:
         rule_errors: list[OrderError] = rule(order=order, store=store)
+        errors.extend(rule_errors)
+    return errors
+
+
+def apply_case_validation(rules: list[Callable], order: Order, store: Store) -> list[CaseError]:
+    errors: list[CaseError] = []
+    for rule in rules:
+        rule_errors: list[CaseError] = rule(order=order, store=store)
         errors.extend(rule_errors)
     return errors
 
--- a/cg/services/order_validation_service/validators/data/rules.py
+++ b/cg/services/order_validation_service/validators/data/rules.py
@@ -4,10 +4,14 @@ from cg.services.order_validation_service.models.errors import (
     CaseSampleError,
     CustomerCannotSkipReceptionControlError,
     CustomerDoesNotExistError,
+    InvalidGenePanelsError,
     OrderError,
     UserNotAssociatedWithCustomerError,
 )
 from cg.services.order_validation_service.models.order import Order
+from cg.services.order_validation_service.validators.data.utils import (
+    validate_panels_for_case,
+)
 from cg.services.order_validation_service.workflows.tomte.models.order import TomteOrder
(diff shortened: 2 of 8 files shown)


cg/services/order_validation_service/utils.py
```diff
@@ -19,6 +20,14 @@ def apply_order_validation(rules: list[Callable], order: Order, store: Store) ->
     return errors
 
 
+def apply_case_validation(rules: list[Callable], order: Order, store: Store) -> list[CaseError]:
```
islean, line 23: Is there a point to having "apply_case_validation" as a new method and not reuse the others?

seallard: The existing division is just based on the different error types the different rules produce. I guess this is the first rule that generates a pure `CaseError` and as such it needs a separate function.

seallard: There might be a better way to do it but it is not apparent to me atm, maybe we can revisit it later?

cg/services/order_validation_service/validators/data/rules.py
```diff
@@ -67,3 +71,13 @@ def validate_application_not_archived(
...
                 errors.append(error)
     return errors
+
+
+def validate_gene_panels_exist(
+    order: TomteOrder, store: Store, **kwargs
+) -> list[InvalidGenePanelsError]:
+    errors: list[InvalidGenePanelsError] = []
+    for case in order.cases:
+        case_errors: list[InvalidGenePanelsError] = validate_panels_for_case(case=case, store=store)
```
seallard, line 81: I don't think the type hints add any value here but keep them if you think so

cg/services/order_validation_service/utils.py
islean: I was thinking that we could have a "apply_validation" and type hint it with a very general error, and then the list of rules will in the end be the only thing dictating the return type. Given that Python doesn't enforce these type hints (as far as I know), that is the current behaviour either way.

seallard (approved):
Nice 🚀

cg/services/order_validation_service/validators/data/rules.py
islean: I'll keep them for now, but it might be more reasonable to type hint depending on category, i.e. CaseError instead of InvalidGenePanelsError.

cg/services/order_validation_service/utils.py
seallard: I don't follow 😞 But since we have this division of errors
```python
class ValidationErrors(BaseModel):
    order_errors: list[OrderError] | None = None
    case_errors: list[CaseError] | None = None
    sample_errors: list[SampleError] | None = None
    case_sample_errors: list[CaseSampleError] | None = None
```

I think it is necessary to have a separate `apply` for all of the different types of errors

islean: I just mean that the `apply` methods can take any of the rules and return any kind of error. The only difference between them are their names and the type hinted return type. To my knowledge we would get exactly the same behaviour from

image

as from 

image

although we of course get warnings when working in the code.

seallard: Maybe we could rewrite the different errors so that we can treat them interchangeably but I don't see how. I'd rather we left it as is now and just have four roughly identical functions 😅

islean: I will leave it as is for now! We could probably have some parent error for it if we are desperate but I am fine as is also.

Issue #2702, opened 2023-11-21
Deliveries and storage of incomplete analyses AND proper tracking of deliveries.


karlnyr:
## Description

Recently we have been having issues with completing analyses for samples received from a specific customer, `cust127`. This issue has then required production to deliver vcf, fastq, bam, etc. to the delivery server manually as the current logic requires the files to be there before storing. I think it is due time that we change this logic - in case this would be needed. The same logic applies when we have had to skip steps for an analysis, resulting in a kind of partial analysis.

However, we cannot simply start to deliver incomplete analyses without also tracking properly what has been delivered. That leads to the next point - updating the delivery report. We should be able to add content to the delivery report in free text. I also think that the delivery report itself can have an additional page, containing each file that is to be included in the delivery and a short explanation of what the purpose of the file is. This would add clarity, and reduce any possible misinformation that could be kept in the ticket rather than in the Scout case or the file deliveries.

## Suggested solution

- Override the current store, allowing the user to store partial cases
   - store information that the analysis stored is partial
- Enable the user to add a comment to the delivery report somehow
- Add a summary of the files that are delivered to the delivery report

## This can be closed when

Above is discussed and implemented or closed

## Blocked by

*If there are any blocking issues/prs/things in this or other repos. Please link to them.*

Vince-janv:
### Decision

### Phase 1
- Somehow set Analysis to "FAILED" (or similar) which should avoid automatic upload
- Ability to generate Delivery-report with force-flag
- Using the force flag should also require a comment flag to be used
- The comment is stored in the Analysis table
### Phase 2
- Creation of delivery-report checks status-db for comment and adds to report
- Add which files are delivered to caesar as a section in the delivery-report
- Short description of the aforementioned files
- Add description of all generated files for all pipelines to an available resource

PR #3500, opened 2024-07-31
Add force store comment

ivadym:
## Description

Context:
- https://github.com/Clinical-Genomics/cg/issues/2702#issuecomment-1822890325

### Added
- Force store comment
- Prevent automatic uploads of force stored cases

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    ```

### How to test
- [x] cg workflow <workflow> <case-id> store --force --comment 

### Expected test outcome

- [x] Analysis can be stored
  - [x] Analysis stored with the provided comment
  - [x] Analysis can be uploaded and delivered

  Screenshot 2024-07-31 at 12 10 40

- [x] Delivery report includes the analysis comment

### Misc
- [x] Cleaning of analysis
  - Not affected. `get_analyses_to_clean` uses `started_at`.
- [x] Scout load config file using `completed_at`.
  - Updated to use `started_at` within `add_common_info_to_load_config` when `completed_at` empty.
- [x] Part of the string representation of the`Analysis(Base)` model.
  - Irrelevant if we have `None` as the `completed_at` field

### Filters
- [x] `AnalysisFilter.VALID_IN_PRODUCTION` (`filter_valid_analyses_in_production`)
  - Used in `get_analyses_to_upload`. It will keep having the expected behaviour (we don't want this cases to be automatically uploaded)
  - `analyses_to_delivery_report`: not affected as we'll generate the reports manually
  - `analyses_to_upload_delivery_reports`: not affected as we'll generate the reports manually
- [x] `AnalysisFilter.COMPLETED` (`filter_completed_analyses`)
  - Used in `get_analyses_to_upload`. It will keep having the expected behaviour (we don't want this cases to be automatically uploaded)
- [x] `AnalysisFilter.ORDER_BY_COMPLETED_AT` (`order_analyses_by_completed_at_asc`)
    - Used in `get_analyses_to_upload`, `analyses_to_delivery_report`, `analyses_to_upload_delivery_reports`. Not affected as these analysis won't be picked up for automatic uploads.
- [x] `CaseFilter.NOT_ANALYSED` (`filter_cases_not_analysed`)
  - Used in `get_not_analysed_cases_by_sample_internal_id`/`add_external_data_to_housekeeper`/`cg add external-hk`. Not relevant for this manual command which uses the ticket as an argument.
- [x] `Case.latest_analyzed`
  - Used to extract `cases_to_analyse` and by `is_case_ready_for_analysis`. To retrieve these cases it also takes into account the CaseActions ( `CaseActions.ANALYZE`) which should be empty after an analysis is forcedly stored. Therefore, the behaviour  should not be affected.

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/workflow/analysis.py
```diff
@@ -250,9 +252,10 @@ def upload_bundle_statusdb(self, case_id: str, dry_run: bool = False) -> None:
         new_analysis: Case = self.status_db.add_analysis(
```
diitaz93, line 252: Should this return an Analysis?
```suggestion
        new_analysis: Analysis = self.status_db.add_analysis(
```

diitaz93 (approved):
Great 🌟

cg/cli/workflow/balsamic/base.py
```diff
@@ -161,11 +166,13 @@ def store_housekeeper(context: CGConfig, case_id: str, force: bool):
...
+        analysis_api.verify_case_config_file_exists(case_id=case_id, dry_run=dry_run)
         analysis_api.verify_deliverables_file_exists(case_id=case_id)
-        analysis_api.upload_bundle_housekeeper(case_id=case_id, force=force)
-        analysis_api.upload_bundle_statusdb(case_id=case_id)
-        analysis_api.set_statusdb_action(case_id=case_id, action=None)
+        analysis_api.upload_bundle_housekeeper(case_id=case_id, dry_run=dry_run, force=force)
+        analysis_api.upload_bundle_statusdb(
+            case_id=case_id, comment=comment, dry_run=dry_run, force=force
+        )
+        analysis_api.set_statusdb_action(case_id=case_id, action=None, dry_run=dry_run)
```
islean, line 175: Probably out of scope for this PR but it would be nice to not have all this logic in the CLI file.

ivadym: Yes I totally agree, but I'm not sure it's worth the effort giving that we'll try to deprecate Balsamic in the next 6-12 months 😓

ivadym:
Deployed!
Screenshot 2024-08-06 at 09 50 10

PR #3503, opened 2024-08-01
Add subject_id different from sample name validation

islean:
## Description

Subject ids need to be different from the sample's name.

### Added

- Validation ensuring that the sample name and subject id can not be the same.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/validators/inter_field/rules.py
+++ b/cg/services/order_validation_service/validators/inter_field/rules.py
@@ -9,6 +9,7 @@ from cg.services.order_validation_service.models.errors import (
     InvalidBufferError,
     OrderError,
     OrderNameRequiredError,
+    SubjectIdSameAsSampleNameError,
     TicketNumberRequiredError,
 )
 from cg.services.order_validation_service.models.order import Order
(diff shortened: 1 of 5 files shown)


cg/services/order_validation_service/validators/inter_field/rules.py
```diff
@@ -70,3 +71,13 @@ def validate_buffers_are_allowed(order: TomteOrder) -> list[CaseSampleError]:
                 error = InvalidBufferError(case_name=case.name, sample_name=sample.name)
                 errors.append(error)
     return errors
+
+
+def validate_subject_ids_different_from_sample_names(order: TomteOrder) -> list[CaseSampleError]:
+    errors = []
+    for case in order.cases:
```
seallard, line 78: Consider adding a method to the `TomteOrder` class called `get_samples` to get rid of these loops here 😄

seallard (approved):
🚀

cg/services/order_validation_service/validators/inter_field/rules.py
islean: I like it! Would you prefer that as a method or a property? order.get_samples() vs order.samples

islean: Also, I need both the case name and the sample name

seallard: Ah, true... Do we want to map in the case name to the samples? I feel like it would make our lives easier :)

seallard: Not possible since a sample can be in multiple cases 😅 Lets leave it

PR #3505, opened 2024-08-01
Add subject_id different from case name validation

islean:
## Description

Subject ids are not allowed to be the same as the case's name.

### Added

- Validation ensuring that the subject id is not the same as the case name.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/workflows/tomte/validation/inter_field/rules.py
+++ b/cg/services/order_validation_service/workflows/tomte/validation/inter_field/rules.py
@@ -1,10 +1,11 @@
 from cg.services.order_validation_service.models.errors import (
+    CaseSampleError,
     FatherNotInCaseError,
     InvalidFatherSexError,
     InvalidMotherSexError,
-    PedigreeError,
     MotherNotInCaseError,
     OccupiedWellError,
+    PedigreeError,
     RepeatedCaseNameError,
     RepeatedSampleNameError,
 )
@@ -22,6 +23,7 @@ from cg.services.order_validation_service.workflows.tomte.validation.inter_field
     get_mother_sex_errors,
     get_repeated_case_name_errors,
     get_repeated_sample_name_errors,
+    validate_subject_ids_in_case,
 )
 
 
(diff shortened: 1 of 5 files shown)


cg/services/order_validation_service/workflows/tomte/validation/inter_field/rules.py
```diff
@@ -81,3 +83,10 @@ def validate_mothers_in_same_case_as_children(order: TomteOrder) -> list[MotherN
         case_errors = get_mother_case_errors(case)
         errors.extend(case_errors)
     return errors
+
+
+def validate_subject_ids_different_from_case_names(order: TomteOrder) -> list[CaseSampleError]:
+    errors = []
+    for case in order.cases:
+        errors.extend(validate_subject_ids_in_case(case))
```
seallard, line 91: In general I think it is preferable to avoid nesting function calls like this

seallard (approved):
👍

cg/services/order_validation_service/workflows/tomte/validation/inter_field/rules.py
```diff
@@ -81,3 +83,10 @@ def validate_mothers_in_same_case_as_children(order: TomteOrder) -> list[MotherN
         case_errors = get_mother_case_errors(case)
         errors.extend(case_errors)
     return errors
+
+
+def validate_subject_ids_different_from_case_names(order: TomteOrder) -> list[CaseSampleError]:
```
seallard, line 88: I think it is better to type the return with the exact type of error the function returns

cg/services/order_validation_service/workflows/tomte/validation/inter_field/rules.py
islean: Fixed

PR #3506, opened 2024-08-02
Add cancelled status to samples

seallard:
Part of resolving https://github.com/Clinical-Genomics/clinical-genomics-ui/issues/431.

We'll add a more robust way of tracking sample actions in a `SampleAction` table but we are starting like this.


islean (approved):
Looks good to me!

PR #3507, opened 2024-08-02
Refactor cancel action

seallard:
The existing cancel action is pretty messy, this PR tidies it up a bit.


cg/services/sample_service/sample_service.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class SampleService:
+
+    def __init__(self, store: Store):
+        self.store = store
+
+    def cancel_sample(self, sample_id: int, user_email: str) -> list[str]:
+        """Cancel a single sample and return a list of all cases it used to be in."""
+        case_ids: list[str] = self.store.get_case_ids_with_sample(sample_id)
```
islean, line 16: The returned list here is very confusing and to me it does not serve that much of a purpose. Could we not let this method only cancel the sample? Then we could just store the case ids before calling it.

cg/services/sample_service/utils.py
```diff
@@ -0,0 +1,20 @@
...
+    message = f"Cancelled {len(sample_ids)} {samples}. "
+
+    if remaining_cases:
+        cases = "case" if len(remaining_cases) == 1 else "cases"
+        case_message = ", ".join(remaining_cases)
+        message += f"Found {len(remaining_cases)} {cases} with additional samples: {case_message}."
+    else:
+        message += "No case contained additional samples."
+
+    return message
```
islean, line 20: This method is very hard to follow. I think it stems from 

1. The `samples` and `cases` being instantiated in a convoluted way. In general it is quite dense with if/elses.
2. `remaining_cases` is a perplexing parameter name. From the context it is "cases with more samples than the one being cancelled" but my first thought would be that it is "cases that have yet to be handled in this flow".
3. The string `f"Found {len(remaining_cases)} {cases} with additional samples: {case_message}."`I would most likely read expecting `case_message` to be the list of additional samples.

cg/store/crud/delete.py
```diff
@@ -36,3 +36,13 @@ def delete_illumina_flow_cell(self, internal_id: str):
             self.session.commit()
         else:
             raise ValueError(f"Illumina flow cell with internal id {internal_id} not found.")
+
+    def delete_sample_from_cases(self, sample_id: int) -> None:
```
islean, line 40:
```suggestion
    def decouple_sample_from_cases(self, sample_id: int) -> None:
```
For me this is a clearer naming.

cg/services/sample_service/sample_service.py
```diff
@@ -0,0 +1,43 @@
...
+from cg.store.models import User
+from cg.store.store import Store
+
+
+class SampleService:
+
+    def __init__(self, store: Store):
+        self.store = store
+
+    def cancel_sample(self, sample_id: int, user_email: str) -> list[str]:
```
islean, line 14: Most likely, the users will specify the internal_id rather than the database entry id.

cg/store/crud/read.py
```diff
@@ -1463,3 +1463,8 @@ def get_cases_for_sequencing_qc(self) -> list[Case]:
                 CaseFilter.HAS_SEQUENCE,
             ],
         ).all()
+
+    def get_case_ids_with_sample(self, sample_id: str) -> list[str]:
```
islean, line 1467:
```suggestion
    def get_case_ids_with_sample(self, sample_id: int) -> list[str]:
```

If you decide to go with the database id, although I would probably use the internal_id instead.

cg/store/crud/update.py
```diff
@@ -21,12 +21,10 @@ def __init__(self, session: Session):
...
 
-    def update_sample_comment(self, sample: Sample, comment: str) -> None:
+    def update_sample_comment(self, sample_id: int, comment: str) -> None:
         """Update comment on sample with the provided comment."""
-        if sample.comment:
-            sample.comment = sample.comment + " " + comment
-        else:
-            sample.comment = comment
+        sample: Sample | None = self.get_sample_by_entry_id(sample_id)
+        sample.comment = f"{sample.comment} {comment}" if sample.comment else comment
```
islean, line 27: Do we need a null check here?

cg/store/crud/update.py
```diff
@@ -71,3 +69,8 @@ def update_sample_sequenced_at(self, internal_id: str, date: datetime):
         sample: Sample = self.get_sample_by_internal_id(internal_id)
         sample.last_sequenced_at = date
         self.session.commit()
+
+    def mark_sample_as_cancelled(self, sample_id: int) -> None:
+        sample: Sample = self.get_sample_by_entry_id(sample_id)
+        sample.is_cancelled = True
```
islean, line 75: null check? Or do we want the error to be raised? Might be fine-

cg/services/sample_service/sample_service.py
```diff
@@ -0,0 +1,43 @@
...
+        """Cancel a single sample and return a list of all cases it used to be in."""
+        case_ids: list[str] = self.store.get_case_ids_with_sample(sample_id)
+
+        self.store.mark_sample_as_cancelled(sample_id)
+        self.store.delete_sample_from_cases(sample_id)
+        self.store.delete_cases_without_samples(case_ids)
```
islean, line 20: I don't think these two steps are reversible, so I do not think that this meets the acceptance criteria listed by @beatrizsavinhas:

image

islean (commented):
I think this action is not reversible and hence, we need to rethink what should happen. Both the Case and CaseSample table contains data which should not be deleted when cancelling a sample.

cg/services/sample_service/sample_service.py
seallard: This PR has nothing to do with those criteria, it just refactors the existing cancel action.

cg/services/sample_service/sample_service.py
seallard: When you select stuff in the admin interface it is the entry id:s by default!

cg/services/sample_service/sample_service.py
seallard: Resolved in https://github.com/Clinical-Genomics/cg/pull/3507/commits/41c8b7b29beffb69bd26234e2ab22bd87627fedb

cg/store/crud/read.py
seallard: Using the `internal_id` is not an option since the selected records are identified with the database id:s by default. Fixed the type hint in https://github.com/Clinical-Genomics/cg/pull/3507/commits/41c8b7b29beffb69bd26234e2ab22bd87627fedb

cg/services/sample_service/sample_service.py
islean: Yeah, I agree, for me this would be

1. `case_ids = [case.internal_id] for case in sample.links`
2. `cancel_sample(sample)`
3. `cancel_empty_cases(case_ids)`
4. `remaining_cases = get_remaining_cases(case_ids)`
5. `return [case.internal_id for case in remaining_cases]`

cg/services/sample_service/sample_service.py
islean: Oh! Forgot that this is via the web api!

cg/store/crud/update.py
seallard: Fixed in https://github.com/Clinical-Genomics/cg/pull/3507/commits/70e66ba5e2c0878722aa412686340482fe7f9c4f

cg/store/crud/update.py
seallard: Fixed in https://github.com/Clinical-Genomics/cg/pull/3507/commits/b907fa5abf6fd1c675fe7f57677f3123a1d2fa13
Raised exception

cg/services/sample_service/utils.py
seallard: Made more readable in https://github.com/Clinical-Genomics/cg/pull/3507/commits/53e15193a73f78c51c9120e88cf7bb96c961ea6b

Feel free to suggest further improvements 🙂

cg/store/crud/delete.py
seallard: https://github.com/Clinical-Genomics/cg/pull/3507/commits/4a6f467fd8a5b4e8b7353631114254c89a3ac718

cg/services/sample_service/sample_service.py
```diff
@@ -0,0 +1,39 @@
...
+        case_ids: list[str] = self.store.get_case_ids_with_sample(sample_id)
+        self.store.mark_sample_as_cancelled(sample_id)
+        self.store.decouple_sample_from_cases(sample_id)
+        self.store.delete_cases_without_samples(case_ids)
+        self._add_cancel_comment(sample_id=sample_id, user_email=user_email)
+
+    def _add_cancel_comment(self, sample_id: int, user_email: str) -> None:
+        if user := self.store.get_user_by_email(user_email):
+            comment = get_cancel_comment(user.name)
+            self.store.update_sample_comment(sample_id=sample_id, comment=comment)
```
islean, line 24: Would we want to validate that the user exists before acting on anything at all? Or that is done by the endpoint already right?

cg/services/sample_service/sample_service.py
```diff
@@ -0,0 +1,39 @@
...
+        for sample_id in sample_ids:
+            self.cancel_sample(sample_id=sample_id, user_email=user_email)
+
+        self.store.delete_cases_without_samples(case_ids)
+        remaining_cases = self.store.filter_cases_with_samples(case_ids)
+
+        return get_confirmation_message(
+            sample_ids=sample_ids,
+            case_ids=remaining_cases,
+        )
```
islean, line 39: Seems as if we are duplicating some logic here. In `self.cancel_sample`, do we not already delete cases without samples?

cg/store/crud/read.py
```diff
@@ -1463,3 +1468,14 @@ def get_cases_for_sequencing_qc(self) -> list[Case]:
                 CaseFilter.HAS_SEQUENCE,
             ],
         ).all()
+
+    def get_case_ids_with_sample(self, sample_id: int) -> list[str]:
+        """Return all case ids with a sample."""
+        sample: Sample = self.get_sample_by_entry_id(sample_id)
+        return [link.case.internal_id for link in sample.links] if sample else []
+
+    def get_case_internal_ids_with_samples(self, sample_ids: list[int]) -> list[str]:
```
islean, line 1477: these methods should either call the internal_ids case_ids or case_internal_ids but they should not be mixed.

islean (approved):
Much cleaner 🥳

cg/services/sample_service/sample_service.py
seallard: Yeah, it is!

cg/services/sample_service/sample_service.py
seallard: True, simplified in https://github.com/Clinical-Genomics/cg/pull/3507/commits/21709d774a10a2494ac924b44fdd136821df3e1d

cg/services/sample_service/sample_service.py
seallard: Cleaned up 👍

cg/store/crud/read.py
seallard: Fixed in https://github.com/Clinical-Genomics/cg/pull/3507/commits/3c8bc540c46fee39b2f6b092705b87d9da440591

seallard:
Tested in stage.

PR #3510, opened 2024-08-02
Hide cancelled samples from customers

seallard:
Closes https://github.com/Clinical-Genomics/clinical-genomics-ui/issues/431

Extract logic for retrieving collaborator samples from the endpoint definition to the sample service. Also add filter for skipping cancelled samples, preventing customers from directly re-ordering cancelled samples.


diitaz93 (approved):
Looks great 👍

cg/server/dto/samples/collaborator_samples_request.py
```diff
@@ -0,0 +1,7 @@
+from pydantic import BaseModel
+
+
+class CollaboratorSamplesRequest(BaseModel):
+    enquiry: str
+    customer: str
+    limit: int = 50
```
islean, line 7: Why do we have a limit here? I would guess that all samples matching the query are relevant.

islean: I mean, if the rest of the parameters result in 60 samples, on what grounds can we hide 10 of them?

seallard: It was there so I kept it. I guess we don't want an infinite list in the drop down and that the search the customers do usually narrows it down to fewer than 50 entries anyway

islean: I think it is fundamentally wrong to limit this query. As you say, the query sent by the customer should already limit the query enough, and if not, there is no good reason to hide some of them. I do not know if this is needed for when the customer starts typing but I hope that there is a better way. Either way, since it was not introduced here, you can consider it out of scope.

seallard:
Deployed and tested in stage.

PR #3511, opened 2024-08-05
Fix revision

seallard:
The default parameter name was incorrect, resulting in the default value not being set for the new `is_cancelled` column in the sample table.


islean (approved):
NIce!

PR #3513, opened 2024-08-05
Extract sample endpoints to separate blueprint

seallard:
Minor refactoring extracting all endpoints related to samples into a separate blueprint.


cg/server/app.py
```diff
@@ -90,7 +90,8 @@ def logged_in(blueprint, token):
     app.register_blueprint(SAMPLES_BLUEPRINT)
     _register_admin_views()
 
-    ext.csrf.exempt(api.BLUEPRINT)  # Protected with Auth header already
+    ext.csrf.exempt(api.BLUEPRINT)
+    ext.csrf.exempt(SAMPLES_BLUEPRINT)
```
islean, line 94: What happens in the exempt? And does it work well with the app.register_blueprint(SAMPLES_BLUEPRINT)?

seallard: It disables CSRF protection, which was done previously. It is not good that we disable it, but I recall we had issues when it was enabled. I will consider enabling it out of scope for this PR.

Yes, it works

diitaz93 (approved):
👍

PR #3518, opened 2024-08-06
Set PacBio sequencing times

diitaz93:
## Description
Closes https://github.com/Clinical-Genomics/add-new-tech/issues/88

This PR parses the datetime for the start of the sequencing and the completion of the sequencing. The dataset metrics hold a proxy for the sequencing completed at date in a variable called `createdAt`. The sequencing started at can be extracted from the movie name, in any of the file names.


cg/services/post_processing/pacbio/metrics_parser/models.py
```diff
@@ -137,6 +140,16 @@ def extract_plate(cls, data: Any):
...
         return data
 
+    @model_validator(mode="before")
+    @classmethod
+    def set_sequencing_completed_at(cls, data: Any):
+        if isinstance(data, dict):
+            movie_name = data.get(SmrtLinkDatabasesIDs.MOVIE_NAME)
+            if movie_name:
+                date: str = movie_name.split("_")[1] + movie_name.split("_")[2]
+                data["run_started_at"] = datetime.strptime(date, "%y%m%d%H%M%S")
```
Vince-janv, line 150: Are there any time zone considerations to be had hare?

cg/services/post_processing/pacbio/data_transfer_service/utils.py
```diff
@@ -18,7 +18,8 @@ def get_sequencing_run_dto(metrics: PacBioMetrics) -> PacBioSequencingRunDTO:
         type=DeviceType.PACBIO,
         well=metrics.dataset_metrics.well,
         plate=metrics.dataset_metrics.plate,
-        movie_time_hours=None,
+        run_started_at=metrics.dataset_metrics.run_started_at,
+        run_completed_at=metrics.dataset_metrics.run_completed_at,
```
Vince-janv, line 22: Same as with the database model. Do we need run_xxxxx?

Vince-janv (approved):
Nice ⭐

cg/services/post_processing/pacbio/metrics_parser/models.py
diitaz93: Agreed to consider timezones when it is fixed in the PacBio server

cg/services/post_processing/pacbio/data_transfer_service/utils.py
diitaz93: implemented

PR #3521, opened 2024-08-06
Add application to response

seallard:
The application information was missing from the samples response when customers select an existing sample when placing an order.

### DIFF
--- a/cg/server/dto/samples/samples_response.py
+++ b/cg/server/dto/samples/samples_response.py
@@ -1,6 +1,52 @@
+from datetime import datetime
 from pydantic import BaseModel
 
 from cg.constants.subject import Sex
+
+
+class ApplicationVersionDTO(BaseModel):
+    id: int
+    version: int | None = (None,)
+    valid_from: datetime | None = None
+    price_standard: int | None = None
+    price_priority: int | None = None
+    price_express: int | None = None
+    price_research: int | None = None
+    price_clinical_trials: int | None = None
+    comment: str | None = None
+    created_at: datetime | None = None
+    updated_at: datetime | None = None
+    application_id: int | None = None
+
+
+class ApplicationDTO(BaseModel):
+    id: int
+    tag: str | None = None
+    prep_category: str | None = None
+    is_external: bool | None = None
+    description: str | None = None
+    is_accredited: bool | None = None
+    turnaround_time: int | None = None
+    minimum_order: int | None = None
+    sequencing_depth: int | None
+    min_sequencing_depth: int | None = None
+    target_reads: int | None = None
+    percent_reads_guaranteed: int | None = None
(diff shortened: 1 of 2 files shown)


islean (commented):
I think there is data missing regarding links and flowcells as well. See the `to_dict` method on the sample level which generated the old response:

image

cg/server/dto/samples/samples_response.py
```diff
@@ -1,19 +1,66 @@
...
+    percent_kth: int | None = None
+    comment: str | None = None
+    is_archived: bool | None = None
+    created_at: datetime | None = None
+    updated_at: datetime | None = None
+
+
 class CustomerDto(BaseModel):
     internal_id: str
     name: str
```
islean, line 54: We returned a lot more customer data before as well

seallard: It is not used for anything though so there is no need to return it

seallard:
> I think there is data missing regarding links and flowcells as well. See the `to_dict` method on the sample level which generated the old response:

I don't think that information is used anywhere but looking into it

islean:
> > I think there is data missing regarding links and flowcells as well. See the `to_dict` method on the sample level which generated the old response:
> 
> I don't think that information is used anywhere but looking into it

I have not looked into it too much either! Just wanted to highlight in case we will run into the same issues again after merge.

seallard:
> I have not looked into it too much either! Just wanted to highlight in case we will run into the same issues again after merge.

Yeah, it is a good point! Looking through the frontend code, the links are not used anywhere for the result from the `/samples` endpoint and I'll leave it for now.

cg/services/sample_service/utils.py
```diff
@@ -43,4 +49,62 @@ def create_sample_dto(sample: Sample) -> SampleDTO:
...
+        price_standard=version.price_standard,
+        price_priority=version.price_priority,
+        price_express=version.price_express,
+        price_research=version.price_research,
+        price_clinical_trials=version.price_clinical_trials,
+        comment=version.comment,
+        created_at=version.created_at,
+        updated_at=version.updated_at,
+        application_id=version.application_id,
     )
```
islean, line 110: Could these not be instantiated with ApplicationDTO(**application) or something? Or does that defeat the purpose?

islean (approved):
Looks good! If anything is missing it should be caught when trying to place orders in stage.

cg/services/sample_service/utils.py
seallard: I don't like is that we need to have the same naming for the fields that way, but I'll check it out in a separate branch. I guess it just discards the database fields not present in the Pydantic model.

seallard:
Placed order in stage without errors. The application is displayed.

PR #3524, opened 2024-08-06
Instantiate TicketHandler with freshdesk client

ahdamin:
### Added

- FreshDesk Client

### Changed

- osticket in `ticket_handler.py`

### DIFF
--- a/cg/clients/freshdesk/freshdesk_client.py
+++ b/cg/clients/freshdesk/freshdesk_client.py
@@ -31,16 +31,16 @@ class FreshdeskClient:
     def create_ticket(self, ticket: TicketCreate, attachments: list[Path] = None) -> TicketResponse:
         """Create a ticket."""
         ticket_data = ticket.model_dump(exclude_none=True)
-        endpint_url = self.base_url + EndPoints.TICKETS
+        endpoint_url = self.base_url + EndPoints.TICKETS
 
         if attachments:
             files = [
                 ("attachments[]", (attachment.name, open(attachment, "rb")))
                 for attachment in attachments
             ]
-            response: Response = self.session.post(url=endpint_url, data=ticket_data, files=files)
+            response: Response = self.session.post(url=endpoint_url, data=ticket_data, files=files)
         else:
-            response: Response = self.session.post(url=endpint_url, json=ticket_data)
+            response: Response = self.session.post(url=endpoint_url, json=ticket_data)
 
         response.raise_for_status()
         return TicketResponse.model_validate(response.json())
--- a/cg/clients/freshdesk/freshdesk_client.py
+++ b/cg/clients/freshdesk/freshdesk_client.py
@@ -1,17 +1,14 @@
 import logging
 from http import HTTPStatus
 from pathlib import Path
-from tempfile import TemporaryDirectory
 
 from requests import Response, Session
 from requests.adapters import HTTPAdapter
 from urllib3 import Retry
 
 from cg.clients.freshdesk.constants import EndPoints
-from cg.clients.freshdesk.utils import handle_client_errors
 from cg.clients.freshdesk.models import TicketCreate, TicketResponse
-from cg.constants.constants import FileFormat
-from cg.io.controller import WriteFile, WriteStream
+from cg.clients.freshdesk.utils import handle_client_errors, prepare_attachments
 
 LOG = logging.getLogger(__name__)
 TEXT_FILE_ATTACH_PARAMS = "data:text/plain;charset=utf-8,{content}"
@@ -31,17 +28,12 @@ class FreshdeskClient:
     def create_ticket(self, ticket: TicketCreate, attachments: list[Path] = None) -> TicketResponse:
         """Create a ticket."""
         ticket_data = ticket.model_dump(exclude_none=True)
-        endpoint_url = self.base_url + EndPoints.TICKETS
+        files = prepare_attachments(attachments) if attachments else None
 
-        if attachments:
-            files = [
-                ("attachments[]", (attachment.name, open(attachment, "rb")))
-                for attachment in attachments
-            ]
-            response: Response = self.session.post(url=endpoint_url, data=ticket_data, files=files)
-        else:
-            response: Response = self.session.post(url=endpoint_url, json=ticket_data)
-
+        LOG.info(ticket_data)
+        response: Response = self.session.post(
+            url=f"{self.base_url}{EndPoints.TICKETS}", data=ticket_data, files=files
+        )
         response.raise_for_status()
         return TicketResponse.model_validate(response.json())
 
@@ -49,8 +41,7 @@ class FreshdeskClient:
     def _get_session(self) -> Session:
         session = Session()
         session.auth = (self.api_key, "X")
-        retries = Retry(total=5, backoff_factor=1, status_forcelist=[429, 500, 502, 503, 504])
-        session.mount("https://", HTTPAdapter(max_retries=retries))
+        self._configure_retries(session)
         return session
 
     @staticmethod
@@ -69,23 +60,3 @@ class FreshdeskClient:
         )
         adapter = HTTPAdapter(max_retries=retry_strategy)
         session.mount("https://", adapter)
-
-    @staticmethod
-    def create_new_ticket_attachment(content: dict, file_name: str) -> dict:
-        return {
-            file_name: TEXT_FILE_ATTACH_PARAMS.format(
-                content=WriteStream.write_stream_from_content(
-                    content=content, file_format=FileFormat.JSON
-                )
-            )
-        }
-
-    @staticmethod
-    def create_connecting_ticket_attachment(content: dict) -> TemporaryDirectory:
-        directory = TemporaryDirectory()
-        WriteFile.write_file_from_content(
-            content=content,
-            file_format=FileFormat.JSON,
-            file_path=Path(directory.name, "order.json"),
-        )
-        return directory
--- a/cg/clients/freshdesk/freshdesk_client.py
+++ b/cg/clients/freshdesk/freshdesk_client.py
@@ -1,6 +1,8 @@
 import logging
 from http import HTTPStatus
-
+from tempfile import TemporaryDirectory
+from cg.constants.constants import FileFormat
+from pathlib import Path
 from pydantic import ValidationError
 from requests import RequestException, Response, Session
 from requests.adapters import HTTPAdapter
@@ -12,8 +14,11 @@ from cg.clients.freshdesk.exceptions import (
     FreshdeskModelException,
 )
 from cg.clients.freshdesk.models import TicketCreate, TicketResponse
+from cg.io.controller import WriteFile, WriteStream
+
 
 LOG = logging.getLogger(__name__)
+TEXT_FILE_ATTACH_PARAMS = "data:text/plain;charset=utf-8,{content}"
 
 
 class FreshdeskClient:
@@ -26,22 +31,37 @@ class FreshdeskClient:
         self.env = env
         self.session = self._get_session()
 
-    def create_ticket(self, ticket: TicketCreate) -> TicketResponse:
+    def create_ticket(self, ticket: TicketCreate, attachments: list[Path] = None) -> TicketResponse:
         """Create a ticket."""
         LOG.info(ticket.model_dump_json(exclude_none=True))
         try:
-            response: Response = self.session.post(
-                url=self._url(EndPoints.TICKETS),
-                json=ticket.model_dump(exclude_none=True),
-            )
+            # Prepare the ticket fields
+            ticket_data = ticket.model_dump(exclude_none=True)
+
+            if attachments:
+                # Multipart form-data
+                files = [('attachments[]', (attachment.name, open(attachment, 'rb'))) for attachment in attachments]
+                response: Response = self.session.post(
+                    url=self._url(EndPoints.TICKETS),
+                    data=ticket_data,  # Form fields
+                    files=files
+                )
+            else:
+                # JSON payload
+                response: Response = self.session.post(
+                    url=self._url(EndPoints.TICKETS),
+                    json=ticket_data  # JSON data
+                )
+
             response.raise_for_status()
             return TicketResponse.model_validate(response.json())
+
         except RequestException as error:
             LOG.error(f"Could not create ticket: {error}")
-            LOG.error(f"Response: {response.text}")
+            LOG.error(f"Response: {error.response.text if error.response else 'No response'}")
             raise FreshdeskAPIException(error) from error
         except ValidationError as error:
-            LOG.error(f"Response from Freshdesk does not fit model: {TicketResponse}.\n{error}")
+            LOG.error(f"Could not validate ticket response: {error}")
             raise FreshdeskModelException(error) from error
 
     def _url(self, endpoint: str) -> str:
@@ -49,11 +69,10 @@ class FreshdeskClient:
         return f"{self.base_url}{endpoint}"
 
     def _get_session(self) -> Session:
-        """Configures and sets a session to be used for requests."""
         session = Session()
-        self._configure_retries(session)
-        session.auth = (self.api_key, "X")
-        session.headers.update({"Content-Type": "application/json"})
+        session.auth = (self.api_key, 'X')
+        retries = Retry(total=5, backoff_factor=1, status_forcelist=[429, 500, 502, 503, 504])
+        session.mount('https://', HTTPAdapter(max_retries=retries))
         return session
 
     @staticmethod
@@ -72,3 +91,23 @@ class FreshdeskClient:
         )
         adapter = HTTPAdapter(max_retries=retry_strategy)
         session.mount("https://", adapter)
+
+    @staticmethod
+    def create_new_ticket_attachment(content: dict, file_name: str) -> dict:
+        return {
+            file_name: TEXT_FILE_ATTACH_PARAMS.format(
+                content=WriteStream.write_stream_from_content(
+                    content=content, file_format=FileFormat.JSON
+                )
+            )
+        }
+
+    @staticmethod
+    def create_connecting_ticket_attachment(content: dict) -> TemporaryDirectory:
+        directory = TemporaryDirectory()
+        WriteFile.write_file_from_content(
+            content=content,
+            file_format=FileFormat.JSON,
+            file_path=Path(directory.name, "order.json"),
+        )
+        return directory
--- a/cg/clients/freshdesk/freshdesk_client.py
+++ b/cg/clients/freshdesk/freshdesk_client.py
@@ -31,26 +31,19 @@ class FreshdeskClient:
     def create_ticket(self, ticket: TicketCreate, attachments: list[Path] = None) -> TicketResponse:
         """Create a ticket."""
         ticket_data = ticket.model_dump(exclude_none=True)
+        endpint_url = self.base_url + EndPoints.TICKETS
 
         if attachments:
             files = [
                 ("attachments[]", (attachment.name, open(attachment, "rb")))
                 for attachment in attachments
             ]
-            response: Response = self.session.post(
-                url=self._url(EndPoints.TICKETS), data=ticket_data, files=files
-            )
+            response: Response = self.session.post(url=endpint_url, data=ticket_data, files=files)
         else:
-            response: Response = self.session.post(
-                url=self._url(EndPoints.TICKETS), json=ticket_data
-            )
+            response: Response = self.session.post(url=endpint_url, json=ticket_data)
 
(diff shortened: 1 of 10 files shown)


cg/meta/orders/api.py
```diff
@@ -57,11 +58,15 @@ def _get_submit_handler(project: OrderType, lims: LimsAPI, status: Store) -> Sub
...
 
-    def __init__(self, lims: LimsAPI, status: Store, osticket: OsTicket):
+    def __init__(
+        self, lims: LimsAPI, status: Store, osticket: OsTicket, freshdesk_client: FreshdeskClient
+    ):
         super().__init__()
         self.lims = lims
         self.status = status
-        self.ticket_handler: TicketHandler = TicketHandler(osticket_api=osticket, status_db=status)
+        self.ticket_handler: TicketHandler = TicketHandler(
```
seallard, line 67: The type hint here is redundant

seallard, line 67: The ticket handler should preferably be passed into the `OrdersAPI` class directly as a dependency instead of being instantiated here, but that is a larger refactoring. Feel free to consider it out of scope.

Just want to make you aware of the existing design issues.

cg/meta/orders/ticket_handler.py
```diff
@@ -46,16 +49,20 @@ def create_ticket(
             project=project,
         )
         attachment: dict = self.create_attachment(order=order)
-        ticket_nr: str | None = self.osticket.open_ticket(
-            name=user_name,
+
+        # Create ticket in Freshdesk
+        freshdesk_ticket = TicketCreate(
```
seallard, line 54: The encapsulation of the fresh desk logic is leaky.

We don't want to have logic mentioning fresh desk spread through out our code base. The usage of the client needs to be wrapped in a service. But I guess this `TicketHandler` is the ticketing service.

Ideally it would look something like this.

```python
class TicketingService:
    def __init__(self, client: FreshDeskAPIClient):
        self.client = client
    
    def create_ticket(self, email: str, name: str, subject: str, attachment: dict) -> int:
        ticket = TicketCreate(
            email=email,
            name=name,
            subject=subject,
            attachment=[attachment]
        )
        ticket_response: TicketResponse = self.freshdesk_client.create_ticket(ticket)
        response = self.client.create_ticket(ticket)
        LOG.info(f"{ticket_response.id}: opened new ticket in Freshdesk")
        return ticket_response.id
```

seallard: But looking at these changes it seems like you want to ensure that fresh desk tickets are created in parallel which is fine.

Feel free to disregard this for now, it is a bigger refactoring of existing code.

Vince-janv: I agree with SA. I think it should be referred to as client in the TicketHandler, and maybe ticketing_client in the OrdersAPI.

tests/meta/orders/test_meta_orders_api.py
```diff
@@ -85,33 +114,25 @@ def test_submit(
...
-            f"cg.meta.orders.{submitter}.process_lims",
-            lambda **kwargs: (lims_project_data, lims_map),
-        )
-
-
 @pytest.mark.parametrize(
     "order_type",
     [OrderType.MIP_DNA, OrderType.MIP_RNA, OrderType.BALSAMIC],
 )
+@patch("cg.clients.freshdesk.freshdesk_client.FreshdeskClient.create_ticket")
```
diitaz93, line 121: Is there a reason why you moved the mocking to a decorator instead of a `with` statement? I think it is clearer using `with`, but is a personal opinion

diitaz93 (commented):
Good start! 👍

cg/meta/orders/ticket_handler.py
ahdamin: Thanks for the clarifications. Will consider that when we remove `osticket` !

tests/meta/orders/test_meta_orders_api.py
```diff
@@ -85,33 +114,25 @@ def test_submit(
...
+    mock_create_ticket,
     all_orders_to_submit,
     orders_api: OrdersAPI,
     order_type: OrderType,
     user_mail: str,
     user_name: str,
 ):
+    # Mock Freshdesk ticket creation to raise TicketCreationError
+    mock_create_ticket.side_effect = TicketCreationError("ERROR")
+
```
Vince-janv, line 132: I'm not a huge fan of using the patch decorators for mocking. It becomes quite confusing when there are multiple of them. Something like this might be clearer. You could alsp extract the instantiation of ticket_response into a smaller function if you prefer.
```suggestion
def test_submit_ticketexception(
    all_orders_to_submit,
    orders_api: OrdersAPI,
    order_type: OrderType,
    user_mail: str,
    user_name: str,
):

# GIVEN the ticketing client providing a valid response
ticket_response=TicketResponse(id=ticket_id,
        email=user_mail,
        priority=1,
        source=2,
        status=2,
        subject="Support needed..",
        created_at=dt.datetime.now(),
        attachments=[],
        tags=[],)

with patch.object(ordes_api.ticket_handler.freshdesk_client, 'create_ticket', return_value=ticket_response) as mock_create_ticket:
    # WHEN submitting an order
    orders_api.submit(
            project=order_type,
            order_in=order_data,
            user_name=user_name,
            user_mail=user_mail,
        )
```

Vince-janv (approved):
Nice ⭐

cg/meta/orders/api.py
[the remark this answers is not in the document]
ahdamin: 👍 Done !

tests/meta/orders/test_meta_orders_api.py
```diff
@@ -35,7 +36,33 @@ def test_too_long_order_name():
         OrderIn(name=long_name, customer="", comment="", samples=[])
 
 
+def monkeypatch_process_lims(monkeypatch, order_data):
```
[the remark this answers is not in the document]
ahdamin: Added ✅ thanks for pointing out

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -38,6 +38,7 @@ def create_ticket(self, ticket: TicketCreate) -> TicketResponse:
             return TicketResponse.model_validate(response.json())
         except RequestException as error:
             LOG.error(f"Could not create ticket: {error}")
+            LOG.error(f"Response: {response.text}")
```
seallard, line 41: This error handling and logging should be extracted to a separate module

cg/meta/orders/ticket_handler.py
```diff
@@ -22,9 +21,13 @@ class TicketHandler:
 
     NEW_LINE = "<br />"
 
-    def __init__(self, osticket_api: OsTicket, freshdesk_client: FreshdeskClient, status_db: Store):
-        self.osticket: OsTicket = osticket_api
-        self.freshdesk_client: FreshdeskClient = freshdesk_client
+    def __init__(self, status_db: Store, base_url: str, api_key: str, order_email_id: int, env: str):
+        self.client=FreshdeskClient(
```
seallard, line 25: How come we are backtracking on passing the client in the constructor and instead create it here?

cg/server/api.py
```diff
@@ -68,7 +67,18 @@
 @BLUEPRINT.route("/submit_order/<order_type>", methods=["POST"])
 def submit_order(order_type):
     """Submit an order for samples."""
-    api = OrdersAPI(lims=lims, status=db, osticket=osticket, freshdesk_client=freshdesk_client)
+    ticket_handler = TicketHandler(
```
seallard, line 70: The `TicketHandler` should not be instantiated in the endpoint. Create it in the `ext.py` module instead and import it.

cg/server/api.py
```diff
@@ -68,7 +67,18 @@
...
+        status_db=db,
+        base_url=ticket_service.base_url,
+        api_key=ticket_service.api_key,
+        order_email_id=ticket_service.order_email_id,
+        env=ticket_service.env,
+    )
+    api = OrdersAPI(
```
seallard, line 77: The `OrdersAPI` should not be instantiated in the endpoint. Create it in the `ext.py` module and import it

cg/server/api.py
seallard: Or rather, since it is only used by the `OrdersAPI`, you can just create the `TicketHandler` and then the `OrdersAPI` in the `ext.py` module and import it here.

cg/server/ext.py
```diff
@@ -85,7 +85,7 @@ def init_app(self, app):
 summary_service = OrderSummaryService(store=db, analysis_client=analysis_client)
 order_service = OrderService(store=db, status_service=summary_service)
 sample_service = SampleService(db)
-freshdesk_client = FreshdeskClient(
+ticket_service = FreshdeskClient(
```
seallard, line 88: This is wrong, the client is a web API client. Not a service.

seallard (changes requested):
Looks great, there are some minor things that should be straightened out!

cg/clients/freshdesk/freshdesk_client.py
ahdamin: 👍

cg/meta/orders/ticket_handler.py
```diff
@@ -20,8 +21,8 @@ class TicketHandler:
 
     NEW_LINE = "<br />"
 
-    def __init__(self, osticket_api: OsTicket, status_db: Store):
-        self.osticket: OsTicket = osticket_api
+    def __init__(self, status_db: Store, client: FreshdeskClient):
```
seallard, line 24: 🏅

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -26,33 +27,42 @@ def __init__(self, base_url: str, api_key: str, order_email_id: int, env: str):
...
+            # Multipart form-data
+            files = [
+                ("attachments[]", (attachment.name, open(attachment, "rb")))
+                for attachment in attachments
+            ]
+            response: Response = self.session.post(
+                url=self._url(EndPoints.TICKETS), data=ticket_data, files=files  # Form fields
+            )
+        else:
+            # JSON payload
```
seallard, line 48: Remove comment

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -71,3 +81,23 @@ def _configure_retries(session: Session) -> None:
...
+        }
+
+    @staticmethod
+    def create_connecting_ticket_attachment(content: dict) -> TemporaryDirectory:
+        directory = TemporaryDirectory()
+        WriteFile.write_file_from_content(
+            content=content,
+            file_format=FileFormat.JSON,
+            file_path=Path(directory.name, "order.json"),
+        )
```
seallard, line 102: This logic does not belong here, the only responsibility of the `FreskAPIClient` is to communicate with the fresh desk web API. It should not be dealing with files etc.

I suggest you move this logic out into the `TicketHandler` instead and remove the `attachments: list[Path]` parameter from the `create_ticket` method and just ensure that the attachments are in the `CreateTicket` model.

cg/meta/orders/ticket_handler.py
```diff
@@ -45,22 +46,33 @@ def create_ticket(
...
+        return ticket_response.id
+
+    def create_attachments(self, order: OrderIn, temp_dir: str) -> List[Path]:
+        """Create attachments for the ticket"""
+        attachments = []
+        order_file_path = Path(temp_dir) / "order.json"
+        with order_file_path.open("w") as order_file:
+            order_file.write(order.json())
+        attachments.append(order_file_path)
+        return attachments
```
seallard, line 75: Is there a reason we create a list here? I think you can simplify this a bit
```python
def create_attachment(self, order: OrderIn, temp_dir: str) -> Path:
    """Create attachments for the ticket."""
    order_file_path = Path(temp_dir) / "order.json"
    order_file_path.write_text(order.json())
    return order_file_path
```

cg/server/api.py
```diff
@@ -67,7 +60,8 @@
 @BLUEPRINT.route("/submit_order/<order_type>", methods=["POST"])
 def submit_order(order_type):
     """Submit an order for samples."""
-    api = OrdersAPI(lims=lims, status=db, osticket=osticket)
+
+    api = orders_api
```
seallard, line 64: This can be removed, no need to rename the `orders_api` that has been imported

cg/meta/orders/ticket_handler.py
```diff
@@ -45,22 +46,33 @@ def create_ticket(
...
-        return self.osticket.create_new_ticket_attachment(
-            content=self.replace_empty_string_with_none(obj=order.dict()), file_name="order.json"
-        )
+            # Create ticket in Freshdesk
+            freshdesk_ticket = TicketCreate(
+                email=user_mail,
+                description=message,
+                name=user_name,
+                subject=order.name,
+                attachments=[],
```
seallard, line 59: Why not pass the attachments here instead? :)

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -26,33 +27,42 @@ def __init__(self, base_url: str, api_key: str, order_email_id: int, env: str):
...
 
     def _url(self, endpoint: str) -> str:
         """Get the full URL for the endpoint."""
         return f"{self.base_url}{endpoint}"
 
+    @handle_client_errors
     def _get_session(self) -> Session:
-        """Configures and sets a session to be used for requests."""
         session = Session()
-        self._configure_retries(session)
```
seallard, line 53: If you no longer need the `_configure_retries` method it should be removed completely. Is this a mistake?

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -26,33 +27,42 @@ def __init__(self, base_url: str, api_key: str, order_email_id: int, env: str):
...
+            ]
+            response: Response = self.session.post(
+                url=self._url(EndPoints.TICKETS), data=ticket_data, files=files  # Form fields
+            )
+        else:
+            # JSON payload
             response: Response = self.session.post(
-                url=self._url(EndPoints.TICKETS),
-                json=ticket.model_dump(exclude_none=True),
+                url=self._url(EndPoints.TICKETS), json=ticket_data  # JSON data
```
seallard, line 50: I'd remove this `_url` helper method, it just makes the code more difficult to follow

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -26,33 +27,42 @@ def __init__(self, base_url: str, api_key: str, order_email_id: int, env: str):
         self.env = env
         self.session = self._get_session()
 
-    def create_ticket(self, ticket: TicketCreate) -> TicketResponse:
+    @handle_client_errors
+    def create_ticket(self, ticket: TicketCreate, attachments: list[Path] = None) -> TicketResponse:
         """Create a ticket."""
-        LOG.debug(ticket.model_dump_json())
-        try:
+        LOG.info(ticket.model_dump_json(exclude_none=True))
```
seallard, line 33: Instead of repeating the dump, log the `ticket_data` below

seallard (commented):
Looks so much better ⭐  I feel like we can make the attachment logic a bit clearer, will think more about it tomorrow

cg/server/api.py
ahdamin: 👍

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -26,33 +27,37 @@ def __init__(self, base_url: str, api_key: str, order_email_id: int, env: str):
...
         return f"{self.base_url}{endpoint}"
 
+    @handle_client_errors
     def _get_session(self) -> Session:
-        """Configures and sets a session to be used for requests."""
         session = Session()
-        self._configure_retries(session)
         session.auth = (self.api_key, "X")
-        session.headers.update({"Content-Type": "application/json"})
+        retries = Retry(total=5, backoff_factor=1, status_forcelist=[429, 500, 502, 503, 504])
```
seallard, line 59: If you configure the retries here you can remove the `_configure_retries` method completely. However, I think the `_configure_retries` is clearer so I would keep it.

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -71,3 +76,23 @@ def _configure_retries(session: Session) -> None:
         )
         adapter = HTTPAdapter(max_retries=retry_strategy)
         session.mount("https://", adapter)
+
+    @staticmethod
+    def create_new_ticket_attachment(content: dict, file_name: str) -> dict:
```
seallard, line 81: This logic does not belong in the `FreshDeskAPIClient`, the client is only responsible for sending requests and parsing responses. It should not expose methods for creating attachments.

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -71,3 +76,23 @@ def _configure_retries(session: Session) -> None:
...
+        return {
+            file_name: TEXT_FILE_ATTACH_PARAMS.format(
+                content=WriteStream.write_stream_from_content(
+                    content=content, file_format=FileFormat.JSON
+                )
+            )
+        }
+
+    @staticmethod
+    def create_connecting_ticket_attachment(content: dict) -> TemporaryDirectory:
```
seallard, line 91: This logic does not belong in the `FreshDeskAPIClient`, the client is only responsible for sending requests and parsing responses. It should not expose methods for creating attachments.

seallard (commented):
Some design stuff that can be improved! Let me know if you want to look at it together

cg/server/api.py
```diff
@@ -67,7 +60,7 @@
 @BLUEPRINT.route("/submit_order/<order_type>", methods=["POST"])
 def submit_order(order_type):
     """Submit an order for samples."""
-    api = OrdersAPI(lims=lims, status=db, osticket=osticket)
```
seallard, line 70: This endpoint is defined in `/server/endpoints/orders.py`, not in this file any longer. I think you missed that when you merged master!

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -26,33 +27,37 @@ def __init__(self, base_url: str, api_key: str, order_email_id: int, env: str):
...
-        try:
+        ticket_data = ticket.model_dump(exclude_none=True)
+
+        if attachments:
+            files = [
+                ("attachments[]", (attachment.name, open(attachment, "rb")))
+                for attachment in attachments
+            ]
+            response: Response = self.session.post(
+                url=self._url(EndPoints.TICKETS), data=ticket_data, files=files
```
seallard, line 41: It would be nice to shift all of this logic out of the client

cg/meta/orders/ticket_handler.py
ahdamin: What I understood after a quick look at Freshdesk ticket creation [API](https://developers.freshdesk.com/api/#create_ticket_with_attachment) is that it handles attachments as a list. Here is an [example](https://github.com/freshworks/fresh-samples/blob/master/Python/create_ticket_attachment.py)

cg/clients/freshdesk/freshdesk_client.py
ahdamin: 👍

cg/clients/freshdesk/freshdesk_client.py
ahdamin: 👍

cg/meta/orders/ticket_handler.py
seallard: Alright, then wrap it in a list when you send the request! 

The code becomes clearer if you shift that logic (that it needs to be a list) to where the request is sent against fresh desk.

cg/clients/freshdesk/models.py
```diff
@@ -20,17 +21,19 @@ class TicketCreate(BaseModel):
...
-    source: int
-    status: int
+    description: str
     subject: str
-    tags: list[str] = []
-    type: str | None = None
+    to_emails: Optional[List[str]] = None
+    status: int
+    priority: int
+    attachments: Optional[List[Attachment]] = None
```
seallard, line 39: It seems like this is not used anywhere and can potentially be removed (unless you plan on using it in future PR:s!)

seallard (approved):
Looks good! Some minor stuff to adress

cg/meta/orders/ticket_handler.py
ahdamin: Thanks ! I think it is clear this way ..

cg/clients/freshdesk/models.py
ahdamin: I deleted it for now

ahdamin:
Tested on cg-stage and created a Freshdesk ticket with an attached JSON file that includes order data

PR #3525, opened 2024-08-06
Fix priority and status sample fields

seallard:


cg/services/sample_service/dto_mappers.py
```diff
@@ -14,14 +14,14 @@ def create_sample_dto(sample: Sample) -> SampleDTO:
...
     return SampleDTO(
         comment=sample.comment,
+        control=sample.control,
         customer=customer,
         internal_id=sample.internal_id,
         name=sample.name,
         phenotype_groups=sample.phenotype_groups,
         phenotype_terms=sample.phenotype_terms,
-        priority=sample.priority,
+        priority=sample.priority_human,
```
islean, line 23: What is the difference between these two? Either way, this seems correct 👍

islean (approved):
💯

cg/services/sample_service/dto_mappers.py
seallard: It seems like the `priority_human` is a "human" readable version of the priority enum - this is how the priority field was populated in the `to_dict` method previously.

PR #3530, opened 2024-08-07
Add sequencing times for PacBio run

diitaz93:
## Description
Add the following fields to the `PacBioSequencingRun` db model:
- run_started_at
- run_completed_at

Remove `movie_time_hours`

Create the alembic migration

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b feat-add-pacbio-dates-db -a
    ```

### How to test
- [x] Run the migration (upgrade)
- [x] Run the downgrade

### Expected test outcome

- [x] Check that the upgrade goes through
- [x] The new columns are shown in the table
Screenshot 2024-08-07 at 10 53 19
- [ ] Check that the downgrade goes through

## Review

- [x] Tests executed by SD
- [x] "Merge and deploy" approved by VJ

## Implementation Plan

- [ ] Deployed to stage:
```shell

```
- [ ] Deployed to production:
```shell

```

### DIFF
--- a/cg/store/models.py
+++ b/cg/store/models.py
@@ -1089,7 +1089,9 @@ class PacBioSequencingRun(InstrumentRun):
     id: Mapped[int] = mapped_column(ForeignKey("instrument_run.id"), primary_key=True)
     well: Mapped[Str32]
     plate: Mapped[int]
-    movie_time_hours: Mapped[int]
+    run_started_at: Mapped[datetime | None]
+    run_completed_at: Mapped[datetime | None]
+    movie_time_hours: Mapped[int | None]
     movie_name: Mapped[Str32]
     hifi_reads: Mapped[BigInt]
     hifi_yield: Mapped[BigInt]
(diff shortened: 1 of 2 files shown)


cg/store/models.py
```diff
@@ -1089,7 +1089,9 @@ class PacBioSequencingRun(InstrumentRun):
     id: Mapped[int] = mapped_column(ForeignKey("instrument_run.id"), primary_key=True)
     well: Mapped[Str32]
     plate: Mapped[int]
-    movie_time_hours: Mapped[int]
+    run_started_at: Mapped[datetime | None]
+    run_completed_at: Mapped[datetime | None]
```
Vince-janv, line 1093: The table is named run, so to starting the column name with run is unnecessary.

cg/store/models.py
```diff
@@ -1089,7 +1089,9 @@ class PacBioSequencingRun(InstrumentRun):
...
+    movie_time_hours: Mapped[int | None]
```
Vince-janv, line 1094: Can we drop this one while we're at it? If we don't plan to populate it from the get go I think we can remove it

Vince-janv (approved):
Nice 👍

cg/store/models.py
diitaz93: done

cg/store/models.py
diitaz93: done

Issue #3520, opened 2024-08-06
Bug: Error when submitting a microbial order


moedarrah:
In the order portal with a microbial order:

`message: "400: Unavailable well: Invalid format for row. Expected numeric but got A"`
 
<img width="812" alt="Screenshot 2024-08-06 at 10 24 00" src="https://github.com/user-attachments/assets/dbf2f46c-1afe-4c39-b6c9-ff86e4ec9d5b">

moedarrah:
Multiple tickets were created in SUSY, despite the frontend app indicating that the order submission had failed.

seallard:
Same underlying cause as in https://github.com/Clinical-Genomics/clinical-genomics-ui/issues/502. Closing this as a duplicate.

The way our deploys are setup prevents me from deploying the old version to stage.

seallard:
@islean could not reproduce the issue in stage, I think this was related to another bug which has been patched now 👍

seallard:
Reproduced with the following order form 
[1603.11.microbial.xlsx](https://github.com/user-attachments/files/16519453/1603.11.microbial.xlsx)

seallard:
Error
```
Aug 07 08:58:16 cg-services-stage.scilifelab.se podman[3117263]: 10.0.2.100 - - [07/Aug/2024:08:58:16 +0200] "POST /api/v1/submit_order/microsalt HTTP/1.0" 500 88 "https://clinical-genomics-stage.web.app/" "Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:128.0) Gecko/20100101 Fi>
Aug 07 08:58:16 cg-services-stage.scilifelab.se podman[3117263]: requests.exceptions.HTTPError: 400: Unavailable well: Invalid format for row. Expected numeric but got A
Aug 07 08:58:16 cg-services-stage.scilifelab.se podman[3117263]:     raise requests.exceptions.HTTPError(message, response=response)
Aug 07 08:58:16 cg-services-stage.scilifelab.se podman[3117263]:   File "/usr/local/lib/python3.11/site-packages/genologics/lims.py", line 202, in validate_response
Aug 07 08:58:16 cg-services-stage.scilifelab.se podman[3117263]:     self.validate_response(response, accept_status_codes)
Aug 07 08:58:16 cg-services-stage.scilifelab.se podman[3117263]:   File "/usr/local/lib/python3.11/site-packages/genologics/lims.py", line 209, in parse_response
[...]

PR #3531, opened 2024-08-07
Fix invalid microbial order form fixture

seallard:
Resolves https://github.com/Clinical-Genomics/cg/issues/3520

The microbial order form fixture was invalid (specifying wells for tube samples).
To avoid confusion in the future, this PR corrects it.

We are adding better order validation in a separate project which would have given an earlier error message.


islean:
It renders very weird for me. Also, would be good to have at least one well plate sample with well_position set to test that behaviour.

image

islean (approved):
💯

PR #3543, opened 2024-08-08
Implement convert errors method

islean:
## Description

This PR implements the `convert_errors` method, translating Pydantics errors from validating the models into our own Error classes.

## Changed
- convert_errors are now implemented and tests have been added.


tests/services/order_validation_service/test_tomte_field_validator.py
```diff
@@ -0,0 +1,24 @@
...
+    tomte_model_validator: TomteModelValidator,
+):
+    # GIVEN a valid order
+    order = valid_order.model_dump(by_alias=True)
+
+    # WHEN parsing the order
+    order, errors = tomte_model_validator.validate(order)
+
+    # THEN the parsed order is returned
+    assert order is not None
```
islean, line 18: Is there a benefit to specifying "is not None" here?

seallard: Does not matter, can skip

seallard (approved):
🥳

PR #3549, opened 2024-08-09
Refactor flow cells endpoint

seallard:


cg/services/flow_cell_service/flow_cell_service.py
```diff
@@ -0,0 +1,12 @@
...
+from cg.store.store import Store
+
+
+class FlowCellService:
+    def __init__(self, store: Store):
+        self.store = store
+
+    def get_sequencing_metrics(self, flow_cell_name: str) -> list[SequencingMetrics]:
+        run = self.store.get_illumina_sequencing_run_by_device_internal_id(flow_cell_name)
+        return create_metrics_dto(run.sample_metrics) if run else []
```
islean, line 12: Are we not losing the "400 bad request" behaviour now? Which was previously returned when the flow cell was not found

islean: And error handling for when sequencing metrics are not found.

tests/models/server/dto/test_sequencing_metrics.py
```diff
@@ -1,33 +0,0 @@
...
-            novaseq_x_flow_cell_id
-        ).sample_metrics
-    )
-    assert sequencing_metrics
-
-    # WHEN parsing the sequencing metrics into a request
-    parsed_metrics = parse_metrics_into_request(sequencing_metrics)
-
-    # THEN the parsed metrics should be equal to the original metrics
-    for pos in range(1, len(sequencing_metrics)):
```
Vince-janv, line 23:
```suggestion
for pos in range(len(sequencing_metrics)):
```
Why are we starting the range with 1?

Vince-janv (approved):
Looks good :+1

tests/models/server/dto/test_sequencing_metrics.py
seallard: No clue, but this test was replaced.

cg/services/flow_cell_service/flow_cell_service.py
seallard: Good point,that behavior is reasonable. The way it is written currently, the response would always be 200 but with an empty list. But it is better to be explicit about the resource not existing 👍

cg/services/flow_cell_service/flow_cell_service.py
```diff
@@ -0,0 +1,12 @@
+from cg.services.flow_cell_service.dtos import SequencingMetrics
+from cg.services.flow_cell_service.utils import create_metrics_dto
+from cg.store.store import Store
+
+
+class FlowCellService:
```
ChrOertlin, line 6:
```suggestion
class IlluminaFlowCellService:
```

We should avoid introducing generic class names for "FlowCell" at this point. It is specific for illumina flow cells here. We started this separation in the post processing flow to avoid mixing run instrument services accidentally.

seallard: What should it be called instead?

cg/services/flow_cell_service/flow_cell_service.py
ChrOertlin: We should likely adjust the CRUD to raise an error if something with an internal ID returns nothing and handle the error here.

cg/store/crud/read.py
```diff
@@ -380,7 +380,7 @@ def get_illumina_metrics_entry_by_device_sample_and_lane(
 
     def get_illumina_sequencing_run_by_device_internal_id(
         self, device_internal_id: str
-    ) -> IlluminaSequencingRun:
+    ) -> IlluminaSequencingRun | None:
```
ChrOertlin, line 383: add error handling if sequencing run not found.

cg/services/flow_cell_service/flow_cell_service.py
seallard: Yeah, I thought about that too! Kind of scary to change the behavior though because the method is called in a lot of places. I think I will consider that out of scope here

cg/services/flow_cell_service/flow_cell_service.py
seallard: Changed naming https://github.com/Clinical-Genomics/cg/pull/3549/commits/002147338af541d15080e57049fcf424931ced07

cg/store/crud/read.py
islean: Should that be here or in the calling function though? I think it is probably up to the caller to decide on what to do with that info?

tests/services/sample_run_metrics_service/conftest.py
```diff
@@ -0,0 +1,11 @@
+import pytest
+
+from cg.services.sample_run_metrics_service.sample_run_metrics_service import (
+    SampleRunMetricsService,
+)
+from cg.store.store import Store
+
+
+@pytest.fixture
+def flow_cell_service(store_with_illumina_sequencing_data: Store):
```
islean, line 10: Should this also be renamed?

cg/store/crud/read.py
seallard: I think for these basic data access functions where we provide a unique identifier, it makes sense to raise an exception - it is really weird if we don't get a hit for the identifier.

islean: I might have written some `is_entry_in_status_db` methods which are basically wrapping the returns of these methods with a bool... Probably bad practice but something that needs to be looked at since they will crash with the try except change.

cg/services/flow_cell_service/flow_cell_service.py
seallard: Fixed in https://github.com/Clinical-Genomics/cg/pull/3549/commits/b793a370e72b6eaf8fceef6108bb62c1fb4ea3ae

cg/store/crud/read.py
seallard: Fixed in separate PR

PR #3551, opened 2024-08-12
refactor(error handling crud)

ChrOertlin:
## Description

Adds raising a value error in the crud when retrieving entires by unique identifiers for illumina metrics and flow cells

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/store/crud/create.py
+++ b/cg/store/crud/create.py
@@ -15,6 +15,7 @@ from cg.services.illumina.data_transfer.models import (
 )
 from cg.store.base import BaseHandler
 from cg.store.database import get_session
+from cg.store.exc import EntryNotFoundError, EntryAlreadyExistsError
 from cg.store.models import (
     Analysis,
     Application,
@@ -400,7 +401,7 @@ class CreateHandler(BaseHandler):
         """Add a new Illumina flow cell to the status database as a pending transaction."""
         try:
             self.get_illumina_flow_cell_by_internal_id(flow_cell_dto.internal_id)
-        except ValueError:
+        except EntryNotFoundError:
             new_flow_cell = IlluminaFlowCell(
                 internal_id=flow_cell_dto.internal_id,
                 type=flow_cell_dto.type,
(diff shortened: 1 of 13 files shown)


cg/store/crud/create.py
```diff
@@ -398,16 +398,18 @@ def link_case_to_order(order_id: int, case_id: int):
...
+        except ValueError:
+            new_flow_cell = IlluminaFlowCell(
+                internal_id=flow_cell_dto.internal_id,
+                type=flow_cell_dto.type,
+                model=flow_cell_dto.model,
+            )
+            self.session.add(new_flow_cell)
+            LOG.debug(f"Flow cell added to status db: {new_flow_cell.internal_id}.")
+            return new_flow_cell
+        raise ValueError(f"Flow cell already exists: {flow_cell_dto.internal_id}")
```
seallard, line 412: I think it is better to raise a custom exception here, like `FlowCellNotFound` or something similar

seallard (approved):
Nice! I think it makes sense in our context to treat a missing record in these methods as an exceptional case rather than having an optional return type. We always expect an entry to exist when we call these methods.

It would be nice to lift out the error handling into a decorator, but lets leave it for now 👍 It would be great to have more descriptive names for the exceptions than just reusing the value error

cg/store/crud/create.py
ChrOertlin: I added an "EntryNotFoundError" and "EntryAlreadyExistsError" where applicable.

ChrOertlin:
> Nice! I think it makes sense in our context to treat a missing record in these methods as an exceptional case rather than having an optional return type. We always expect an entry to exist when we call these methods.
> 
> It would be nice to lift out the error handling into a decorator, but lets leave it for now 👍 It would be great to have more descriptive names for the exceptions than just reusing the value error

I can add the decorators in another PR!

PR #3566, opened 2024-08-15
Patch(clinical-delivery) Remove -p flag for rsync

Vince-janv:
## Description
Currently the permissions set when using rsync to transfer files to the delivery server are wrong (see [this issue](https://github.com/Clinical-Genomics/servers/issues/1430)).

This PR removes the permission flags when used for rsync. This together with changing the umask in the bashrc creates files with the correct permissions.

### Changed

- Removed -p flag for rsync


karlnyr (approved):
Nice! ✨

Vince-janv:

PR #3567, opened 2024-08-15
Operate on new samples and cases

seallard:


cg/services/order_validation_service/models/case.py
```diff
@@ -13,8 +13,20 @@ class Case(BaseModel):
...
         return enumerate(self.samples)
 
-    model_config = ConfigDict(str_min_length=1)
+    @property
+    def is_new(self) -> bool:
+        return bool(self.internal_id)
+
+    @property
+    def enumerated_new_samples(self):
+        samples = []
```
islean, line 28: This should be type hinted

cg/services/order_validation_service/workflows/tomte/models/order.py
```diff
@@ -8,3 +8,11 @@ class TomteOrder(Order):
     @property
     def enumerated_cases(self) -> enumerate[TomteCase]:
         return enumerate(self.cases)
+
+    @property
+    def enumerated_new_cases(self) -> enumerate[TomteCase]:
+        cases = []
```
islean, line 14: Type hint

islean (approved):
💯

PR #3574, opened 2024-08-15
refactor(submitters)

ChrOertlin:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/upload/gisaid/constants.py
```diff
@@ -14,7 +14,7 @@
 ]
 
 HEADERS = [
-    "submitter",
+    "submitters",
```
ChrOertlin, line 17:
```suggestion
    "submitter",
```

cg/services/orders/order_lims_service/order_lims_service.py
```diff
@@ -0,0 +1,39 @@
...
+
+    def __init__(self, lims_api: LimsAPI):
+        self.lims_api = lims_api
+
+    @staticmethod
+    def _build_lims_sample(customer: str, samples: list[OrderInSample]) -> list[LimsSample]:
+        """Convert order input to lims interface input."""
+        samples_lims = []
+        for sample in samples:
+            dict_sample = sample.__dict__
```
islean, line 21: This conversion to a dict seems weird to me. Is there a reason for it?

ChrOertlin: Legacy code, this refactoring is not mean to address to code at that level at the moment. For now this refactoring is to inject dependencies and make future reworking easier.

cg/meta/orders/api.py
```diff
@@ -11,44 +11,36 @@
...
+        OrderType.MICROSALT: MicrobialOrderSubmitter,
+        OrderType.MIP_DNA: GenericOrderSubmitter,
+        OrderType.MIP_RNA: GenericOrderSubmitter,
+        OrderType.RML: PoolOrderSubmitter,
+        OrderType.RNAFUSION: GenericOrderSubmitter,
+        OrderType.SARS_COV_2: MicrobialOrderSubmitter,
+        OrderType.TOMTE: GenericOrderSubmitter,
     }
     if project in submitters:
         return submitters[project](lims=lims, status=status)
```
ChrOertlin, line 46: I am unsure how to build these here. Since the Submitters require dependencies, they need to be built.
we can either use a factory using multiple OrderType -> service maps 
or we use a registry where we pre-instantiate the submitters. 

I think we are a bit limited here since this all happens in flask and is bit disjoint from the rest of CG as it seems

cg/meta/orders/api.py
```diff
@@ -11,47 +11,150 @@
...
+
+    def create_order_submitter(self, order_type: OrderType) -> OrderSubmitter:
+        service_classes = order_service_mapping.get(order_type)
+
+        if not service_classes:
+            raise ValueError(f"No services found for order type: {order_type}")
+
+        lims_service, validate_service_class, store_service_class, submitter_class = service_classes
+
+        # Instantiate the necessary services
```
ChrOertlin, line 146:
```suggestion

```

cg/meta/orders/api.py
```diff
@@ -11,47 +11,150 @@
...
+        lims_service: OrderLimsService = lims_service(self.lims_api)
+        validate_service: ValidateOrderService = validate_service_class(self.status_db)
+        store_service: StoreOrderService = store_service_class(
+            store=self.status_db, lims=lims_service
+        )
+
+        # Instantiate and return the corresponding OrderSubmitter
```
ChrOertlin, line 153:
```suggestion

```

cg/meta/orders/api.py
```diff
@@ -11,47 +11,148 @@
...
+
+        lims_service: OrderLimsService = lims_service(self.lims_api)
+        validate_service: ValidateOrderService = validate_service_class(self.status_db)
+        store_service: StoreOrderService = store_service_class(
+            store=self.status_db, lims=lims_service
+        )
+
+        return submitter_class(
+            validate_order_service=validate_service,
+            store_order_service=store_service,
```
ChrOertlin, line 154: This is a factory approach, alternatively we can use a registry where we pre-initialise the different submitters elsewhere. 

It seems like flask is a bit weird where it gets the lims and statusdb connection and there is no config that we pass.

cg/services/orders/order_lims_service/order_lims_service.py
islean: Expected as much. Some comments on the PR might just be ranting about stuff that's out of scope...

seallard (commented):
I think this seems less confusing than before 👍 

Long term, I'd encourage taking a step back and thinking about why we need multiple separate "store_order_services". I think the actual variation is introduced in how the different orders are parsed. In the end, the same database records are created

cg/meta/orders/api.py
```diff
@@ -11,47 +11,148 @@
...
+    ValidateMicrobialOrderService,
+)
+from cg.services.orders.validate_order_services.validate_pool_order import ValidatePoolOrderService
 from cg.store.store import Store
 
 LOG = logging.getLogger(__name__)
 
+order_service_mapping = {
+    OrderType.BALSAMIC: (
+        OrderLimsService,
```
diitaz93, line 53: It seems like the LimsService is the same for all of them, why not instantiate it in the same way for all orders?

cg/meta/orders/api.py
```diff
@@ -11,47 +11,148 @@
...
+
+    def create_order_submitter(self, order_type: OrderType) -> OrderSubmitter:
+        service_classes = order_service_mapping.get(order_type)
+
+        if not service_classes:
+            raise ValueError(f"No services found for order type: {order_type}")
+
+        lims_service, validate_service_class, store_service_class, submitter_class = service_classes
+
+        lims_service: OrderLimsService = lims_service(self.lims_api)
```
diitaz93, line 146: so regarding my previous comment, would this be possible for all order types?
```suggestion
        lims_service = OrderLimsService(self.lims_api)
```

ChrOertlin:
> I think this seems less confusing than before 👍
> 
> Long term, I'd encourage taking a step back and thinking about why we need multiple separate "store_order_services". I think the actual variation is introduced in how the different orders are parsed. In the end, the same database records are created

Yes, I think that is the next step. To see and identify what is clutter and can be simplified. I think one of the main issues is that the OrderIn comes with different models based on OrderType hat not all have the same information which requires a different handling. (i.e. generic order, fastq, pool etc.).

cg/meta/orders/api.py
ChrOertlin: yes, I am not sure about this at the moment. Something we should discuss. I am also unsure about using a factory instead of a registry.

cg/services/orders/store_order_services/store_generic_order.py
```diff
@@ -1,99 +1,40 @@
...
-        for sample in samples:
-            if self._is_rerun_of_existing_case(sample=sample):
-                continue
-            if self.status.get_case_by_name_and_customer(
-                customer=customer, case_name=sample.family_name
-            ):
-                raise OrderError(f"Case name {sample.family_name} already in use")
-
-    def submit_order(self, order: OrderIn) -> dict:
+class StoreGenericOrderService(StoreOrderService):
```
islean, line 17: From this naming I would expect this service to work for all order flows. I think the previous name (`CaseSubmitter)` was chosen to reflect that it was used for workflows where the customers provide data about the cases as well as the samples. It was not a perfect naming but maybe it would be possible to choose a name implying something similar?

cg/meta/orders/api.py
diitaz93: I like more a pre-initialisation approach, but if Flask complains this is alternative is ok

cg/services/orders/submitters/order_submitter.py
```diff
@@ -0,0 +1,43 @@
+"""Abstract base classes for order submitters."""
+
+from abc import ABC, abstractmethod
+
+from cg.models.orders.order import OrderIn
+from cg.services.orders.order_lims_service.order_lims_service import OrderLimsService
+from cg.store.store import Store
+
+
+class ValidateOrderService(ABC):
```
islean, line 10: Should this not be in the validate_order_services folder?

cg/services/orders/submitters/order_submitter.py
```diff
@@ -0,0 +1,43 @@
...
+    @abstractmethod
+    def __init__(self, status_db: Store):
+        self.status_db = status_db
+
+    @abstractmethod
+    def validate_order(self, order_in: OrderIn):
+        pass
+
+
+class StoreOrderService(ABC):
```
islean, line 20: Should this not be in the store_order_services folder?

cg/services/orders/store_order_services/store_fastq_order_service.py
```diff
@@ -126,11 +132,28 @@ def store_items_in_status(
...
+                self.status_db.session.add_all([case, new_relationship])
 
-        self.status.session.add_all(new_samples)
-        self.status.session.commit()
+        self.status_db.session.add_all(new_samples)
+        self.status_db.session.commit()
         return new_samples
+
+    @staticmethod
+    def _fill_in_sample_ids(samples: list[dict], lims_map: dict, id_key: str = "internal_id"):
```
islean, line 145: Was there a reason for moving this from the parent class? Was it to add clarity?

cg/services/orders/submitters/fastq_order_submitter.py
```diff
@@ -0,0 +1,23 @@
+from cg.models.orders.order import OrderIn
+from cg.services.orders.store_order_services.store_fastq_order_service import StoreFastqOrderService
+from cg.services.orders.submitters.order_submitter import OrderSubmitter
+from cg.services.orders.validate_order_services.validate_fastq_order import (
+    ValidateFastqOrderService,
+)
+
+
+class FastqOrderSubmitter(OrderSubmitter):
```
islean, line 9: They get so clean now 🥳

cg/services/orders/store_order_services/store_generic_order.py
```diff
@@ -1,99 +1,40 @@
...
+
+    def __init__(
+        self,
+        status_db: Store,
+        lims_service: OrderLimsService,
+    ):
+        self.status_db = status_db
+        self.lims = lims_service
+
+    def store_order(self, order: OrderIn) -> dict:
```
diitaz93, line 37: Why does a `store` function return a dict? This is a bit confusing, maybe the storing should be one function and another "getter" function should be implemented if an order in a dict form is required

cg/services/orders/submitters/generic_order_submitter.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(
+        self,
+        order_validation_service: ValidateGenericOrderService,
+        order_store_service: StoreGenericOrderService,
+    ):
+        self.order_validation_service = order_validation_service
+        self.order_store_service = order_store_service
+
+    def submit_order(self, order_in: OrderIn) -> dict:
+        """Submit a generic order."""
```
diitaz93, line 32: What is submitting exactly? validating, storing and returning a dict?

cg/meta/orders/api.py
islean: A bit uncertain if I understand but I guess we could instantiate them in ext.py maybe? Uncertain if that is preferable but I think we can get the lims and statusdb connection there if we want.

diitaz93 (commented):
I really like the simplification of the submission process into Validation and Storing, I think this is the way to go 💯 . It is however confusing to me that the store methods return something. Seems to me that a store method should only store data in the database, not return anything. Maybe there should be an extra step of creating an order in a dict form before validating and storing. So basically divide the submission into order creation, validation and storing.

islean (approved):
I do think this is a massive improvement!

cg/services/orders/store_order_services/store_generic_order.py
islean: The dict return object is one of the most confusing objects that I have seen and should definitely be rewritten at some point. Especially because the structure of the dict depends on the workflow. Probably not in scope though?

cg/meta/orders/api.py
ChrOertlin: This is what happens now basically. but I think with a registry it just gets instantiated once and then fetched when needed. The factory instantiates and returns on very call.

cg/services/orders/store_order_services/store_generic_order.py
ChrOertlin: Yes I agree, right now it returns something which should not be in the future. It should just take a model, translate it into database models and store the stuff. Yet, currently the submitters all return a dict. To not break functionality for now I wanted to extract the existing logic as is and improve upon it later on.

cg/services/orders/submitters/generic_order_submitter.py
ChrOertlin: Ideally submitting an order later on would be:
1. Consuming a json order / submitted order via the front end
2. Validating the order using a validator service
3. Translating the order into relevant database models via a transfer service
4. Storing these models in the database by ideally a single store service
5. Returning a Response to the end point that can be evaluated there.

This way we can encapsulate all concerns into their own modules with a single responsibility. Right now it is a bit away from that still

cg/services/orders/submitters/order_submitter.py
ChrOertlin: Yes, it was here for now just to display the overall design

cg/services/orders/submitters/order_submitter.py
ChrOertlin: Yes, it was here for now just to display the overall design

cg/services/orders/store_order_services/store_fastq_order_service.py
ChrOertlin: The overall idea is explained here:
https://github.com/Clinical-Genomics/cg/pull/3574#discussion_r1719593949

Ideally the store service will be the same for all later on with different data transfer services as required. Right now this implementation is not there as the code needs to be further disentangled.

cg/meta/orders/api.py
```diff
@@ -11,66 +11,36 @@
...
+        lims: LimsAPI,
+        status: Store,
+        osticket: OsTicket,
+        submitter_registry: OrderSubmitterRegistry,
+    ):
         super().__init__()
         self.lims = lims
         self.status = status
         self.ticket_handler: TicketHandler = TicketHandler(osticket_api=osticket, status_db=status)
+        self.submitter_registry = submitter_registry
```
ChrOertlin, line 36: opted for a registry that gets instantiated with the flask app. It is more efficient if submitters get re-used, which is the case.

cg/meta/orders/api.py
ChrOertlin: re worked into a registry

ChrOertlin:
# tests 

## Generic order submitter using mip.json

## new order portal
image

## Susy

image

## Lims

image

## Status
### samples
image
### case sample
image
### case
image
### order
image

ChrOertlin:
# tests

## pool order submitter

### order portal
image
### Susy 
image
### Lims 
image
### status
#### sample
image
#### pool
image
#### case
image
#### case sample
image
#### order
image

ChrOertlin:
I highlighted the further process to improve the order submitters in this issue here:
https://github.com/Clinical-Genomics/cg/issues/3592

cg/services/orders/validate_order_services/validate_microbial_order.py
```diff
@@ -1,29 +1,34 @@
...
+        self.status_db = status_db
 
     def validate_order(self, order: OrderIn) -> None:
-        """Validate order sample names."""
-        super().validate_order(order=order)
-        self._validate_sample_names_are_available(samples=order.samples, customer_id=order.customer)
+        if order.order_type == OrderType.SARS_COV_2:
+            self._validate_sample_names_are_available(
+                samples=order.samples, customer_id=order.customer
+            )
```
Vince-janv, line 19: Does this not warrant a `ValidateSarsCov2OrderService`? I feel like having an if-statement here defeats the purpose of the nice registry you set up earlier

cg/services/orders/store_order_services/store_fastq_order_service.py
Vince-janv: But it seems like we have a method, `_fill_in_sample_ids` which is identical in all child classes. What prevents it from being moved to `StoreOrderService`?

Vince-janv (approved):
Nice! This turned out even better than i expected ⭐ I do feel like we should test the other order types as well on stage though, before merging.

cg/services/orders/store_order_services/store_generic_order.py
ChrOertlin: All of the order generate Cases, hence why I thought Case is not really meaningful. I used Generic here since the class would be used for the most use cases we have, that is all major workflows. I clarified the docstring what it is used for.

That being said since both of you think Case is clearer I will rename it.

cg/services/orders/store_order_services/store_fastq_order_service.py
ChrOertlin: moved it, Wanted to keep the ABC classes clean with the future refactor of the store service in mind which would hopefully reduce it to a single service. But we can tackle that later on.

cg/services/orders/validate_order_services/validate_microbial_order.py
ChrOertlin: In the Improve-order-flow new backend validation services are made. I put it here with that in mind. So in the future the more "generic" validation services would be replaced and the current one would be phased out.

PR #3576, opened 2024-08-15
Feat/add freshdesk properties

ahdamin:
### Added

- Ticket properties:
1. Type
2. Tags: data analysis
3. custom field: environment

### DIFF
--- a/cg/clients/freshdesk/freshdesk_client.py
+++ b/cg/clients/freshdesk/freshdesk_client.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from requests import Response, Session
 from requests.adapters import HTTPAdapter
 from urllib3 import Retry
+from typing import Dict, Union
 
 from cg.clients.freshdesk.constants import EndPoints
 from cg.clients.freshdesk.models import TicketCreate, TicketResponse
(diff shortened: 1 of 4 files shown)


cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -37,6 +37,24 @@ def create_ticket(self, ticket: TicketCreate, attachments: list[Path] = None) ->
         response.raise_for_status()
         return TicketResponse.model_validate(response.json())
 
+    def _convert_ticket_to_multipart_data(
```
seallard, line 40: This logic does not belong in the `FreshdeskClient`. The client should be really stupid, just sending requests and returning responses parsed into models.

I suggest moving it into the service wrapping around the client, the ticket handler and then passing the correctly formatted data into the client.

seallard: Additionally, assuming that the `_convert_ticket_to_multipart_data` belongs here in the client, it does not require any state from the class and as such, it just bloats the client class with unnecessary implementation details. It should be extracted to a standalone function to reduce clutter in the class.

This point is moot given that the logic does not belong in the client.

ahdamin:
Moved multipart handling to `TicketCreate` model

PR #3578, opened 2024-08-16
Validate existing samples do in fact exist

seallard:


tests/services/order_validation_service/test_data_validators.py
```diff
@@ -120,4 +123,22 @@ def test_case_internal_ids_does_not_exist(
...
+    # GIVEN an order with a sample marked as existing but which does not exist in the database
+    valid_order.cases[0].samples[0].internal_id = "Non-existent sample"
+
+    # WHEN validating that the samples exists
+    errors = validate_samples_exist(order=valid_order, store=store_with_multiple_cases_and_samples)
+
+    # THEN an error should be returned
+    assert errors
+
+    # THEN the error should concern the non-existent case
```
islean, line 143:
```suggestion
    # THEN the error should concern the non-existent sample
```

cg/services/order_validation_service/models/case.py
```diff
@@ -30,3 +30,11 @@ def enumerated_new_samples(self):
             if sample.is_new:
                 samples.append((sample_index, sample))
         return samples
+
+    @property
+    def enumerated_existing_samples(self):
```
islean, line 35:
```suggestion
    def enumerated_existing_samples(self) -> list[tuple[int, Sample]]:
```

islean (approved):
Mega

PR #3581, opened 2024-08-16
Ensure fields are nullable

seallard:


cg/services/order_validation_service/models/case.py
```diff
@@ -7,11 +7,11 @@
 
 
 class Case(BaseModel):
-    data_delivery: DataDelivery
```
islean, line 10: On what level is data_delivery even sent? Since Cases don't exist in the FE request currently...

cg/services/order_validation_service/models/case.py
```diff
@@ -7,11 +7,11 @@
...
+    data_delivery: DataDelivery | None = None
     internal_id: str | None = None
-    name: str = Field(pattern=NAME_PATTERN, min_length=2, max_length=128)
-    priority: PriorityTerms = PriorityTerms.STANDARD
-    samples: list[Sample]
+    name: str | None = Field(default=None, pattern=NAME_PATTERN, min_length=2, max_length=128)
```
islean, line 12: Should this not be required for now? It is what the user will always provide, since internal_ids are not known by them.

cg/services/order_validation_service/models/case.py
```diff
@@ -7,11 +7,11 @@
...
+    priority: PriorityTerms | None = None
```
islean, line 13: Depends on what we want but currently priority is required in the Frontend but not used in the backend for existing cases

cg/services/order_validation_service/models/case.py
```diff
@@ -7,11 +7,11 @@
...
+    samples: list[Sample] | None = None
```
islean, line 14: This should always be required

seallard:
Unclear requirements, lets patch later

PR #3583, opened 2024-08-16
Validate existing fathers

seallard:


cg/services/order_validation_service/workflows/tomte/models/sample.py
```diff
@@ -10,9 +10,9 @@ class TomteSample(Sample):
     age_at_sampling: float | None = None
     control: ControlEnum | None = None
     elution_buffer: str | None = None
-    father: str | None = Field(None, pattern=NAME_PATTERN)
+    father: str = Field(pattern=NAME_PATTERN)
     formalin_fixation_time: int | None = None
-    mother: str | None = Field(None, pattern=NAME_PATTERN)
+    mother: str = Field(pattern=NAME_PATTERN)
```
islean, line 15: I think they are still nullable (a sample may not have a mother/father in the case)

cg/services/order_validation_service/workflows/tomte/models/case.py
```diff
@@ -21,7 +21,7 @@ def get_sample(self, sample_name: str) -> TomteSample | None:
                 return sample
 
     def get_samples_with_father(self) -> list[tuple[TomteSample, int]]:
-        return [(sample, index) for index, sample in self.enumerated_new_samples if sample.father]
+        return [(sample, index) for index, sample in self.samples if sample.father]
```
islean, line 24:
```suggestion
        return [(sample, index) for index, sample in self.enumerated_samples if sample.father]
```

PR #3594, opened 2024-08-19
(Improve order flow) Add order validation response test

islean:
## Description

This PR adds an integration test for order validation and fixes previous tests which were never run due to a faulty file name.

### Added

- Integration tests for the Order validation service

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/order_validation_service.py
+++ b/cg/services/order_validation_service/order_validation_service.py
@@ -5,5 +5,5 @@ from cg.services.order_validation_service.models.errors import ValidationErrors
 
 class OrderValidationService(ABC):
     @abstractmethod
-    def validate(self, order_json: str) -> ValidationErrors:
+    def _get_errors(self, order_json: str) -> dict:
         pass
--- a/cg/services/order_validation_service/order_validation_service.py
+++ b/cg/services/order_validation_service/order_validation_service.py
@@ -1,9 +1,7 @@
 from abc import ABC, abstractmethod
 
-from cg.services.order_validation_service.models.errors import ValidationErrors
-
 
 class OrderValidationService(ABC):
     @abstractmethod
-    def _get_errors(self, order_json: str) -> dict:
+    def validate(self, order_json: str) -> dict:
         pass
--- a/cg/services/order_validation_service/order_validation_service.py
+++ b/cg/services/order_validation_service/order_validation_service.py
@@ -3,5 +3,5 @@ from abc import ABC, abstractmethod
 
 class OrderValidationService(ABC):
     @abstractmethod
-    def validate(self, order_json: str) -> dict:
+    def validate(self, raw_order: dict) -> dict:
         pass
--- a/cg/services/order_validation_service/models/order.py
+++ b/cg/services/order_validation_service/models/order.py
@@ -9,7 +9,7 @@ TICKET_PATTERN = r"^#\d{4,}"
 class Order(BaseModel):
     comment: str | None = None
     connect_to_ticket: bool = False
-    customer_internal_id: str = Field(alias="customer")
+    customer: str
     delivery_type: DataDelivery
     name: str
     skip_reception_control: bool = False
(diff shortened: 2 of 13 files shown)


cg/services/order_validation_service/order_validation_service.py
```diff
@@ -5,5 +5,5 @@
 
 class OrderValidationService(ABC):
     @abstractmethod
-    def validate(self, order_json: str) -> ValidationErrors:
+    def _get_errors(self, order_json: str) -> dict:
```
seallard, line 8: This should be called `validate` (just rename it in this file)

seallard (approved):
🚀

ChrOertlin:
As a general comment not for this PR.

Can you move this whole service under `cg/services/orders` ?

cg/services/order_validation_service/order_validation_service.py
seallard: Also, it would be great if we returned an actual pydantic object instead of a dict. It makes the type signature of the function a bit mysterious and says nothing about the structure of it. Lets save that for later though.

cg/services/order_validation_service/order_validation_service.py
```diff
@@ -1,9 +1,7 @@
 from abc import ABC, abstractmethod
 
-from cg.services.order_validation_service.models.errors import ValidationErrors
-
 
 class OrderValidationService(ABC):
     @abstractmethod
-    def validate(self, order_json: str) -> ValidationErrors:
+    def validate(self, order_json: str) -> dict:
```
ChrOertlin, line 6: also, you should start thinking about where in the order submission this is supposed to go. Will this really be a JSON or something of type OrderIn that is created in the submit_order endpoint

seallard: Good point! We should think about that, but we have a lot of work to do still before starting to integrate it.

Maybe we should return something indicating whether the validation passed/failed along side the validation response. Something like
```python
class ValidationResult(BaseModel):
    passed: bool
    data: dict

islean: I do think that this type hint is outdated, it should be a dict in my mind. I would also like us to move away from the OrderIn object at some point so I do not think it is necessarily useful to constrain ourselves to it. In practice, the TomteOrder, TomteCase and TomteSample models in this project should be better suited and swapped in for the old OrderIn objects. 

We do need to put some thought into how much would need to change in the submit_order endpoint for that to happen. But I really do not wish the OrderIn objects to stay around for long (they do not even reflect the structure in the frontend and seem only built to be flexible for each workflow). Also, I do not think there should be any logic converting the request to Pydantic models in the endpoint, especially since that is part of the validation. So I see no world in which we would want a pydantic model as input to this method. But there are indeed areas which we need to focus on with regards to the existing flow.

islean: We would ideally like to rewrite the submit_order logic altogether.

ChrOertlin: Indeed, not for this PR.

The validate function could also return nothing and raise an error that can be caught in the relevant place.

seallard: That is also an option 👍 As long as we cleanly encapsulate any logic related to the validation process in the validation service.

islean: Also - we do not wish to raise exceptions in this case, actually. The response will be used in the frontend to showcase to any customers what they need to change. So we would need to return a mapped error message to the correct part of the request stating what is wrong with it.

islean: I will resolve for now since it is out of scope for this PR.

PR #3595, opened 2024-08-19
Add microsalt order validation skeleton

seallard:


cg/services/order_validation_service/workflows/microsalt/validation/field/utils.py
```diff
@@ -0,0 +1,7 @@
+from pydantic_core import ValidationError
+
+from cg.services.order_validation_service.models.errors import ValidationErrors
+
+
+def convert_errors(pydantic_errors: ValidationError) -> ValidationErrors:
```
seallard, line 6: Not sure about this one, I think we should move the existing `convert_errors` up from the tomte module and extend it instead.

cg/services/order_validation_service/workflows/microsalt/models/sample.py
```diff
@@ -0,0 +1,20 @@
...
+class MicroSaltSample(Sample):
+    control: ControlEnum | None = None
+    elution_buffer: str | None = None
+    extraction_method: str | None = None
+    organism: str | None = Field(default=None, max_length=32)
+    priority: PriorityEnum | None = None
+    quantity: int | None = None
+    reference_genome: str | None = Field(default=None, max_length=255)
+    sample_concentration: float | None = None
+    subject_id: str = Field(pattern=NAME_PATTERN, max_length=128)
```
islean, line 20: Remove subject_id. It can not be set in MicroSALT

islean (approved):
Looks good!

islean (approved):
Nice! Now we need to add some rules!

PR #3601, opened 2024-08-19
add(fetch delivery files service)

ChrOertlin:
## Description

Adds services to fetch files to delivery for:
only fastq delivery
analysis files delivery
combined fastq and analysis delivery

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/file_delivery/fetch_file_service/fetch_analysis_files_service.py
```diff
@@ -0,0 +1,76 @@
...
+        sample_tags_with_sample_id: list[set[str]] = [tag | {sample_id} for tag in sample_tags]
+        sample_files: list[File] = self.hk_api.get_files_from_latest_version_containing_tags(
+            bundle_name=case_id, tags=sample_tags_with_sample_id
+        )
+        return [
+            SampleFile(case_id=case_id, sample_id=sample_id, file_path=sample_file.full_path)
+            for sample_file in sample_files
+        ]
+
+    def _get_analysis_sample_delivery_files(self, case: Case) -> list[SampleFile]:
```
ivadym, line 52: We'll be using the sample and case-specific delivery files for the delivery report, unless we decide to change the logic there. And maybe we should, unless we want to handle the formatting at this stage. 

So we would need to rely on the DeliveryFiles model for the delivery report, using the case and sample files as needed. Sounds good to me 👍

cg/services/file_delivery/fetch_file_service/fetch_analysis_files_service.py
```diff
@@ -0,0 +1,76 @@
...
+        analysis_sample_files: list[SampleFile] = self._get_analysis_sample_delivery_files(
+            case=case
+        )
+        return DeliveryFiles(case_files=analysis_case_files, sample_files=analysis_sample_files)
+
+    def _get_case(self, case_id: str):
+        """Get the case."""
+        if case := self.status_db.get_case_by_internal_id(internal_id=case_id):
+            return case
+        raise EntryNotFoundError(f"Case {case_id} not found.")
```
ivadym, line 36: Doesn't feel right to have this general method here, but I guess it's ok for now. Seems like this checks will end up in a service of its own in the future

cg/services/file_delivery/fetch_file_service/fetch_fastq_analysis_files_service.py
```diff
@@ -0,0 +1,28 @@
...
+        self.hk_api = hk_api
+
+    def get_files_to_deliver(self, case_id: str) -> DeliveryFiles:
+        fetch_fastq_service = FetchFastqDeliveryFilesService(self.status_db, self.hk_api)
+        fetch_analysis_service = FetchAnalysisDeliveryFilesService(self.status_db, self.hk_api)
+        fastq_files = fetch_fastq_service.get_files_to_deliver(case_id)
+        analysis_files = fetch_analysis_service.get_files_to_deliver(case_id)
+        return DeliveryFiles(
+            case_files=analysis_files.case_files,
+            sample_files=fastq_files.sample_files + analysis_files.sample_files,
```
ivadym, line 27: I think you would need to make a distinction between sample and fastq files unless we want them to be delivered to the same folder, which is not how it's currently implemented.

cg/services/file_delivery/fetch_file_service/fetch_analysis_files_service.py
ChrOertlin: Yes, this is a weird one. Ideally this check is part of the CRUD. However, this would require a refactoring of all the places to handle the error properly. I don't like having this here.

cg/services/file_delivery/fetch_file_service/fetch_delivery_files_service.py
ivadym, line 1: I'm a bit confused about this. What is the purpose of this service, and how is it different from the analysis and FASTQ services?

cg/services/file_delivery/fetch_file_service/fetch_delivery_files_service.py
```diff
@@ -0,0 +1,34 @@
...
+        """Get the files to deliver."""
+        pass
+
+    @staticmethod
+    def get_analysis_sample_tags_for_workflow(workflow: Workflow) -> list[set[str]]:
+        return PIPELINE_ANALYSIS_TAG_MAP[workflow]["sample_tags"]
+
+    @staticmethod
+    def get_analysis_case_tags_for_workflow(workflow: Workflow) -> list[set[str]]:
+        return PIPELINE_ANALYSIS_TAG_MAP[workflow]["case_tags"]
```
seallard, line 34: I really like this interface 🌞 🚀 

However, would it be possible to skip the `get_analysis_sample_tags_for_workflow` and `get_analysis_case_tags_for_workflow`? They seem like very specific methods that might not need to be part of the interface but rather hidden away somewhere internally.

It might make a lot of sense to have them here though, it just made me hesitant.

cg/services/file_delivery/fetch_file_service/fetch_delivery_files_service.py
```diff
@@ -0,0 +1,34 @@
...
+    Abstract class that encapsulates the logic required for managing the files to deliver.
+    1. Get a workflow and data delivery option
+    2. Fetch the files based on tags from Housekeeper
+    3. Return the files to deliver
+    """
+
+    @abstractmethod
+    def __init__(self, status_db: Store, hk_api: HousekeeperAPI):
+        self.status_db = status_db
+        self.hk_api = hk_api
```
seallard, line 21: There is no need to specify a constructor for an abstract class

cg/meta/delivery/delivery.py
```diff
@@ -144,6 +144,24 @@ def get_analysis_case_delivery_files(self, case: Case) -> list[DeliveryFile]:
...
+    def get_delivery_files(self, case: Case, force: bool = False) -> list[DeliveryFile]:
+        """Return a complete list of files to be delivered to the customer's inbox."""
+        delivery_files: list[DeliveryFile] = []
+        if self.is_fastq_delivery(case.data_delivery):
+            fastq_files: list[DeliveryFile] = self.get_fastq_delivery_files(case=case, force=force)
+            delivery_files.extend(fastq_files)
+        if self.is_analysis_delivery(case.data_delivery):
+            analysis_case_files: list[DeliveryFile] = self.get_analysis_case_delivery_files(case)
+            analysis_sample_files: list[DeliveryFile] = self.get_analysis_sample_delivery_files(
+                case=case
```
seallard, line 156: no need for a named arg here

cg/services/file_delivery/fetch_file_service/fetch_analysis_files_service.py
```diff
@@ -0,0 +1,76 @@
...
+    def __init__(self, status_db: Store, hk_api: HousekeeperAPI):
+        self.status_db = status_db
+        self.hk_api = hk_api
+
+    def get_files_to_deliver(self, case_id: str) -> DeliveryFiles:
+        """Return a list of analysis files to be delivered for a case."""
+        case: Case = self._get_case(case_id)
+        analysis_case_files: list[CaseFile] = self.get_analysis_case_delivery_files(case)
+        analysis_sample_files: list[SampleFile] = self._get_analysis_sample_delivery_files(
+            case=case
```
seallard, line 28: redundant named arg

cg/services/file_delivery/fetch_file_service/fetch_analysis_files_service.py
seallard: It might be better to accept the current design where the return can be none instead of spreading helper methods like this, your call

cg/services/file_delivery/fetch_file_service/fetch_delivery_files_service.py
seallard: Alternatively, it seems like those two methods might make more sense to put in a separate abstract class? A Python class can inherit from multiple abstract classes, so it might be better to split it into two separate ones since the functionality is quite different (get files to deliver vs get case/sample tags).

ivadym: Also, just for the record! This tag map is to be removed and deprecated: https://github.com/Clinical-Genomics/hermes/issues/102. In the future we'll only use one tag, `clinical-delivery` to extract the files we want to rsync

seallard (approved):
Nice 👍 I think the interface can be improved, even though it looks nice! Consider removing the constructor and also splitting it into two abstract classes

cg/services/file_delivery/fetch_file_service/fetch_delivery_files_service.py
ChrOertlin: This here is the implementation of the abstract interface. Its purpose is to define a common interface for all its subclasses. Basically every subclass of FetchFeliveryFilesService have to implement the get_delivery_files methods that returns a DeliveryFiles model. So the subclasses can interchanged in a common flow depending on what logic is required.

Here we have deviating logic based on the data-analysis types, fastq, analysis and fastq and analysis.

cg/services/file_delivery/fetch_file_service/fetch_delivery_files_service.py
ChrOertlin: Thanks for the info @ivadym. And agreed @seallard, I will split it off!

cg/services/file_delivery/fetch_file_service/fetch_delivery_files_service.py
ivadym: Thanks for the clarification 💯

cg/services/file_delivery/fetch_file_service/fetch_delivery_files_service.py
islean: Doesn't Python support hybrid classes quite well though? Or is that a big no-no?

islean (commented):
The refactoring of delivery -> file_delivery introduces a lot of weird doc strings and potentially some bugs. Was the extent of the refactoring intended?

islean:
> The refactoring of delivery -> file_delivery introduces a lot of weird doc strings and potentially some bugs. Was the extent of the refactoring intended?

To be clear, I have not gone through all the files yet, so there might very well be more issues with the refactoring.

ChrOertlin:
> > The refactoring of delivery -> file_delivery introduces a lot of weird doc strings and potentially some bugs. Was the extent of the refactoring intended?
> 
> To be clear, I have not gone through all the files yet, so there might very well be more issues with the refactoring.

it is a mistake that I need to revert

alembic/versions/2021_04_19_1dadcefd3bbf_feat_contacts_as_emails.py
```diff
@@ -106,7 +106,7 @@ def downgrade():
     count = 0
     # replace email addresses on customers with connection between customer and user using users email
     for customer in session.query(Customer):
-        # delivery contact
+        # file_delivery contact
```
ChrOertlin, line 109:
```suggestion
        # delivery contact
```

alembic/versions/2021_09_28_c494649637d5_add_data_delivery_options.py
```diff
@@ -1,4 +1,4 @@
-"""Add data delivery options
+"""Add data file_delivery options
```
ChrOertlin, line 1:
```suggestion
"""Add data delivery options
```

cg/services/file_delivery/fetch_file_service/fetch_delivery_files_service.py
ChrOertlin: > There is no need to specify a constructor for an abstract class and mark it as abstract.
> 
 Unless you want to use them in a factory /registry method that assumes a certain format. Then this abstractmethod check if all subclasses have the same constructor is what I understand from it.

PR #3602, opened 2024-08-20
Validate microsalt app compatibility

seallard:


cg/services/order_validation_service/workflows/microsalt/models/sample.py
```diff
@@ -4,7 +4,7 @@
 from cg.services.order_validation_service.models.sample import Sample
 
 
-class MicroSaltSample(Sample):
+class MicrosaltSample(Sample):
```
islean, line 7: 😌

cg/services/order_validation_service/workflows/microsalt/validation_rules.py
```diff
@@ -18,6 +20,4 @@
 ]
 SAMPLE_RULES: list[callable] = [
     validate_application_compatibility,
-    validate_application_exists,
```
islean, line 21: Hey! Don't remove my validation 😤

islean: But the import should probably be changed...

islean (approved):
Looks good!

PR #3604, opened 2024-08-20
Add microsalt max one sample per well validation

seallard:


cg/services/order_validation_service/workflows/microsalt/validation/inter_field/utils.py
```diff
@@ -0,0 +1,35 @@
...
+                conflicting_samples.extend(samples_indices[1:])
+        return get_well_errors(conflicting_samples)
+
+
+def get_well_errors(sample_indices: list[int]) -> list[OccupiedWellError]:
+    return [OccupiedWellError(sample_index=sample_index) for sample_index in sample_indices]
+
+
+def is_sample_on_plate(sample: MicrosaltSample) -> bool:
+    return sample.container == ContainerEnum.plate
```
islean, line 35: Why are we not reusing the functionality? This could be a method operating on the parent `Sample` class since the container is already set there.

cg/services/order_validation_service/workflows/microsalt/validation/inter_field/utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def _initialize_map(self, order: MicrosaltOrder):
+        for sample_index, sample in order.enumerated_new_samples:
+            if is_sample_on_plate(sample):
+                key: tuple[str, str] = (sample.container_name, sample.well_position)
+                if not self.wells.get(key):
+                    self.wells[key] = []
+                self.wells[key].append(sample_index)
+
+    def get_occupied_well_errors(self) -> list[OccupiedWellError]:
```
islean, line 21: For me it is not super clear that this method should be in this class, at least not from the naming. But feel free to keep it. Just thought I'd mention it.

islean: The class name sounds like a data class to me

cg/services/order_validation_service/workflows/microsalt/validation/inter_field/utils.py
```diff
@@ -0,0 +1,35 @@
+from cg.models.orders.sample_base import ContainerEnum
+from cg.services.order_validation_service.errors.sample_errors import OccupiedWellError
+from cg.services.order_validation_service.workflows.microsalt.models.order import MicrosaltOrder
+from cg.services.order_validation_service.workflows.microsalt.models.sample import MicrosaltSample
+
+
+class PlateSamples:
```
islean, line 7: Nice! We might want to think about refactoring to something similar for the case_sample flow.

cg/services/order_validation_service/workflows/microsalt/validation/inter_field/utils.py
```diff
@@ -0,0 +1,35 @@
...
+
+    def __init__(self, order: MicrosaltOrder):
+        self.wells: dict = {}
+        self._initialize_map(order)
+
+    def _initialize_map(self, order: MicrosaltOrder):
```
islean, line 13: I think a docstring is useful for this map, since the structure is a bit convoluted.

islean (approved):
Nice!

cg/services/order_validation_service/workflows/microsalt/validation/inter_field/utils.py
seallard: I renamed it to PlateSamplesValidator. I think it is a bit better.

PR #3619, opened 2024-08-21
Add dev endpoints

seallard:


islean (approved):
🥳 We just need to remember to remove them before the master merge

PR #3621, opened 2024-08-21
Remove tomte from general rules

seallard:


cg/services/order_validation_service/workflows/order_validation_rules.py
```diff
@@ -0,0 +1,16 @@
...
+    validate_customer_can_skip_reception_control,
+    validate_customer_exists,
+    validate_user_belongs_to_customer,
+)
+from cg.services.order_validation_service.validators.inter_field.rules import (
+    validate_ticket_number_required_if_connected,
+)
+
+
+ORDER_RULES: list[callable] = [
```
seallard, line 11: These rules will be the same no matter which workflow is used.

Vince-janv (approved):
💯

islean (approved):
Looks great!

PR #3630, opened 2024-08-23
(Improve order flow) Patch min string length

islean:
## Description

The FE sends both null and empty strings for fields which are not set. Hence we can not set a global min length on strings.

### Added

-

### Changed

-

### Fixed

- Min string length is set on field level.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/models/sample.py
+++ b/cg/services/order_validation_service/models/sample.py
@@ -1,10 +1,10 @@
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, Field
 
 from cg.models.orders.sample_base import NAME_PATTERN, ContainerEnum
 
 
 class Sample(BaseModel):
-    application: str
+    application: str = Field(min_length=1)
     comment: str | None = None
     container: ContainerEnum
     container_name: str | None = None
@@ -13,8 +13,6 @@ class Sample(BaseModel):
     require_qc_ok: bool
     volume: int | None = None
     well_position: str | None = None
-
-    model_config = ConfigDict(str_min_length=1)
 
     @property
     def is_new(self) -> bool:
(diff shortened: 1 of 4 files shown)


cg/services/order_validation_service/models/sample.py
```diff
@@ -1,10 +1,10 @@
-from pydantic import BaseModel, ConfigDict, Field
+from pydantic import BaseModel, Field
 
 from cg.models.orders.sample_base import NAME_PATTERN, ContainerEnum
 
 
 class Sample(BaseModel):
-    application: str
+    application: str = Field(min_length=1)
```
seallard, line 7: Is this correct if an existing sample is used?

cg/services/order_validation_service/workflows/tomte/models/sample.py
```diff
@@ -20,6 +20,6 @@ class TomteSample(Sample):
     sex: SexEnum | None = None
     source: str | None = None
     status: StatusEnum | None = None
-    subject_id: str = Field(pattern=NAME_PATTERN, max_length=128)
+    subject_id: str = Field(pattern=NAME_PATTERN, min_length=1, max_length=128)
```
seallard, line 23: Is this correct if an existing sample is used?

seallard (approved):
👍

cg/services/order_validation_service/models/sample.py
islean: Pretty certain but remains to be tested in a separate PR. The behaviour does not change for the field for now though

cg/services/order_validation_service/workflows/tomte/models/sample.py
islean: Remains to be seen! Might change in a future PR

PR #3641, opened 2024-08-26
Feat(rsync) Remove -t flag for sars-cov

Vince-janv:
### Description
For some reason the rsync command used for covid samples contains the `-t` flag which preserves the timestamp on the files. This however seems to have the side effect of setting the the group owning the file to users.

There is no documentation in the PR or anywhere else on why the -t flag was added, so I'm removing it so that our rsyncs behave the same way.

### Fixes
- Files delivered now correctly belongs to the customers group


islean (approved):
Looks good to me!

Vince-janv:
Deployed to production

PR #3644, opened 2024-08-27
Perform validation before ticket creation

### LINKED ISSUE #3642: Bug: Tickets are submitted despite incorrect case names
**Describe the bug**

When a customer orders a case that already exists, but with a different sample, the validation returns a "case already exists". This is correct, as we do not allow customers to change the constellation of a family. The case is not created in statusdb, there is no order created. However, there is a ticket generated from this:

image

The notification from the email above contains information about the order that was placed. This causes a lot of confusion for us in production, making it difficult to see which one we should deliver.

First of all, the customers are unaware that they are spamming us with tickets, because the validation occurs later. I think that part could be fixed when we have a better validation in place, but currently, this is confusing for the customer, and production users.

**To Reproduce**
1. Place an order for an existing case name

**Expected behavior**

No ticket should be generated.

islean:
## Description

Closes #3642. The bug appears often enough that a temporary fix (which is to be removed in the improve-order-flow project) is in order.

### Added

-

### Changed

-

### Fixed

- order validation is performed before submission.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
Tested in stage with no ticket created:

image

karlnyr (approved):
Jank, but works well. Could we somehow make sure that this logic is removed in the future? Assuming that this is not the vision you are heading for with the order logic

islean:
Deployed to `cg-services-prod`

PR #3648, opened 2024-08-27
Add models for existing samples and cases

islean:
## Description

This PR separates the new and existing cases and samples into separate Pydantic models.

### Added

- ExistingSample
- ExistingCase

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/models/discriminators.py
+++ b/cg/services/order_validation_service/models/discriminators.py
@@ -1,0 +1,7 @@
+from typing import Any
+
+
+def has_internal_id(v: Any) -> str:
+    if isinstance(v, dict):
+        return "existing" if "internal_id" in v.keys() else "new"
+    return "existing" if getattr(v, "internal_id", None) else "new"
--- a/cg/services/order_validation_service/models/existing_case.py
+++ b/cg/services/order_validation_service/models/existing_case.py
@@ -1,0 +1,10 @@
+from pydantic import BaseModel
+
+
+class ExistingCase(BaseModel):
+    internal_id: str
+    panels: list[str]
+
+    @property
+    def is_new(self) -> bool:
+        return False
--- a/cg/services/order_validation_service/models/existing_sample.py
+++ b/cg/services/order_validation_service/models/existing_sample.py
@@ -1,0 +1,14 @@
+from pydantic import BaseModel, Field
+
+from cg.models.orders.sample_base import NAME_PATTERN, StatusEnum
+
+
+class ExistingSample(BaseModel):
+    father: str | None = Field(None, pattern=NAME_PATTERN)
+    internal_id: str
+    mother: str | None = Field(None, pattern=NAME_PATTERN)
+    status: StatusEnum | None = None
+
+    @property
+    def is_new(self) -> bool:
+        return False
--- a/cg/services/order_validation_service/models/existing_sample.py
+++ b/cg/services/order_validation_service/models/existing_sample.py
@@ -7,7 +7,7 @@ class ExistingSample(BaseModel):
     father: str | None = Field(None, pattern=NAME_PATTERN)
     internal_id: str
     mother: str | None = Field(None, pattern=NAME_PATTERN)
-    status: StatusEnum | None = None
+    status: StatusEnum
 
     @property
     def is_new(self) -> bool:
--- a/cg/services/order_validation_service/models/order_with_cases.py
+++ b/cg/services/order_validation_service/models/order_with_cases.py
@@ -1,9 +1,19 @@
+from pydantic import Discriminator, Tag
+from typing_extensions import Annotated
+
 from cg.services.order_validation_service.models.case import Case
+from cg.services.order_validation_service.models.discriminators import has_internal_id
+from cg.services.order_validation_service.models.existing_case import ExistingCase
 from cg.services.order_validation_service.models.order import Order
 
 
 class OrderWithCases(Order):
-    cases: list[Case]
+    cases: list[
+        Annotated[
+            Annotated[Case, Tag("new")] | Annotated[ExistingCase, Tag("existing")],
+            Discriminator(has_internal_id),
+        ]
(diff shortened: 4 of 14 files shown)


cg/services/order_validation_service/models/order_with_cases.py
```diff
@@ -1,9 +1,19 @@
...
 
 
 class OrderWithCases(Order):
-    cases: list[Case]
+    cases: list[
+        Annotated[
+            Annotated[Case, Tag("new")] | Annotated[ExistingCase, Tag("existing")],
+            Discriminator(has_internal_id),
+        ]
+    ]
```
islean, line 16: Ugly typing - should I break it up?

cg/services/order_validation_service/models/existing_sample.py
```diff
@@ -0,0 +1,14 @@
+from pydantic import BaseModel, Field
+
+from cg.models.orders.sample_base import NAME_PATTERN, StatusEnum
+
+
+class ExistingSample(BaseModel):
+    father: str | None = Field(None, pattern=NAME_PATTERN)
+    internal_id: str
+    mother: str | None = Field(None, pattern=NAME_PATTERN)
+    status: StatusEnum | None = None
```
islean, line 10: Had to make this optional in the end since we are using this for existing microsalt samples as well.

Vince-janv: I don't think existing microsalt samples exist...

islean: Removed!

cg/services/order_validation_service/models/discriminators.py
```diff
@@ -0,0 +1,7 @@
+from typing import Any
+
+
+def has_internal_id(v: Any) -> str:
```
seallard, line 4: Should this not return a bool?

cg/services/order_validation_service/models/existing_case.py
```diff
@@ -0,0 +1,10 @@
+from pydantic import BaseModel
+
+
+class ExistingCase(BaseModel):
+    internal_id: str
+    panels: list[str]
```
seallard, line 6: To me it is really strange that we require the client to provide a list of panels when they use an existing case. Sure, they are allowed to provide a new set of panels, but it should not be required.

We decided to postpone that discussion though, so disregard this.

cg/services/order_validation_service/models/existing_sample.py
```diff
@@ -0,0 +1,14 @@
+from pydantic import BaseModel, Field
+
+from cg.models.orders.sample_base import NAME_PATTERN, StatusEnum
+
+
+class ExistingSample(BaseModel):
+    father: str | None = Field(None, pattern=NAME_PATTERN)
+    internal_id: str
+    mother: str | None = Field(None, pattern=NAME_PATTERN)
+    status: StatusEnum
```
seallard, line 10: Same here, I don't think it makes sense to have the status as a required field when providing an existing sample.
But disregard for now unless you agree.

cg/services/order_validation_service/models/discriminators.py
islean: It needs to return a matching "Tag". But maybe the naming should reflect that? `get_tag` or something?

cg/services/order_validation_service/models/existing_case.py
islean: For me, given that the frontend sends the old panels as default, I think sending `None` indicates an active effort by the customer to remove the panels. But let's postpone the discussion

cg/services/order_validation_service/models/order_with_cases.py
seallard: I think it is fine, if you prefer something like this go for it
```python
NewCaseType: TypeAlias = Annotated[Case, Tag("new")]
ExistingCaseType: TypeAlias = Annotated[ExistingCase, Tag("existing")]
DiscriminatedCaseType: TypeAlias = Annotated[Union[NewCaseType, ExistingCaseType], Discriminator(has_internal_id)]

cases: List[DiscriminatedCaseType]
```

cg/services/order_validation_service/models/existing_sample.py
islean: The status is set on the `case_sample` entries which are new even for existing cases (provided they are within a new case). So I don't think it would work to set it as optional given that we don't know where to fetch it from.

seallard (approved):
I like that we parse the data into the correct models 🚀 

That way, we can actually use the basic pydantic validation to a greater extent in the validation logic which I think removes some bloat.

cg/services/order_validation_service/models/discriminators.py
seallard: Sounds good!

cg/services/order_validation_service/models/existing_sample.py
seallard: Sounds like the status does not belong on the case sample model but rather on the sample model then. Definitely out of scope here, but that is some weird data modelling

cg/services/order_validation_service/models/order_with_cases.py
islean: I'll go for a hybrid! I like `NewCaseType` and `ExistingCaseType` but it might make sense to see that there is a union of the two classes still

cg/services/order_validation_service/models/discriminators.py
islean: Resolved in Slack - keeping naming for now since it gives more info on what decides what model to parse

PR #3664, opened 2024-08-29
Remove unnecessary rules

islean:
## Description

Since we introduced models for existing cases and samples, we do not need to validate that certain fields are filled in via rules. We can mark them as required in the Pydantic models instead.


seallard (approved):
🚀

PR #3665, opened 2024-08-29
Add balsamic sample sex validation

seallard:


cg/services/order_validation_service/workflows/balsamic/rules/case_sample/utils.py
```diff
@@ -0,0 +1,5 @@
+from cg.services.order_validation_service.workflows.balsamic.models.sample import BalsamicSample
+
+
+def has_sex_and_subject(sample: BalsamicSample) -> bool:
+    return sample.subject_id and sample.sex
```
islean, line 5: I don't think this check works. The Atlas page states that if the sample has a subject_id and a _known_ gender - but sample.sex is an Enum which has a value "unknown". With this check, samples with unknown sex would still be picked up and validated.

cg/store/crud/read.py
```diff
@@ -1562,3 +1562,16 @@ def get_case_ids_for_samples(self, sample_ids: list[int]) -> list[str]:
...
+        self,
+        customer_internal_id: str,
+        subject_id: str,
+        sex: str,
+    ) -> bool:
+        samples: list[Sample] = self.get_samples_by_customer_and_subject_id(
+            customer_internal_id=customer_internal_id,
+            subject_id=subject_id,
+        )
+        existing_sex: set[str] = {sample.sex for sample in samples}
```
islean, line 1576: In essence, a subject_id corresponds to one individual. Hence, one could argue that if this set contains anything other than one element, the data is off. A bit uncertain if it can be a mix of a known sex and unknown though.

islean (approved):
A bit uncertain on how to handle unknown sexes. But apart from that, looks good!

cg/store/crud/read.py
islean: So this check is not perfect. If the sex in the db is "unkown" and they provide a "female", then the check should not fail.

cg/services/order_validation_service/workflows/balsamic/rules/case_sample/rules.py
```diff
@@ -0,0 +1,30 @@
+from cg.services.order_validation_service.errors.case_sample_errors import SubjectIdGenderError
+from cg.services.order_validation_service.workflows.balsamic.models.order import BalsamicOrder
+from cg.services.order_validation_service.workflows.balsamic.rules.case_sample.utils import (
+    has_sex_and_subject,
+)
+from cg.store.store import Store
+
+
+def validate_subject_sex_consistency(
```
islean, line 9: Should we add a test?

cg/services/order_validation_service/errors/case_sample_errors.py
```diff
@@ -143,3 +143,8 @@ class InvalidVolumeError(CaseSampleError):
 class InvalidBufferError(CaseSampleError):
     field: str = "elution_buffer"
     message: str = "The chosen buffer is not allowed when skipping reception control"
+
+
+class SubjectIdGenderError(CaseSampleError):
+    field: str = "subject_id"
```
islean, line 149: Probably should be the sex field which should have an error

PR #3668, opened 2024-08-29
Dev freshdesk

ahdamin:
### Changed

- replace `osticket` with `freshdesk`


cg/clients/freshdesk/models.py
```diff
@@ -1,36 +1,60 @@
...
 
 from cg.clients.freshdesk.constants import Priority, Source, Status
 
 
 class TicketCreate(BaseModel):
     """Freshdesk ticket."""
 
-    attachments: list[dict[str, str]] = []
-    email: str
+    attachments: list[Union[str, bytes]] = Field(default_factory=list)
```
islean, line 10:
```suggestion
    attachments: list[str | bytes] = Field(default_factory=list)
```

cg/clients/freshdesk/freshdesk_client.py
```diff
@@ -1,56 +1,40 @@
...
 
 class FreshdeskClient:
     """Client for communicating with the freshdesk REST API."""
 
-    def __init__(self, base_url: str, api_key: str):
-        self.base_url = base_url
-        self.api_key = api_key
+    def __init__(self, config: dict):
+        self.base_url = config["freshdesk_url"]
+        self.api_key = config["freshdesk_api_key"]
```
Vince-janv, line 18: Passing around dicts on constructors hides what parameters are needed for the class to be instantiated and creates "magic strings" for fetching the correct keys. I think we should explicitly pass the url and the key

cg/clients/freshdesk/models.py
```diff
@@ -1,36 +1,60 @@
...
     email_config_id: int | None = None
+    description: str
     name: str
     priority: int = Priority.LOW
     source: int = Source.EMAIL
     status: int = Status.OPEN
     subject: str
     tags: list[str] = []
     type: str | None = None
+    custom_fields: dict[str, Union[str, int, float, None]] = Field(default_factory=dict)
```
islean, line 21:
```suggestion
    custom_fields: dict[str, str | int | float | None] = Field(default_factory=dict)
```

cg/meta/orders/ticket_handler.py
```diff
@@ -20,15 +19,17 @@ class TicketHandler:
 
     NEW_LINE = "<br />"
 
-    def __init__(self, osticket_api: OsTicket, status_db: Store):
-        self.osticket: OsTicket = osticket_api
-        self.status_db: Store = status_db
+    def __init__(self, config: dict, db: Store, client: FreshdeskClient):
```
Vince-janv, line 22: Same here regarding passing dicts around

cg/clients/freshdesk/models.py
```diff
@@ -1,36 +1,60 @@
...
+
+    def to_multipart_data(self) -> list[Tuple[str, Union[str, int, bytes]]]:
```
islean, line 23:
```suggestion
    def to_multipart_data(self) -> list[Tuple[str, str | int | bytes]]:
```

cg/server/app_config.py
```diff
@@ -4,23 +4,25 @@
 class AppConfig(BaseSettings):
```
Vince-janv, line 4: Can we add a doc string to not confuse people with the dummy values?
```suggestion
class AppConfig(BaseSettings):
"""Default values are overridden by environment variable at run-time."""
```

Vince-janv (approved):
Nice work! Some comments about class instantiation

cg/clients/freshdesk/freshdesk_client.py
ahdamin: Thanks for the clarification! ✅

cg/meta/orders/ticket_handler.py
ahdamin: ✅

ahdamin:
@Vince-janv please review and test

PR #3673, opened 2024-09-02
fix(new delivery service calls in CLI)

ChrOertlin:
## Description

Implements the new delivery service in the CLI calls
1. 'cg deliver analysis' changed to 'cg deliver case' -> now only delivers files on a case basis to the customer inbox
2. 'cg deliver ticket' changed behaviour, does no longer RSYNC files automatically.
3. Removed 'cg deliver concatenate' as concatenation is done automatically now where it is required.

Motivation: 
Previously cg deliver analysis had functionality for both ticket and case delivery in one even though a cg deliver ticket cli command exists.
Streamlined the functionaliy of the cg deliver case/ticket, now both function just deliver files to the customer inbox on the HPC. A separate Rsync call is now required.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


karlnyr:
In production we need to have one command which does the full delivery, meaning both to the customer inbox on hasta and then from hasta to Caesar. Having this command would reduce the possibility that on would forget to run the rsync command.
Otherwise, I think that this logic looks a lot cleaner and easier to understand. However, I was wondering about the removal of the `--ignore-missing-bundles` and `--force-all` flag. Will we now always ignore missing bundles? And will we always transfer all data available?

The merge of this PR would require that you change following document to be up to date: https://atlas.scilifelab.se/production/data_analysis/workflows/raw_data_delivery/

ChrOertlin:
> In production we need to have one command which does the full delivery, meaning both to the customer inbox on hasta and then from hasta to Caesar. Having this command would reduce the possibility that one would forget to run the rsync command.

Alright, then I will add the rsync to both.

> Otherwise, I think that this logic looks a lot cleaner and easier to understand. However, I was wondering about the removal of the `--ignore-missing-bundles` and `--force-all` flag. Will we now always ignore missing bundles? And will we always transfer all data available?

It fetches what is available and delivers it. If a bundle is missing it will be skipped by default. 
The command will overwrite existing data in the HPC on hasta and with the rsync change will start the sync

> The merge of this PR would require that you change the following document to be up to date: https://atlas.scilifelab.se/production/data_analysis/workflows/raw_data_delivery/

cg/cli/deliver/base.py
```diff
@@ -133,60 +51,56 @@ def rsync(context: CGConfig, ticket: str, dry_run: bool):
...
+    "-c",
+    "--case-id",
+    required=True,
+    help="Deliver the files for a specific case",
+)
+@click.pass_obj
+def deliver_analysis(
+    context: CGConfig,
+    case_id: str,
+    delivery_type: list[str],
```
Vince-janv, line 66: delivery_type given here is "mid-dna", "balsamic" etc. It doesn't make much sense but it's what prod uses, we'll have to ask them if they want to keep it

cg/cli/deliver/base.py
```diff
@@ -133,60 +51,56 @@ def rsync(context: CGConfig, ticket: str, dry_run: bool):
...
-    deliver_ticket_api = DeliverTicketAPI(config=cg_context)
-    deliver_ticket_api.concatenate_fastq_files(ticket=ticket, dry_run=dry_run)
+    Deliver all case files based on delivery type to the customer inbox on the HPC
+    """
+    inbox: str = context.delivery_path
+    service_builder = DeliveryServiceFactory(
+        store=context.status_db, hk_api=context.housekeeper_api
+    )
+    case: Case = context.status_db.get_case_by_internal_id(internal_id=case_id)
+    for delivery in delivery_type:
```
Vince-janv, line 76: If they are ok with not giving workflows as delivery types we can skip this loop!

cg/services/file_delivery/deliver_files_service/deliver_files_service.py
```diff
@@ -24,15 +27,27 @@ def __init__(
...
         delivery_files: DeliveryFiles = self.file_manager.get_files_to_deliver(case_id)
         moved_files: DeliveryFiles = self.file_mover.move_files(
             delivery_files=delivery_files, delivery_base_path=delivery_base_path
         )
         self.file_formatter.format_files(moved_files)
+
+    def deliver_files_for_ticket(self, ticket_id: str, delivery_base_path: Path) -> None:
+        """Deliver the files for all cases in a ticket to the customer folder."""
+        cases: list[Case] = self.status_db.get_cases_by_ticket_id(ticket_id)
+        if not cases:
```
Vince-janv, line 48: Clean!

Vince-janv (changes requested):
Looks really nice! Let's make sure all of the weird logic is covered though. I'll post a little list

Vince-janv (commented):
Did not mean to request changes 😅

Vince-janv:
> It fetches what is available and delivers it. If a bundle is missing it will be skipped by default. The command will overwrite existing data in the HPC on hasta and with the rsync change will start the sync

I really like the logic described above. I think simplicity is something we should really strive for here. @karlnyr Do you see any potential problems with it?

cg/cli/deliver/base.py
diitaz93: I would say change it to workflow

Vince-janv:
## Here's a list of strange things we need to include from the top of my head.
## Concatenation of microSALT cases
- [ ] Cases with data-analysis microSALT should have their FASTQ file concatenated when delivered with the delivery type FASTQ (and FASTQ_QC, FASTQ_QC+ANALYSIS)
## "Missing bundles"
- [ ] Cases of type FASTQ, microSALT, and Mutant should not raise an exception if a one of its samples is missing a bundle
## CLI commands
- [ ] A CLI command exists to deliver (just link to cust inbox on hasta) a whole ticket just a case.
    - It should take a delivery type as input as well (ie one should be able to specify delivery for a case that has data_delivery=scout)
- [ ] A CLI command exists that which takes a ticket and, then links the correct files and starts an rsync job. If this ticket is a covid ticket extra logic needs to be applied in the rsync part (should be able to just swap out the first part of this command)

cg/cli/deliver/base.py
ChrOertlin: Removed the option. Delivery type and workflow are set on case. I will fetch it from there.

cg/cli/deliver/base.py
ChrOertlin: Yes, I removed it and also the option to pass delivery type. As it is set on the case

ChrOertlin:
Merging into dev branch - copied the required tests into the dev brancnh as well.

PR #3683, opened 2024-09-04
(Improve order flow) Add Balsamic validation rules

islean:
## Description

This PR adds validation rules for Balsamic and also hooks some rules up which were written but never used in the workflows.

### Added

- Validation of capture kits being required for TGS samples.

### Fixed

- Well position requirement validated for Case orders
- Container name requirement validated for case orders 

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


diitaz93 (approved):
LGTM 👍 I'm wondering if you could use the same validation rules for different pipelines and define them in a single place so that Sonarcloud does not complain, but I lack the domain knowledge for that, maybe is better to have them separate. Also, would be good that all utils functions have unit tests

diitaz93:
I think is ok to ignore the sonar cloud warning for now

islean:
> LGTM 👍 I'm wondering if you could use the same validation rules for different pipelines and define them in a single place so that Sonarcloud does not complain, but I lack the domain knowledge for that, maybe is better to have them separate. Also, would be good that all utils functions have unit tests

We have had discussions in the project and decided that having an overview of "these are all the rules performed for this workflow" was more important than the code duplication. So in this case I would not like to refactor regardless of Sonarcloud. Regarding tests, the convention we have settled for thus far is to have tests for all the rules but not necessarily for the all the utils. For me it would not add too much value but if it there is a consensus that it needs to be done, it can be added. Would take quite some time though. 

I will definitely add a test for the rule introduced in this PR though.

PR #3687, opened 2024-09-04
add raredisease pedigree check

peterpru:
## Description
The RareDisease requires an additional metric from the multiQC data file with respect to the other Nextflow pipelines, which is the pedigree check. This is a pair-wise sample  metric, meaning there is a value for every pair of samples combination in the case. To parse them, the method `get_multiqc_json_metrics` from `NfAnalysisAPI` is overridden in `RarediseaseAnalysisAPI`.

### Added

- Pedigree check in raredisease QC: parent_error_ped_check

### Changed

- MEDIAN_TARGET_COVERAGE threshold to >25

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b add-raredisease-pedigree-check
    ```

### How to test
- [x] Do nano /home/proj/stage/analysis/cases/funkitten/multiqc/multiqc_data/multiqc_data.json and MEDIAN_TARGET_COVERAGE set to 26 for all three samples
- [x] Do `cg workflow raredisease metrics-deliver funkitten` with MEDIAN_TARGET_COVERAGE set to 26.

- [x] Do  `sed -i 's/"parent_error_ped_check": "False",/"parent_error_ped_check": "True",/g' /home/proj/stage/analysis/cases/funkitten/multiqc/multiqc_data/multiqc_data.json`
- [x] Do `cg workflow raredisease metrics-deliver funkitten`

### Expected test outcome

- [x] Check that there are no errors, and command completes as expected.
image

- [x] When setting parent error to True, the command should fail. Error needs to be propagated to trailblazer
image

- [x] See that the check is visible in the deliverables.yaml file:
image

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/meta/workflow/nf_analysis.py
```diff
@@ -631,11 +631,10 @@ def get_workflow_metrics(self, metric_id: str) -> dict:
...
         """Return search patterns for MultiQC. Each key is a search pattern and each value
         corresponds to the metric ID to set in the metrics deliverables file.
-        Multiple search patterns can be added. Ideally patterns used should be sample ids, e.g.
+        Multiple search patterns can be added. Ideally, used patterns should be sample ids, e.g.
         {sample_id_1: sample_id_1, sample_id_2: sample_id_2}."""
-        sample_ids: Iterator[str] = self.status_db.get_sample_ids_by_case_id(case_id=case_id)
-        search_patterns: dict[str, str] = {sample_id: sample_id for sample_id in sample_ids}
-        return search_patterns
+        sample_ids: Iterator[str] = self.status_db.get_sample_ids_by_case_id(case_id)
+        return {sample_id: sample_id for sample_id in sample_ids}
```
islean, line 637: I get confused by this. Why do we have a long doc string describing something much more complex than taking all the sample ids from a case and creating a dictionary with the same keys and values from them? Also, why do we need a dictionary here if it simply maps a sample id to a sample id?

cg/meta/workflow/nf_analysis.py
```diff
@@ -649,11 +648,13 @@ def get_deduplicated_metrics(metrics: list[MetricsBase]) -> list[MetricsBase]:
                 deduplicated_metrics.append(metric)
         return deduplicated_metrics
 
+    def get_multiqc_data_json(self, case_id: str) -> MultiqcDataJson:
+        """Return a MultiqcDataJson object."""
```
islean, line 652: I think this docstring just duplicates the return type written a line above.

cg/meta/workflow/raredisease.py
```diff
@@ -149,11 +152,58 @@ def get_managed_variants(self) -> list[str]:
         return self._get_managed_variants(genome_build=GenePanelGenomeBuild.hg19)
 
     def get_workflow_metrics(self, sample_id: str) -> dict:
+        """Return Raredisease workflow metric conditions for a sample."""
         sample: Sample = self.status_db.get_sample_by_internal_id(internal_id=sample_id)
-        metric_conditions: dict[str, dict[str, Any]] = dict(RAREDISEASE_METRIC_CONDITIONS)
-        self.set_order_sex_for_sample(sample, metric_conditions)
+        if "-" not in sample_id:
```
islean, line 157: Is this check needed? In all of statusdb there are only three samples, all from 2017, whose internal ids contain a dash. Or is the sample id the name?

cg/meta/workflow/raredisease.py
```diff
@@ -149,11 +152,58 @@ def get_managed_variants(self) -> list[str]:
...
+                metric_name=metric_name,
+                metric_value=sample_pair_metrics[metric_name],
+                metric_id=pair_sample_ids,
+            )
+
+    def get_multiqc_json_metrics(self, case_id: str) -> list[MetricsBase]:
+        """Return a list of the metrics specified in a MultiQC json file."""
+        multiqc_json: MultiqcDataJson = self.get_multiqc_data_json(case_id=case_id)
+        metrics = []
+        for search_pattern, metric_id in self.get_multiqc_search_patterns(case_id).items():
```
islean, line 189: So was this loop already done in the method we are now overriding? Or was this check also added now?

cg/meta/workflow/raredisease.py
```diff
@@ -149,11 +152,58 @@ def get_managed_variants(self) -> list[str]:
...
+                    search_pattern=search_pattern,
+                    multiqc_json=multiqc_json,
+                    metric_id=metric_id,
+                    exact_match=self.is_multiqc_pattern_search_exact,
+                )
+            )
+            metrics.extend(metrics_for_pattern)
+        for sample_pair in self._get_sample_pair_patterns(case_id):
+            if parent_error_metric := self.get_parent_error_ped_check_metric(
+                pair_sample_ids=sample_pair, multiqc_raw_data=multiqc_json.report_saved_raw_data
```
islean, line 201: Is there a way of baking this into the `self.get_metrics_from_multiqc_json_with_pattern`? Seems to me as if we are getting metrics from the multiqc json using a pattern.

islean (approved):
Looks good! A bit confused over certain parts of the code.

cg/meta/workflow/nf_analysis.py
diitaz93: I had the same doubts, but it is likely that this will be overrriden in some nf analysis apis to have different key, value pairs

diitaz93: The docstring could be simplified though, but this is the analysis apis, it is obscure territory

cg/meta/workflow/nf_analysis.py
diitaz93: lets remove the docstring @peterpru

cg/meta/workflow/raredisease.py
diitaz93: This is not always a sample id. We added this check because we use this function with str like `sample1-sample2`. This is a function overridden from the NfAnalysisAPI so renaming the parameter will be a bit invasive

cg/meta/workflow/raredisease.py
diitaz93: yes and this is bothering me a bit. However, if we use `super().get_multiqc_json_metrics()` we are parsing the multiqc_data.json twice. This file is huge, and I wanted to avoid that

islean: Oh, that is no issue for me, I was just trying to understand what was what.

cg/meta/workflow/raredisease.py
islean: Then I do not agree that we want to use this method? Or why not have two separate methods given that the behaviour differs depending on the input?

diitaz93: This function will still be called from a parent class using the pattern as "sample id". I think the best I can think of is doing like:
```python
def get_workflow_metrics(self, sample_id: str) -> dict:
    return self._some_other_function(combined_sample_ids=sample_id)
```
It is worth mentioning that the whole QC parsing is planned to be taken out of the Analysis API in the future

peterpru:
Thank you all for your comments @ChrOertlin, @islean, after multiple discussions with @diitaz93 and @jemten we have decided to keep the code as is, and add test functions to make sure it works as intended. I have marked most conversations as resolved. We can refactor the nextflow QC code in the future, or turn it into a microservice where we have it outside of cg.

diitaz93 (approved):
Looks good 🚀

peterpru:
Successfully deployed in stage:
image

Deployed in production:
image

Issue #3020, opened 2024-03-05
Add validation of well positions


islean:
## Description

When customers submit order forms containing faulty well_positions, the error is caught by Lims. But that happens after a ticket is opened in SuSy, meaning that the customer can go on thinking that all went well. We will not have any records in Lims or status db though.

## Suggested solution

Implement validation for this upon order submission (and/or when parsing the order form)

## This can be closed when

No such orders result in tickets being created.

Karl-Svard:
### Allowed values in lims:

#### Tube container
Only `1:1`. One sample per container.

#### 96 well plate container
`A:1` to `H:12`

Vince-janv:
According to @islean the "A:1 to H:12" validation is missing from the `improve-order-flow` branch

diitaz93:
## Technical refinement
- Implement in a PR to be merged into the dev branch of the implement order flow project
- There are two rules to be implemented:
  - Container names for tubes should be unique among samples
  - Well-plate positions should be formatted as A-H:1-12

eliottBo:
- [x] Add well position validation for sample
- [ ] Add well position validation for case_sample
- [ ] Add tube name validation for sample
- [ ] Add tube name validation for case_sample

diitaz93:
Finished

PR #3716, opened 2024-09-12
Add validation tube names case

eliottBo:
## Description
This PR addresses the second part of https://github.com/Clinical-Genomics/cg/issues/3020. It is part of improve-order-flow-main.
### Added

- Add validation for tube names for cases orders.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


eliottBo:
Note: two samples in different cases of the same order can have the same tube name. Lmk if the behaviour should be that tube names across all cases in an order must be unique.

cg/services/order_validation_service/errors/case_sample_errors.py
```diff
@@ -143,3 +143,8 @@ class CaptureKitMissingError(CaseSampleError):
 class WellFormatError(CaseSampleError):
     field: str = "well_position"
     message: str = "Well position must follow the format A-H:1-12"
+
+
+class ContainerNameRepeatedError(CaseSampleError):
+    field: str = "container_name"
+    message: str = "Container name must be unique among samples"
```
islean, line 150: Thought that hit me now and which should be patched for the samples PR as well - we should have the message highlighting that tube names needs to be unique. Container names are allowed to be repeated, especially for well plate samples.

cg/services/order_validation_service/rules/case_sample/utils.py
```diff
@@ -294,3 +296,17 @@ def is_invalid_plate_well_format(sample: Sample) -> bool:
     if sample.is_on_plate:
         return not bool(re.match(correct_well_position_pattern, sample.well_position))
     return False
+
+
+def is_tube_container_name_redundant(sample: Sample, counter: Counter) -> bool:
```
islean, line 301: Is "redundant" the word we are looking for here? It seems to return true if the tube container name is reused within a case. Maybe "is_tube_name_reused" is more descriptive?

cg/services/order_validation_service/rules/case_sample/utils.py
```diff
@@ -294,3 +296,17 @@ def is_invalid_plate_well_format(sample: Sample) -> bool:
...
+    """Check if a tube container name is reused across samples."""
+    return sample.container == ContainerEnum.tube and counter.get(sample.container_name) > 1
+
+
+def get_repeated_tube_names(case: Case) -> list[int]:
```
islean, line 306: The naming here could be slightly better since it currently implies that a list of names would be returned, but in fact a list of indices are returned. Maybe `get_indices_for_repeated_tube_names`?

cg/services/order_validation_service/rules/case_sample/utils.py
```diff
@@ -294,3 +296,17 @@ def is_invalid_plate_well_format(sample: Sample) -> bool:
...
+    counter = Counter([sample[1].container_name for sample in case.enumerated_new_samples])
```
islean, line 307: This counter needs to take into account all samples within the order, not only the samples within a case.

cg/services/order_validation_service/rules/case_sample/utils.py
```diff
@@ -33,6 +33,8 @@
 from cg.store.models import Application
 from cg.store.store import Store
 
+import pdb
```
islean, line 36: What is this import for?

cg/services/order_validation_service/rules/case_sample/rules.py
```diff
@@ -339,3 +341,15 @@ def validate_well_position_format(order: OrderWithCases, **kwargs) -> list[WellF
...
                 errors.append(error)
     return errors
+
+
+def validate_tube_container_name_unique(
+    order: OrderWithCases, **kwargs
+) -> list[ContainerNameRepeatedError]:
+    errors: list[ContainerNameRepeatedError] = []
+    for case_index, case in order.enumerated_new_cases:
+        repeated_tube_name_indices: list[str] = get_repeated_tube_names(case=case)
```
islean, line 351: I think we need to check this for all samples within the order so I am unsure if we can pass the case here?

islean (approved):
Looks good but there are some quirks we need to remedy - in particular we need to make sure that all samples are considered in unison

cg/services/order_validation_service/rules/case_sample/utils.py
eliottBo: It was for testing purpose I forgot to remove it

islean (approved):
Mega

PR #3721, opened 2024-09-13
add(bam delivery support)

ChrOertlin:
## Description
closes: https://github.com/Clinical-Genomics/add-new-tech/issues/103
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/constants/delivery.py
```diff
@@ -78,10 +78,11 @@
 
 CLINICAL_DELIVERY_TAGS: list[set[str]] = [{HermesFileTag.CLINICAL_DELIVERY}]
 
-FASTQ_ANALYSIS_CASE_TAGS: list[set[str]] = []
+RAW_DATA_ANALYSIS_CASE_TAGS: list[set[str]] = []
 
-FASTQ_ANALYSIS_SAMPLE_TAGS: list[set[str]] = [
+RAW_DATA_ANALYSIS_SAMPLE_TAGS: list[set[str]] = [
     {"fastq"},
+    {"bam"},
```
ChrOertlin, line 85: hmm not sure if I like how this could end up ultimately... it would start pulling BAM files for fastq-analysis if these are there... 

Maybe we should re think what raw data really means.

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
```diff
@@ -112,6 +117,7 @@ def _validate_delivery_type(delivery_type: DataDelivery):
             DataDelivery.FASTQ,
             DataDelivery.ANALYSIS_FILES,
             DataDelivery.FASTQ_ANALYSIS,
+            DataDelivery.BAM,
         ]:
             return
         raise DeliveryTypeNotSupported(
```
diitaz93, line 123: update error message to include BAM

cg/services/deliver_files/delivery_file_tag_fetcher_service/bam_delivery_tags_fetcher.py
```diff
@@ -0,0 +1,27 @@
...
+
+class BamDeliveryTagsFetcher(FetchDeliveryFileTagsService):
+    """Class to fetch tags for bam files to deliver."""
+
+    @handle_tag_errors
+    def fetch_tags(self, workflow: Workflow) -> DeliveryFileTags:
+        """Fetch the tags for the bam files to deliver."""
+        return DeliveryFileTags(
+            case_tags=None,
+            sample_tags=[{"bam"}],
```
diitaz93, line 19: make "bam" a constant, maybe in `housekeeper_tags`

cg/services/deliver_files/delivery_file_tag_fetcher_service/bam_delivery_tags_fetcher.py
```diff
@@ -0,0 +1,27 @@
...
+        )
+
+    @staticmethod
+    def _validate_workflow(workflow: Workflow):
+        """Validate the workflow."""
+        if not workflow == Workflow.FASTQ:
```
diitaz93:
```suggestion
        if workflow != Workflow.FASTQ:
```

ChrOertlin, line 25:
```suggestion
        if workflow != Workflow.FASTQ:
```

cg/constants/delivery.py
```diff
@@ -78,9 +78,9 @@
 
 CLINICAL_DELIVERY_TAGS: list[set[str]] = [{HermesFileTag.CLINICAL_DELIVERY}]
 
-FASTQ_ANALYSIS_CASE_TAGS: list[set[str]] = []
+RAW_DATA_ANALYSIS_CASE_TAGS: list[set[str]] = []
 
-FASTQ_ANALYSIS_SAMPLE_TAGS: list[set[str]] = [
+RAW_DATA_ANALYSIS_SAMPLE_TAGS: list[set[str]] = [
```
diitaz93, line 83: This is a bit confusing as raw data includes both fastq and bam. Maybe we should treat raw data differently and not include that workflow in this dictionary, would that be feasible? I can't think of any other workaround right now but maybe it is worth discussing

diitaz93 (commented):
It looks good 👍 Although I am a bit hesitant about raw data tags having only fastq

cg/constants/delivery.py
ChrOertlin: Yes it is a design problem that fastq is a workflow and a delivery type I think. Not sure how to address it

cg/services/deliver_files/delivery_file_fetcher_service/raw_data_delivery_file_fetcher.py
```diff
@@ -19,7 +22,7 @@
 from housekeeper.store.models import File
 
 
-class FastqDeliveryFileFetcher(FetchDeliveryFilesService):
+class RawDataDeliveryFileFetcher(FetchDeliveryFilesService):
```
Vince-janv, line 25: There are a lot of instances of "fastq" in both the methods, variables and docstrings in this class. Pleas refactor them

cg/services/deliver_files/delivery_file_tag_fetcher_service/bam_delivery_tags_fetcher.py
```diff
@@ -0,0 +1,28 @@
...
+            case_tags=None,
+            sample_tags=[{AlignmentFileTag.BAM}],
+        )
+
+    @staticmethod
+    def _validate_workflow(workflow: Workflow):
+        """Validate the workflow."""
+        if workflow != Workflow.FASTQ:
+            raise ValueError(f"Workflow {workflow} is not supported for only BAM file delivery.")
+        return workflow
```
Vince-janv, line 28: I don't really see the need for this method. Also it seems to never be called

cg/constants/delivery.py
Vince-janv: One idea could be to remove Workflow.FASTQ from `SampleAndCaseDeliveryTagsFetcher`. Since we wan't to stop treating FASTQ/RAW-DATA as a Workflow like the others, I think it would simplify some logic

Vince-janv, line 83: But these tags will only be used for cases with data_delivery=fastq, right? Maybe we can leave them as is?

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
```diff
@@ -79,15 +82,17 @@ def _get_file_tag_fetcher(delivery_type: DataDelivery) -> FetchDeliveryFileTagsS
             DataDelivery.FASTQ: SampleAndCaseDeliveryTagsFetcher,
             DataDelivery.ANALYSIS_FILES: SampleAndCaseDeliveryTagsFetcher,
             DataDelivery.FASTQ_ANALYSIS: SampleAndCaseDeliveryTagsFetcher,
+            DataDelivery.BAM: BamDeliveryTagsFetcher,
```
Vince-janv, line 85: Adding a whole new delivery flow by just adding a few lines in this file is so nice!

Vince-janv (approved):
Nice to see the fruits of the refactoring. SD brings up a good point about what to do to do with Workflow.FASTQ. Out of scope for this PR but what do you think is the cleanest solution?

cg/services/deliver_files/delivery_file_tag_fetcher_service/bam_delivery_tags_fetcher.py
ChrOertlin: added the call now. In cas someone tries to deliver bam files for something that is not RawData this should raise an error since that workflow might not contain any or doesn't have it on the sample level. So this is a user-error check

cg/constants/delivery.py
ChrOertlin: Right now this will be in line with the other workflows. i..e fastq will be raw data.

cg/services/deliver_files/delivery_file_fetcher_service/raw_data_delivery_file_fetcher.py
ChrOertlin: think I got them all

ChrOertlin:
> Nice to see the fruits of the refactoring. SD brings up a good point about what to do to do with Workflow.FASTQ. Out of scope for this PR but what do you think is the cleanest solution?

I think the cleanest solution is to step away from the workflow first delivery tag hierarchy.


Now it is : Workflow{case,sample}
I think having: DataDeliveryType{workflow{case, sample}} might solve this problem. Then we can tailor the tags for each data delivery type for each workflow separately. Might make things easier as well. Would need to play a bit with this idea though. If you want me to spend time on it.

cg/constants/delivery.py
ChrOertlin: I explained an option below. Not sure why we want to treat WorkFlow.RawData differently, perhaps I missed this. Resolving this for now.

ChrOertlin:
Added an issue to track the tag/workflow discussion here. https://github.com/Clinical-Genomics/cg/issues/3729

cg/constants/delivery.py
henrikstranneheim: I think raw data should at least be unaligned, hence BAM should not  be part of it.

PR #3744, opened 2024-09-18
Fix(Invoicing) Fix crashes when prices are 0

Vince-janv:
## Description
AZ reported that samples with the price 0 crashes the entire invoicing flow

### Changed

- Return 0 instead of None if price is 0


ChrOertlin (approved):

Vince-janv:
Working locallt

PR #3631, opened 2024-08-23
Improve delivery report pytests

ivadym:
### Changed
- Removed meta class dependency from the delivery reports
- Updated delivery report pytests
- Added delivery report pytests for all supported workflows
- Some method have been renamed
- Removed mocked delivery report related APIs

### How to prepare for test

- [ ] Ssh to relevant server (depending on type of change)
- [ ] Use stage: `us`
- [ ] Paxa the environment: `paxa`
- [ ] Install on stage (example for Hasta):
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b [THIS-BRANCH-NAME] -a
    ```

### Tests

| Workflows   | Store Case | Generate Delivery Report | Upload to Scout |
|:-----------:|:--------------:|:--------------------------:|:---------------:|
| raredisease  | X | X | X (manual `delivery-report-to-scout` command) |
| balsamic    | X | X | X |
| mip-dna     | X | X | - |
| rnafusion   | X | X | X |
| taxprofiler  | X | X | - |
| tomte       | X | X | - |

## Review

- [ ] Tests executed by
- [ ] "Merge and deploy" approved by
Thanks for filling in who performed the code review and the test!

### This [version](https://semver.org/) is a

- [ ] **MAJOR** - when you make incompatible API changes
- [ ] **MINOR** - when you add functionality in a backwards compatible manner
- [ ] **PATCH** - when you make backwards compatible bug fixes or documentation/instructions

## Implementation Plan

- [ ] Document in ...
- [ ] Deploy this branch on ...
- [ ] Inform to ...

PR #3774, opened 2024-09-26
Revert "Improve delivery report pytests"

Vince-janv:
Reverts Clinical-Genomics/cg#3631

Prod has reported that delivery reports cannot be generated for balsamic cases


karlnyr (approved):
🙏

PR #3775, opened 2024-09-26
Fix(Balsamic Report) Parse correct type

Vince-janv:
## Description
After merging #3631 production discovered a bug where WGS cases could not have their delivery reports generated.
This PR adds a test replicating the bug and patches it

### Added

- Test to parse files for WGS case

### Changed

- Changed BalsamicAnalysis.sample_metrics to be dict[str, BalsamicQCMetrics]
 instead of a union.


diitaz93 (approved):
👍

PR #3776, opened 2024-09-26
Improve delivery report pytests (second attempt)

Vince-janv:
## Description
Second attempt to merge the changes in #3631. A bug was discovered after merge which resulted in a revert. The bug was patched in #3775.

See #3631 for changes


Vince-janv:
Deployed in production

PR #3777, opened 2024-09-26
Feat(MicroFASTQ) Allow new excel sheet

Vince-janv:
### Added

- Support for uploading new orderform for OrderType.MICROBIAL_FASTQ


cg/constants/orderforms.py
```diff
@@ -28,6 +28,7 @@ class Orderform(StrEnum):
     RNAFUSION: str = "1508"
     RML: str = "1604"
     SARS_COV_2: str = "2184"
+    MICROBIAL_FASTQ: str = "microbial_sequencing"
```
diitaz93, line 31: Isn't there a way to assign also a number to this? Unclear how this works but it seems to me that the new orderform should be standardised with the previous convention

tests/fixtures/orderforms/microbial_sequencing.1.xlsx
diitaz93, line 1: Regarding the standardisation I mentioned above, it seems that the version of the form goes before the name, not after

tests/fixture_plugins/orders_fixtures/order_form_fixtures.py
```diff
@@ -153,6 +153,15 @@ def microbial_orderform(orderforms: Path) -> str:
     ).as_posix()
```
diitaz93, line 153: maybe you can rename this fixture as `microbial_typing_orderform` to make the distinction

tests/apps/orderform/test_excel_orderform_parser.py
```diff
@@ -249,3 +249,17 @@ def test_get_data_delivery(microbial_order_parser):
...
     # THEN no errors should be raised and a data_delivery string should be returned
     assert data_delivery
+
+
+def test_parse_microbial_sequencing_orderform(microbial_sequencing_orderform: str):
+    """Test the parsing of a microbial sequencing orderform."""
+    # GIVEN a microbial sequencing orderform in excel format
+
+    # WHEN parsing the orderform
+    order_form_parser: ExcelOrderformParser = ExcelOrderformParser()
```
diitaz93, line 259:
```suggestion
    order_form_parser = ExcelOrderformParser()
```

diitaz93 (approved):
In general looks good, but I have doubts about the numbering of the formal and made some comments about standardising the naming, but all are disregardable

cg/constants/orderforms.py
Vince-janv: The numbers are basically from our old document-system that had an ID for each document. Since we no longer use it I don't see a point in making up a new number. @islean Do you have an opinion?

islean: I much prefer naming it this way, although I wonder if this will make customer's try this order form for microsalt as well? 
image

tests/fixture_plugins/orders_fixtures/order_form_fixtures.py
islean: Or just "microsalt" orderform so that the layman (me) does not need to know what the connection is?

Vince-janv: Changed to microsalt

Vince-janv:
Tests work on stage with the new ordeform

PR #3783, opened 2024-09-27
Fix raredisease sample pairs in delivery report

peterpru:
## Description
For `cg generate delivery report`, for nextflow pipelines, metrics are collected from `get_multiqc_json_metrics`. All other nextflow pipelines use the original method, but for raredisease this gets overridden, and pairwise sample search patterns are added in addition to the sample IDs. 

This creates a bug when the delivery report is generated, as only sample IDs should be included.

Picture of the error:
image

### Added

-

### Changed

- Raredisease multiQC processing is now raredisease-specific. 

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b fix-raredisease-sample-pairs-list
    ```

### How to test

- [x] Do `cg generate delivery-report amusingmarmoset -f`

### Expected test outcome

- [x] Check that there is no 117 validation errors for NextflowAnalysis.
- [x] Take a screenshot and attach or copy/paste the output. **EDIT**: updated the screenshot with the newest commit, same result, but included date time command as proof.
image

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


diitaz93 (commented):
I think the correct way of fixing this would be fixing the method in raredisease?  Maybe add a method that removes the double sample ids in raredisease. This calling of the class inside itself looks shady to me

cg/meta/workflow/nf_analysis.py
```diff
@@ -904,5 +904,5 @@ def parse_analysis(self, qc_metrics_raw: list[MetricsBase], **kwargs) -> Nextflo
 
     def get_latest_metadata(self, case_id: str) -> NextflowAnalysis:
         """Return analysis output of a Nextflow case."""
-        qc_metrics: list[MetricsBase] = self.get_multiqc_json_metrics(case_id)
+        qc_metrics: list[MetricsBase] = NfAnalysisAPI.get_multiqc_json_metrics(self, case_id)
```
diitaz93, line 907: I would advise against this, as it will remove the possibility of other nextflow pipelines of override this method

peterpru: That is true, I will have another look at the raredisease method. 👍

peterpru: How about this? I have now reverted the `NfAnalysisAPI.get_latest_metadata` to its original state, so that the other pipelines are unaffected. The `RarediseaseAnalysisAPI.get_multiqc_json_metrics` remains as is and is used for doing the QC check, including the sample pairs. And then in the `RarediseaseAnalysisAPI` I explicitly refer to the `NfAnalysisAPI.get_multiqc_json_metrics` in the `get_latest_metadata` step when collecting the metrics to generate the delivery report, so that it only looks for samples, and not sample pairs.

diitaz93 (approved):
👍

cg/meta/workflow/nf_analysis.py
diitaz93: resolve if the new approach works

peterpru:
Still works as expected, thanks @diitaz93 
image

peterpru:
Successfully deployed on stage:
image

peterpru:
Successfully installed in prod:
image

PR #3810, opened 2024-10-07
reduce-log-level-compress

Vince-janv:
## Description
The compress and clean commands for FASTQ are some of the most bloated logs at CG. This PR aims to reduce the spam in this log.

### Changed

- Reduced the log level of several compress related commands

### DIFF
--- a/cg/apps/housekeeper/hk.py
+++ b/cg/apps/housekeeper/hk.py
@@ -7,7 +7,11 @@ from pathlib import Path
 
 from housekeeper.include import checksum as hk_checksum
 from housekeeper.include import include_version
-from housekeeper.store.database import create_all_tables, drop_all_tables, initialize_database
+from housekeeper.store.database import (
+    create_all_tables,
+    drop_all_tables,
+    initialize_database,
+)
 from housekeeper.store.models import Archive, Bundle, File, Tag, Version
 from housekeeper.store.store import Store
 from sqlalchemy.orm import Query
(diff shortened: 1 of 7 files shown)


cg/apps/housekeeper/hk.py
```diff
@@ -273,7 +277,7 @@ def get_latest_bundle_version(self, bundle_name: str) -> Version | None:
         """Get the latest version of a Housekeeper bundle."""
         last_version: Version = self.last_version(bundle_name)
         if not last_version:
-            LOG.warning(f"No bundle found for {bundle_name} in Housekeeper")
+            LOG.debug(f"No bundle found for {bundle_name} in Housekeeper")
```
Vince-janv, line 280: This one I'm unsure of... Do you think it should warn if no bundle exists?

islean: I am confused about the whole method tbh. So we have a wrapper for `self.last_version` but what is the difference between the two apart from the added logging?

Probably out of scope but I do not think there is a need to have both methods present. If we are keeping both, I would say we should log a warning if that is the only difference between the methods.

islean (approved):
👍

cg/apps/housekeeper/hk.py
Vince-janv: I assume there is proper error handling with relevant logging down the line if the method expects anything else than `None`

Vince-janv:
Deployed in stage and prod

PR #3818, opened 2024-10-07
add cli command post process all

diitaz93:
## Description
Add CLI command to post-process all sequencing runs for all instruments. Currently only works for Pacbio.

### Added

- CLI command `cg post-process all`
- Function `post_process_all` in abstract post-processing service
- Tests and fixtures

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b feat-post-process-all -a
    ```

### How to test

- [ ] Do `cg -l DEBUG post-process all --dry-run` 

### Expected test outcome

- [ ] Check that all functions are properly called

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


diitaz93:
## Tests on stage
Comand exists
```shell
$ cg -l DEBUG post-process --help
Usage: cg post-process [OPTIONS] COMMAND [ARGS]...

  Post-process sequencing runs from the sequencing instruments.

Options:
  --help  Show this message and exit.

Commands:
  all  Post-process all sequencing runs from the sequencing instruments.
  run  Post-process a sequencing run from the PacBio instrument.
$ cg -l DEBUG post-process all --help
Running cg post-processing.
Usage: cg post-process all [OPTIONS]

  Post-process all sequencing runs from the sequencing instruments.

Options:
  --dry-run  Runs the command without making any changes
  --help     Show this message and exit
```
Dry run
```shell

```

cg/cli/post_process/post_process.py
```diff
@@ -35,4 +35,19 @@ def post_process_sequencing_run(context: CGConfig, run_name: str, dry_run: bool)
...
+@click.pass_obj
+def post_process_all_sequencing_runs(context: CGConfig, dry_run: bool) -> None:
+    """Post-process all sequencing runs from the sequencing instruments."""
+    services: PostProcessingServices = context.post_processing_services
+    are_all_services_successful: bool = True
+    for service in [services.pacbio]:
+        is_service_successful: bool = service.post_process_all(dry_run=dry_run)
+        are_all_services_successful = are_all_services_successful and is_service_successful
+    if not are_all_services_successful:
+        raise click.Abort
```
ChrOertlin, line 49: not everything will be a sequencing run -> post_process_instrument_run perhaps?

cg/services/run_devices/abstract_classes.py
```diff
@@ -83,6 +86,23 @@ def post_process(self, run_name: str, dry_run: bool = False):
...
+        """Post-process all PacBio runs in the sequencing directory."""
+        LOG.info("Starting PacBio post-processing for all runs")
+        is_post_processing_successful = True
+        for run_name in self._get_all_run_names():
+            try:
+                self.post_process(run_name=run_name, dry_run=dry_run)
+            except Exception as error:
+                LOG.error(f"Could not post-process {run_name}: {error}")
+                is_post_processing_successful = False
+        return is_post_processing_successful
```
ChrOertlin, line 99: Can we make this raise an error instead of swapping a bool when something goes wrong? We can then except this specific error in the CLI and handle it there

cg/cli/post_process/post_process.py
ChrOertlin, line 49: Catching the error here (see comment below) would prevent the bool juggling in line 47

cg/services/run_devices/pacbio/post_processing_service.py
```diff
@@ -49,11 +50,20 @@ def __init__(
...
         self._touch_post_processing_complete(run_data)
+
+    def _get_all_run_names(self) -> list[str]:
+        """Get all run names in the Pacbio sequencing directory."""
+        run_names = []
+        for run_folder in Path(self.sequencing_dir).iterdir():
+            if run_folder.is_dir():
+                for cell_folder in run_folder.iterdir():
+                    run_names.append(f"{run_folder.name}/{cell_folder.name}")
+        return run_names
```
ChrOertlin, line 69: this will crash if some random folder is added to the pacbio directory. I guess we should sanitise the directories somehow.

cg/services/run_devices/pacbio/run_validator/pacbio_run_validator.py
```diff
@@ -33,11 +36,19 @@ def __init__(
...
+        4. Touches a file to indicate that the run is validated
+        5. Skips validation if the run is already validated
+
+        Raises:
+            PostProcessingError: If the run has already been post-processed
         """
+        if Path(run_data.full_path, POST_PROCESSING_COMPLETED).exists():
+            message: str = f"Run {run_data.full_path} is already post-processed, skipping"
+            LOG.warning(message)
+            raise PostProcessingError(message)
```
ChrOertlin, line 51: I don't think this should raise an error. merely skipping to the next (i.e. returning) should be enough. An error should happen when something goes wrong. Here we just want nothing to happen.

cg/services/run_devices/pacbio/post_processing_service.py
ChrOertlin: Will end up with a reg ex check then I suppose...

ChrOertlin: current folder in prod btw: 
image

cg/services/run_devices/pacbio/run_validator/pacbio_run_validator.py
diitaz93: Right now, if it returns None, the service will start the post-processing and that is what we don't want. We want this to stop here. The error is caught anyway

cg/cli/post_process/post_process.py
diitaz93: fixed

cg/services/run_devices/abstract_classes.py
diitaz93: The problem I see with raising an error here is it would terminate the iteration early and some smart cells would not start their post-processing despite being able to

cg/cli/post_process/post_process.py
```diff
@@ -35,4 +35,19 @@ def post_process_sequencing_run(context: CGConfig, run_name: str, dry_run: bool)
...
+@click.pass_obj
+def post_process_all_runs(context: CGConfig, dry_run: bool) -> None:
+    """Post-process all runs from the instruments."""
+    services: PostProcessingServices = context.post_processing_services
+    are_all_services_successful: bool = True
+    for service in [services.pacbio]:
+        is_service_successful: bool = service.post_process_all(dry_run=dry_run)
+        are_all_services_successful = are_all_services_successful and is_service_successful
+    if not are_all_services_successful:
+        raise click.Abort
```
islean, line 49: What does this do exactly? We are already on the last line of the command so aborting should not do anything right?

cg/services/run_devices/abstract_classes.py
islean: For me the most reasonable would probably be to raise an error in post_process_run which post_process_all catches. And if any run has crashed in the post_process_all, then it can raise an error which we can catch in the CLI. It does sound like the CLI might be doing a bit too much at that stage though..

cg/services/run_devices/pacbio/post_processing_service.py
islean: Should we not be able to rely on the data being correctly formatted in prod? To me I think we should remove the runs which break this logic.

cg/cli/post_process/post_process.py
diitaz93: It finishes with status != 0 and prints that in stdout. I have not stopped to question this before, maybe is not needed but I will leave it for now

cg/services/run_devices/pacbio/run_validator/pacbio_run_validator.py
islean: Maybe we can add the check to the `_get_all_runs` methods instead? And rename it to `_get_all_runs_to_postprocess` or something?

islean: An error does seem a bit weird to raise if this is to be expected

cg/services/run_devices/pacbio/run_validator/pacbio_run_validator.py
```diff
@@ -58,9 +68,13 @@ def ensure_post_processing_can_start(self, run_data: PacBioRunData) -> None:
...
-        Path(run_path, PacBioDirsAndFiles.RUN_IS_VALID).touch()
+    def _is_post_processed(run_path: Path) -> bool:
+        return Path(run_path, PacBioDirsAndFiles.POST_PROCESSING_COMPLETED).exists()
 
     @staticmethod
     def _is_validated(run_path: Path) -> bool:
         return Path(run_path, PacBioDirsAndFiles.RUN_IS_VALID).exists()
+
+    @staticmethod
+    def _touch_is_validated(run_path: Path):
```
islean, line 79:
```suggestion
    def _touch_is_validated(run_path: Path) -> None:
```

tests/fixture_plugins/pacbio_fixtures/service_fixtures.py
```diff
@@ -58,14 +74,15 @@ def pac_bio_post_processing_service(
     pac_bio_run_data_generator: PacBioRunDataGenerator,
     pac_bio_housekeeper_service: PacBioHousekeeperService,
     pac_bio_store_service: PacBioStoreService,
-    pacbio_barcoded_sequencing_run_name: str,
+    mock_pacbio_run_validator: PacBioRunValidator,
```
islean, line 77: Do we need this? Seems as if we do not pass it anywhere

cg/cli/post_process/post_process.py
islean: But are these not the patterns we want to break when writing new code?

cg/cli/post_process/post_process.py
```diff
@@ -35,4 +35,19 @@ def post_process_sequencing_run(context: CGConfig, run_name: str, dry_run: bool)
...
 
-post_process_group.add_command(post_process_sequencing_run)
+@post_process_group.command(name="all")
+@DRY_RUN
+@click.pass_obj
+def post_process_all_runs(context: CGConfig, dry_run: bool) -> None:
+    """Post-process all runs from the instruments."""
+    services: PostProcessingServices = context.post_processing_services
+    are_all_services_successful: bool = True
+    for service in [services.pacbio]:
```
islean, line 45: I guess what we would aim for in the end is to just go "for service in services" here? It is just not possible currently

cg/services/run_devices/pacbio/run_validator/pacbio_run_validator.py
diitaz93: or we can make it return a bool, renaming to is_post_processing_ready

cg/services/run_devices/pacbio/post_processing_service.py
Vince-janv: There are currently two folders in prod that do not seem to be tied to any specific run (`355`, `TransferSchemeTests`). We could tell the sync to hasta to ignore them and just assume that the folder will only contain runs.

That simplifies the logic but might make the command a bit brittle. What do you think?

cg/services/run_devices/abstract_classes.py
ChrOertlin: > The problem I see with raising an error here is it would terminate the iteration early and some smart cells would not start their post-processing despite being able to

Yes, the post_process_all function makes things more complicated that just iterating through the dirs and running the post process function in the CLI. I think that would prevent the need for all this extra stuff.

Should we re think the approach?

cg/services/run_devices/pacbio/post_processing_service.py
ChrOertlin: I think making sure we only process pac bio runs in the backend is more robust that relying on an assumption that can easily be violated by anyone who can run mkdir in prod.

cg/services/run_devices/pacbio/run_validator/pacbio_run_validator.py
ChrOertlin: I think the check is not in the proper place. Ensure_post_processing_can_start has should check if it can start and not if it has already been processed. I think this goes against SRP.

diitaz93: Agree, moving to the post-processing service

cg/cli/post_process/post_process.py
diitaz93: I don't understand, what do you suggest here?

cg/cli/post_process/post_process.py
diitaz93: Is it? This is how the Illumina post-processing works now and I don't think there is any complain about it. Why is this pattern undesirable?

cg/services/run_devices/abstract_classes.py
islean: To me the tradeoff would still be that we would need to go

Fetch device from directory name -> Fetch expected directory structure from device -> get run name -> from run name get device (which we in principle already have determined and uses a regex) -> post process run

To me, handling errors in the current approach is still more attractive and simpler, but we could certainly go for another refinement.

cg/services/run_devices/pacbio/post_processing_service.py
ChrOertlin: in this case the check would be to ignore folders that are not are not named properly like a run.
It would make sure the code should not reach a state where it breaks due to malformed data, which i think is more of a plus in an automated system. If the data within the expected directories is malformed then it should break of course but not due to that a wrong directory is read.

Vince-janv: I think I agree with Isak ☝️ Given that we proceed with the next run either way I think simplicity is prio 1

cg/services/run_devices/abstract_classes.py
diitaz93: closing after refinement

cg/cli/post_process/post_process.py
```diff
@@ -35,4 +35,25 @@ def post_process_sequencing_run(context: CGConfig, run_name: str, dry_run: bool)
...
-post_process_group.add_command(post_process_sequencing_run)
+@post_process_group.command(name="all")
+@DRY_RUN
+@click.pass_obj
+def post_process_all_runs(context: CGConfig, dry_run: bool) -> None:
+    """Post-process all runs from the instruments."""
+    services: PostProcessingServices = context.post_processing_services
+    names_services: RunDirectoryNamesServices = context.run_directory_names_services
+    are_all_services_successful: bool = True
+    for service, name_service in [(services.pacbio, names_services.pacbio)]:
```
diitaz93, line 46: I would like a better way to do this in the future when we have more devices in this flow

cg/cli/post_process/post_process.py
islean: Resolved IRL

cg/services/run_devices/pacbio/post_processing_service.py
```diff
@@ -49,7 +50,10 @@ def __init__(
         to_raise=PostProcessingError,
     )
     def post_process(self, run_name: str, dry_run: bool = False) -> None:
-        LOG.info(f"Starting PacBio post-processing for run: {run_name}")
+        if self._is_run_post_processed(run_name):
```
islean, line 53: Did we not say that this should check should be where we pick up the runs to iterate over?

diitaz93: Good catch, I forgot about that. I also double checked and if we try to post-process an already existing smrt cell it will fail before writing anything into the database or saving stuff to Housekeeper so this is not needed here

cg/cli/post_process/post_process.py
```diff
@@ -35,4 +35,25 @@ def post_process_sequencing_run(context: CGConfig, run_name: str, dry_run: bool)
...
+        try:
+            service.post_process_all(
+                run_directory_names=name_service.get_run_names(), dry_run=dry_run
+            )
+        except Exception as error:
```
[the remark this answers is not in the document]
diitaz93: I agree that the error handling is too much, but in the <do-something>-all commands we always except all Exceptions, otherwise it will halt production if there is a, for example, Pydantic error or something else

cg/services/run_devices/pacbio/post_processing_service.py
diitaz93: fixed

cg/cli/post_process/post_process.py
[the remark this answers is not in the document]
islean: Then I do not see the point in wrapping our errors at all. The whole point is to be able to perform clearer error handling?

PR #3822, opened 2024-10-08
fix(no delivery files check)

ChrOertlin:
## Description

Moves the no delivery files check to the delivery file fetcher.
adds customer inbox path to delivery meta data, this was needed to fetch the inbox path from the model instead of sample files. As with the new changes empty sample files were allowed, this would have crashed.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/deliver_files/delivery_file_fetcher_service/delivery_file_fetcher_service.py
```diff
@@ -28,3 +32,13 @@ def __init__(
     def get_files_to_deliver(self, case_id: str) -> DeliveryFiles:
         """Get the files to deliver."""
         pass
+
+    @staticmethod
+    def _validate_delivery_has_content(delivery_files: DeliveryFiles) -> DeliveryFiles:
```
islean, line 37: Slight preference for having this method not return anything and just raising an error if all files are missing, but feel free to go either way

cg/services/deliver_files/delivery_file_fetcher_service/error_handling.py
```diff
@@ -14,10 +15,20 @@ def handle_missing_bundle_errors(func):
...
             return func(*args, **kwargs)
-        except HousekeeperBundleVersionMissingError as error:
-            LOG.error(error)
+        except HousekeeperBundleVersionMissingError:
+            msg: str = f"Missing bundles detected for:"
+            if case_id:
+                msg += f" case {case_id}"
+            if sample_id:
+                msg += f" sample {sample_id}"
+            if case:
```
islean, line 29: A bit confused by the possibility of both a case_id and a case here. But if not important, feel free to resolve.

islean (commented):
I like it! But do we actually set the customer_inbox at any point?

cg/services/deliver_files/delivery_file_fetcher_service/delivery_file_fetcher_service.py
ChrOertlin: I understand, I will keep it this way then. The other one would require an extra line of code :P

ChrOertlin:
# test

ticket with cases that are not in HK
image

The messages seem a lot, but this is a ticket where nothing is in HK. So now all messages are written, this is not the case when just one sample is missing in a case for example

islean:
> I like it! But do we actually set the customer_inbox at any point?

Just pinging this question since it was the bulk of the review

diitaz93:
In `cg/cli/deliver/utils.py:deliver_raw_data_for_analyses` we will set the uploaded_at date to `datetime.now()` even if there are no files to deliver. This is the current behaviour, (so not a consequence of the code in this PR) but I still think it should not be that way. This happens because when there are no files the function `deliver_files_for_case` exits successfully and returns None. The same happens for all the pipeline's UploadAPI method `upload`.

cg/services/deliver_files/delivery_file_fetcher_service/error_handling.py
ChrOertlin: two decorators that get called in the appropraite place

ChrOertlin:
> > I like it! But do we actually set the customer_inbox at any point?
> 
> Just pinging this question since it was the bulk of the review

yes it is set in the move service

islean (approved):
💯

cg/services/deliver_files/delivery_file_fetcher_service/error_handling.py
```diff
@@ -14,10 +15,20 @@ def handle_missing_bundle_errors(func):
...
+                msg += f" case {case.internal_id}"
+            LOG.error(msg)
```
diitaz93, line 31: 💪

diitaz93 (approved):
Looks good! Still concerned about the uploaded at date updated if there are no files to upload, maybe it is worth taking a look

cg/services/deliver_files/delivery_file_fetcher_service/error_handling.py
```diff
@@ -14,10 +15,20 @@ def handle_missing_bundle_errors(func):
...
     def wrapper(*args, **kwargs):
+        case_id: str = kwargs.get("case_id")
+        sample_id: str = kwargs.get("sample_id")
+        case: Case = kwargs.get("case")
         try:
             return func(*args, **kwargs)
-        except HousekeeperBundleVersionMissingError as error:
-            LOG.error(error)
+        except HousekeeperBundleVersionMissingError:
+            msg: str = f"Missing bundles detected for:"
```
diitaz93, line 24:
```suggestion
            msg: str = "Missing bundles detected for:"
```

PR #3830, opened 2024-10-10
fix(set case priority research bug)

ChrOertlin:
## Description

The Priority enum links the input research to 0.
In the set case there is a check for `if not any(options)` . In the case for priority research the options look like. `[0]` 

According to the following:
```
>>> test = [0]
>>> any(test)
False
```

So setting priority to research would exit here with nothing to change, which is a bug.

Changed the check to evaluate the length of the options, which is a list[str] but can also be a list of int it seems.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/cli/set/case.py
+++ b/cg/cli/set/case.py
@@ -84,7 +84,7 @@ def set_case(
 
 
 def abort_on_empty_options(options: list[str]) -> None:
-    if not any(options):
+    if len(options) > 1:
         LOG.error("Nothing to change")
         raise click.Abort
 


cg/cli/set/case.py
```diff
@@ -84,7 +84,7 @@ def set_case(
 
 
 def abort_on_empty_options(options: list[str]) -> None:
-    if not any(options):
+    if len(options) > 1:
```
islean, line 87: Maybe
```suggestion
    if len(options) < 1:
```

cg/cli/set/case.py
```diff
@@ -77,18 +80,25 @@ def set_case(
...
+    if isinstance(priority, Priority):
         update_priority(case=case, priority=priority)
 
     status_db.session.commit()
 
 
-def abort_on_empty_options(options: list[str]) -> None:
-    if not any(options):
+def abort_on_empty_options(
+    options: list[str], bypass_check: Callable[[any], bool], priority: Priority | None
```
islean, line 90: I get that it is expandable but the bypass_check parameter is quite hard to understand. Why are we allowing for this to be passed as a parameter rather than always just checking if the priority is research?

cg/cli/set/case.py
```diff
@@ -77,18 +80,25 @@ def set_case(
...
+) -> None:
+    # Abort if options are empty and bypass_check for priority returns False
```
islean, line 92:
```suggestion
    """Abort if options are empty and bypass_check for priority returns False"""
```

cg/cli/set/case.py
```diff
@@ -77,18 +80,25 @@ def set_case(
...
+    if not any(options) and not bypass_check(priority):
```
islean, line 93: This check seems unclearer to me than just saying `if len(options) == 0:`. Is there a reason for writing it like this?

cg/cli/set/case.py
```diff
@@ -77,18 +80,25 @@ def set_case(
     if panel_abbreviations:
         update_panels(case=case, panel_abbreviations=panel_abbreviations, status_db=status_db)
 
-    if priority:
+    if isinstance(priority, Priority):
```
islean, line 83: I would probably keep this check and change the check in abort on empty options to if len(options) == 0:. Would those changes miss anything?

islean (commented):
I think the change does what it wants but it does so in a quite convoluted way - could we try to make it clearer?

cg/cli/set/case.py
islean: We could probably do an `if priority is not None` here as well, if that is easier to read

cg/cli/set/case.py
ChrOertlin: len option is never smaller than one, since it is a list full of stuff. I tried that initially but it broke loads of tests.

cg/cli/set/case.py
```diff
@@ -1,6 +1,7 @@
 """Set case attributes in the status database."""
 
 import logging
+from typing import Callable
```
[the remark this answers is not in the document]
diitaz93: but I would rather refactor the whole `abort_on_empty_options` to not use the bypassing, so no need for callable

diitaz93 (commented):
why not checking `options == []`??

cg/cli/set/case.py
```diff
@@ -77,18 +80,25 @@ def set_case(
...
-    if priority:
+    if isinstance(priority, Priority):
         update_priority(case=case, priority=priority)
 
     status_db.session.commit()
 
 
-def abort_on_empty_options(options: list[str]) -> None:
-    if not any(options):
+def abort_on_empty_options(
```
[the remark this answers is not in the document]
ChrOertlin: the list never seems to be just empty.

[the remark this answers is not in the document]
diitaz93: noted

cg/cli/set/case.py
```diff
@@ -77,14 +78,15 @@ def set_case(
...
         update_priority(case=case, priority=priority)
 
     status_db.session.commit()
 
 
-def abort_on_empty_options(options: list[str]) -> None:
-    if not any(options):
+def abort_on_empty_options(options: list[str], priority: Priority | None) -> None:
+    """Abort if options are empty and bypass_check for priority returns False"""
+    if not any(options) and not priority == Priority.research:
```
islean, line 89: Works potentially but is it not more a `if all(option is None for option in options)` we want?

diitaz93 (approved):
The current solution works too

cg/cli/set/case.py
```diff
@@ -77,14 +78,15 @@ def set_case(
...
         LOG.error("Nothing to change")
         raise click.Abort
```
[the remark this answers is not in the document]
diitaz93: Feel free to not implement, I think the current solution also works

cg/cli/set/case.py
islean: I do see, we will have a long list with multiple None entries

cg/cli/set/case.py
ChrOertlin: No, there is an empty tuple in there, it wont work

PR #3832, opened 2024-10-10
feat - new post process all command

diitaz93:
## Description
Add a service to fetch instrument run names for PacBio.

### Added

- Abstract service RunNamesServie to expand for all instruments
- PacBioRunNamesService
- Tests and fictures

### How to prepare for test

- no tests on stage, no CLI command nor production code affected. Tests will be done for #3818 

## Review

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/run_devices/run_names/pacbio.py
```diff
@@ -0,0 +1,20 @@
+from pathlib import Path
+
+from cg.services.run_devices.run_names.service import RunDirectoryNamesService
+
+
+class PacbioRunDirectoryNamesService(RunDirectoryNamesService):
+
+    def get_run_names(self) -> list[str]:
+        """
+        Get all the run names from a run directory, assuming the run directory has only sequencing
```
diitaz93, line 10:
```suggestion
        Get all the run directory names from the PacBio sequencing directory, assuming the run directory has only sequencing
```

cg/services/run_devices/run_names/service.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class RunDirectoryNamesService(ABC):
+
+    def __init__(self, run_directory: str):
+        self.run_directory = run_directory
+
+    @abstractmethod
+    def get_run_names(self) -> list[str]:
+        """Get all the run names from a run directory."""
```
diitaz93, line 11:
```suggestion
        """Get all the run directory names from a data directory."""
```

cg/models/cg_config.py
```diff
@@ -633,6 +634,17 @@ def pdc_service(self) -> PdcService:
             self.pdc_service_ = service
         return service
 
+    @property
+    def run_directory_names_services(self) -> RunDirectoryNamesServices:
+        services = self.__dict__.get("run_directory_names_services_")
```
islean, line 639: What is the difference between this and `self.run_directory_names_services_`?

cg/models/cg_config.py
```diff
@@ -633,6 +634,17 @@ def pdc_service(self) -> PdcService:
...
+        if services is None:
+            LOG.debug("Instantiating run directory names services")
+            services = RunDirectoryNamesServices(
+                pacbio=PacbioRunDirectoryNamesService(self.run_instruments.pacbio.data_dir)
```
islean, line 643:
```suggestion
            pacbio_service = PacbioRunDirectoryNamesService(self.run_instruments.pacbio.data_dir)
            services = RunDirectoryNamesServices(pacbio=pacbio_service)
```

cg/services/run_devices/run_names/pacbio.py
```diff
@@ -0,0 +1,20 @@
...
+        folders with SMRTcells inside. The run names are in the form <sequencing_run>/<SMRTcell>:
+        Example:
+            r84202_20240913_121403/1_C01
+        """
```
islean, line 14: I think this docstring just adds confusion by trying to excuse us expecting a correct folder structure on Hasta. To me, the only relevant info is the structure of the run names, since we already know that we are getting the pacbio run names by the class and method name

cg/services/run_devices/run_names/pacbio.py
```diff
@@ -0,0 +1,20 @@
...
+        run_names = []
+        for run_folder in Path(self.run_directory).iterdir():
+            if run_folder.is_dir():
```
islean, line 17:
```suggestion
```
Just a suggestion that if we are to be relying on the structure being correct, this check should be unnecessary right? If we want it, it might be that we do not want to trust the structure at all and redesign a bit. But I am not a fan of half measures

cg/services/run_devices/run_names/service.py
islean: What is the difference? Is the run directory name the run name + well?

islean (approved):
Does what we want! I do think that we should decide to either fully trust the directory structure or implement proper pruning though.

cg/services/run_devices/run_names/pacbio.py
diitaz93: done

cg/models/cg_config.py
islean: Does it crash with my suggestion? Otherwise I would probably change it

cg/services/run_devices/run_names/pacbio.py
diitaz93: That is a good point. But, as it is now, if there is a **directory** with an unexpected structure, an incorrect name would be generated and this will raise an error in the post-processing, which will be caught. However, If there is a file in the sequencing directory. the `run_folder.iterdir()` will raise a `NotADirectoryError` which we currently don't catch. This is an easy solution to that.

cg/services/run_devices/run_names/pacbio.py
diitaz93: fixed

cg/services/run_devices/run_names/service.py
diitaz93: true, dismissing

cg/models/cg_config.py
```diff
@@ -360,6 +353,13 @@ class RunInstruments(BaseModel):
     illumina: IlluminaConfig
 
 
+class RunDirectoryNamesServices(BaseModel):
+    pacbio: PacbioRunDirectoryNamesService
+
+    class Config:
+        arbitrary_types_allowed = True
```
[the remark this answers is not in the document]
diitaz93: fixed

cg/models/cg_config.py
diitaz93: changed

cg/models/cg_config.py
diitaz93: For me both are basically the same, but the original looks more expandable (only adding one line instead of two), I will keep it the way it is

islean: Do so!

cg/services/run_devices/run_names/pacbio.py
islean: I would say that having a file in this directory is enough cause for concern to not catch the error though? Or why would we want to disregard that?

cg/models/cg_config.py
```diff
@@ -360,11 +353,14 @@ class RunInstruments(BaseModel):
     illumina: IlluminaConfig
 
 
+class RunDirectoryNamesServices(BaseModel):
+    pacbio: PacbioRunDirectoryNamesService
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
Vince-janv, line 358: Why do we need `arbitrary_types_allowed`?

Vince-janv (approved):
Looks good!

cg/models/cg_config.py
diitaz93: It only happens when the types are inherited types, same happened with the PostProcessingService

Vince-janv: Could you elaborate? What happens and which object are you referring to as inherited?

islean: As far as I know this has nothing to do with inheritance. By default your own custom classes are not allowed as a field in a BaseModel - they should be standard types or other BaseModels (might be some caveats): https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.arbitrary_types_allowed

cg/services/run_devices/run_names/pacbio.py
diitaz93: in person: decided to keep it

cg/models/cg_config.py
diitaz93: But this does not happen with the RunInstruments class in line 350

Vince-janv: That's because they inherit from BaseModel

diitaz93: So the attributes of the `RunInstruments` class are pydantic objects that inherit directly from BaseModel, while the atreibutes of `RunDirectoryNamesServices` are pydantic objects that inherit directly from `RunNamesService`. That is where the issue is I think

islean: As far as I can tell, the attributes of RunDirectoryNamesServices are not pydantic objects at all

cg/services/run_devices/abstract_classes.py
```diff
@@ -83,6 +87,27 @@ def post_process(self, run_name: str, dry_run: bool = False):
...
         pass
 
+    def post_process_all(self, run_names: list[str], dry_run: bool = False) -> bool:
+        """Post-process all PacBio runs in the sequencing directory."""
+        LOG.info("Starting post-processing for all runs")
+        is_post_processing_successful = True
+        for run_name in run_names:
+            if self._is_run_post_processed(run_name):
+                LOG.info(f"Run {run_name} has already been post-processed")
+                continue
```
ChrOertlin, line 97: move this into the self.post_process

cg/services/run_devices/pacbio/post_processing_service.py
```diff
@@ -39,6 +39,7 @@ def __init__(
         self.hk_service: PacBioHousekeeperService = hk_service
         self.store_service: PacBioStoreService = store_service
         self.sequencing_dir: str = sequencing_dir
+        self.instrument: str = "PacBio"
```
ChrOertlin, line 42:
```suggestion
        self.instrument: str = "Pacbio"
```

cg/models/cg_config.py
diitaz93: Ohh yes my bad, they inherit from ABC

cg/services/run_devices/abstract_classes.py
```diff
@@ -83,6 +87,27 @@ def post_process(self, run_name: str, dry_run: bool = False):
         """Store sequencing metrics in StatusDB and relevant files in Housekeeper."""
         pass
 
+    def post_process_all(self, run_names: list[str], dry_run: bool = False) -> bool:
```
ChrOertlin, line 90: I forgot what we decided. Did we want to keep the post_process_all now ?

cg/services/run_devices/abstract_classes.py
diitaz93: I receive conflicting arguments now, as @islean said we agreed to put this in the post_process_all. I think there is merit in keeping it here

ChrOertlin: ah I see it is on the ABC... I don't think this is the right approach more and more. We just seem to need to find work arounds and copy logic into places.

the self._is_run_post_processed should also happen in the self.post_post_process.

Getting more and more in favour of fetching the services in the CLI and just going through the dirs and call self.post_process.

cg/services/run_devices/abstract_classes.py
diitaz93: yes, what we agreed was to improve the error handling and reduce the boolean juggling

cg/services/run_devices/abstract_classes.py
islean: To clarify - I do not want this in the post_process_all - I want this functionality in `RunNamesService.get_run_names`. I was also not envisioning the implementation of `post_process_all` to be put on the parent `PostProcessingService`, as I do agree that this is a limiting design choice. But on the other hand, I can not see what steps in the parts of `post_process_all` which would differ between the devices if the functionality of the post_process_all would be (excluding error handling)

```
runs = self.run_names_service.get_run_names
for run in runs:
    self.post_process(run)
```

ChrOertlin: `I want this functionality in `RunNamesService.get_run_names` the run names should just return the list of directories and should not care about its state of post processing. Otherwise you require extra functionality to fetch directories for cleaning that does not have this check in place for example.

cg/services/run_devices/abstract_classes.py
```diff
@@ -83,6 +87,27 @@ def post_process(self, run_name: str, dry_run: bool = False):
...
+                LOG.error(f"Could not post-process {self.instrument} run {run_name}: {error}")
+                is_post_processing_successful = False
+        if not is_post_processing_successful:
+            raise PostProcessingError(f"Could not post-process all {self.instrument} runs")
+
+    def _is_run_post_processed(self, run_name: str) -> bool:
+        """Check if a run has been post-processed."""
+        processing_complete_file = Path(self.sequencing_dir, run_name, POST_PROCESSING_COMPLETED)
+        return processing_complete_file.exists()
+
```
ChrOertlin, line 110: Also both of these functions access a `self.sequencing_dir` and `self.insturment` This is a concrete implementation that requires all the interfaces that implement this ABC to have these parameters. we should not do this, as it introduces a dependecy.

cg/services/run_devices/abstract_classes.py
islean: Good point - then I want another method in the service called `get_runs_to_post_process` or something which does care about the state of post processing. 

I do think that the functionality ensuring that all previously post processed runs are not post processed again should be happening either in
1. picking up runs, or
2. ensuring that post processing can start.

I am pretty certain that it was suggested to be added in 2. but it was rebuked. What is the benefit of adding it in the actual post_process command?

cg/services/run_devices/pacbio/post_processing_service.py
diitaz93: fixed

cg/services/run_devices/abstract_classes.py
diitaz93: started new branch with this option, resolving

cg/services/run_devices/abstract_classes.py
diitaz93: I think it is the base of post-processing services to depend on a `sequencing_dir` where to find the runs to post-process. 

Regarding the `self.instrument`, I don't see a problem with having it but it is not crucial, it is just for the logs, it could be removed

cg/models/cg_config.py
diitaz93: Can I resolve this? @Vince-janv @islean

Vince-janv: Yes

diitaz93:
Closing in favour of https://github.com/Clinical-Genomics/cg/pull/3845

PR #3855, opened 2024-10-16
feat(deliver raw data files for single samples)

ChrOertlin:
## Description

Add a CLI command to deliver raw data files for single samples.

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/cli/deliver/base.py
+++ b/cg/cli/deliver/base.py
@@ -133,6 +133,54 @@ def deliver_ticket(
     )
 
 
+@deliver.command(
+    name="sample",
+    help="Deliver all case files based on delivery type to the customer inbox on the HPC "
+    "and start an Rsync job to clinical-delivery. "
+    "NOTE: the dry-run flag will copy files to the customer inbox on Hasta, "
+    "but will not perform the Rsync job.",
+)
+@click.pass_obj
+@click.option(
+    "-c",
+    "--case-id",
+    required=True,
+    help="Deliver the files for a specific case.",
+)
+@click.option("-s", "--sample-id", required=True, help="Deliver the files for a specific sample.")
+@click.option(
+    "-d",
+    "--delivery-type",
+    required=False,
+    help="The delivery type to use.",
+    type=click.Choice(choices=["fastq", "bam"]),
+)
+@DRY_RUN
+def deliver_sample_raw_data(
+    context: CGConfig,
+    case_id: str,
+    sample_id: str,
+    delivery_type: FileDeliveryOption,
+    dry_run: bool,
+):
+    """
+    Deliver all case files based on delivery type to the customer inbox on the HPC
+    """
+    inbox: str = context.delivery_path
+    service_builder: DeliveryServiceFactory = context.delivery_service_factory
+    case: Case = context.status_db.get_case_by_internal_id(internal_id=case_id)
+    if not case:
+        LOG.error(f"Could not find case with id {case_id}")
+        return
+    delivery_service: DeliverFilesService = service_builder.build_delivery_service(
+        delivery_type=delivery_type if delivery_type else case.data_delivery,
+        workflow=case.data_analysis,
+    )
+    delivery_service.deliver_files_for_sample(
+        case=case, sample_id=sample_id, delivery_base_path=Path(inbox), dry_run=dry_run
+    )
+
+
 @deliver.command(name="auto-raw-data")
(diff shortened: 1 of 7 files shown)


cg/cli/deliver/base.py
```diff
@@ -133,6 +133,54 @@ def deliver_ticket(
...
+    help="The case the sample is on.",
+)
+@click.option("-s", "--sample-id", required=True, help="Deliver the files for a specific sample.")
+@click.option(
+    "-d",
+    "--delivery-type",
+    required=True,
+    help="The delivery type to use.",
+    type=click.Choice(choices=["fastq", "bam"]),
+)
```
islean, line 157: Should we not get this from the case?

ChrOertlin: No, if the case is not deliverytype fastq or bam you couldn't deliver raw files for it that way.

islean: And we want this functionality to work for cases which does not have those delivery types? If so, please resolve

ChrOertlin: We want to be able to deliver raw-data for single samples when we want to even if the original case has only "analysis" as delivery type.

diitaz93: I also feel that this could be extracted, it is not as a sample could be delivered in any of the two ways. Why not getting this from the application version in the sample? Would we need the PR of the new table for this? @islean

ChrOertlin: The application version has no relation to this part of the code.

If a customer requests delivery of fastq files for a sample tjat was initially ordered as "analysis" you will need to provide the requested delivery type in order to override it. This is a feature requested by prod.

islean: My concerns are mainly centered around what happens if the provided case can be misaligned with the options provided by the production member in any way

cg/cli/deliver/base.py
```diff
@@ -133,6 +133,54 @@ def deliver_ticket(
...
+):
+    """
+    Deliver fastq or bam files for a single sample to the customer inbox on the HPC
+    """
+    inbox: str = context.delivery_path
+    service_builder: DeliveryServiceFactory = context.delivery_service_factory
+    case: Case = context.status_db.get_case_by_internal_id(internal_id=case_id)
+    if not case:
+        LOG.error(f"Could not find case with id {case_id}")
+        return
```
diitaz93, line 174:
```suggestion
        click.Abort
```

tests/services/file_delivery/file_filter/test_sample_filter_service.py
```diff
@@ -0,0 +1,22 @@
+from cg.services.deliver_files.file_fetcher.models import DeliveryFiles
+from cg.services.deliver_files.file_filter.sample_service import SampleFileFilter
+
+
+def test_filter_delivery_files(expected_fastq_delivery_files: DeliveryFiles, sample_id: str):
+    """Test to filter delivery files."""
+
+    # GIVEN a delivery files object with multiple sample ids, a sample id and a filter delivery files service
```
diitaz93, line 8:
```suggestion
    # GIVEN a delivery files object with multiple sample ids and a filter delivery files service
```

diitaz93 (approved):
Looks good 👍 I would say that creating a whole family of services to filter files by sample id was a bit of an overkill. But it works

cg/services/deliver_files/file_filter/abstract.py
```diff
@@ -0,0 +1,10 @@
+from abc import abstractmethod, ABC
+
+from cg.services.deliver_files.file_fetcher.models import DeliveryFiles
+
+
+class FilterDeliveryFilesService(ABC):
```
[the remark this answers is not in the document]
ChrOertlin: leaves room for expandability in case we require other types of filtering down the line.

cg/cli/deliver/base.py
ChrOertlin: it is a manual command here an error will be logged not sure if click abort adds any value

ChrOertlin:
# tested in stage

PR #3844, opened 2024-10-14
Add order type application table

islean:
## Description

First part of https://github.com/Clinical-Genomics/add-new-tech/issues/123. Adds a new table for mapping order types to their allowed applications

### Added

- OrderTypeApplication table

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Ssh to relevant server (depending on type of change)
- [ ] Use stage: `us`
- [ ] Paxa the environment: `paxa`
- [ ] Install on stage (example for Hasta):
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b add-ordertype-application-table -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...
- [ ] Take a screenshot and attach or copy/paste the output.

## Review

- [ ] Tests executed by
- [ ] "Merge and deploy" approved by
Thanks for filling in who performed the code review and the test!

### This [version](https://semver.org/) is a

- [ ] **MAJOR** - when you make incompatible API changes
- [ ] **MINOR** - when you add functionality in a backwards compatible manner
- [ ] **PATCH** - when you make backwards compatible bug fixes or documentation/instructions

## Implementation Plan

- [ ] Document in ...
- [ ] Deploy this branch on ...
- [ ] Inform to ...

PR #3856, opened 2024-10-16
Hook up ordertype_application creation

islean:
## Description

With the table added in #3844 we need a way to add entries via the web UI. This allows a multiple choice option when adding or editing an entry in the application table

### Added

- Option to specify order types that an application is applicable for

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/server/admin.py
+++ b/cg/server/admin.py
@@ -7,9 +7,12 @@ from flask_admin.actions import action
 from flask_admin.contrib.sqla import ModelView
 from flask_dance.contrib.google import google
 from markupsafe import Markup
+from wtforms.fields.choices import SelectMultipleField
 
 from cg.constants.constants import NG_UL_SUFFIX, CaseActions, DataDelivery, Workflow
+from cg.models.orders.constants import OrderType
 from cg.server.ext import db, sample_service
+from cg.store.models import OrderTypeApplication
 from cg.utils.flask.enum import SelectEnumField
 
 
@@ -151,6 +154,11 @@ class ApplicationView(BaseView):
     column_filters = ["prep_category", "is_accredited"]
     column_searchable_list = ["tag", "prep_category"]
     form_excluded_columns = ["category", "versions"]
+    form_extra_fields = {
+        "suitable_order_types": SelectMultipleField(
+            "Order Types", choices=[(choice, choice.name) for choice in OrderType]
+        )
+    }
 
     @staticmethod
     def view_application_link(unused1, unused2, model, unused3):
@@ -167,6 +175,17 @@ class ApplicationView(BaseView):
             if model.application
             else ""
         )
+
+    def on_model_change(self, form, model, is_created):
+        """Override to persist entries to the OrderTypeApplication table"""
+        super(ApplicationView, self).on_model_change(form, model, is_created)
+        db.session.query(OrderTypeApplication).filter_by(application_id=model.id).delete()
+        selected_order_types: list[OrderType] = form.suitable_order_types.data
+        for order_type in selected_order_types:
+            print(f"Trying to add entry with {model.id} and {order_type}")
+            db.session.add(OrderTypeApplication(application=model, order_type=order_type))
+
+        db.session.commit()
 
 
 class ApplicationVersionView(BaseView):
--- a/cg/server/admin.py
+++ b/cg/server/admin.py
@@ -8,11 +8,12 @@ from flask_admin.contrib.sqla import ModelView
 from flask_dance.contrib.google import google
 from markupsafe import Markup
 from wtforms.fields.choices import SelectMultipleField
+from wtforms.form import Form
 
 from cg.constants.constants import NG_UL_SUFFIX, CaseActions, DataDelivery, Workflow
 from cg.models.orders.constants import OrderType
 from cg.server.ext import db, sample_service
-from cg.store.models import OrderTypeApplication
+from cg.store.models import Application
 from cg.utils.flask.enum import SelectEnumField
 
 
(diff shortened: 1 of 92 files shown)


islean:
How it is showcased currently:

image

Would we want tick boxes instead? And do we want to showcase the chosen order types in the application table?

islean:
Looks horrendous with alignment if we just use a Pythonic way to add tick boxes:

image

ChatGPT suggests some css wizardry to make the alignment better - do we think it is worth it?

cg/store/crud/update.py
```diff
@@ -106,3 +110,11 @@ def update_analysis_upload_started_at(
         analysis = self.get_analysis_by_entry_id(analysis_id)
         analysis.upload_started_at = upload_started_at
         self.session.commit()
+
+    def update_order_type_applications(self, application: Application, form: Form):
+        """Updates the order_type_application table based on the updated Application object"""
```
diitaz93, line 115:
```suggestion
        """Updates the order_type_application table based on the updated Application."""
```

cg/store/crud/update.py
```diff
@@ -106,3 +110,11 @@ def update_analysis_upload_started_at(
...
+        self._get_query(OrderTypeApplication).filter_by(application_id=application.id).delete()
+        selected_order_types: list[OrderType] = form["suitable_order_types"].data
+        for order_type in selected_order_types:
```
diitaz93, line 118: and then you can call `store.link_order_types_to_application` (not yet merged)

cg/store/crud/update.py
```diff
@@ -106,3 +109,12 @@ def update_analysis_upload_started_at(
         analysis = self.get_analysis_by_entry_id(analysis_id)
         analysis.upload_started_at = upload_started_at
         self.session.commit()
+
+    def update_order_type_applications(
+        self, application: Application, order_types: list[OrderType]
+    ):
+        """Updates the order_type_application table based on the updated Application object"""
+        self._get_query(OrderTypeApplication).filter_by(application_id=application.id).delete()
```
ChrOertlin, line 117: are we not deleting stuff here? If so, that should be handled in the DeleteHandler.

islean: Yes, I have put way too much thought into where to put stuff. I think nothing should be in the update handler. I am refactoring

cg/server/admin.py
```diff
@@ -168,6 +191,27 @@ def view_application_link(unused1, unused2, model, unused3):
...
 
+    def on_model_change(self, form: Form, model: Application, is_created: bool):
+        """Override to persist entries to the OrderTypeApplication table"""
+        super(ApplicationView, self).on_model_change(form=form, model=model, is_created=is_created)
+        order_types: list[OrderType] = form["suitable_order_types"].data
+        db.delete_order_type_applications_by_application_id(model.id)
+        db.session.add_all(
+            db.link_order_types_to_application(application=model, order_types=order_types)
+        )
+        db.session.commit()
```
islean, line 202: I would say that we either keep these here or we move them to the Application service introduced by @diitaz93 in #3864.

cg/server/admin.py
```diff
@@ -143,14 +160,20 @@ class ApplicationView(BaseView):
...
+        "order_types": view_order_types,
         "sample_concentration_minimum": view_sample_concentration_minimum,
         "sample_concentration_maximum": view_sample_concentration_maximum,
         "sample_concentration_minimum_cfdna": view_sample_concentration_minimum_cfdna,
         "sample_concentration_maximum_cfdna": view_sample_concentration_maximum_cfdna,
     }
     column_filters = ["prep_category", "is_accredited"]
     column_searchable_list = ["tag", "prep_category"]
-    form_excluded_columns = ["category", "versions"]
+    form_excluded_columns = ["category", "versions", "order_types"]
```
islean, line 171: Apparently relationships are by default shown in the edit/create form in flask, meaning that we need to exclude the default order types

islean: And instead add them via tick boxes in the form_extra_fields

cg/server/admin.py
```diff
@@ -113,6 +128,8 @@ def view_sample_concentration_maximum_cfdna(unused1, unused2, model, unused3):
 class ApplicationView(BaseView):
     """Admin view for Model.Application"""
 
+    column_list = [column_name for column_name in inspect(Application).columns] + ["order_types"]
```
Vince-janv, line 131: neat

ChrOertlin (approved):
Looks good, the admin view is a wild west. We should move the database logic out of this layer though.

cg/server/admin.py
```diff
@@ -168,6 +191,27 @@ def view_application_link(unused1, unused2, model, unused3):
...
+        """Override to persist entries to the OrderTypeApplication table"""
+        super(ApplicationView, self).on_model_change(form=form, model=model, is_created=is_created)
+        order_types: list[OrderType] = form["suitable_order_types"].data
+        db.delete_order_type_applications_by_application_id(model.id)
+        db.session.add_all(
+            db.link_order_types_to_application(application=model, order_types=order_types)
+        )
+        db.session.commit()
+
+    def edit_form(self, obj=None):
```
Vince-janv, line 204: Can we get a doc string for this one?

Vince-janv (approved):
Nice 👍 I don't think we need to spend way too much time polishing admin.py.

cg/server/admin.py
diitaz93: Maybe wrap the db operations inside a try-except to avoid pending transactions. 

There is another method in the CaseView (`set_action_for_cases`) which handles a session and commits. Not a fan either but I can live with that. There is also some delivered message in this function to the user, maybe something to consider??

diitaz93: And I think that there is no need to involve the ApplicationsWebService

diitaz93 (approved):
Good job! The design for this one was hard! I think this will work 💪 Maybe tray to catch db exceptions before comitting

cg/server/admin.py
```diff
@@ -168,6 +191,27 @@ def view_application_link(unused1, unused2, model, unused3):
             else ""
         )
 
+    def on_model_change(self, form: Form, model: Application, is_created: bool):
+        """Override to persist entries to the OrderTypeApplication table"""
```
diitaz93, line 195:
```suggestion
        """Override to persist entries to the OrderTypeApplication table."""
```

cg/server/admin.py
islean: Getting conflicting advice here. Unless @Vince-janv is also strongly against, I will move it to the application service. I think there is a bigger issue with having the database persisting in the admin views which are already complicated enough than the scope of the service changing

Vince-janv: I'm for moving 🤚

cg/store/crud/update.py
islean: done

cg/store/crud/update.py
islean: Moved it to the ApplicationService

PR #3860, opened 2024-10-17
fix(raise error when sample not found)

ChrOertlin:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/store/crud/create.py
+++ b/cg/store/crud/create.py
@@ -462,6 +462,7 @@ class CreateHandler(BaseHandler):
         """
         sample: Sample = self.get_sample_by_internal_id(metrics_dto.sample_id)
         if not sample:
+            self.session.rollback()
             raise EntryNotFoundError(f"Sample not found: {metrics_dto.sample_id}")
         new_metric = IlluminaSampleSequencingMetrics(
             sample=sample,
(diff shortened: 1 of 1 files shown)


cg/store/crud/create.py
```diff
@@ -541,6 +541,9 @@ def create_pac_bio_sample_sequencing_run(
         sample_id: str = sample_run_metrics_dto.sample_internal_id
         LOG.debug(f"Creating Pacbio sample sequencing metric for sample {sample_id}")
         sample: Sample = self.get_sample_by_internal_id(sample_id)
+        if not sample:
+            self.session.rollback()
+            raise EntryNotFoundError(f"Sample not found: {sample_id}")
```
islean, line 546: I get that we have not seen the error thus far but would it make sense to do the same thing with the sequencing run?

islean (approved):
💯

cg/store/crud/create.py
ChrOertlin: the sequencing run is an aggregate of the samples and is called donwstream of this. It will fail here first and not proceed further. 

I will leave this as is for now as the plan is to move the actual fail into the crud layer instead of layers using the crud. So then it will be checked and required in one place only. The current situation is not DRY and coupled.

PR #3864, opened 2024-10-18
add application ordertype endpoint (#3864)(minor)

diitaz93:
## Description
Closes https://github.com/Clinical-Genomics/add-new-tech/issues/130
Adds the endpoint requesting applications given the order types

### Added

- Endpoint function `get_application_order_types` in `cg/server/endpoints/applications.py`
- Error handler for endpoint in `cg/server/endpoints/error_handler.py`
- Aplication web service that returns the response to the endpoint in `cg/services/application/service.py`
- ApplicationResponse model
- Join application order query function `_get_join_application_ordertype_query` in `cg/store/base.py`
- CRUD function `link_order_types_to_application`. to create `OrderTypeApplication` entries (create)
- CRUF function `get_active_applications_by_order_type` (read)
- Status filter module `cg/store/filters/status_ordertype_application_filters.py` for OrderTypeApplication table and one filter function
- Attribute `order_types` to the `Appliacation` model in the database
- Attribute `application` to the `OrderTypeApplication` model in the database
- Tests for the filter function
- Test for the CRUD (read) function
- Fixtures
- Store helpers function

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] See below

### Expected test outcome

- [ ] Check that fetching an application with an existing order type returns an application
- [ ] Chack that fetching an application with a non-existing order type raises an informative error

## Review

- [ ] Tests executed by SD, IO
- [ ] "Merge and deploy" approved by VJ, IO, CO

## Implementation Plan

- [ ] Deployed to stage:
```shell

```
- [ ] Deployed to production:
```shell

```

### DIFF
--- a/cg/server/endpoints/applications.py
+++ b/cg/server/endpoints/applications.py
@@ -20,7 +20,7 @@ def get_applications():
         message: str = "No applications found for the given order type"
         return abort(make_response(jsonify(message=message), HTTPStatus.NOT_FOUND))
     app_tags: list[str] = [application.tag for application in applications]
-    return jsonify(applications=parsed_applications)
+    return jsonify(applications=app_tags)
 
 
 @APPLICATIONS_BLUEPRINT.route("/applications/<order_type>")
--- a/cg/server/endpoints/applications.py
+++ b/cg/server/endpoints/applications.py
@@ -5,10 +5,7 @@ from flask import Blueprint, abort, jsonify, make_response
 
 from cg.server.endpoints.utils import before_request, is_public
 from cg.server.ext import db
-from cg.store.models import (
-    Application,
-    ApplicationLimitations,
-)
+from cg.store.models import Application, ApplicationLimitations
 
 APPLICATIONS_BLUEPRINT = Blueprint("applications", __name__, url_prefix="/api/v1")
 APPLICATIONS_BLUEPRINT.before_request(before_request)
@@ -19,6 +16,14 @@ APPLICATIONS_BLUEPRINT.before_request(before_request)
 def get_applications():
     """Return application tags."""
     applications: list[Application] = db.get_applications_is_not_archived()
+    parsed_applications: list[dict] = [application.to_dict() for application in applications]
+    return jsonify(applications=parsed_applications)
+
+
+@APPLICATIONS_BLUEPRINT.route("/applications/<order_type>")
+def get_application_order_types(order_type: str):
+    """Return application order types.."""
+    applications: list[Application] = db.get_applications_by_ordertype(order_type)
     parsed_applications: list[dict] = [application.to_dict() for application in applications]
     return jsonify(applications=parsed_applications)
 
--- a/cg/server/endpoints/applications.py
+++ b/cg/server/endpoints/applications.py
@@ -3,7 +3,8 @@ from typing import Any
 
 from flask import Blueprint, abort, jsonify, make_response
 
-from cg.server.endpoints.error_handler import handle_endpoint_errors
+from cg.models.orders.constants import OrderType
+from cg.server.endpoints.error_handler import handle_missing_entries
 from cg.server.endpoints.utils import before_request, is_public
 from cg.server.ext import applications_service, db
 from cg.services.application.models import ApplicationResponse
@@ -23,8 +24,8 @@ def get_applications():
 
 
 @APPLICATIONS_BLUEPRINT.route("/applications/<order_type>")
-@handle_endpoint_errors
-def get_application_order_types(order_type: str):
+@handle_missing_entries
+def get_application_order_types(order_type: OrderType):
     """Return application order types."""
     applications: ApplicationResponse = applications_service.get_valid_applications(order_type)
     return jsonify(applications.model_dump())
--- a/cg/server/endpoints/applications.py
+++ b/cg/server/endpoints/applications.py
@@ -16,19 +16,16 @@ APPLICATIONS_BLUEPRINT.before_request(before_request)
 def get_applications():
     """Return application tags."""
     applications: list[Application] = db.get_applications_is_not_archived()
-    if not applications:
-        message: str = "No applications found for the given order type"
-        return abort(make_response(jsonify(message=message), HTTPStatus.NOT_FOUND))
-    app_tags: list[str] = [application.tag for application in applications]
-    return jsonify(applications=app_tags)
+    parsed_applications: list[dict] = [application.to_dict() for application in applications]
+    return jsonify(applications=parsed_applications)
 
 
 @APPLICATIONS_BLUEPRINT.route("/applications/<order_type>")
 def get_application_order_types(order_type: str):
     """Return application order types.."""
     applications: list[Application] = db.get_active_applications_by_order_type(order_type)
-    parsed_applications: list[dict] = [application.to_dict() for application in applications]
-    return jsonify(applications=parsed_applications)
+    app_tags: list[str] = [application.tag for application in applications]
+    return jsonify(applications=app_tags)
 
 
 @APPLICATIONS_BLUEPRINT.route("/applications/<tag>")
--- a/cg/server/endpoints/applications.py
+++ b/cg/server/endpoints/applications.py
@@ -16,14 +16,17 @@ APPLICATIONS_BLUEPRINT.before_request(before_request)
 def get_applications():
     """Return application tags."""
     applications: list[Application] = db.get_applications_is_not_archived()
-    parsed_applications: list[dict] = [application.to_dict() for application in applications]
+    if not applications:
+        message: str = "No applications found for the given order type"
+        return abort(make_response(jsonify(message=message), HTTPStatus.NOT_FOUND))
+    app_tags: list[str] = [application.tag for application in applications]
     return jsonify(applications=parsed_applications)
 
 
 @APPLICATIONS_BLUEPRINT.route("/applications/<order_type>")
 def get_application_order_types(order_type: str):
     """Return application order types.."""
-    applications: list[Application] = db.get_applications_by_ordertype(order_type)
+    applications: list[Application] = db.get_active_applications_by_order_type(order_type)
     parsed_applications: list[dict] = [application.to_dict() for application in applications]
     return jsonify(applications=parsed_applications)
 
--- a/cg/server/endpoints/applications.py
+++ b/cg/server/endpoints/applications.py
@@ -25,7 +25,7 @@ def get_applications():
 @APPLICATIONS_BLUEPRINT.route("/applications/<order_type>")
 @handle_endpoint_errors
 def get_application_order_types(order_type: str):
-    """Return application order types.."""
+    """Return application order types."""
     applications: ApplicationResponse = applications_service.get_applications_by_order_type(
         order_type
     )
--- a/cg/server/endpoints/applications.py
+++ b/cg/server/endpoints/applications.py
@@ -26,9 +26,7 @@ def get_applications():
 @handle_endpoint_errors
 def get_application_order_types(order_type: str):
     """Return application order types."""
-    applications: ApplicationResponse = applications_service.get_applications_by_order_type(
-        order_type
-    )
+    applications: ApplicationResponse = applications_service.get_valid_applications(order_type)
     return jsonify(applications.model_dump())
 
 
--- a/cg/server/endpoints/applications.py
+++ b/cg/server/endpoints/applications.py
@@ -3,8 +3,10 @@ from typing import Any
 
 from flask import Blueprint, abort, jsonify, make_response
 
+from cg.server.endpoints.error_handler import handle_endpoint_errors
 from cg.server.endpoints.utils import before_request, is_public
-from cg.server.ext import db
+from cg.server.ext import applications_service, db
+from cg.services.application.models import ApplicationResponse
 from cg.store.models import Application, ApplicationLimitations
 
 APPLICATIONS_BLUEPRINT = Blueprint("applications", __name__, url_prefix="/api/v1")
@@ -21,14 +23,13 @@ def get_applications():
 
 
 @APPLICATIONS_BLUEPRINT.route("/applications/<order_type>")
+@handle_endpoint_errors
 def get_application_order_types(order_type: str):
     """Return application order types.."""
-    applications: list[Application] = db.get_active_applications_by_order_type(order_type)
-    if not applications:
-        message: str = "No applications found for the given order type"
-        return abort(make_response(jsonify(message=message), HTTPStatus.NOT_FOUND))
-    app_tags: list[str] = [application.tag for application in applications]
-    return jsonify(applications=app_tags)
+    applications: ApplicationResponse = applications_service.get_applications_by_order_type(
+        order_type
+    )
+    return jsonify(applications.model_dump())
 
 
 @APPLICATIONS_BLUEPRINT.route("/applications/<tag>")
--- a/cg/server/endpoints/applications.py
+++ b/cg/server/endpoints/applications.py
@@ -24,6 +24,9 @@ def get_applications():
 def get_application_order_types(order_type: str):
     """Return application order types.."""
     applications: list[Application] = db.get_active_applications_by_order_type(order_type)
+    if not applications:
+        message: str = "No applications found for the given order type"
+        return abort(make_response(jsonify(message=message), HTTPStatus.NOT_FOUND))
     app_tags: list[str] = [application.tag for application in applications]
     return jsonify(applications=app_tags)
 
--- a/cg/server/endpoints/error_handler.py
+++ b/cg/server/endpoints/error_handler.py
@@ -1,0 +1,27 @@
+import logging
+from functools import wraps
+from http import HTTPStatus
+
+from flask import jsonify
+
+from cg.store.exc import EntryNotFoundError
+
+LOG = logging.getLogger(__name__)
+
+
+def handle_endpoint_errors(func):
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except EntryNotFoundError as error:
+            LOG.error(error)
+            return jsonify(error=str(error)), HTTPStatus.NOT_FOUND
+        except Exception as error:
+            LOG.error(f"Unexpected error in endpoint: {error}")
+            return (
+                jsonify(error="An error occurred while processing your request."),
+                HTTPStatus.INTERNAL_SERVER_ERROR,
+            )
+
+    return wrapper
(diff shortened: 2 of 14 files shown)


cg/store/models.py
```diff
@@ -181,6 +181,7 @@ class Application(Base):
     pipeline_limitations: Mapped[list["ApplicationLimitations"]] = orm.relationship(
         back_populates="application"
     )
+    order_types = orm.relationship("OrderTypeApplication", back_populates="application")
```
islean, line 184:
```suggestion
    order_types: Mapped[list["OrderTypeApplication"]] = orm.relationship("OrderTypeApplication", back_populates="application")
```

cg/store/crud/read.py
```diff
@@ -838,6 +822,18 @@ def get_applications(self) -> list[Application]:
...
+        """Return all possible not archived applications for an order type."""
+        non_archived_applications: Query = apply_application_filter(
+            applications=self._get_join_application_ordertype_query(),
+            filter_functions=[ApplicationFilter.IS_NOT_ARCHIVED],
+        )
+        return apply_ordertype_application_filter(
+            ordertype_applications=non_archived_applications,
+            filter_functions=[OrderTypeApplicationFilter.BY_ORDER_TYPE],
+            order_type=order_type,
+        ).all()
```
ChrOertlin, line 835: I think we should start raising errors when something cannot be returned that is expected to have entries. In this case it would be a missing link between an order type and application. Which should be checked for as we shouldn't support an ordertype that has no application yet as the application is inherently tied to invoicing.

cg/server/endpoints/applications.py
```diff
@@ -23,6 +20,14 @@ def get_applications():
     return jsonify(applications=parsed_applications)
 
 
+@APPLICATIONS_BLUEPRINT.route("/applications/<order_type>")
+def get_application_order_types(order_type: str):
+    """Return application order types.."""
+    applications: list[Application] = db.get_applications_by_ordertype(order_type)
+    parsed_applications: list[dict] = [application.to_dict() for application in applications]
```
islean, line 27:
```suggestion
    app_tags: list[str] = [application.tag for application in applications]
```
Do we need to return more data? In the current response, only the app tags are returned

cg/server/endpoints/applications.py
```diff
@@ -23,6 +20,14 @@ def get_applications():
...
+    return jsonify(applications=parsed_applications)
```
ChrOertlin, line 28: add error handling for when no application tags can be found for the ordertype

cg/store/crud/read.py
```diff
@@ -838,6 +822,18 @@ def get_applications(self) -> list[Application]:
             .all()
         )
 
+    def get_applications_by_ordertype(self, order_type: str) -> list[Application]:
```
islean, line 825: We could consider a naming which makes it clearer that they are not archived as well, something like `get_active_applications_by_order_type`. Not sure what is better

islean, line 825: We might want to split ordertype into order type though

cg/store/filters/status_ordertype_application_filters.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def filter_applications_by_order_type(
+    ordertype_applications: Query, order_type: str, **kwargs
+) -> Query:
+    """Return application by order type."""
+    return ordertype_applications.filter(OrderTypeApplication.order_type.is_(order_type))
+
+
+def apply_ordertype_application_filter(
```
islean, line 15:
```suggestion
def apply_order_type_application_filter(
```

cg/store/filters/status_ordertype_application_filters.py
```diff
@@ -0,0 +1,33 @@
...
+    filter_functions: list[callable],
+    ordertype_applications: Query,
+    order_type: str = None,
```
islean, line 18:
```suggestion
    order_type: OrderType = None,
```
There is an enum which is suitable for this

cg/store/filters/status_ordertype_application_filters.py
```diff
@@ -0,0 +1,33 @@
+from enum import Enum
+
+from sqlalchemy.orm import Query
+
+from cg.store.models import OrderTypeApplication
+
+
+def filter_applications_by_order_type(
+    ordertype_applications: Query, order_type: str, **kwargs
```
islean, line 9:
```suggestion
    ordertype_applications: Query, order_type: OrderType, **kwargs
```

islean (approved):
Looks good! I think we can return a bit less data though

cg/server/endpoints/applications.py
islean: Do we consider this relevant even if we change the order_type parameter to an enum? To me, being able to parse the order type to the enum fulfills a similar purpose

ChrOertlin: Changing order type to an enum would make the check wheter the order type is supported at all. That it is supported by the back end does not immediatly mean that there are also app tags for it in the database. One could argue that there always should be. However, it might get missed with the next implementation which would lead to non sense errors in the front end.

islean: If we have already checked the input and deemed it okay (as one can interpret being able to parse it into an enum) I think it could be considered normal and simply the case that there are no applications for this order type configured yet.

islean: I am fine with either - maybe it is just a matter of interpretation whether we consider "no applications found" an error or not. I do think that the type hint should reuse the OrderType enum though

ChrOertlin: Yes what would happen when you get an ordertype for which the applications are not yet configured? The endpoint would return nothing for a field that is required and thus cannot be filled. Right?

islean: Would the response not just look like

```Json
{
    "applications": []
}
```
which is sort of fine by me - it would result in a drop down with no entries in the front end. Kind of correct but maybe confusing?

cg/store/crud/read.py
diitaz93: fixed

cg/store/filters/status_ordertype_application_filters.py
diitaz93: fixed

cg/store/filters/status_ordertype_application_filters.py
diitaz93: fixed

cg/store/filters/status_ordertype_application_filters.py
diitaz93: fixed

cg/server/endpoints/applications.py
diitaz93: fixed

cg/store/crud/read.py
```diff
@@ -838,6 +822,18 @@ def get_applications(self) -> list[Application]:
             .all()
         )
 
+    def get_active_applications_by_order_type(self, order_type: str) -> list[Application]:
+        """Return all possible not archived applications for an order type."""
+        non_archived_applications: Query = apply_application_filter(
+            applications=self._get_join_application_ordertype_query(),
+            filter_functions=[ApplicationFilter.IS_NOT_ARCHIVED],
+        )
+        return apply_order_type_application_filter(
```
diitaz93, line 831: add error raising

cg/server/endpoints/applications.py
ChrOertlin: Yes, as a user getting a drop down menu that is empty without indication that something is wrong is not a great experience in my opinion.

cg/services/application/service.py
```diff
@@ -0,0 +1,20 @@
...
+def create_application_response(app_tags: list[str]) -> ApplicationResponse:
+    return ApplicationResponse(applications=app_tags)
+
+
+class ApplicationsService:
+
+    def __init__(self, store: Store):
+        self.store = store
+
+    def get_applications_by_order_type(self, order_type: str) -> ApplicationResponse:
```
diitaz93, line 15: ??
```suggestion
    def request_applications_by_order_type(self, order_type: str) -> ApplicationResponse:
```

cg/server/endpoints/applications.py
diitaz93: Will resolve this in favour of a new discussion

cg/store/crud/read.py
diitaz93: fixed

cg/services/application/service.py
Vince-janv:
```suggestion
    def get_valid_applications(self, order_type: str) -> ApplicationResponse:
```

cg/server/endpoints/error_handler.py
```diff
@@ -0,0 +1,27 @@
...
+from http import HTTPStatus
+
+from flask import jsonify
+
+from cg.store.exc import EntryNotFoundError
+
+LOG = logging.getLogger(__name__)
+
+
+def handle_endpoint_errors(func):
```
Vince-janv, line 12: This naming feels very general, whereas it is written to handle errors in a specific endpoint I would rename it to reflect that

Vince-janv (approved):
Looks good 👍

cg/server/endpoints/error_handler.py
diitaz93: I made it to be used in any endpoint

cg/services/application/service.py
diitaz93: fixed

cg/server/endpoints/error_handler.py
Vince-janv: But that assumes the same error raising in all endpoints it is used on right?

ChrOertlin: I think this makes sense for end points

islean: I do not see it being an endpoint as being a strong enough argument. While converting errors into a corresponding HTTP status code and error message is only wished for in endpoints it is not sufficient reasoning given that one endpoint could need completely different error handling despite the same error raised.

That being said, I think you could convince me to go either way with this. But if you go with having a vague naming and intending it to be reused for all endpoints, we should beware its limitations. Especially I think we should be ready to move away from the pattern if we run into a situation where the same exception needs to be handled in different ways for different endpoints. Writing this down, I am struggling to find good reasoning for aiming for a general error handler - what are the benefits?

tests/store/filters/test_status_order_type_application_filters.py
```diff
@@ -0,0 +1,36 @@
...
+
+    # WHEN filtering applications by order type
+    order_type_applications: Query = applications_store._get_query(table=OrderTypeApplication)
+    filtered_order_type_applications: Query = filter_applications_by_order_type(
+        order_type_applications=order_type_applications, order_type=order_type
+    )
+
+    # THEN assert that only the applications with the given order type are returned
+    assert filtered_order_type_applications.count() == 1
+    assert filtered_order_type_applications.first().order_type == order_type
```
islean, line 36: I am confused here - how does this test not fail? Is the left hand side not an OrderTypeApplication entry and the left hand is an OrderType enum?

cg/services/application/service.py
```diff
@@ -0,0 +1,20 @@
...
+def create_application_response(app_tags: list[str]) -> ApplicationResponse:
+    return ApplicationResponse(applications=app_tags)
+
+
+class ApplicationsService:
+
+    def __init__(self, store: Store):
+        self.store = store
+
+    def get_valid_applications(self, order_type: str) -> ApplicationResponse:
```
islean, line 15:
```suggestion
    def get_valid_applications(self, order_type: OrderType) -> ApplicationResponse:
```

cg/store/crud/read.py
```diff
@@ -838,6 +822,25 @@ def get_applications(self) -> list[Application]:
             .all()
         )
 
+    def get_active_applications_by_order_type(self, order_type: str) -> list[Application]:
```
islean, line 825:
```suggestion
    def get_active_applications_by_order_type(self, order_type: OrderType) -> list[Application]:
```

tests/store/crud/read/test_read_application.py
```diff
@@ -0,0 +1,23 @@
+import pytest
+
+from cg.models.orders.constants import OrderType
+from cg.store.exc import EntryNotFoundError
+from cg.store.models import Application
+from cg.store.store import Store
+
+
+def test_get_active_applications_by_order_type_no_application(store: Store):
```
islean, line 9: Might be nice with a test for where there is an application to return

islean (approved):
Looks great! 🥳  I think we should be forceful in expecting an order type which is one of the choices in our enum.

cg/server/endpoints/applications.py
```diff
@@ -23,6 +22,14 @@ def get_applications():
     return jsonify(applications=parsed_applications)
 
 
+@APPLICATIONS_BLUEPRINT.route("/applications/<order_type>")
+@handle_endpoint_errors
+def get_application_order_types(order_type: str):
```
islean, line 27: Can this also be type hinted as OrderType already?

diitaz93:
## Tests on stage
```shell
$ 'https://cg-statusdb-stage.scilifelab.se/api/v1/applications/mip-dna'
{"applications":["MyTestTag"]}
$ 'https://cg-statusdb-stage.scilifelab.se/api/v1/applications/tomte'
{"error":"No applications found for order type tomte"}
```
LOGS
```shell
Oct 22 10:20:07 cg-services-stage.scilifelab.se podman[3073726]: [2024-10-22 10:20:07 +0200] [8] [DEBUG] Closing connection.
/api/v1/applications/mip-dna HTTP.....
Oct 22 10:20:07 cg-services-stage.scilifelab.se podman[3073726]: [2024-10-22 10:20:07 +0200] [8] [DEBUG] GET /api/v1/applications/mip-dna
Oct 22 10:19:51 cg-services-stage.scilifelab.se systemd[1524]: Started deploy service cg.
Oct 22 10:19:51 cg-services-stage.scilifelab.se systemd[1524]: Started deploy service success report.

Oct 22 10:28:08 cg-services-stage.scilifelab.se podman[3073726]: [2024-10-22 10:28:08 +0200] [11] [DEBUG] Closing connection.
/api/v1/applications/tomte HTTP...
Oct 22 10:28:08 cg-services-stage.scilifelab.se podman[3073726]: cg.server.endpoints.error_handler[11] ERROR No applications found for order type tomte
Oct 22 10:28:08 cg-services-stage.scilifelab.se podman[3073726]: [2024-10-22 10:28:08 +0200] [11] [DEBUG] GET /api/v1/applications/tomte
```

cg/server/endpoints/error_handler.py
diitaz93: I will rename to `handle_missing_entries`

cg/server/endpoints/applications.py
diitaz93: wish granted

tests/store/crud/read/test_read_application.py
diitaz93: I had one, I realised it is exactly the same test as `test_filter_applications_by_order_type` so I removed it

islean: fair

cg/server/endpoints/error_handler.py
islean: I discussed a bit with chatgpt and seems as if the most common error handling could be in a common decorator and you could have separate endpoint specific ones if you need anything more special. So I think there are more than two options here.

My only point is that it is not clear to me that we will always want the same HTTP status returned for the same exception being caught. In this case it might make sense to have a 404 not found for the NoEntryFound error (or a 403 bad request or something else completely) but in another scenario the NoEntryFound might signal something different, or be something expected. 

Unless we are absolutely certain that all caught NoEntryFound errors should result in a 404 error code (which I am not), it just seems more adaptable for the future to not have it in a super general handler.

ChrOertlin: > My only point is that it is not clear to me that we will always want the same HTTP status returned for the same exception being caught. In this case it might make sense to have a 404 not found for the NoEntryFound error (or a 403 bad request or something else completely) but in another scenario the NoEntryFound might signal something different, or be something expected.

My argument is, the logic or decisions to raise this error or what this error means should not be the responsibility of the end point.

I think the responsibility of the end point should be:
get a request
send a response
The response can be either data or the error message.

If for some reason a NoEntryFound error is expected given the request, then the layer below the endpoint should deal with this, not the end point.

islean: And where would you want the mapping between the exceptions and the corresponding status code to take place? Because this seems like a rebuttal of the need for this decorator altogether to me and that all this logic should take place in the service but then we are tying the service to only being used for a web service

ChrOertlin: The mapping would be in the decorator.

The logic to do something different in instances where an EntryNotFound error should be re-interpreted  in the service

I actually like the idea of the ApplicationService becoming an ApplicationWebService - it reduces its scope and responsibility to only that, which I think is a good thing.

cg/services/application/service.py
```diff
@@ -0,0 +1,21 @@
...
+from cg.services.application.models import ApplicationResponse
+from cg.store.models import Application
+from cg.store.store import Store
+
+
+def create_application_response(app_tags: list[str]) -> ApplicationResponse:
+    return ApplicationResponse(applications=app_tags)
+
+
+class ApplicationsService:
```
ChrOertlin, line 11: In light of the discussion @islean and I had in this PR (see the other comments). the aim is to reduce the scope of this service to only WebAPI interactions. A narrow scope means less code to write and less things to consider/ test. 

```suggestion
class ApplicationsWebService:
```

ChrOertlin (approved):
Nice work!

cg/services/application/service.py
diitaz93: I totally agree

cg/server/endpoints/error_handler.py
diitaz93: I will resolve this and we can continue the discussion about error handlers somewhere else

islean: I agree much more with you than I think my comments convey - in general it is probably bad design if each endpoint has a different mapping for the same exception, say if the error type is too broad. And I am completely fine with this being a general error handler for the time being. I just think that there's legitimacy to questioning if we can use the same decorator for all methods in our API. Couple that with my failure to see any drawbacks with the alternative - to introduce endpoint specific handling in instances where it is suitable - I am just worried that we are looking at this from the wrong perspective. In my mind it is very much a situation where there are arguments for both sides, and the most problematic outcome would be if start writing code where good alternatives have been branded as something to be avoided. Definitely a discussion out of scope for this PR though.


Regarding doing a more narrow service, I could agree or disagree with you depending on what you want to include in it. The narrow scope I am positive of, but I would not want to see HTTP code conversions in it, since it would create an artificial hindrance for its functionality to be useful in other parts of the code.

PR #3883, opened 2024-10-24
Fix(microFASTQ) Set internal_id when ordering

Vince-janv:
## Description
A bug described in https://github.com/Clinical-Genomics/bug-reports/issues/11 causes samples created from microbial-fastq orders to have incorrect internal_ids

### Fixed
- internal_id is now set to the LIMS-ID when submitting an order

### DIFF
--- a/cg/services/orders/store_order_services/store_microbial_fastq_order_service.py
+++ b/cg/services/orders/store_order_services/store_microbial_fastq_order_service.py
@@ -40,6 +40,7 @@ class StoreMicrobialFastqOrderService(StoreOrderService):
                 {
                     "application": sample.application,
                     "comment": sample.comment,
+                    "internal_id": sample.internal_id,
                     "data_analysis": sample.data_analysis,
                     "data_delivery": sample.data_delivery,
                     "name": sample.name,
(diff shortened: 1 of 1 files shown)


Vince-janv:
Local tests successfull

diitaz93 (approved):
👍

diitaz93:
@Vince-janv The samples dictionary generated in `cg/services/orders/store_order_services/store_microbial_fastq_order_service.py:order_to_status` doe snot seem to have the field `internal_id`

cg/services/orders/store_order_services/store_microbial_fastq_order_service.py
```diff
@@ -113,6 +113,7 @@ def _create_sample(
             customer=customer,
             sex=SexOptions.UNKNOWN,
             comment=sample_dict["comment"],
+            internal_id=sample_dict["internal_id"],
```
diitaz93, line 116: To make the tests pass, but this may not solve the problem as it will return None
```suggestion
            internal_id=sample.get("internal_id"),
```

Vince-janv: Solved using the default None

Vince-janv:
Deployed to prod VM

PR #3833, opened 2024-10-10
Fix clashing ticket ids

islean:
## Description

This PR reworks filtering on ticket ids to go via the Order table instead of the tickets column found in the Case table. Closes #3702 

### Added

-

### Changed

- Filtering on ticket id is done in the Order table

### Fixed

-

### How to prepare for test

- [ ] Ssh to relevant server (depending on type of change)
- [ ] Use stage: `us`
- [ ] Paxa the environment: `paxa`
- [ ] Install on stage (example for Hasta):
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b fix-ticket-clashes -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...
- [ ] Take a screenshot and attach or copy/paste the output.

## Review

- [ ] Tests executed by
- [ ] "Merge and deploy" approved by
Thanks for filling in who performed the code review and the test!

### This [version](https://semver.org/) is a

- [ ] **MAJOR** - when you make incompatible API changes
- [ ] **MINOR** - when you add functionality in a backwards compatible manner
- [ ] **PATCH** - when you make backwards compatible bug fixes or documentation/instructions

## Implementation Plan

- [ ] Document in ...
- [ ] Deploy this branch on ...
- [ ] Inform to ...

PR #3884, opened 2024-10-24
Rework order storing

islean:
## Description

When I reworked the filter cases on ticket id in #3833, we started being unable to connect an order to its cases. This PR ensures that the order storing is not dependent on the ticket being in the case table

### Added

-

### Changed

-

### Fixed

- Orders are connected to their cases

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/orders/store_order_services/store_fastq_order_service.py
+++ b/cg/services/orders/store_order_services/store_fastq_order_service.py
@@ -1,14 +1,21 @@
 import logging
 from datetime import datetime
 
-from cg.constants import Workflow, DataDelivery, GenePanelMasterList, Priority
+from cg.constants import DataDelivery, GenePanelMasterList, Priority, Workflow
 from cg.constants.constants import CustomerId, PrepCategory
 from cg.exc import OrderError
 from cg.models.orders.order import OrderIn
 from cg.models.orders.sample_base import StatusEnum
 from cg.services.orders.order_lims_service.order_lims_service import OrderLimsService
 from cg.services.orders.submitters.order_submitter import StoreOrderService
-from cg.store.models import Case, CaseSample, Sample, Customer, ApplicationVersion
+from cg.store.models import (
+    ApplicationVersion,
+    Case,
+    CaseSample,
+    Customer,
+    Order,
+    Sample,
+)
 from cg.store.store import Store
 
 LOG = logging.getLogger(__name__)
@@ -61,7 +68,7 @@ class StoreFastqOrderService(StoreOrderService):
         }
         return status_data
 
-    def create_maf_case(self, sample_obj: Sample) -> None:
+    def create_maf_case(self, sample_obj: Sample, order: Order) -> None:
         """Add a MAF case to the Status database."""
         case: Case = self.status_db.add_case(
             data_analysis=Workflow(Workflow.MIP_DNA),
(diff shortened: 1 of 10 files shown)


tests/services/orders/order_store_service/test_generic_order_store_service.py
```diff
@@ -194,106 +194,3 @@ def test_store_cancer_samples(
     assert new_link.sample.application_version.application.tag == "WGSPCFC030"
     assert new_link.sample.comment == "other Elution buffer"
     assert new_link.sample.is_tumour
-
-
-def test_store_existing_single_sample_from_trio(
```
islean, line 199: Both removed tests in this file started failing because they called the same storing method twice, resulting in a uniqueness constraint being broken in the Order table. Given that these methods should not be called more than once with the same ticket id I saw them fit to be removed. They can be reworked but it would be a bit annoying

cg/services/orders/store_order_services/store_pacbio_order_service.py
```diff
@@ -65,6 +65,12 @@ def _store_samples_in_statusdb(
         customer: Customer = self.status_db.get_customer_by_internal_id(
             customer_internal_id=customer_id
         )
+        status_db_order = Order(
+            customer=customer,
+            order_date=datetime.now(),
+            ticket_id=int(ticket_id),
+            workflow=Workflow(samples[0]["data_analysis"]),
```
diitaz93, line 72: I just add this as in the microbial fastq is this way, but disregard if irrelevant
```suggestion
            workflow=Workflow.RAW_DATA,
```

cg/services/orders/store_order_services/store_fastq_order_service.py
```diff
@@ -61,7 +68,7 @@ def order_to_status(order: OrderIn) -> dict:
         }
         return status_data
 
-    def create_maf_case(self, sample_obj: Sample) -> None:
+    def create_maf_case(self, sample_obj: Sample, order: Order) -> None:
```
islean, line 71: I guess the MAF case should get ties to the order?

islean:
Orders submitted successfully, please double check if the amount of cases seem off - the fastq order has a weird count due to MAF:

image

cg/services/orders/store_order_services/store_pacbio_order_service.py
islean: That is probably a better way. We really should rework the whole flow soon though

diitaz93 (approved):
Looks like a great improvement 💯 It makes so much sense to add the order in the submitter and not in a separate statement. I didn't spot any flaws but I guess that the tests have to be a bit rigorous. I understand the deletion of the tests and I think it is ok to merge without them, but I think it is important to replace them ASAP to not leave this part untested

islean:
Rerun okay as well (order with two old cases and one new case with old samples):

image

cg/services/orders/store_order_services/store_fastq_order_service.py
islean: yes

islean:
Deployed to cg-services-stage and cg-services-prod

PR #3902, opened 2024-10-29
Add mip rna messages

beatrizsavinhas:
## Description

Closes https://github.com/Clinical-Genomics/cigrid-ui/issues/575

### Added

- crud functions to retrieve dna cases related to an case.
- Logic for generation of delivery messages for mip-rna.

### How to prepare for test

- [X] Ssh to relevant server `cg-services-stage`
- [X] Paxa the environment: `paxa`
- [X] Install and test using curl commands

### How to test

- [X] Deploy in `cg-services-stage`
- [X] Generate the delivery message for a completed case, several completed cases from the same order and generate a delivery message for the order.

### Expected test outcome

- [X] Check that the content of the message (rna case name and corresponding dna scout link) is correct.

## Review

- [X] Tests executed by @beatrizsavinhas and @islean 

## Implementation Plan

- [ ] Deploy this branch on hasta and cg services prod
- [ ] Inform prodbioinfo


cg/services/delivery_message/messages/rna_analysis_scout_message.py
```diff
@@ -0,0 +1,59 @@
...
+)
+from cg.store.models import Case
+
+
+class RNAAnalysisScoutMessage(DeliveryMessage):
+    def __init__(self, store):
+        super().__init__()
+        self.store = store
+
+    def create_message(self, cases: list[Case]) -> str:
```
beatrizsavinhas, line 14: I added these `get_case_message` and `get_cases_message` as part of the class since they require the store (instead of passing it as an argument). Let me know if this should be changed!

cg/store/crud/read.py
```diff
@@ -1604,3 +1637,50 @@ def get_related_cases(
...
+
+    def get_uploaded_related_dna_cases(self, rna_case: Case) -> list[Case]:
+        """Returns all uploaded DNA cases related to the specified RNA case."""
+        dna_cases: set[Case] = self.get_related_dna_cases_from_rna_case(rna_case=rna_case)
+        uploaded_dna_cases: list[Case] = []
+        for dna_case in dna_cases:
+            if dna_case.is_uploaded:
+                uploaded_dna_cases.append(dna_case)
+            else:
+                LOG.warning(f"Related DNA case {dna_case.internal_id} has not been completed.")
```
beatrizsavinhas, line 1649: Are these logs no longer relevant here?

cg/services/delivery_message/messages/rna_analysis_scout_message.py
diitaz93: For me it is ok

cg/server/endpoints/cases.py
```diff
@@ -62,7 +68,7 @@ def get_cases_delivery_message():
             delivery_message_request
         )
         return jsonify(response.model_dump()), HTTPStatus.OK
-    except (CaseNotFoundError, OrderMismatchError) as error:
+    except (CaseNotFoundError, OrderMismatchError, CgDataError) as error:
```
diitaz93, line 71: What would be the consequence of excepting all here @islean ?

cg/services/delivery_message/delivery_message_service.py
```diff
@@ -54,5 +54,5 @@ def _get_validated_analyses(
     def _get_delivery_message(self, case_ids: set[str]) -> str:
         cases: list[Case] = self.store.get_cases_by_internal_ids(case_ids)
         validate_cases(cases=cases, case_ids=case_ids)
-        message: str = get_message(cases)
+        message: str = get_message(cases, self.store)
```
diitaz93, line 57: add named args

cg/server/endpoints/cases.py
islean: If we get a connection issue when querying the database or if something else completely unforeseen happens, we would return a bad request status. I think we should handle all the errors which we have control over. So if we know that we want a BAD REQUEST status if we can not find a case, then do so. If we can not write down more explicitly what goes wrong than a general exception, I think it should result in an internal server error. 

In general the 4XX HTTP status codes are reserved for errors done by the user, while 5XX codes are errors on our end.

cg/store/crud/read.py
```diff
@@ -1604,3 +1637,50 @@ def get_related_cases(
...
+                f"{rna_sample.subject_id}. Number of matches: {len(related_dna_samples)} "
+            )
+        dna_sample: Sample = related_dna_samples[0]
+
+        dna_cases: list[Case] = self.get_related_cases(
+            sample_internal_id=dna_sample.internal_id,
+            workflows=DNA_WORKFLOWS_WITH_SCOUT_UPLOAD,
+            collaborators=collaborators,
+        )
+        return dna_cases
```
ChrOertlin, line 1686:
```suggestion
    def get_uploaded_dna_cases_related_to_rna_case(self, rna_case_id: str, prep_categories: list[str], workflows: list[str]) -> list[Case]:
    """Get all uploaded DNA cases related to a given RNA case using a database query."""
    return (
        self.session.query(Case)
        .join(Sample, Sample.case_id == Case.id)
        .filter(
            Sample.prep_category.in_(prep_categories),
            Sample.subject_id.isnot(None),
            Case.workflow.in_(workflows),
            Case.is_uploaded.is_(True),
            Case.rna_case_id == rna_case_id
        )
        .all()
    )
```

ChrOertlin: We do not need to loop though cases or samples, we likely can get the answer that we want from the database with a direct query. 

Copying the same functions from the service into the crud does not solve the efficiency problem with extracting the data and looping of the database rows in python.

ChrOertlin (changes requested):

cg/store/crud/read.py
beatrizsavinhas: Changed this.

cg/services/delivery_message/utils.py
```diff
@@ -24,20 +24,24 @@
...
 
-def get_message(cases: list[Case]) -> str:
-    message_strategy: DeliveryMessage = get_message_strategy(cases[0])
+def get_message(cases: list[Case], store: Store) -> str:
+    message_strategy: DeliveryMessage = get_message_strategy(cases[0], store)
     return message_strategy.create_message(cases)
 
 
-def get_message_strategy(case: Case) -> DeliveryMessage:
+def get_message_strategy(case: Case, store: Store) -> DeliveryMessage:
```
beatrizsavinhas, line 35: An alternative to adding a store here and RNAMessage classes is adding the crud function as a property of `Case`.

cg/services/delivery_message/messages/rna_scout_message.py
```diff
@@ -0,0 +1,45 @@
...
+            return self.get_case_message(cases[0])
+        return self.get_cases_message(cases)
+
+    def get_case_message(self, case: Case) -> str:
+
+        related_uploaded_dna_cases: list[Case] = self.store.get_uploaded_related_dna_cases(
+            rna_case=case
+        )
+        scout_links: list[str] = [get_scout_link(case) for case in related_uploaded_dna_cases]
+        scout_links_row_separated: str = "\n".join(scout_links)
```
ChrOertlin, line 22: similar to above

cg/services/delivery_message/messages/rna_scout_message.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+
+    def get_cases_message(self, cases: list[Case]) -> str:
+        message: str = "Hello,\n\n"
+        for case in cases:
+            related_uploaded_dna_cases: list[Case] = self.store.get_uploaded_related_dna_cases(
+                rna_case=case
+            )
+            scout_links: list[str] = [get_scout_link(case) for case in related_uploaded_dna_cases]
+            scout_links_row_separated: str = "\n".join(scout_links)
```
ChrOertlin, line 37: similar to above

cg/services/delivery_message/utils.py
islean: Then I prefer passing the store here. Although perhaps not elegant, I think that has more to do with the way we link RNA and DNA cases than code design.

cg/services/delivery_message/utils.py
```diff
@@ -24,20 +24,24 @@
...
     if case.data_analysis == Workflow.MICROSALT:
         return get_microsalt_message_strategy(case)
 
     if case.data_analysis == Workflow.MUTANT:
         return CovidMessage()
 
+    if case.data_analysis == Workflow.MIP_RNA:
+        return get_rna_message_strategy_from_data_delivery(case, store)
```
diitaz93, line 43: named args

cg/store/crud/read.py
```diff
@@ -1608,27 +1614,47 @@ def get_related_samples(
...
+                cases=dna_samples_cases_query,
+                workflows=DNA_WORKFLOWS_WITH_SCOUT_UPLOAD,
+                customer_entry_ids=[customer.id for customer in collaborators],
+                filter_functions=[
+                    CaseFilter.BY_WORKFLOWS,
+                    CaseFilter.BY_CUSTOMER_ENTRY_IDS,
+                ],
+            )
+
+            dna_samples_cases_analysis: Query = dna_samples_cases_query.join(Analysis)
```
ChrOertlin, line 1649: This join be be done in line 1635 as well it

cg/services/delivery_message/utils.py
diitaz93: I like the store here

cg/store/crud/read.py
```diff
@@ -1608,27 +1614,47 @@ def get_related_samples(
...
+
+            dna_samples_cases_analysis: Query = apply_analysis_filter(
```
ChrOertlin, line 1651:
```suggestion
            uploaded_dna_samples: list[Case] = apply_analysis_filter(
```

cg/store/crud/read.py
```diff
@@ -1608,27 +1614,47 @@ def get_related_samples(
...
+                analyses=dna_samples_cases_analysis,
+                filter_functions=[AnalysisFilter.IS_UPLOADED],
+            )
```
ChrOertlin, line 1654:
```suggestion
            )with_entities(Case).all()
```

cg/store/crud/read.py
```diff
@@ -1608,27 +1614,47 @@ def get_related_samples(
...
+
+            dna_cases: list[Case] = dna_samples_cases_analysis.with_entities(Case).all()
```
ChrOertlin, line 1656:
```suggestion

```

cg/store/crud/read.py
```diff
@@ -1608,27 +1614,47 @@ def get_related_samples(
...
+
+            related_dna_cases.extend([case for case in dna_cases])
```
ChrOertlin, line 1658:
```suggestion
            related_dna_cases.extend([case for case in uploaded_dna_cases])
```

cg/store/crud/read.py
```diff
@@ -1608,27 +1614,47 @@ def get_related_samples(
...
+
+            related_dna_samples: Query = self._get_related_samples_query(
+                sample=rna_sample,
+                prep_categories=DNA_PREP_CATEGORIES,
+                collaborators=collaborators,
+            )
+
+            dna_samples_cases_query = related_dna_samples.join(
+                CaseSample, Case.id == CaseSample.case_id
+            ).join(Case, CaseSample.case_id == Case.id)
```
ChrOertlin, line 1637:
```suggestion
            ).join(Case, CaseSample.case_id == Case.id).join(Analysis)
```

cg/services/delivery_message/messages/rna_analysis_scout_message.py
```diff
@@ -0,0 +1,59 @@
...
+        if len(cases) == 1:
+            return self.get_case_message(cases[0])
+        return self.get_cases_message(cases)
+
+    def get_case_message(self, case: Case) -> str:
+
+        related_uploaded_dna_cases: list[Case] = self.store.get_uploaded_related_dna_cases(
+            rna_case=case
+        )
+        scout_links: list[str] = [get_scout_link(case) for case in related_uploaded_dna_cases]
```
ChrOertlin, line 24: similar here

cg/services/delivery_message/messages/rna_analysis_scout_message.py
```diff
@@ -0,0 +1,59 @@
...
+        )
+
+    def get_cases_message(self, cases: list[Case]) -> str:
+        message: str = "Hello,\n\n"
+        for case in cases:
+            related_uploaded_dna_cases: list[Case] = self.store.get_uploaded_related_dna_cases(
+                rna_case=case
+            )
+            scout_links: list[str] = [get_scout_link(case) for case in related_uploaded_dna_cases]
+            scout_links_row_separated: str = "\n".join(scout_links)
```
ChrOertlin, line 43: similar here

cg/services/delivery_message/utils.py
```diff
@@ -60,6 +64,31 @@ def get_message_strategy_from_data_delivery(case: Case) -> DeliveryMessage:
...
 
 
+def get_rna_message_strategy_from_data_delivery(case: Case, store: Store) -> DeliveryMessage:
+
+    if case.data_delivery in scout_data_delivery:
+        return rna_message_map[case.data_delivery](store)
+    return rna_message_map[case.data_delivery]()
+
+
+scout_data_delivery: list[DataDelivery] = [
```
diitaz93, line 74: It would be ice to define all these constant maps at the top of this file, just below the imports. This includes the map not added in this PR, just to be consistent. Also, lets name them with capital letters so it is clear that is a constant dictionary

cg/services/delivery_message/messages/rna_fastq_analysis_scout_message.py
```diff
@@ -0,0 +1,59 @@
...
+            return self.get_case_message(cases[0])
+        return self.get_cases_message(cases)
+
+    def get_case_message(self, case: Case) -> str:
+
+        related_uploaded_dna_cases: list[Case] = self.store.get_uploaded_related_dna_cases(
+            rna_case=case
+        )
+        scout_links: list[str] = [get_scout_link(case) for case in related_uploaded_dna_cases]
+        scout_links_row_separated: str = "\n".join(scout_links)
```
ChrOertlin, line 25: similar here

ChrOertlin (commented):
Looks good! We can reduce some logic in the query by combining some steps and extract some re-used logic into smaller functions. That helps with code readability. Instead of needing to understand the for loop to get scout links you can just read the function name "get_scout_links" at that point you then don't need to know  or understand how they are retrieved.

cg/services/delivery_message/utils.py
ChrOertlin: Having it on the case model in the database will introduce more business logic to the database layer. We will run into issues with separation of concerns. Also it adds more pressure on the case model in regards to testing- you always need to make sure that you get the right message strategy back from the case model. So it will be harder to test things where the model is involved.

We should strive to keep the database layer only to manage and persist data.

cg/store/crud/read.py
islean, line 1637:
```suggestion
            dna_samples_cases_query = related_dna_samples.join(Sample.links).join(CaseSample.case)
```
just a bit more concise. Up to you what you go with. I do second Chris' suggestion to do all the joins prior to any filtering though

cg/services/delivery_message/messages/rna_fastq_scout_message.py
```diff
@@ -0,0 +1,60 @@
+from cg.services.delivery_message.messages.delivery_message import DeliveryMessage
+from cg.services.delivery_message.messages.utils import (
+    get_caesar_delivery_path,
+    get_scout_link,
+)
+from cg.store.models import Case
+from cg.store.store import Store
```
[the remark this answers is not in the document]
islean: Should be type hinted in the init though 🙈

cg/services/delivery_message/messages/rna_fastq_analysis_scout_message.py
```diff
@@ -0,0 +1,59 @@
...
+
+        delivery_path: str = get_caesar_delivery_path(case)
+        return (
+            f"Hello,\n\n"
+            f"The analysis for case {case.name} has been uploaded to the corresponding DNA case(s) on Scout at:\n\n"
+            f"{scout_links_row_separated}\n\n"
+            f"The analysis files are currently being uploaded to your inbox on Caesar:\n\n"
```
diitaz93, line 32: should this include fastq files as well?

cg/services/delivery_message/messages/rna_fastq_scout_message.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+        scout_links: list[str] = [get_scout_link(case) for case in related_uploaded_dna_cases]
+        scout_links_row_separated: str = "\n".join(scout_links)
+
+        delivery_path: str = get_caesar_delivery_path(case)
+        return (
+            f"Hello,\n\n"
+            f"The analysis for case {case.name} has been uploaded to the corresponding DNA case(s) on Scout at:\n\n"
+            f"{scout_links_row_separated}\n\n"
+            f"The analysis files are currently being uploaded to your inbox on Caesar:\n\n"
```
diitaz93, line 33: should this be only fastq files instead of analysis files?

cg/services/delivery_message/utils.py
beatrizsavinhas: It stays here then!

cg/services/delivery_message/messages/rna_fastq_scout_message.py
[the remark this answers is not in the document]
beatrizsavinhas: Removed!

cg/services/delivery_message/messages/rna_fastq_analysis_scout_message.py
beatrizsavinhas: Very much so! Well spotted and thank you! Changed now.

cg/services/delivery_message/messages/rna_fastq_scout_message.py
beatrizsavinhas: Yes! Thanks! Changed now.

cg/services/delivery_message/utils.py
beatrizsavinhas: Added!

cg/services/delivery_message/messages/rna_fastq_scout_message.py
```diff
@@ -0,0 +1,60 @@
...
+            return self.get_case_message(cases[0])
+        return self.get_cases_message(cases)
+
+    def get_case_message(self, case: Case) -> str:
+
+        related_uploaded_dna_cases: list[Case] = self.store.get_uploaded_related_dna_cases(
+            rna_case=case
+        )
+        scout_links: list[str] = [get_scout_link(case) for case in related_uploaded_dna_cases]
+        scout_links_row_separated: str = "\n".join(scout_links)
```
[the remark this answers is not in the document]
beatrizsavinhas: This was done because all the other scout delivery messages had this structure. Do you think I should refactor all of them in this PR? Or just add this `_get_scout_links_row_separated` function only for the messages I introduce here and leave the others as they are?

cg/services/delivery_message/messages/rna_scout_message.py
beatrizsavinhas: See my question in the comment above!

cg/services/delivery_message/messages/rna_analysis_scout_message.py
beatrizsavinhas: See my question in the comment above!

cg/server/endpoints/cases.py
```diff
@@ -1,9 +1,15 @@
 import logging
 from http import HTTPStatus
+
 from flask import Blueprint, abort, g, jsonify, request
-from cg.exc import CaseNotFoundError, OrderMismatchError
-from cg.server.dto.delivery_message.delivery_message_request import DeliveryMessageRequest
-from cg.server.dto.delivery_message.delivery_message_response import DeliveryMessageResponse
+
+from cg.exc import CaseNotFoundError, CgDataError, OrderMismatchError
+from cg.server.dto.delivery_message.delivery_message_request import (
```
diitaz93, line 7: these imports are still wrongly linted

cg/services/delivery_message/messages/rna_analysis_scout_message.py
beatrizsavinhas: See my question in the comment above!

cg/services/delivery_message/messages/rna_fastq_analysis_scout_message.py
beatrizsavinhas: See my question in the comment above!See my question in the comment above!

cg/store/crud/read.py
```diff
@@ -1581,14 +1586,13 @@ def get_case_ids_for_samples(self, sample_ids: list[int]) -> list[str]:
...
         self,
-        sample_internal_id: str,
+        sample: Sample,
         prep_categories: list[SeqLibraryPrepCategory],
         collaborators: set[Customer],
-    ) -> list[Sample]:
-        """Returns a list of samples with the same subject_id, tumour status and within the collaborators of a given sample and within the given list of prep categories."""
-        sample: Sample = self.get_sample_by_internal_id(internal_id=sample_internal_id)
+    ) -> Query:
+        """Returns a sample query with the same subject_id, tumour status and within the collaborators of a given sample and within the given list of prep categories."""
```
diitaz93, line 1595: break this docstring into two lines

cg/services/delivery_message/messages/rna_fastq_scout_message.py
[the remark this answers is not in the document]
islean: If you extract it into a utils method, it makes sense to fix it for the other scout delivery messages as well. Not the end of the world but it would be a nice touch

cg/store/crud/read.py
diitaz93: We wanted to filter early to avoid big intermediate Queries that could lead to slow perfomance

cg/store/crud/read.py
diitaz93: What does this do?

cg/store/crud/read.py
islean: A bit uncertain of how SQLAlchemy works "under the hood" but I thought the filtering was invoked when running the `.all()` or `.first()` etc. Until you basically add a command which "executes" such as these, we are only building the query. So this should not really affect performance as far as I know. Might be wrong though

islean: Seems as if the order does matter. If there is no noticeable performance gain I would still do the joins first. But feel free to go with whatever.

cg/services/delivery_message/utils.py
beatrizsavinhas: Haven't pushed the commit yet hahaha!

cg/services/delivery_message/utils.py
beatrizsavinhas: However, these maps are used in the functions directly above them and maybe it is easier to reads if they are closed to where they are used (and not used anywhere else). What do you think?

cg/store/crud/read.py
beatrizsavinhas: @diitaz93 it retrieves all the objects from the query with a structure of `Case`. Since the query is a joined table of samples, cases and analyses, you can specify that you want to extract the cases only.

beatrizsavinhas:
@ChrOertlin @diitaz93 @islean 
About this question concerning the crud query in `get_uploaded_related_dna_cases` and doing either:

a) All the joins first and all the filtering afterwards
	Pros:
		- Less lines of code
		- Possibly more readable
	Cons:
		- Higher risk of all the joins creating a large dataset first that is only filtered afterwards *

b) A sequence of a join and a filter related to it:
	Pros:
		- Avoid large datasets by filtering as we go *
		- The logic can be more understandable as every join has a subsequent filter related to it - i.e. when we join cases, we then filter on cases
	Cons:
		- More lines of code
		
		
*In this particular circumstance, we usually only have like 1 or 2 dna samples related to the rna sample, which results in like max 5 related cases and maybe 10-20 analyses in total. As such, the results from `_get_related_samples_query` are already small and from the subsequent joins and filters we also expect a small number of results. So I don't think there will be that much of a difference in performance in either of the options. Also, as @islean mentioned, I'm not sure if any data fetching is actually done interim or if we're just putting together a query and only fetching data when calling `.all()` - in the later case, there would be no difference in performance between the two options.

In principle and in general, I think it is better to filter as we go even if it doesn't make a significant difference in this particular case. Also, having the joins together with the related filters makes makes the login a bit more clear.

You might have some more information, arguments or answers to my doubts, though! So, what do you think?

islean:
> @ChrOertlin @diitaz93 @islean About this question concerning the crud query in `get_uploaded_related_dna_cases` and doing either:
> 
> **a) All the joins first and all the filtering afterwards** Pros: - Less lines of code - Possibly more readable Cons: - Higher risk of all the joins creating a large dataset first that is only filtered afterwards *
> 
> **b) A sequence of a join and a filter related to it:** Pros: - Avoid large datasets by filtering as we go * - The logic can be more understandable as every join has a subsequent filter related to it - i.e. when we join cases, we then filter on cases Cons: - More lines of code
> 
> *In this particular circumstance, we usually only have like 1 or 2 dna samples related to the rna sample, which results in like max 5 related cases and maybe 10-20 analyses in total. As such, the results from `_get_related_samples_query` are already small and from the subsequent joins and filters we also expect a small number of results. So I don't think there will be that much of a difference in performance in either of the options. Also, as @islean mentioned, I'm not sure if any data fetching is actually done interim or if we're just putting together a query and only fetching data when calling `.all()` - in the later case, there would be no difference in performance between the two options.
> 
> In principle and in general, I think it is better to filter as we go even if it doesn't make a significant difference in this particular case. Also, having the joins together with the related filters makes makes the login a bit more clear.
> 
> You might have some more information, arguments or answers to my doubts, though! So, what do you think?

I don't really know if there is a "one size fits all" for these situations. The performance can definitely be impacted by the order of our filters/joins, but it is seldom the limiting factor in our system. I think the guiding principle should be to keep it as simple and understandable as possible. For me that usually means that everything is placed intuitively. Whether that means "keep the joins and filters separate" or "keep the filters related to the join as close as possible" is debatable. I would trust your judgement in this. 

I would only like to stress that unless you see a significant performance issue when you test it, I really do think that simplicity should be what guides you.

cg/services/delivery_message/messages/rna_fastq_scout_message.py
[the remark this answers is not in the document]
ChrOertlin: 💯

ChrOertlin:
I agree with Isak that in the CG database performance is likely not a limiting factor since we do not have so much data to begin with. 

Here I think code block that unify similar logic might improve code comprehension overall (i.e. option a), at least it does for me. I'd even argue the join should be treated as any other join we do and separated into a join function that we have in the cg `cg/store/base.py`. I'd recommend you to have a look at these.

cg/store/crud/read.py
beatrizsavinhas: Added.

cg/store/crud/read.py
beatrizsavinhas: Changed.

cg/store/crud/read.py
beatrizsavinhas: Added.

cg/store/crud/read.py
beatrizsavinhas: Changed this.

cg/store/crud/read.py
beatrizsavinhas: Changed this.

cg/store/crud/read.py
beatrizsavinhas: Changed this.

cg/store/crud/read.py
beatrizsavinhas: Changed this.

cg/store/crud/read.py
beatrizsavinhas: Changed this.

beatrizsavinhas:
OK, I have made the changes according to the suggestions - i.e. **a) All the joins first and all the filtering afterwards**!

ChrOertlin (approved):

PR #3911, opened 2024-10-31
Add case names to the delivery messages

### LINKED ISSUE #3846: Add information of what cases are being delivered in the delivery message for analysis files deliveries
As a production member,
I want the delivery message to be altered when we are delivering only analysis files so that it includes information on what cases are being delivered,
So that we do not have to fetch that information from elsewhere or allow us to forget to inform the customer what cases they should expect to be transferred. Today, it would also allow for easier closing of tickets, as we would enforce the information about which cases have been delivered in the message.

# Work impact

Answer the following questions:
- Is there currently a workaround for this issue? If so, what is it?
    - Yes, one can find out the case name from other sources, but it does require you to leave the cigrid view
- How much time would be saved by implementing this feature on a weekly basis?
    - Maybe ten seconds per case?
- How many users are affected by this issue?
    - All production members
- Are customers affected by this issue?
    - Yes and no. Yes if we forget to do this manually.

# Acceptance Criteria

Given that a case(s) have delivery options: `fastq, fastq+analysis, analysis`:
- [ ] When we copy the delivery message for a case/cases the delivery message should first state which `case.name(s)` we have selected and started the transfer for, then give the path that we are transferring those cases to (as we currently do)

Suggested message for a single case:

```plaintext
Hi,

The analysis files for the following case are currently being uploaded to your inbox on Caesar:

<customer_case_name>

Available under:

/home/<customer>/inbox/<ticket_id>
```

Suggested message for multiple cases:

```plaintext
Hi,

The analysis files for the following cases are currently being uploaded to your inbox on Caesar:

<customer_case_name_1>
<customer_case_name_2>

Available under:

/home/<customer>/inbox/<ticket_id>
```

# Notes

- Additional information.
- Dependencies.
- Related user stories.

islean:
## Description

Closes #3846. 

### Added

- Information on what cases the delivery message was generated for.

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/delivery_message/messages/fastq_message.py
Vince-janv, line 1: Looking at this I'm not sure the fastq_message is within the scope of the issue. If it is I think we should list the files and not the cases.

The argument is that fastq-only deliveries are only allowed for raw-data cases, ie cases with no analyses. Then it does not make sense to report the made-up case names we have created in our database

Vince-janv (approved):
Looks good! Small comment regarding the fastq message. Might need to send a question back to prod

cg/services/delivery_message/messages/fastq_message.py
islean: I was wondering about that when I was writing it! Would it be sample names that would be more appropriate and can we then make sure that all cases have one sample only? Or is that why you suggest the file names?

islean: @Clinical-Genomics/prodbioinfo Any input?

islean: I guess not - should I loop through all the samples in the cases instead then?

islean: Pinging @Clinical-Genomics/prodbioinfo

karlnyr: This is fastq+analysis. So we should definitely use the case names. For the "FastqMessage" below, I think sample names are more appropriate :)

ChrOertlin: Indeed since we make cases with a non-user provided case name we should use the sample names. That being said I think there is merit to first finish the task one case per fastq sample, and then implement the messaging strategy. Otherwise we end up doing double work here.

islean: Regardless of the number of samples in a case, won't we end up writing it like this?

```Python
sample_names: str = "\n".join([sample.name for case in cases for sample in case.samples])
```

Or am I missing some rewrite or where this PR is affected by the number of samples in a case?

ChrOertlin: Should not change this logic here, nvm me

islean:
Tested with delivery fastq bulk action:

Hello,

The fastq files for the following samples have been uploaded to your inbox on Caesar:

<name_1>
<name_2>
<name_3>

Available under: 

/home/<cust>/inbox/<ticket>

Tested with fastq+analysis order action:

Hello,

The fastq and analysis files for the following case are currently being uploaded to your inbox on Caesar:

<case_name>

Available under: 
/home/<cust>/inbox/<ticket>

Tested with analysis order action:

Hello,

The analysis files for the following cases are currently being uploaded to your inbox on Caesar:

<case_1>
<case_2>

Available under: 

/home/<cust>/inbox/<ticket>

islean:
Deployed to the VMs

PR #3913, opened 2024-11-04
add fastq order and sample models

diitaz93:
## Description
Closes https://github.com/Clinical-Genomics/improve-order-flow/issues/67

### Added

- Adds the FastqSample and FastqOrder models to the Order validation service

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/order_validation_service/models/order_with_samples.py
```diff
@@ -1,11 +1,10 @@
-from cg.services.order_validation_service.models.aliases import NonHumanSample
 from cg.services.order_validation_service.models.order import Order
 from cg.services.order_validation_service.models.sample import Sample
 
 
 class OrderWithNonHumanSamples(Order):
```
diitaz93, line 5: Should this be renamed to `OrderWithSamples`?

cg/services/order_validation_service/models/aliases.py
```diff
@@ -1,12 +1,7 @@
...
 from cg.services.order_validation_service.workflows.mip_dna.models.sample import MipDnaSample
-from cg.services.order_validation_service.workflows.mutant.models.sample import MutantSample
 from cg.services.order_validation_service.workflows.tomte.models.case import TomteCase
 from cg.services.order_validation_service.workflows.tomte.models.sample import TomteSample
 
-
 SampleWithRelatives = TomteSample | MipDnaSample
 CaseContainingRelatives = TomteCase | MipDnaCase
-
-NonHumanSample = MutantSample | MicrosaltSample
```
islean, line 12: I agree that the name is a bit clunky but is there a reason for this being removed? I am pretty sure that this was added due to certain validation rules only being applicable for these workflows. If we change the type hints in the rules, will we not refer to properties which are not necessarily present in the sample?

cg/services/order_validation_service/models/order_with_samples.py
islean: Maybe? At least the current naming is not correct, since we do have human orders without cases (such as fastq). Just wondering if OrderWithSamples will cause confusion since all orders have samples, it is just a matter of whether they have cases or not. OrderWithSanples should be better than the current naming though

cg/services/order_validation_service/workflows/fastq/models/order.py
```diff
@@ -0,0 +1,8 @@
+from cg.constants import DataDelivery
+from cg.services.order_validation_service.models.order_with_samples import OrderWithNonHumanSamples
+from cg.services.order_validation_service.workflows.fastq.models.sample import FastqSample
+
+
+class FastqOrder(OrderWithNonHumanSamples):
```
islean, line 6: The OrderWithNonHumanSamples is a typing alias - not a parent class. So I do not think that this inheritance will work. I would also like all order types to only inherit from the parent class Order, so as to make the code as readable as possible and avoid the OrderInSample stacking of inheritances.

cg/services/order_validation_service/workflows/fastq/models/sample.py
```diff
@@ -0,0 +1,13 @@
+from cg.models.orders.sample_base import PriorityEnum, SexEnum
+from cg.services.order_validation_service.models.sample import Sample
+
+
+class FastqSample(Sample):
+    concentration_ng_ul: float | None = None
+    elution_buffer: str
```
islean, line 7:
```suggestion
    elution_buffer: ElutionBufferEnum
```

cg/services/order_validation_service/workflows/fastq/models/sample.py
```diff
@@ -0,0 +1,13 @@
...
+    priority: PriorityEnum
+    quantity: int | None = None
+    require_qc_ok: bool
+    sex: SexEnum
+    source: str  # Required if no internal id
```
islean, line 12: There are no internal ids sent from the front end for fastq orders - I suspect the Atlas info might be outdated. Internal Ids are sent for existing samples/reruns which we do not support for fastq.

cg/services/order_validation_service/workflows/fastq/models/sample.py
```diff
@@ -0,0 +1,14 @@
+from cg.models.orders.sample_base import PriorityEnum, SexEnum
+from cg.services.order_validation_service.constants import ElutionBuffer
+from cg.services.order_validation_service.models.sample import Sample
+
+
+class FastqSample(Sample):
+    concentration_ng_ul: float | None = None
+    elution_buffer: ElutionBuffer
```
diitaz93, line 8: @islean ? Atlas says it is optional
```suggestion
    elution_buffer: ElutionBuffer | None = None
```

cg/services/order_validation_service/workflows/fastq/models/sample.py
diitaz93: So this means that source is optional or that it should not be an attribute to consider?

cg/services/order_validation_service/models/aliases.py
diitaz93: I misunderstood the inheritance model, I will put this back

cg/services/order_validation_service/models/order_with_samples.py
diitaz93: Ok but now I understand that this is not involved in the inheritance. Therefore the renaming is out of scope

cg/services/order_validation_service/workflows/fastq/models/order.py
diitaz93: The `OrderWithNonHumanSamples` is defined as a class in `cg/services/order_validation_service/models/order_with_samples.py` and it is used as a parent class for the Microbial and Mutant orders already

islean: Sorry, this was a very confused comment added yesterday where I got the Sample alias confused - you are completely correct

cg/services/order_validation_service/workflows/fastq/models/sample.py
islean: I see no way in which you can submit an order without the buffer set in this order type so I think it should be considered required: 
image

Atlas will have some confusing statements, since our different softwares are not always in agreement. For fastq orders, I interpret the Atlas page as saying "this is required in the FE but the BE gives it a default value as well". Simply having it as required in the BE makes more sense to me

cg/services/order_validation_service/workflows/fastq/models/sample.py
islean: I was mainly thinking of the comment you added. The source seems to be correctly type hinted 💪

cg/services/order_validation_service/models/order_with_samples.py
islean: I do think we need to have a think about what the proposed change with how we type hint the samples will mean. See for example the rule `validate_organism_exists` - for all the samples which the "NonHumanSample" encompass (Mutant and Microsalt thus far), we will have a sample.organism attribute that we can access. If we change the type hint here to `Sample` instead of `NonHumanSample` there is no guarantee that there actually is an `organism`attribute.

I think I like your proposed change since Fastq orders will not be nonhuman, and maybe we should look into adding some order aliases to remedy the failing attribute accessing? So we could add `NonHumanOrderTypes = MicrosaltOrder | MutantOrder` and change the type hint in `validate_organism_exists` to use that alias?

cg/services/order_validation_service/workflows/fastq/models/sample.py
```diff
@@ -0,0 +1,14 @@
...
+    priority: PriorityEnum
+    quantity: int | None = None
+    require_qc_ok: bool
+    sex: SexEnum
+    source: str  # Required if no internal id
+    tumour: bool = False
```
islean, line 14: Not for this PR, but this reminded me that the attribute tumour seems to be incorrectly named is_tumour for the balsamic sample.

cg/services/order_validation_service/workflows/fastq/models/sample.py
```diff
@@ -0,0 +1,14 @@
+from cg.models.orders.sample_base import PriorityEnum, SexEnum
+from cg.services.order_validation_service.constants import ElutionBuffer
+from cg.services.order_validation_service.models.sample import Sample
+
+
+class FastqSample(Sample):
```
islean, line 6: We seem to be missing SubjectId (seems to be missing in Atlas as well).

diitaz93: Should that be in the parent class instead?

cg/services/order_validation_service/workflows/fastq/models/sample.py
diitaz93: I can change it in this PR

diitaz93, line 14: @islean Should I override the `enumerated_samples` property with the correct return type as it is done in Microsalt (but not in Mutant 👀 )? I considered it unnecessary but maybe there is some functionality I didn't consider

cg/services/order_validation_service/workflows/fastq/models/sample.py
```diff
@@ -0,0 +1,14 @@
...
+    concentration_ng_ul: float | None = None
+    elution_buffer: ElutionBuffer
+    priority: PriorityEnum
+    quantity: int | None = None
+    require_qc_ok: bool
+    sex: SexEnum
+    source: str  # Required if no internal id
```
diitaz93, line 13:
```suggestion
    source: str 
```

cg/services/order_validation_service/workflows/fastq/models/sample.py
```diff
@@ -0,0 +1,14 @@
+from cg.models.orders.sample_base import PriorityEnum, SexEnum
+from cg.services.order_validation_service.constants import ElutionBuffer
+from cg.services.order_validation_service.models.sample import Sample
+
+
+class FastqSample(Sample):
+    concentration_ng_ul: float | None = None
```
diitaz93, line 7: as in Microsalt @islean?
```suggestion
    sample_concentration: float | None = None
```

cg/services/order_validation_service/workflows/fastq/models/sample.py
islean: I also prefer it, but I do consider renaming of variables sent to us by the FE to be out of scope and should be a separate issue/work item.

cg/services/order_validation_service/workflows/fastq/models/sample.py
islean: Nope, a subject id refers to a person, and for our nonhuman order types, e.g. mutant, you will not be able to send a subject id

cg/services/order_validation_service/workflows/fastq/models/sample.py
islean: Humm, good question. I think we override it if it makes the type hints and attributes work better, which we will see when we add rules for the workflow. My guess is that we have some rule which is only applicable for Microsalt (out of the as of yet implemented order types either way), so when accessing attributes it only made sense if the .samples was certain to return a list of MicrosaltSamples.

cg/services/order_validation_service/workflows/fastq/models/sample.py
islean: Nope, actually two different fields (although I agree that this is a great source for confusion). The concentration_ng_ul can be seen here:

image

while the sample_concentration can be seen here:

image

Come to think of it - I think the sample_concentration has been removed from Microsalt due to the recent request by Fay?

diitaz93: yes it has been removed

islean: Hmm, I'll remove the field from microsalt in a PR then

diitaz93: unify names in another PR, leave as concentration_ng_ul

cg/services/order_validation_service/workflows/fastq/models/sample.py
diitaz93: leave tumour, fix in another PR

diitaz93: add overriding of enumerated_samples

diitaz93: add for mutant too

cg/services/order_validation_service/workflows/fastq/models/sample.py
diitaz93: I can remove it from Microsalt

cg/services/order_validation_service/workflows/fastq/models/sample.py
diitaz93: fixed

cg/services/order_validation_service/rules/sample/rules.py
```diff
@@ -77,7 +71,7 @@ def validate_required_volume(
 
 
 def validate_organism_exists(
-    order: OrderWithNonHumanSamples, store: Store, **kwargs
+    order: OrderWithSamples, store: Store, **kwargs
```
islean, line 74: Are we planning to address the introduced warnings in a separate PR? I can do it after you merge.

islean (approved):
Looks good! Should I do some alias tinkering with the sample rules to make sure all attributes exist?

cg/services/order_validation_service/rules/sample/rules.py
diitaz93: do it in a separate PR, not sure how to solve them

PR #3914, opened 2024-11-04
Add mip-rna order validation

ahdamin:
### Added

- Models for `mip-rna` order, case, and sample validation
- Related validation rules and validation service


diitaz93:
I see you also implemented the validation rules, so you can also close https://github.com/Clinical-Genomics/improve-order-flow/issues/79 in this PR 💪

diitaz93 (commented):
The models look good to me, I have no knowledge about the validation rules though

islean:
I know that we have not been super consistent with it in the project but it would be super nice if we had an integration test for this. So tests/services/order_validation_service/workflows you can find `test_valid_order` for tomte and balsamic and having something similar to test this PR with would benefit us greatly (we should probably have that for all order types). 

It is quite cumbersome to set up such a test though, and one needs to create a whole valid order for the order type to pass through the whole validation. So I am also fine with that being a separate PR.

cg/services/order_validation_service/workflows/mip_rna/models/order.py
```diff
@@ -0,0 +1,16 @@
...
+from cg.services.order_validation_service.models.order_with_cases import OrderWithCases
+from cg.services.order_validation_service.workflows.mip_rna.models.case import (
+    MipRnaCase,
+)
+
+NewCase = Annotated[MipRnaCase, Tag("new")]
+OldCase = Annotated[ExistingCase, Tag("existing")]
+
+
+class MipRnaOrder(OrderWithCases):
```
islean, line 15: You probably want to override the delivery type with only the allowed delivery types for this specific order type

cg/services/order_validation_service/workflows/mip_rna/models/sample.py
```diff
@@ -0,0 +1,20 @@
...
+from cg.models.orders.sample_base import NAME_PATTERN, ControlEnum, SexEnum
+from cg.services.order_validation_service.constants import TissueBlockEnum
+from cg.services.order_validation_service.models.sample import Sample
+
+
+class MipRnaSample(Sample):
+    age_at_sampling: float | None = None
+    concentration_ng_ul: float | None = None
+    control: ControlEnum
+    elution_buffer: str | None = None
```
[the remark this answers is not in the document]
islean: Whether the buffer is optional is dependent on the order type - for MIP-RNA it seems to be optional (although I can't think of a scenario where it is required for this workflow, maybe it should be dropped even? But even if so, definitely not in this PR.). Easiest way to check is in the order portal. 

So I would write this as `elution_buffer: ElutionBuffer | None = None`.

cg/services/order_validation_service/workflows/mip_rna/models/sample.py
```diff
@@ -0,0 +1,20 @@
+from pydantic import Field
+
+from cg.models.orders.sample_base import NAME_PATTERN, ControlEnum, SexEnum
+from cg.services.order_validation_service.constants import TissueBlockEnum
+from cg.services.order_validation_service.models.sample import Sample
+
+
+class MipRnaSample(Sample):
```
islean, line 8: You seem to have started this branch without the latest commits in improve-order-flow-main. Or at least in your branch the parent `Sample` does still have the require_qc_ok attribute, which has since been deemed to be order type specific. So you will need to add the require_qc_ok in the MipRnaSample

cg/services/order_validation_service/workflows/mip_rna/validation_rules.py
```diff
@@ -0,0 +1,50 @@
...
+    validate_case_names_available,
+    validate_case_names_not_repeated,
+]
+
+CASE_SAMPLE_RULES: list[callable] = [
+    validate_application_compatibility,
+    validate_application_exists,
+    validate_application_not_archived,
+    validate_buffer_skip_rc_condition,
+    validate_required_volume,
```
islean, line 37: Pet peeve but could we place validate_required_volume in its alphabetical place in the list? 🥺

cg/services/order_validation_service/workflows/mip_rna/validation_rules.py
```diff
@@ -0,0 +1,50 @@
...
+    validate_wells_contain_at_most_one_sample,
+)
+
+CASE_RULES: list[callable] = [
+    validate_case_internal_ids_exist,
+    validate_case_names_available,
+    validate_case_names_not_repeated,
+]
+
+CASE_SAMPLE_RULES: list[callable] = [
```
islean, line 32: We seem to have missed implementing this rule 

image

Seems to have been missed for other workflows as well so it might warrant its own PR where we add it to all relevant order types in one go?

islean (approved):
Nice work!

cg/services/order_validation_service/workflows/mip_rna/validation_rules.py
islean: It seems to exist for Balsamic so we just need to widen its scope!

cg/services/order_validation_service/workflows/mip_rna/validation_rules.py
ahdamin: @islean resolved?

cg/services/order_validation_service/workflows/mip_rna/models/sample.py
ahdamin: @islean Added as a required field..

cg/services/order_validation_service/workflows/mip_rna/models/order.py
ahdamin: @islean Could you please check if this is addressed by the new commits?

cg/services/order_validation_service/rules/case_sample/rules.py
```diff
@@ -390,3 +390,27 @@ def validate_tube_container_name_unique(
                 error = ContainerNameRepeatedError(case_index=case_index, sample_index=sample_index)
                 errors.append(error)
     return errors
+
+
+def validate_subject_sex_consistency(
```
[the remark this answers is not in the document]
diitaz93: yes, see line 332

cg/services/order_validation_service/workflows/mip_rna/models/order.py
diitaz93: It looks good @ahdamin

cg/services/order_validation_service/workflows/mip_rna/validation_rules.py
diitaz93: I can see that it is implemented 🙌 Could you make this list in alphabetical order so that it is easier to read?

diitaz93 (approved):
Looks great! Just delete the new `validate_subject_sex_consistency` function

cg/services/order_validation_service/rules/case_sample/rules.py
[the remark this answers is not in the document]
diitaz93: @ahdamin could you add the function in line 332? so that all are in alphabetical order

diitaz93 (approved):
👍

PR #3929, opened 2024-11-11
technical debt( rework order.workflow column)

ChrOertlin:
## Description

Removes the order.workflow column in the database.

Reason: 
While the worfklow column on the order table is an accurate representation of how orders are placed currently it is not optimal for the following reasons.

1. Introduces data redudancy
2. Limits flexibility between orders and cases if requirements change

### Changed

- Removed workflow column from order table
- Changed get_orders() crud function to perform a case join and filter on the workflow there instead
-  Added property workflow to orders model, which gets the workflow of the first case associated with it. NOTE: if requirements change where we have multiple different workflows per order this will require attention

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/alembic/versions/2024_11_11_05ffb5e13d7b_remove_order_workflow.py
+++ b/alembic/versions/2024_11_11_05ffb5e13d7b_remove_order_workflow.py
@@ -1,0 +1,51 @@
+"""remove_order_workflow
+
+Revision ID: 05ffb5e13d7b
+Revises: 3503b53a9bc3
+Create Date: 2024-11-11 09:35:33.020098
+
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "05ffb5e13d7b"
+down_revision = "3503b53a9bc3"
+branch_labels = None
+depends_on = None
+workflows = [
+    "balsamic",
+    "balsamic-pon",
+    "balsamic-qc",
+    "balsamic-umi",
+    "demultiplex",
+    "raw-data",
+    "fluffy",
+    "microsalt",
+    "mip-dna",
+    "mip-rna",
+    "mutant",
+    "raredisease",
+    "rnafusion",
+    "rsync",
+    "spring",
+    "taxprofiler",
+    "sars-cov-2",
+    "tomte",
+]
+
+
+def upgrade():
+    op.drop_column(table_name="order", column_name="workflow")
+
+
+def downgrade():
+    op.add_column(
+        "order",
+        sa.Column(
+            "workflow",
+            type_=sa.Enum(*workflows),
+        ),
+    )
--- a/cg/store/filters/status_order_filters.py
+++ b/cg/store/filters/status_order_filters.py
@@ -9,12 +9,6 @@ from cg.server.dto.orders.orders_request import OrderSortField, SortOrder
 from cg.store.models import Customer, Order
 
 
-def filter_orders_by_workflow(orders: Query, workflow: str | None, **kwargs) -> Query:
-    if workflow == Workflow.BALSAMIC:
-        return orders.filter(Order.workflow.startswith(workflow))
-    return orders.filter(Order.workflow == workflow) if workflow else orders
-
-
 def filter_orders_by_id(orders: Query, id: int | None, **kwargs) -> Query:
     return orders.filter(Order.id == id) if id else orders
 
@@ -65,7 +59,6 @@ class OrderFilter(Enum):
     BY_IDS: Callable = filter_orders_by_ids
     BY_SEARCH: Callable = filter_orders_by_search
     BY_TICKET_ID: Callable = filter_orders_by_ticket_id
-    BY_WORKFLOW: Callable = filter_orders_by_workflow
     BY_OPEN: Callable = filter_orders_by_is_open
     PAGINATE: Callable = apply_pagination
     SORT: Callable = apply_sorting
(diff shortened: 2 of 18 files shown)


alembic/versions/2024_11_11_05ffb5e13d7b_remove_order_workflow.py
```diff
@@ -0,0 +1,51 @@
...
+    "microsalt",
+    "mip-dna",
+    "mip-rna",
+    "mutant",
+    "raredisease",
+    "rnafusion",
+    "rsync",
+    "spring",
+    "taxprofiler",
+    "sars-cov-2",
```
islean, line 35:
```suggestion
    "sars-cov-2",
```
Was that never removed as an option?

cg/store/filters/status_order_filters.py
```diff
@@ -9,12 +9,6 @@
 from cg.store.models import Customer, Order
 
 
-def filter_orders_by_workflow(orders: Query, workflow: str | None, **kwargs) -> Query:
-    if workflow == Workflow.BALSAMIC:
-        return orders.filter(Order.workflow.startswith(workflow))
```
islean, line 14: We should test how this functionality works with this change. This ensured that balsamic-umi and balsamic-qc orders shows up under the balsamic tab in cigrid. We might need to add something similar for the get_orders endpoint

alembic/versions/2024_11_11_05ffb5e13d7b_remove_order_workflow.py
```diff
@@ -0,0 +1,51 @@
...
+    "tomte",
+]
```
diitaz93, line 37: missing "jasen" and "raw-data", and should not include "sars-cov-2", see the definition of the Workflow strenum

cg/store/models.py
```diff
@@ -991,9 +991,12 @@ class Order(Base):
     customer: Mapped[Customer] = orm.relationship(foreign_keys=[customer_id])
     order_date: Mapped[datetime] = mapped_column(default=datetime.now)
     ticket_id: Mapped[int] = mapped_column(unique=True, index=True)
-    workflow: Mapped[str] = mapped_column(types.Enum(*(workflow.value for workflow in Workflow)))
     is_open: Mapped[bool] = mapped_column(default=True)
 
+    @property
+    def workflow(self) -> Workflow:
+        return self.cases[0].data_analysis
```
islean, line 998: Potential issue with regards to MAF cases: For each fastq order we will have only one of its cases (the "real" one) being of the fastq workflow, while we have a number of MAF cases under MIP-DNA. This implementation would make the fastq orders have a workflow which quite often says MIP-DNA

ahdamin (approved):
Looks good ..!

cg/store/models.py
```diff
@@ -991,9 +991,12 @@ class Order(Base):
     customer: Mapped[Customer] = orm.relationship(foreign_keys=[customer_id])
     order_date: Mapped[datetime] = mapped_column(default=datetime.now)
     ticket_id: Mapped[int] = mapped_column(unique=True, index=True)
-    workflow: Mapped[str] = mapped_column(types.Enum(*(workflow.value for workflow in Workflow)))
```
islean, line 994: I think the OrdersResponse in one of the endpoints (maybe more response DTOs as well) just renders the database entries. We might need to add the workflow to the response DTOs to not break functionality in the FE

islean (approved):
Looks good! Added some considerations for some edge cases that we should look at!

cg/store/models.py
ChrOertlin: I think the property takes care of that. It was accessing orders.workflow (which now still is there)

diitaz93 (approved):
Looks good 👍

cg/store/models.py
ChrOertlin: I think a new case is created attached to a different order for this ( I will check). So ideally these things should not mingle.

alembic/versions/2024_11_11_05ffb5e13d7b_remove_order_workflow.py
ChrOertlin: Will fix it

cg/store/models.py
ChrOertlin: I just checked. Indeed we put the MAF case onto the original order instead of making a separate one...

islean: Yeah. I still don't quite understand MAF cases but I think it makes sense to have them in the same order so you know what cases they are tied to?

cg/store/models.py
islean: Oh, I got confused by the DatabaseOrder and the Order name clashes. You are most likely correct

cg/store/models.py
islean: I would be very happy if they could be in another order so yeah?

cg/store/filters/status_order_filters.py
ChrOertlin: I moved this logic part to the order service.

ChrOertlin:
# tests

- [ ] Database adapted
- [ ] Trailblazer can see orders per workflow
- [ ] Orders can be placed properly

PR #3933, opened 2024-11-12
fix(balsmic workflow fetching order view)

ChrOertlin:
## Description

Fetching of all orders for balsamic workflows was not implemented with changing the column residency.

The problem:
Trailblazer has one balsamic tab for three different workflows.

The old flow had business logic in the crud layer that translated the balsamic call to include balsamic umi and balsamic qc.
This PR aims to re-introduce this logic

image

### Added

- OrderQueryParams model
Reason: This model now translates the front end request into query parameters that takes care of the businiess logic PRIOR to sending it to the CRUD. This way we can keep the crud clean of this logic
Furthermore, we solve a coupling issue between the front end and the crud layer where we separate OrderRequest from crud functionality. If we where to require end point changes these do now not propagate to the CRUD layer.

### Changed

- Moved OrderService specific logic into the Service to improve encapsulation and logic coherence. While working with this class it was confusing to jump between files to find logic that is essential for only this class.

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/orders/order_service/order_service.py
```diff
@@ -25,7 +26,8 @@ def get_order(self, order_id: int) -> OrderResponse:
         return create_order_response(order=order, summary=summary)
 
     def get_orders(self, orders_request: OrdersRequest) -> OrdersResponse:
-        orders, total_count = self.store.get_orders(orders_request)
+        order_query_params: OrderQueryParams = self._get_order_query_params(orders_request)
```
islean, line 29: If we are making this change, we should probably do it in the OrdersRequest already? I do not see the gain in reading the args into a pydantic object to cast into a slightly different one moments after?

cg/services/orders/order_service/models.py
```diff
@@ -0,0 +1,22 @@
...
+    sort_order: str | None = Field(default=SortOrder.DESC)
+    search: str | None = None
+    workflows: list[str] | None = []
+    is_open: bool | None = None
+
+    @field_validator("workflows", mode="before")
+    def expand_balsamic_workflow(cls, value):
+        if value and Workflow.BALSAMIC in value:
+            value = list(
+                set(value) | {Workflow.BALSAMIC, Workflow.BALSAMIC_UMI, Workflow.BALSAMIC_QC}
```
islean, line 20: This line is quite cryptic to me - does it do anything different from value = {Workflow.BALSAMIC, Workflow.BALSAMIC_UMI, Workflow.BALSAMIC_UMI} would?

islean: Will it not always just return list(set(value)) when the provided workflow contains "balsamic"?

islean (commented):
I do not really see the need for the introduced Pydantic object and would prefer if we do it in the OrderRequest object which already decouples the FE and the CRUD. But the validator is a much better approach than having this functionality in the filter!

cg/services/orders/order_service/order_service.py
ChrOertlin: I added a diagram showing the communication between the layers. 

Passing the OrderRequest, dependent on requirements from the outside, all the way to the CRUD you are forced to change logic in all layers where the OrderRequest interacts with something. I do not think the OrderRequest achieves the decoupling, it rather couples multiple layers.

Having the Request translated into a different model that is passed on deeper into the system keeps the request logic isolated in the OrderService.

The tradeoff: More complexity (to what degree is arguable) for better separation of concerns and more flexibility of the modules.

islean: My arguments are the following:

This service is simple and the only thing it does is get a request from TB which basically consists of all the filtering parameters in the backend. While I do agree that the request from TB in principle is different from the database filtering options, _today_ the TB request is basically the database parameters . Adding extra complexities for a hypothetical flexibility which there currently is no need for is something I would prefer we wait with until we have a concrete need for it. To me, there are no immediate benefits and this separation of concerns is better acted upon when there is an actual difference. 

In short: I do not see any gain from the added flexibility and I have heard of no plans requiring such a change. I will gladly put this decoupling in place when there is a clear need for it, but to me it just adds confusion currently.

cg/services/orders/order_service/order_service.py
```diff
@@ -22,25 +19,68 @@ def __init__(self, store: Store, status_service: OrderSummaryService) -> None:
...
         elif not order.is_open:
             self.set_open(order_id=order_id, open=True)
+
+    @staticmethod
+    def _get_order_query_params(orders_request: OrdersRequest) -> OrderQueryParams:
+        return OrderQueryParams(
+            page=orders_request.page,
+            page_size=orders_request.page_size,
+            search=orders_request.search,
+            open=orders_request.open,
```
diitaz93, line 52:
```suggestion
            open=orders_request.is_open,
```

cg/services/orders/order_service/order_service.py
```diff
@@ -22,25 +19,68 @@ def __init__(self, store: Store, status_service: OrderSummaryService) -> None:
...
+    @staticmethod
+    def _add_summaries(orders: list[Order], summaries: list[OrderSummary]) -> list[Order]:
+        order_map = {order.id: order for order in orders}
+        for summary in summaries:
+            order = order_map[summary.order_id]
+            order.summary = summary
+        return orders
+
+    @staticmethod
+    def _order_is_closed(case_count: int, delivered_analyses: int) -> bool:
```
diitaz93, line 85:
```suggestion
    def _is_order_closed(case_count: int, delivered_analyses: int) -> bool:
```

cg/services/orders/order_service/order_service.py
```diff
@@ -22,25 +19,68 @@ def __init__(self, store: Store, status_service: OrderSummaryService) -> None:
...
+            id=order.id,
+            is_open=order.is_open,
+            workflow=order.workflow,
+            summary=summary,
+        )
+
+    def _create_orders_response(
+        self, orders: list[DatabaseOrder], summaries: list[OrderSummary], total: int
+    ) -> OrdersResponse:
+        orders: list[Order] = [self.create_order_response(order) for order in orders]
```
diitaz93, line 72:
```suggestion
        orders: list[Order] = [self._create_order_response(order) for order in orders]
```

cg/services/orders/order_service/order_service.py
ChrOertlin: This PR, if not done in this way will result in what I explained.
We would need to modify the OrderRequest to allow for the behaviour and it would trickle down. I'd argue this exact situation already shows the need.

cg/services/orders/order_service/models.py
```diff
@@ -0,0 +1,20 @@
...
+    page_size: int | None = Field(default=50)
+    page: int | None = Field(default=1)
+    sort_field: str | None = Field(default=OrderSortField.ORDER_DATE)
+    sort_order: str | None = Field(default=SortOrder.ASC)
+    search: str | None = None
+    workflows: list[str] | None = []
+    is_open: bool | None = None
+
+    @field_validator("workflows", mode="before")
+    def expand_balsamic_workflow(cls, value):
```
diitaz93, line 16: I see @islean now, implementing this in the OrdersRequest class would be much simpler. Here we are adding a new model with exactly the same attributes but with an extra validation. I also understand @ChrOertlin concern on decoupling layers. Decoupling makes changes easier but maybe we doon't want to incorporate changes here that often? This is an interesting design pattern that already has been considered in the order flow project and we could discuss better

ChrOertlin:
# test
Balsamic-umi orders are now showing

cg/services/orders/order_service/order_service.py
islean: Summarizing a discussion from Slack:

IO: Technically yes - or at least it shows that there is a difference between what is in the request and how the filtering is done. Does it warrant a new object to me? No. If I wrote the PR I would probably add `workflows: Annotated[list[Workflow], BeforeValidator(set_workflows)] = Field(alias="workflow")` to the OrderRequest and be done with that. The difference between the FE workflow parameter and the filtering done on workflows is "close enough" in my mind.

To me a clearer need for the separation would be if we in the FE request suddenly had some parameters which would go towards Housekeeper and some parameters which would go to statusdb. At that point it would make no sense to send the OrderRequest object to both the store and to the housekeeper api for them to only use a fraction of the parameters in the passed object.

CO: To me, the OrderRequest violates SRP.
I also just like that the flow is documented more explictly:
you get a request, that requires a query, and you translate that request to params for the store
Thematically the "Request" and "Response" syntax is more something of the Front end layer to me

IO: And to me we are adding unneeded complexity if we add a marginally different object to remedy that
To me it is mostly flavour, though. I can't argue against it (CO's design) being more flexible and robust

CO: Yes, I think neither is wrong. And we just agree to disagree

At this point, unless anyone else has strong opinions, I would leave it up to CO to decide. I have made my points, he has made his. I think there are pros and cons with both options, and the choice will come down to what weight you give them.

tests/fixtures/orderforms/~$1508.31.mip.xlsx
diitaz93, line 1: This looks like a debris file

cg/services/orders/order_service/models.py
```diff
@@ -0,0 +1,20 @@
+from pydantic import BaseModel, field_validator, Field
+from cg.constants import Workflow
+from cg.server.dto.orders.orders_request import OrderSortField, SortOrder
+
+
+class OrderQueryParams(BaseModel):
+    page_size: int | None = Field(default=50)
+    page: int | None = Field(default=1)
+    sort_field: str | None = Field(default=OrderSortField.ORDER_DATE)
+    sort_order: str | None = Field(default=SortOrder.ASC)
```
islean, line 10: Should the defaults be removed from the OrderRequest object?

cg/services/orders/order_service/order_service.py
diitaz93: My opinion: If this is a hotfix it should not include any refactoring. Bug in prod = fix ASAP, make a bare minimum patch that does not introduce any bugs and merge. Here there is some refactoring that is causing debate and delaying the bug patch. The changes introduced here should have been discussed in a refinement session. Having said that, it seems this solution solves the issue, so I will approve it, although I agree with @islean that a much simpler solution would have sufficed

diitaz93 (approved):
👍

islean (approved):
Looks good!

cg/services/orders/order_service/models.py
ChrOertlin: I would leave it as is.

PR #3936, opened 2024-11-12
(Improve order flow) Minimal implementation FASTQSubmission

Vince-janv:
## Description
This implementation does not change any of the logic in the storing. It just replaces the OrderIn and the weird dict by the pydantic models generated in the validation

This PR reworks the `FastqStoringService` and allows for Fastq orders to be submitted using our new validation flow. Main points:

- `FastqValidationService` now has two entries; `validate` which performs all the validation on a dict and returns a dict with attached errors (same as before) and `parse_and_validate` which runs all the validation and returns a parsed `FastqOrder` if successful. What should happen otherwise?
- `StoreFastqOrderService` now works with the `FastqOrder` object instead of the `OrderIn` object. Although not relevant here, we need to discuss how to populate family id for the orders with cases at some point.
- The `submit_order` endpoint uses the new `FastqValidationService`. The placement of it is very wrong though. The FastqSubmitter has it injected but never uses it. I imagine that we ideally do not even want the submitters any more, or that we move logic currently in the endpoint into them. Probably the latter but would require some more work.
- Tests for storing fastq orders are enabled again.

To decide:
- Should we have same response structure?

### DIFF
--- a/cg/services/order_validation_service/workflows/fastq/validation_service.py
+++ b/cg/services/order_validation_service/workflows/fastq/validation_service.py
@@ -1,4 +1,5 @@
-from cg.exc import OrderError, OrderFormError
+from cg.exc import OrderSubmissionError
+from cg.services.order_validation_service.errors.order_errors import OrderError
 from cg.services.order_validation_service.errors.sample_errors import SampleError
 from cg.services.order_validation_service.errors.validation_errors import ValidationErrors
 from cg.services.order_validation_service.model_validator.model_validator import ModelValidator
@@ -48,5 +49,5 @@ class FastqValidationService(OrderValidationService):
         if parsed_order:
             errors: ValidationErrors = self._perform_rule_validation(order=parsed_order)
         if not errors.is_empty:
-            raise OrderFormError(message="Order form contained errors")
+            raise OrderSubmissionError(message="Order contained errors")
         return parsed_order
--- a/cg/exc.py
+++ b/cg/exc.py
@@ -174,6 +174,12 @@ class OrderError(CgError):
     """
 
 
+class OrderSubmissionError(CgError):
+    """
+    Exception related to order submission.
+    """
+
+
 class OrderFormError(CgError):
     """
     Exception related to the order form.


tests/services/orders/order_store_service/test_fastq_order_service.py
```diff
@@ -161,44 +102,11 @@ def test_store_fastq_samples_non_wgs_as_fastq(
...
-    # GIVEN a basic store with no samples and a fastq order
-    assert not base_store._get_query(table=Sample).first()
-    assert base_store._get_query(table=Case).count() == 0
-
-    # GIVEN a non-existing application tag
-    for sample in fastq_status_data["samples"]:
-        sample["application"] = "nonexistingtag"
-
-    # THEN it should raise OrderError
-    with pytest.raises(OrderError):
```
islean, line 196: This validation is not performed in the storing part anymore

cg/server/endpoints/orders.py
```diff
@@ -175,6 +175,9 @@ def submit_order(order_type: OrderType):
                 content=request_json, file_format=FileFormat.JSON
             ),
         )
+        request.json["delivery_type"] = "fastq"
```
islean, line 178: This should be solved in the FE in https://github.com/Clinical-Genomics/clinical-genomics-ui/issues/606

cg/server/endpoints/orders.py
```diff
@@ -175,6 +175,9 @@ def submit_order(order_type: OrderType):
...
+        request_json["project_type"] = order_type
+        request_json["user_id"] = g.current_user.id
```
islean, line 180: These should be solved in https://github.com/Clinical-Genomics/improve-order-flow/issues/100

cg/services/order_validation_service/workflows/fastq/validation_service.py
```diff
@@ -46,5 +44,10 @@ def _get_errors(self, raw_order: dict) -> ValidationErrors:
...
         )
 
-    def parse_and_validate(self, raw_order: dict) -> Order:
-        pass
+    def parse_and_validate(self, raw_order: dict) -> FastqOrder | dict:
+        parsed_order, errors = ModelValidator.validate(order=raw_order, model=FastqOrder)
+        if parsed_order:
+            errors: ValidationErrors = self._perform_rule_validation(order=parsed_order)
+        if not errors.is_empty:
+            return create_order_validation_response(raw_order=raw_order, errors=errors)
```
islean, line 52: A bit uncertain of what should happen here - should an error be raised or what do we want the FE to look like?

cg/services/orders/submitters/fastq_order_submitter.py
```diff
@@ -1,24 +1,23 @@
...
-)
 
 
 class FastqOrderSubmitter(OrderSubmitter):
     """Submitter for fastq orders."""
 
     def __init__(
         self,
-        order_validation_service: ValidateFastqOrderService,
+        order_validation_service: FastqValidationService,
```
islean, line 14: The whole thing is built a bit backwards now since the order_validation_service is never used within this class

cg/services/orders/store_order_services/store_fastq_order_service.py
```diff
@@ -22,51 +21,26 @@ def __init__(self, status_db: Store, lims_service: OrderLimsService):
...
-                    "tumour": sample.tumour,
-                    "volume": sample.volume,
-                }
-                for sample in order.samples
-            ],
-        }
-        return status_data
+        self._fill_in_sample_ids(samples=order.samples, lims_map=lims_map)
+        new_samples = self.store_items_in_status(order=order)
+        return {"samples": new_samples, "project_data": project_data}
```
islean, line 37: Should this still be records?

cg/services/order_validation_service/errors/validation_errors.py
```diff
@@ -11,3 +11,7 @@ class ValidationErrors(BaseModel):
     case_errors: list[CaseError] = []
     sample_errors: list[SampleError] = []
     case_sample_errors: list[CaseSampleError] = []
+
+    @property
+    def is_empty(self) -> bool:
```
diitaz93, line 16:
```suggestion
    def is_empty(self) -> bool:
        """Return True if there are no errors in any of the attributes."""
```

cg/services/orders/submitters/fastq_order_submitter.py
diitaz93: then when is the validation done?

cg/services/orders/store_order_services/store_fastq_order_service.py
islean: Yes, at least for now

cg/server/endpoints/orders.py
```diff
@@ -178,6 +178,8 @@ def submit_order(order_type: OrderType):
                 content=request_json, file_format=FileFormat.JSON
             ),
         )
+        request_json["project_type"] = order_type
```
diitaz93, line 181: so this line will be removed right?

cg/services/order_validation_service/workflows/fastq/validation_service.py
```diff
@@ -21,14 +20,12 @@ def __init__(self, store: Store):
         self.store = store
 
     def validate(self, raw_order: dict) -> dict:
-        errors: ValidationErrors = self._get_errors(raw_order)
+        parsed_order, errors = ModelValidator.validate(order=raw_order, model=FastqOrder)
+        if parsed_order:
```
diitaz93, line 24: So we only want to apply the rules when the pydantic validation has passed?

diitaz93 (approved):
Looks good 👍

cg/services/order_validation_service/workflows/fastq/validation_service.py
islean: The rules all take a parsed Pydantic BaseModel as input so there is no other option. It will be a bit weird for customers that they do not get all the errors at the same time, but without a massive rework I do not think we have a choice

islean: Like if they send a request which looks like `{"message": "Hello world!"}` there is not much validation that can be performed apart from stating that we can not parse it as an order

cg/server/endpoints/orders.py
islean: yes!

islean: Or a bit unsure why the FE would need to send it actually? This should always be correct

cg/services/order_validation_service/workflows/fastq/validation_service.py
```diff
@@ -46,5 +43,10 @@ def _get_errors(self, raw_order: dict) -> ValidationErrors:
...
         )
 
-    def parse_and_validate(self, raw_order: dict) -> Order:
-        pass
+    def parse_and_validate(self, raw_order: dict) -> FastqOrder:
+        parsed_order, errors = ModelValidator.validate(order=raw_order, model=FastqOrder)
+        if parsed_order:
+            errors: ValidationErrors = self._perform_rule_validation(order=parsed_order)
+        if not errors.is_empty:
+            raise OrderFormError(message="Order form contained errors")
```
Vince-janv, line 51: I think `OrderFormError` is used when parsing excel sheets and JSONs. Maybe just `OrderError` is ok?

cg/services/orders/store_order_services/store_fastq_order_service.py
```diff
@@ -22,51 +21,26 @@ def __init__(self, status_db: Store, lims_service: OrderLimsService):
...
+        new_samples = self.store_items_in_status(order=order)
+        return {"records": new_samples, "project_data": project_data}
 
     def create_maf_case(self, sample_obj: Sample, order: Order) -> None:
         """Add a MAF case to the Status database."""
         case: Case = self.status_db.add_case(
-            data_analysis=Workflow(Workflow.MIP_DNA),
-            data_delivery=DataDelivery(DataDelivery.NO_DELIVERY),
+            data_analysis=Workflow.MIP_DNA,
+            data_delivery=DataDelivery.NO_DELIVERY,
```
Vince-janv, line 43: Funky 😂

cg/services/orders/store_order_services/store_fastq_order_service.py
```diff
@@ -81,56 +55,49 @@ def create_maf_case(self, sample_obj: Sample, order: Order) -> None:
...
-                    comment=sample["comment"],
-                    internal_id=sample.get("internal_id"),
-                    order=order,
-                    ordered=ordered,
+                    name=sample.name,
+                    sex=sample.sex or "unknown",
+                    comment=sample.comment,
+                    internal_id=sample._generated_lims_id,
+                    order=order.name,
+                    ordered=datetime.now(),
```
Vince-janv, line 81: What was the "ordered" variable we're replacing with datetime.now?

cg/services/orders/store_order_services/store_fastq_order_service.py
```diff
@@ -81,56 +55,49 @@ def create_maf_case(self, sample_obj: Sample, order: Order) -> None:
...
-                    raise OrderError(f"Invalid application: {sample['application']}")
-                new_sample.application_version: ApplicationVersion = application_version
+                new_sample.application_version = application_version
                 new_samples.append(new_sample)
                 if not case:
                     case = self.status_db.add_case(
-                        data_analysis=Workflow(submitted_case["data_analysis"]),
-                        data_delivery=DataDelivery(submitted_case["data_delivery"]),
+                        data_analysis=Workflow.RAW_DATA,
+                        data_delivery=DataDelivery.FASTQ,
```
Vince-janv, line 98: I remember making this change. To me using Workflow and DataDelivery makes sense here. Do you agree @islean?

Vince-janv (commented):
Great effort ⭐  (Can't approve my own PR tho...)

cg/services/orders/store_order_services/store_fastq_order_service.py
Vince-janv: (meaning the previous version of these lines)

cg/services/order_validation_service/workflows/fastq/validation_service.py
islean: Works! I will need to fix the naming clashes since we already have another form of OrderError in this class

cg/services/orders/store_order_services/store_fastq_order_service.py
islean: Good catch! Seems to have been the "date" returned when submitting the data to LIMS....

image

image

Would you prefer we keep it that way?

cg/services/order_validation_service/workflows/fastq/validation_service.py
diitaz93: fixed

cg/services/orders/store_order_services/store_fastq_order_service.py
Vince-janv: hahaha, this logic 🔫 Let's use datatime.now()

cg/services/orders/store_order_services/store_fastq_order_service.py
islean: I think so too! Although we need to make sure that the storing service is not used for other workflows I guess. Potentially we should use the one sent on the order object

islean: I think this is fine for now though

PR #3944, opened 2024-11-15
Fix(Isort) Add isort config to pyproject.toml file

Vince-janv:
## Description
To avoid differing configurations between different installation, this PR adds the isort configuration to the `pyproject.toml` file

### Added

- isort config to `pyproject.toml` file


diitaz93 (approved):
🙌

PR #3946, opened 2024-11-18
(Improve order flow) Rework fill in sample ids

islean:
## Description

Closes https://github.com/Clinical-Genomics/improve-order-flow/issues/95. 

### Added

-

### Changed

- _fill_in_sample_ids operates on the Sample object introduced in the improve-order-flow project

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/models/sample.py
+++ b/cg/services/order_validation_service/models/sample.py
@@ -8,7 +8,7 @@ class Sample(BaseModel):
     comment: str | None = None
     container: ContainerEnum
     container_name: str | None = None
-    generated_lims_id: str | None = PrivateAttr(default=None)  # Will be populated by LIMS
+    _generated_lims_id: str | None = PrivateAttr(default=None)  # Will be populated by LIMS
     name: str = Field(pattern=NAME_PATTERN, min_length=2, max_length=128)
     volume: int | None = None
     well_position: str | None = None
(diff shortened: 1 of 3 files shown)


cg/services/orders/submitters/order_submitter.py
```diff
@@ -34,14 +33,13 @@ def store_order(self, order_in: OrderIn):
         pass
 
     @staticmethod
-    def _fill_in_sample_ids(samples: list[dict], lims_map: dict, id_key: str = "internal_id"):
+    def _fill_in_sample_ids(samples: list[Sample], lims_map: dict):
```
islean, line 36: The `Sample` is only used for new samples in the order, meaning that all samples in the input will miss their lims id. So we should be able to remove the check if the sample already has one set

cg/services/order_validation_service/models/sample.py
```diff
@@ -8,6 +8,7 @@ class Sample(BaseModel):
     comment: str | None = None
     container: ContainerEnum
     container_name: str | None = None
+    generated_lims_id: str | None = None  # Will be populated by LIMS and stored as the internal id
```
islean, line 11: Technically a tech-savvy user with curl knowledge can set their own internal ids with this change. Is that something we want to fix now or is that for the future?

diitaz93 (approved):
👍

cg/services/order_validation_service/models/sample.py
diitaz93: fixed with the private attribute

PR #3958, opened 2024-11-20
add pseudo workflow to delivery

diitaz93:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/cli/deliver/base.py
+++ b/cg/cli/deliver/base.py
@@ -6,7 +6,7 @@ from pathlib import Path
 import click
 
 from cg.apps.tb import TrailblazerAPI
-from cg.cli.deliver.utils import deliver_raw_data_for_analyses
+from cg.cli.deliver.utils import deliver_raw_data_for_analyses, get_pseudo_workflow
 from cg.cli.utils import CLICK_CONTEXT_SETTINGS
 from cg.constants import Workflow
 from cg.constants.cli_options import DRY_RUN
@@ -18,9 +18,7 @@ from cg.services.deliver_files.deliver_files_service.deliver_files_service impor
 from cg.services.deliver_files.deliver_files_service.deliver_files_service_factory import (
     DeliveryServiceFactory,
 )
-from cg.services.deliver_files.rsync.service import (
-    DeliveryRsyncService,
-)
+from cg.services.deliver_files.rsync.service import DeliveryRsyncService
 from cg.store.models import Analysis, Case
 
 LOG = logging.getLogger(__name__)
@@ -89,9 +87,10 @@ def deliver_case(
     if not case:
         LOG.error(f"Could not find case with id {case_id}")
         return
+    pseudo_workflow: str = get_pseudo_workflow(case=cases[0])
     delivery_service: DeliverFilesService = service_builder.build_delivery_service(
         delivery_type=delivery_type if delivery_type else case.data_delivery,
-        workflow=case.data_analysis,
+        workflow=pseudo_workflow,
     )
     delivery_service.deliver_files_for_case(
         case=case, delivery_base_path=Path(inbox), dry_run=dry_run
@@ -124,9 +123,10 @@ def deliver_ticket(
     if not cases:
         LOG.error(f"Could not find case connected to ticket {ticket}")
         return
+    pseudo_workflow: str = get_pseudo_workflow(case=cases[0])
     delivery_service: DeliverFilesService = service_builder.build_delivery_service(
         delivery_type=delivery_type if delivery_type else cases[0].data_delivery,
-        workflow=cases[0].data_analysis,
+        workflow=pseudo_workflow,
     )
     delivery_service.deliver_files_for_ticket(
         ticket_id=ticket, delivery_base_path=Path(inbox), dry_run=dry_run
--- a/cg/cli/deliver/utils.py
+++ b/cg/cli/deliver/utils.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from pathlib import Path
 
 from cg.constants import Workflow
+from cg.constants.constants import MICROBIAL_APP_TAGS
 from cg.services.deliver_files.deliver_files_service.deliver_files_service import (
     DeliverFilesService,
 )
@@ -43,3 +44,11 @@ def deliver_raw_data_for_analyses(
             )
             LOG.error(f"Could not deliver files for analysis {analysis.id}: {error}")
             continue
+
+
+def get_pseudo_workflow(case: Case) -> str:
+    """Return the literal '' if the case is Microbial-fastq, otherwise return the case workflow."""
+    tag: str = case.samples[0].application_version.application.tag
+    if case.data_analysis == Workflow.RAW_DATA and tag in MICROBIAL_APP_TAGS:
+        return "microbial-fastq"
+    return case.data_analysis
--- a/cg/constants/constants.py
+++ b/cg/constants/constants.py
@@ -258,6 +258,13 @@ class MicrosaltAppTags(StrEnum):
     PREP_CATEGORY: str = "mic"
 
 
+MICROBIAL_APP_TAGS = [
+    MicrosaltAppTags.MWRNXTR003,
+    MicrosaltAppTags.MWXNXTR003,
+    MicrosaltAppTags.VWGNXTR001,
+]
+
+
 class MutantQC:
     EXTERNAL_NEGATIVE_CONTROL_READS_THRESHOLD: int = 100000
     INTERNAL_NEGATIVE_CONTROL_READS_THRESHOLD: int = 2000
--- a/cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
+++ b/cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
@@ -1,60 +1,38 @@
 """Module for the factory of the deliver files service."""
 
 from typing import Type
+
 from cg.apps.housekeeper.hk import HousekeeperAPI
 from cg.apps.tb import TrailblazerAPI
-from cg.constants import Workflow, DataDelivery
+from cg.constants import DataDelivery, Workflow
 from cg.services.analysis_service.analysis_service import AnalysisService
-from cg.services.deliver_files.file_filter.sample_service import SampleFileFilter
-from cg.services.deliver_files.tag_fetcher.bam_service import (
-    BamDeliveryTagsFetcher,
-)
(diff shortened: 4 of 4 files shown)


cg/constants/constants.py
```diff
@@ -258,6 +258,13 @@ class MicrosaltAppTags(StrEnum):
     PREP_CATEGORY: str = "mic"
 
 
+MICROBIAL_APP_TAGS = [
+    MicrosaltAppTags.MWRNXTR003,
+    MicrosaltAppTags.MWXNXTR003,
+    MicrosaltAppTags.VWGNXTR001,
+]
```
ChrOertlin, line 265: With my suggestion above this is not required

```suggestion
```

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
```diff
@@ -104,10 +82,10 @@ def _get_file_fetcher(self, delivery_type: DataDelivery) -> FetchDeliveryFilesSe
...
     @staticmethod
     def _get_sample_file_formatter(
-        workflow: Workflow,
+        workflow: str,
     ) -> SampleFileFormatter | SampleFileConcatenationFormatter:
         """Get the file formatter service based on the workflow."""
-        if workflow in [Workflow.MICROSALT]:
+        if workflow in [Workflow.MICROSALT, "microbial-fastq"]:
             return SampleFileConcatenationFormatter(FastqConcatenationService())
         return SampleFileFormatter()
```
ChrOertlin, line 90: this is not needed with my suggestion above

cg/cli/deliver/base.py
```diff
@@ -89,9 +87,10 @@ def deliver_case(
     if not case:
         LOG.error(f"Could not find case with id {case_id}")
         return
+    pseudo_workflow: str = get_pseudo_workflow(case=cases[0])
```
ChrOertlin, line 90: this is not needed with my suggestion

With the current implementation this would also be needed in the upload api

cg/cli/deliver/utils.py
```diff
@@ -43,3 +44,11 @@ def deliver_raw_data_for_analyses(
...
             LOG.error(f"Could not deliver files for analysis {analysis.id}: {error}")
             continue
+
+
+def get_pseudo_workflow(case: Case) -> str:
+    """Return the literal '' if the case is Microbial-fastq, otherwise return the case workflow."""
+    tag: str = case.samples[0].application_version.application.tag
+    if case.data_analysis == Workflow.RAW_DATA and tag in MICROBIAL_APP_TAGS:
+        return "microbial-fastq"
+    return case.data_analysis
```
ChrOertlin, line 54: add this function to the factory class in: cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py

```suggestion
def convert_workflow(case: Case) -> Workflow:
    """Converts a workflow.With the introduction of the microbial-fastq delivery type an unsupported combination of delivery type and workflow setup is required. This function makes sure that a raw data workflow with microbial fastq delivery type is treated as a microsalt workflow so that the microbial-fastq sample files can be concatenated."""
    tag: str = case.samples[0].application_version.application.tag
    if case.data_analysis == Workflow.RAW_DATA and case.delivery_type == "microbial-fastq"
        return Workflow.MICROSALT
    return case.data_analysis
```
 add this line to the function build_delivery_service in cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py

```suggestion
workflow: Workflow = self.convert_workflow(workflow,delivery_type)
```

Reasoning: 

We also use this factory in `cg/meta/upload/upload_api.py` with the current implementation we would need to also fix the workflow fetching in that place. 

On the other hand, if we let the factory handle the conversion other places in the code do not need to know about that a conversion is happening. I think this would lead to the least changes overall and the logic is kept within the delivery module.

diitaz93: The problem I see here is that "microbial-fastq" is not a case DataDelivery option, as these refer to Scout, Analysis-fastq etc. We would then still need to test the tag

ChrOertlin: Yes, you are right! The tags would still be required then!

cg/constants/constants.py
ChrOertlin: still required as microbial-fastq is not a delivery option so it will not be available to the factory

cg/cli/deliver/utils.py
diitaz93: implemented in factory class

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
diitaz93: implemented

cg/cli/deliver/base.py
diitaz93: implemented the changes

tests/services/file_delivery/delivery_file_service/test_service_builder.py
```diff
@@ -1,73 +1,77 @@
...
+from cg.services.deliver_files.file_formatter.utils.sample_service import SampleFileFormatter
+from cg.services.deliver_files.file_mover.service import DeliveryFilesMover
+from cg.services.deliver_files.tag_fetcher.abstract import FetchDeliveryFileTagsService
+from cg.services.deliver_files.tag_fetcher.sample_and_case_service import (
+    SampleAndCaseDeliveryTagsFetcher,
 )
+from cg.store.store import Store
+
+
+class DeliveryScenario(BaseModel):
```
diitaz93, line 31: add docstring

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
```diff
@@ -149,6 +134,17 @@ def build_delivery_service(
             analysis_service=self.analysis_service,
         )
 
+    @staticmethod
+    def _convert_workflow(case: Case) -> Workflow:
```
diitaz93, line 138: Im planning to add a unit test for this

ChrOertlin (approved):
Looks good.

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
islean, line 138: Should this not be handled in the `get_sample_file_formatter`? This way is a bit hard to understand for me.

diitaz93: good point

cg/constants/constants.py
islean, line 265: Should these not be fetched from statusdb? Otherwise we need to update this whenever an app tag is archived or introduced?

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
diitaz93: maybe is not that important if the public function is tested

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
```diff
@@ -122,18 +102,23 @@ def _validate_delivery_type(delivery_type: DataDelivery):
...
-        self, workflow: Workflow, delivery_type: DataDelivery
+        self, case: Case, delivery_type: DataDelivery | None = None
     ) -> DeliverFilesService:
-        """Build a delivery service based on the workflow and delivery type."""
-        delivery_type: DataDelivery = self._sanitise_delivery_type(delivery_type)
+        """Build a delivery service based on a case."""
+        delivery_type: DataDelivery = self._sanitise_delivery_type(
+            delivery_type if delivery_type else case.data_delivery
+        )
         self._validate_delivery_type(delivery_type)
```
islean, line 118: "sanitise" and "validate" seem to share some responsibility? If something has been sanitised, it should not need to be validated

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
```diff
@@ -102,12 +82,13 @@ def _get_file_fetcher(self, delivery_type: DataDelivery) -> FetchDeliveryFilesSe
...
-        workflow: Workflow,
+        self,
+        case: Case,
     ) -> SampleFileFormatter | SampleFileConcatenationFormatter:
         """Get the file formatter service based on the workflow."""
-        if workflow in [Workflow.MICROSALT]:
+        converted_workflow: Workflow = self._convert_workflow(case)
+        if converted_workflow in [Workflow.MICROSALT]:
             return SampleFileConcatenationFormatter(FastqConcatenationService())
         return SampleFileFormatter()
```
islean, line 93:
```suggestion
        needs_fastq_concatenation: bool = self._needs_fastq_concatenation(case)
        if needs_fastq_concatenation:
            return SampleFileConcatenationFormatter(FastqConcatenationService())
        return SampleFileFormatter()
```

islean (approved):
Added some comments! I think it makes sense to check apptags in statusdb since then we do not need to apply changes to the code when entries are changed. I also think we can clarify the intent of a method by changing its return type to bool

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
diitaz93: out of scope

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
diitaz93: This is a valid point that we discussed with @ChrOertlin. We agreed that the changes that Chris and Eliot are working on regarding Mutant will kind of override this, as more logic will be introduced regarding when to concatenate. We decided to leave it as "converted workflow"

cg/constants/constants.py
diitaz93: Good point, I will take a look at this

cg/services/deliver_files/deliver_files_service/deliver_files_service_factory.py
ChrOertlin: Input sanitisation: cleaning or transforming input
Input validation: making sure input is semantically correct

Related but not the same

cg/constants/constants.py
diitaz93: fixed

PR #3964, opened 2024-11-25
refactor samples endpoint

diitaz93:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/server/endpoints/samples.py
```diff
@@ -46,21 +44,30 @@ def parse_sample_in_collaboration(sample_id):
...
     else:
-        customers: list[Customer] | None = (
-            None if g.current_user.is_admin else g.current_user.customers
-        )
-        samples: list[Sample] = db.get_samples_by_customers_and_pattern(
-            pattern=request.args.get("enquiry"), customers=customers
-        )
-    limit = int(request.args.get("limit", 50))
+        return abort(HTTPStatus.BAD_REQUEST)
+    limit: int = request.page_size or 50
```
ChrOertlin, line 71: Have a look at the `cg/services/sample_service/sample_service.py`

I would decouple this endpoint from the db as much as possible and let the service handle the db interaction. 
There is also a function called `create_samples_response` that has a sample_dto. 

If you consider this out of scope then ignore. However, since most of it is already setup time required might be minimal for a major improvement.

cg/store/crud/read.py
```diff
@@ -612,8 +612,8 @@ def get_ready_made_library_expected_reads(self, case_id: str) -> int:
 
     def get_samples_by_customers_and_pattern(
         self, *, customers: list[Customer] | None = None, pattern: str = None
-    ) -> list[Sample]:
-        """Get samples by customer and sample internal id  or sample name pattern."""
+    ) -> tuple[list[Sample], int]:
+        """Get samples by customer and sample internal id or sample name pattern."""
```
ChrOertlin, line 616:
```suggestion
        """Get samples by customer and sample internal id or sample name pattern.
        This query is used in the front end next to the samples it also return the total number of samples used to determine the pagination.
        """
```

ChrOertlin (approved):
If suggestion is deemed out of scope the PR looks good otherwise.

cg/server/endpoints/samples.py
diitaz93: moving to sample service

cg/store/crud/read.py
islean: Is this usually part of the documentation? To me, it does not make sense that the backend should know what the FE will use it for. I would probably write `Also returns the total number of samples to support pagination`.

cg/store/crud/read.py
```diff
@@ -623,12 +623,13 @@ def get_samples_by_customers_and_pattern(
...
             filter_functions.extend([SampleFilter.BY_INTERNAL_ID_OR_NAME_SEARCH])
         filter_functions.append(SampleFilter.ORDER_BY_CREATED_AT_DESC)
-        return apply_sample_filter(
+        samples: list[Sample] = apply_sample_filter(
             samples=samples,
             customers=customers,
             search_pattern=pattern,
             filter_functions=filter_functions,
         ).all()
+        return samples, len(samples)
```
islean, line 632: The int we return should not be the same as the number of samples we return. The number of samples we return should always be the same as the page size (if specified) while the total should always refer to the total number of samples matching the filter parameters

cg/store/crud/read.py
```diff
@@ -612,8 +612,8 @@ def get_ready_made_library_expected_reads(self, case_id: str) -> int:
 
     def get_samples_by_customers_and_pattern(
```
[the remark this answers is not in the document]
diitaz93: done

cg/server/endpoints/samples.py
```diff
@@ -46,21 +44,30 @@ def parse_sample_in_collaboration(sample_id):
...
+    samples_request = SamplesRequest.model_validate(request.args.to_dict())
+    if samples_request.status:
+        if not g.current_user.is_admin:
+            return abort(HTTPStatus.FORBIDDEN)
+        else:
+            return _get_samples_handled_from_lims(request=samples_request)
+    customers: list[Customer] | None = None if g.current_user.is_admin else g.current_user.customers
+    samples, total = db.get_samples_by_customers_and_pattern(
+        pattern=samples_request.enquiry, customers=customers
+    )
```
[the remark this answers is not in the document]
diitaz93: done

cg/store/crud/read.py
diitaz93: I added the docstring but I think needs some polishing, will look at it later

islean (commented):
We still need to pass the parameters to the CRUD but looks good!

cg/server/dto/samples/requests.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+class CollaboratorSamplesRequest(BaseModel):
+    enquiry: str
+    customer: str
+    limit: int = 50
+
+
+class SamplesRequest(BaseModel):
+    status: str | None = None
```
islean, line 11:
```suggestion
    status: SampleStatus | None = None
```

cg/exc.py
```diff
@@ -104,6 +104,10 @@ class FlowCellError(CgError):
     """Raised when there is a problem with a flow cell."""
 
 
+class FrontendForbiddenTransactionError(CgError):
```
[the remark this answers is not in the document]
diitaz93: I like it, implementing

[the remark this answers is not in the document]
islean: Even better

cg/server/dto/samples/requests.py
islean: Fair. Nothing will happen if it is not one of them though

islean: but keep it as a string for now

cg/store/crud/read.py
diitaz93: yes, fixed

cg/store/crud/read.py
```diff
@@ -611,9 +611,17 @@ def get_ready_made_library_expected_reads(self, case_id: str) -> int:
...
-        self, *, customers: list[Customer] | None = None, pattern: str = None
-    ) -> list[Sample]:
-        """Get samples by customer and sample internal id  or sample name pattern."""
+        self,
+        *,
+        customers: list[Customer] | None = None,
+        pattern: str = None,
+        limit: int = 50,
+        offset: int = 0,
```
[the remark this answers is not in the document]
diitaz93: This function was only used in the endpoint an in tests. In the endpoint, it gets filtered to 50 anyway, so the default behaviour towards the endpoint is not modified.

cg/constants/constants.py
```diff
@@ -309,3 +309,13 @@ class SequencingQCStatus(Enum):
...
+
+
+class SampleStatus(StrEnum):
+    INCOMING = "incoming"
+    LABPREP = "labprep"
+    SEQUENCING = "sequencing"
+
+    @classmethod
+    def statuses(cls) -> list[str]:
+        return [cls.INCOMING, cls.LABPREP, cls.SEQUENCING]
```
diitaz93, line 321: how about now? @islean

PR #3969, opened 2024-11-25
add pacbio models

diitaz93:
### Added

- PacBio Sample and Pacbio Order
- Pacbio delivery types

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/workflows/pacbio_long_read/models/sample.py
+++ b/cg/services/order_validation_service/workflows/pacbio_long_read/models/sample.py
@@ -12,4 +12,4 @@ class PacbioSample(Sample):
     sex: SexEnum
     source: str
     subject_id: str | None = Field(pattern=NAME_PATTERN, max_length=128)
-    tumour: bool = False
+    tumour: bool
--- a/cg/services/order_validation_service/workflows/pacbio_long_read/models/sample.py
+++ b/cg/services/order_validation_service/workflows/pacbio_long_read/models/sample.py
@@ -1,0 +1,15 @@
+from pydantic import Field
+
+from cg.models.orders.sample_base import NAME_PATTERN, PriorityEnum, SexEnum
+from cg.services.order_validation_service.models.sample import Sample
+
+
+class PacbioSample(Sample):
+    concentration_ng_ul: float | None = None
+    priority: PriorityEnum
+    quantity: int | None = None
+    require_qc_ok: bool
+    sex: SexEnum
+    source: str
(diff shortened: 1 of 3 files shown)


cg/services/order_validation_service/workflows/pacbio_long_read/models/sample.py
```diff
@@ -0,0 +1,15 @@
...
+
+class PacbioSample(Sample):
+    concentration_ng_ul: float | None = None
+    priority: PriorityEnum
+    quantity: int | None = None
+    require_qc_ok: bool
+    sex: SexEnum
+    source: str
+    subject_id: str | None = Field(pattern=NAME_PATTERN, max_length=128)
+    tumour: bool = False
```
ChrOertlin, line 15: should we initialise this to False here? What if the customer forgets to set tumour type, will it then automatically be False? 
It might be safer to fail in such cases in my opinion- that way we can always be sure a this attribute is well set.

ChrOertlin (approved):
Looks good!

cg/services/order_validation_service/workflows/pacbio_long_read/models/sample.py
islean: I do agree that we should probably not give a default here. I can try to give a reason as to why we give a default in the old flow which I realised yesterday when testing the fastq orders in my PR:

For variables which are rendered as toggles in the FE (such as this one or require_qc_ok), the FE is not necessarily consistent with what it sends to the backend. Sometimes the toggle being switched off is sent as false, and sometimes as null. Potentially at some point there was an issue with the untoggled is_tumour being sent as null, whence this was deemed a good solution. 

With all that being said - I have not seen this behaviour for Balsamic orders (I just tried it out and it seems to be sent as false), and even if it was the case, it should be changed in the FE.

islean: Would be interesting to see what happens in Pacbio, but the dev branch does not seem to have been updated with the pacbio flow

diitaz93: I will make it mandatory and keep this in mind for the dev branch

PR #3987, opened 2024-11-29
Improve mip rna messages

beatrizsavinhas:
### Added

- Error raising for when no matching DNA cases are found.

### Changed

- Simplified code structure.

### How to prepare for test

- [x] Deploy using git hub action
- [x] Log in to cigrid stage

### How to test

- [x] Deploy in cg services stage
- [x] Test in Trailblazer

### Expected test outcome

- [x] Check that the correct delivery messages are formed.

## Review

- [x] Tests executed by BSV

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/delivery_message/messages/rna_delivery_message.py
```diff
@@ -51,31 +52,23 @@ def __init__(self, store, strategy: RNAUploadMessageStrategy):
...
     def create_message(self, cases: list[Case]) -> str:
-        if len(cases) == 1:
-            return self._get_case_message(cases[0])
         return self._get_cases_message(cases)
 
-    def _get_case_message(self, case: Case) -> str:
+    def _get_scout_message_for_case(self, case: Case) -> str:
         related_uploaded_dna_cases = self.store.get_uploaded_related_dna_cases(rna_case=case)
+        if not related_uploaded_dna_cases:
+            raise CaseNotFoundError(f"No matching DNA cases for case {case.name}.")
```
ChrOertlin: We can have this in the CRUD layer I reckon. If we try to fetch something that doesn't exist of which we assume it does it should break. so in the get_uploaded_related_dna_cases

cg/services/delivery_message/messages/rna_delivery_message.py
```diff
@@ -51,31 +52,23 @@ def __init__(self, store, strategy: RNAUploadMessageStrategy):
         self.strategy = strategy
 
     def create_message(self, cases: list[Case]) -> str:
-        if len(cases) == 1:
-            return self._get_case_message(cases[0])
         return self._get_cases_message(cases)
 
-    def _get_case_message(self, case: Case) -> str:
+    def _get_scout_message_for_case(self, case: Case) -> str:
```
ChrOertlin, line 57:
```suggestion
    def _get_case_message(self, case: Case) -> str:
```
We do not need to kn ow whether it is for scout or not

cg/services/delivery_message/messages/rna_delivery_message.py
```diff
@@ -51,31 +52,23 @@ def __init__(self, store, strategy: RNAUploadMessageStrategy):
...
         message = "Hello,\n\n"
         for case in cases:
-            related_uploaded_dna_cases = self.store.get_uploaded_related_dna_cases(rna_case=case)
-            scout_links = get_scout_links_row_separated(cases=related_uploaded_dna_cases)
-            message += (
-                f"The analysis for case {case.name} has been uploaded to the corresponding DNA case(s) on Scout at:\n\n"
-                f"{scout_links}\n\n"
-            )
+            scout_message = self._get_scout_message_for_case(case=case)
+            message += scout_message
```
ChrOertlin, line 73: Nice!

cg/services/delivery_message/messages/rna_delivery_message.py
beatrizsavinhas: Don't we?
Now the logic is that for each case, we get a scout message and a file delivery message. This function only gives the part for scout, not for the whole case.

cg/services/delivery_message/messages/rna_delivery_message.py
beatrizsavinhas: Changing this!

cg/services/delivery_message/messages/rna_delivery_message.py
beatrizsavinhas: Thank youuu!

cg/services/delivery_message/messages/rna_delivery_message.py
```diff
@@ -51,31 +51,21 @@ def __init__(self, store, strategy: RNAUploadMessageStrategy):
         self.strategy = strategy
 
     def create_message(self, cases: list[Case]) -> str:
-        if len(cases) == 1:
-            return self._get_case_message(cases[0])
         return self._get_cases_message(cases)
```
islean, line 54: Can we not move the method body of _get_cases_message into this method now? I don't see the point in splitting them

cg/services/delivery_message/messages/rna_delivery_message.py
islean: I agree, I think it makes sense to mention Scout in the name

islean (approved):
Nice!

cg/services/delivery_message/messages/rna_delivery_message.py
beatrizsavinhas: Will keep it!

PR #3994, opened 2024-12-03
add pacbio service

diitaz93:
## Description
Closes https://github.com/Clinical-Genomics/improve-order-flow/issues/85

### Added

- Pacbio validation service
- Sample rules for pacbio
- instantiation of the service in `ext.py`.
- Include service in endpoint

### Changed

- docstrings of some functions
- Turned if statements into elifs

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/server/endpoints/orders.py
+++ b/cg/server/endpoints/orders.py
@@ -44,6 +44,7 @@ from cg.server.ext import (
     mutant_validation_service,
     order_service,
     order_submitter_registry,
+    pacbio_long_read_validation_service,
     rna_fusion_validation_service,
     ticket_handler,
     tomte_validation_service,
@@ -274,16 +275,18 @@ def validate_order(order_type: OrderType):
     response = {}
     if order_type == OrderType.BALSAMIC:
         response = balsamic_validation_service.validate(raw_order)
-    if order_type == OrderType.MICROBIAL_FASTQ:
+    elif order_type == OrderType.MICROBIAL_FASTQ:
         response = microbial_fastq_validation_service.validate(raw_order)
-    if order_type == OrderType.MICROSALT:
+    elif order_type == OrderType.MICROSALT:
         response = microsalt_validation_service.validate(raw_order)
-    if order_type == OrderType.MIP_DNA:
+    elif order_type == OrderType.MIP_DNA:
         response = mip_dna_validation_service.validate(raw_order)
-    if order_type == OrderType.SARS_COV_2:
+    elif order_type == OrderType.PACBIO_LONG_READ:
+        response = pacbio_long_read_validation_service.validate(raw_order)
+    elif order_type == OrderType.SARS_COV_2:
         response = mutant_validation_service.validate(raw_order)
-    if order_type == OrderType.RNAFUSION:
+    elif order_type == OrderType.RNAFUSION:
         response = rna_fusion_validation_service.validate(raw_order)
-    if order_type == OrderType.TOMTE:
+    elif order_type == OrderType.TOMTE:
         response = tomte_validation_service.validate(raw_order)
     return jsonify(response), HTTPStatus.OK
--- a/cg/server/ext.py
+++ b/cg/server/ext.py
@@ -27,6 +27,9 @@ from cg.services.order_validation_service.workflows.mip_dna.validation_service i
 from cg.services.order_validation_service.workflows.mutant.validation_service import (
     MutantValidationService,
 )
+from cg.services.order_validation_service.workflows.pacbio_long_read.validation_service import (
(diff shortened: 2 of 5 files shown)


cg/services/order_validation_service/workflows/pacbio_long_read/validation_service.py
```diff
@@ -0,0 +1,48 @@
...
+)
+from cg.services.order_validation_service.workflows.order_validation_rules import ORDER_RULES
+from cg.services.order_validation_service.workflows.pacbio_long_read.models.order import PacbioOrder
+from cg.services.order_validation_service.workflows.pacbio_long_read.validation_rules import (
+    SAMPLE_RULES,
+)
+from cg.store.store import Store
+
+
+class PacbioLongReadValidationService(OrderValidationService):
```
diitaz93, line 19: too long for a name service? should it be only `PacbioValidationService`?

cg/server/endpoints/orders.py
```diff
@@ -274,16 +275,18 @@ def validate_order(order_type: OrderType):
...
+    elif order_type == OrderType.PACBIO_LONG_READ:
+        response = pacbio_long_read_validation_service.validate(raw_order)
+    elif order_type == OrderType.SARS_COV_2:
         response = mutant_validation_service.validate(raw_order)
-    if order_type == OrderType.RNAFUSION:
+    elif order_type == OrderType.RNAFUSION:
         response = rna_fusion_validation_service.validate(raw_order)
-    if order_type == OrderType.TOMTE:
+    elif order_type == OrderType.TOMTE:
         response = tomte_validation_service.validate(raw_order)
```
ChrOertlin, line 291: can we make a map for this ? 🥲

cg/server/ext.py
```diff
@@ -123,6 +126,7 @@ def init_app(self, app):
 microsalt_validation_service = MicroSaltValidationService(store=db)
 mip_dna_validation_service = MipDnaValidationService(store=db)
 mutant_validation_service = MutantValidationService(store=db)
+pacbio_long_read_validation_service = PacbioLongReadValidationService(store=db)
```
ChrOertlin, line 129: We should move these into a registry

cg/services/order_validation_service/workflows/pacbio_long_read/validation_service.py
ChrOertlin: I think it is "ok", either works!

ChrOertlin (approved):
👍

cg/services/order_validation_service/workflows/pacbio_long_read/validation_service.py
islean: Either way works for me. If the order type is called PacbioLongRead it might be nice to have it like this for now.

cg/server/ext.py
islean: Yes, we definitely should tidy up how these are instantiated. My hunch is that we might not want these services order type specific in the long run even. If they accepted the lists of rules and model (e.g. TomteOrder) as input in the .validation method we could have one service fits all order types.

islean: But I would probably consider that a separate issue from the Pacbio service addition

cg/server/endpoints/orders.py
islean: Would be much better! I think this endpoint will be around for when the customer saves cases/samples as well so it should definitely be done. 

I think when we created this it was intended to be a test endpoint that we would remove before going live, but I do not think that is the case anymore

cg/server/ext.py
ChrOertlin: > Yes, we definitely should tidy up how these are instantiated. My hunch is that we might not want these services order type specific in the long run even. If they accepted the lists of rules and model (e.g. TomteOrder) as input in the .validation method we could have one service fits all order types.

You mean have one "validator" in which you inject a ruleset?

islean: Yeah pretty much! I guess we would have some sort of registry where you would provide the order type and get a ruleset and a Basemodel to validate towards? And then inject them into a common "validator"

diitaz93: closing as out of scope. There is a lot to polish but this will be done in the dev branch

cg/server/endpoints/orders.py
diitaz93: We should refactor this before we go live but not in this PR

PR #4035, opened 2024-12-19
(Improve order flow) Rework store case order service

islean:
## Description

This PR reworks the StoreCaseOrderService to use an OrderWithCases as input. It is not so beautiful, mainly because it handles orders of different types, such as balsamic and mip-dna which can have vastly different fields. Without a larger rework (say one storing service per order type), the workaround suggested here is to use `getattr` and pass the models as key word arguments.

### Added

-

### Changed

- The StoreCaseOrderService works with OrderWithCases inputs

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
Managed to submit a mip order:
image

islean:
To note - this PR does not include the familyID to LIMS patch

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -1,14 +1,18 @@
...
 from cg.services.orders.order_lims_service.order_lims_service import OrderLimsService
 from cg.services.orders.submitters.order_submitter import StoreOrderService
-from cg.store.models import ApplicationVersion, Case, CaseSample, Customer, Order, Sample
+from cg.store.models import Case as DbCase
+from cg.store.models import CaseSample, Customer
+from cg.store.models import Order as DbOrder
+from cg.store.models import Sample as DbSample
 from cg.store.store import Store
 
 LOG = logging.getLogger(__name__)
```
diitaz93, line 18: line 30: Should RNAFusion be part of the list too?

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -34,263 +38,182 @@ def __init__(
         self.status_db = status_db
         self.lims = lims_service
 
-    def store_order(self, order: OrderIn) -> dict:
+    def store_order(self, order: OrderWithCases) -> dict:
         """Submit a batch of samples for sequencing and analysis."""
         return self._process_case_samples(order=order)
```
diitaz93, line 43: I think this is an unnecessary wrapping. The `store_order` of the other services handle everything directly in their body

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -34,263 +38,182 @@ def __init__(
...
+            customer_internal_id=order.customer
         )
-        new_cases: list[Case] = []
-        status_db_order = Order(
-            customer=customer,
-            order_date=datetime.now(),
-            ticket_id=int(ticket_id),
+        new_cases: list[DbCase] = []
+        status_db_order = DbOrder(
+            customer=customer, order_date=datetime.now(), ticket_id=order._generated_ticket_id
```
diitaz93, line 71: Refactoring: consider wrapping this into `self._create_db_order` as in the fastq store service

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -34,263 +38,182 @@ def __init__(
...
+        return link
 
-    def _create_sample(self, case, customer_obj, order, ordered, sample, ticket):
-        sample_obj = self.status_db.add_sample(
-            name=sample["name"],
-            comment=sample["comment"],
-            control=sample["control"],
-            internal_id=sample["internal_id"],
-            order=order,
+    def _create_sample(
```
diitaz93, line 178: just for clarity and consistency with the other services
```suggestion
    def _create_db_sample(
```

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -34,263 +38,182 @@ def __init__(
...
 
-    def _create_case(self, case: dict, customer_obj: Customer, ticket: str):
-        case_obj = self.status_db.add_case(
-            cohorts=case["cohorts"],
-            data_analysis=Workflow(case["data_analysis"]),
-            data_delivery=DataDelivery(case["data_delivery"]),
-            name=case["name"],
-            priority=case["priority"],
-            synopsis=case["synopsis"],
+    def _create_case(
```
diitaz93, line 204: just for clarity and consistency with the other services
```suggestion
    def _create_db_case(
```

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -34,263 +38,182 @@ def __init__(
...
-        sample_obj.customer = customer_obj
+        db_sample.customer = customer
         with self.status_db.session.no_autoflush:
-            application_tag = sample["application"]
-            sample_obj.application_version: ApplicationVersion = (
-                self.status_db.get_current_application_version_by_tag(tag=application_tag)
+            application_tag = sample.application
+            db_sample.application_version = self.status_db.get_current_application_version_by_tag(
+                tag=application_tag
             )
```
diitaz93, line 200: the customer and application version can be added as parameter of the `self.status_db.add_sample` function

diitaz93: and the autoflush can be called in the outer function

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -34,263 +38,182 @@ def __init__(
...
+        link = self.status_db.relate_sample(
+            case=case,
+            sample=family_samples[sample.name],
+            status=getattr(sample, "status", None),
+            mother=mother,
+            father=father,
         )
-        self.status_db.session.add(link_obj)
-        return link_obj
+        self.status_db.session.add(link)
```
diitaz93, line 175: I have no strong opinion in adding to the db transaction inside the `_create` functions or outside, but whatever we choose we should be consistent along all storing services. In the ones that I made, I have included them outside the functions. Do we have any pros/cons with that?

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -34,263 +38,182 @@ def __init__(
...
-                )
+                    self._update_relationship(
+                        father=sample_father,
+                        link=case_sample,
+                        mother=sample_mother,
+                        sample=sample,
+                    )
             self.status_db.session.add_all(new_cases)
             self.status_db.session.add(status_db_order)
             self.status_db.session.commit()
```
diitaz93, line 134: @ChrOertlin suggested decoupling the storing services from the third-party db handling library (SQLAlchemy) by using the functions defined in `cg/store/base.py`. I think it is a good point

cg/meta/orders/ticket_handler.py
```diff
@@ -206,7 +206,7 @@ def create_case_xml_sample_list(self, order, message: str) -> str:
                             message=message, case_name=case.name
                         )
                         message = self.add_sample_priority_to_message(
-                            message=message, priority=sample.priority
+                            message=message, priority=case.priority
```
diitaz93, line 209: will this work with orders with samples?

cg/store/models.py
```diff
@@ -815,6 +815,14 @@ class Sample(Base, PriorityMixin):
         back_populates="sample", cascade="all, delete"
     )
 
+    def __init__(self, **kwargs):
```
diitaz93, line 818: why this?

tests/fixture_plugins/orders_fixtures/status_data_fixtures.py
```diff
@@ -128,18 +113,66 @@ def rml_status_data(
     return store_pool_order_service.order_to_status(order=order)
 
 
-@pytest.fixture
```
diitaz93, line 131: I did this in my branch too 😅

diitaz93 (approved):
Looks good  👍 I would push for some refactoring in the storing service but I leave that to your criteria

cg/services/orders/store_order_services/store_case_order.py
islean: But do they not commit already? This way we can revert the commit if it crashes down the line

cg/store/models.py
islean: Because otherwise it crashed when I passed keyword arguments to it which does not exist in the database model. So concentration and stuff which we only have in LIMS

cg/services/orders/store_order_services/store_case_order.py
islean: I can refactor it to the one you have

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -34,263 +39,192 @@ def __init__(
...
+                            order_name=order.name,
+                            ordered=datetime.now(),
+                            sample=sample,
+                            ticket=str(order._generated_ticket_id),
+                        )
+                    else:
+                        db_sample: DbSample = self.status_db.get_sample_by_internal_id(
+                            sample.internal_id
+                        )
+                    case_samples[db_sample.name] = db_sample
```
diitaz93, line 90: maybe all of this can be inside the `self._createdbd_case` function

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -34,263 +41,200 @@ def __init__(
...
+
+            case_sample: CaseSample = self._create_link(
+                case=db_case,
+                db_sample=db_sample,
+                father=db_sample_father,
+                mother=db_sample_mother,
+                sample=sample,
+            )
+
+            self._update_relationship(
```
islean, line 213: Should this method exist? When would it be used?

cg/services/orders/store_order_services/store_case_order.py
```diff
@@ -34,263 +41,189 @@ def __init__(
...
+            case_sample: CaseSample = self._create_link(
+                case=db_case,
+                db_sample=db_sample,
+                father=db_sample_father,
+                mother=db_sample_mother,
+                sample=sample,
+            )
+            self.status_db.add_item_to_store(case_sample)
+
+    def _create_db_sample_dict(
```
Vince-janv, line 209: why do we need this step? Is it to accommodate existing samples?

cg/store/models.py
Vince-janv: Hmm, I hope we don't override some SQLalchemy `init`. Let's see in the tests

Vince-janv (approved):
Nice! This should be the biggest one 👍

cg/services/orders/store_order_services/store_case_order.py
islean: Sort of - one could potentially redesign this from scratch, but we need data such as the internal ids for all samples when linking them to cases. The internal id only exists on the database model. So this creates a dict to house all database sample entries which is accessible via the sample names.

cg/services/orders/store_order_services/store_case_order.py
islean: removed it

cg/services/orders/store_order_services/store_case_order.py
islean: I think it would clutter the purpose of the create_db_case method. I am all for a future refactoring but I think it should be a new layer instead of incorporating it into create_db_case if so.

islean:
Tested:

- [x] MIP-DNA with a mixture of new and old samples
- [x] Balsamic order 
- [ ]

cg/services/orders/store_order_services/store_case_order.py
islean: We can discuss it in future PRs

cg/store/models.py
islean: Worked with submitting orders

PR #4037, opened 2024-12-19
refactor fastq and microbial fastq storing services

diitaz93:
## Description
Refactored the existing new storing services using a new template: implementing the addition of samples, cases and order in different hidden methods within the service. This improves readability and separation of concerns

### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/constants/orderforms.py
+++ b/cg/constants/orderforms.py
@@ -1,18 +1,10 @@
 from enum import StrEnum
 
 from cg.constants import ANALYSIS_SOURCES, METAGENOME_SOURCES
-from cg.models.orders.order import OrderType
 
 SEX_MAP = {"male": "M", "female": "F", "unknown": "unknown"}
 REV_SEX_MAP = {value: key for key, value in SEX_MAP.items()}
-CONTAINER_TYPES = ["Tube", "96 well plate"]
 SOURCE_TYPES = set().union(METAGENOME_SOURCES, ANALYSIS_SOURCES)
-
-CASE_PROJECT_TYPES = [
-    OrderType.MIP_DNA,
-    OrderType.BALSAMIC,
-    OrderType.MIP_RNA,
-]
 
 
 class Orderform(StrEnum):
--- a/cg/services/orders/store_order_services/store_fastq_order_service.py
+++ b/cg/services/orders/store_order_services/store_fastq_order_service.py
@@ -4,7 +4,7 @@ from datetime import datetime
 from cg.constants import DataDelivery, GenePanelMasterList, Priority, Workflow
 from cg.constants.constants import CustomerId
 from cg.constants.sequencing import SeqLibraryPrepCategory
-from cg.models.orders.sample_base import StatusEnum
+from cg.models.orders.sample_base import SexEnum, StatusEnum
 from cg.services.order_validation_service.workflows.fastq.models.order import FastqOrder
 from cg.services.order_validation_service.workflows.fastq.models.sample import FastqSample
 from cg.services.orders.order_lims_service.order_lims_service import OrderLimsService
@@ -99,7 +99,7 @@ class StoreFastqOrderService(StoreOrderService):
         )
         return self.status_db.add_sample(
             name=sample.name,
-            sex=sample.sex or "unknown",
+            sex=sample.sex or SexEnum.unknown,
             comment=sample.comment,
             internal_id=sample._generated_lims_id,
             ordered=datetime.now(),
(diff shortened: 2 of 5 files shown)


cg/constants/orderforms.py
```diff
@@ -1,19 +1,11 @@
...
 
 from cg.constants import ANALYSIS_SOURCES, METAGENOME_SOURCES
-from cg.models.orders.order import OrderType
 
 SEX_MAP = {"male": "M", "female": "F", "unknown": "unknown"}
 REV_SEX_MAP = {value: key for key, value in SEX_MAP.items()}
-CONTAINER_TYPES = ["Tube", "96 well plate"]
 SOURCE_TYPES = set().union(METAGENOME_SOURCES, ANALYSIS_SOURCES)
 
-CASE_PROJECT_TYPES = [
```
diitaz93, line 11: The removal of these unused constants fixed a circular import error

islean: They seem misleading either way

cg/services/orders/store_order_services/store_fastq_order_service.py
```diff
@@ -33,84 +33,109 @@ def store_order(self, order: FastqOrder) -> dict:
...
+        self, sample: FastqSample, order_name: str, customer: Customer, ticket_id: str
+    ) -> Sample:
+        """Return a Sample database object."""
+        application_version: ApplicationVersion = (
+            self.status_db.get_current_application_version_by_tag(tag=sample.application)
         )
-        status_db_order: Order = self.status_db.add_order(
+        return self.status_db.add_sample(
+            name=sample.name,
+            sex=sample.sex or "unknown",
```
islean, line 102:
```suggestion
            sex=sample.sex or SexEnum.UNKNOWN,
```

cg/store/crud/create.py
```diff
@@ -398,12 +397,13 @@ def add_organism(
...
-    def add_order(self, customer: Customer, ticket_id: int) -> Order:
+    def add_order(self, customer: Customer, ticket_id: int, **kwargs) -> Order:
         """Build a new Order record."""
         order = Order(
             customer=customer,
             order_date=datetime.now(),
             ticket_id=ticket_id,
+            **kwargs,
         )
         return order
```
ChrOertlin, line 408: this only creates an order object but does not add it to the DB. this misses the self.session.add()

cg/services/orders/store_order_services/store_fastq_order_service.py
```diff
@@ -33,84 +33,109 @@ def store_order(self, order: FastqOrder) -> dict:
...
-        self.status_db.session.add_all(new_samples)
-        self.status_db.session.commit()
-        return new_samples
+
+    def _create_maf_case(self, db_sample: Sample, db_order: Order) -> None:
+        """
+        Add a MAF case to the current Status database transaction only if the given sample is
+        non-tumour and  WGS. Return None otherwise without adding anything to the transaction.
+        This function does not commit to the database.
+        """
```
islean, line 121: The method seems to be both creating and linking the case so maybe we should mention that in the docstring. Also - would we want to call it "ensure_maf_case" or something given that it might not add anything?
```suggestion
    def _create_maf_case(self, db_sample: Sample, db_order: Order) -> None:
        """
        Add a MAF case to the current Status database transaction and link it to the given sample, but only if the given sample is
        non-tumour and  WGS. 
        This function does not commit to the database.
        """
```

ChrOertlin (approved):
Looks good. Just a comment regarding teh crud function add_order(). Cn be out of scope for this PR. But I highlighted this before. Our add_ functions are not creating anything in the database they merely return objects at this time.

cg/store/crud/create.py
```diff
@@ -398,12 +397,13 @@ def add_organism(
             **kwargs,
         )
 
-    def add_order(self, customer: Customer, ticket_id: int) -> Order:
+    def add_order(self, customer: Customer, ticket_id: int, **kwargs) -> Order:
```
islean, line 400: Why did this change?

islean (approved):
Super nice!

cg/store/crud/create.py
diitaz93: I think all our Create functions return only the object without adding them to the transaction. So I tried to make this function as similar as the others. We should discuss if we want these functions to add to the transaction too.

cg/store/crud/create.py
diitaz93: I made this function more similar to the others in this module (specially add_sample). This allows you to add any valid argument to the model as parameter to the function. Not currently used as we only use ticket and customer to initialise Orders, but I think this could be used in the future

PR #4045, opened 2025-01-02
(Improve order flow) Fix ExistingSample

islean:
## Description

Existing sample has Status as a mandatory field, even though some order types supporting reruns can't provide it. This PR makes it optional.

### Added

-

### Changed

-

### Fixed

- Status is an optional field for existing samples

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/order_validation_service/models/existing_sample.py
+++ b/cg/services/order_validation_service/models/existing_sample.py
@@ -7,7 +7,7 @@ class ExistingSample(BaseModel):
     father: str | None = Field(None, pattern=NAME_PATTERN)
     internal_id: str
     mother: str | None = Field(None, pattern=NAME_PATTERN)
-    status: StatusEnum
+    status: StatusEnum | None = None
 
     @property
     def is_new(self) -> bool:


ChrOertlin:
shouldn't the existingsample diverge into one that supports it and one that doesn't? I think we need to enforce status for the samples that require it. would this not make it so that existing samples requiring a status now can pass without it in theory?

islean:
> shouldn't the existingsample diverge into one that requires a status and one that doesn't? I think we need to enforce status for the samples that require it. would this not make it so that existing samples requiring a status now can pass without it in theory?

We could definitely consider creating order type specific existing samples, but I am uncertain of whether it is worth it. In part, I do not know if we can perform any validation since the criteria have changed through CG's history; and in part because there are very few fields one can set for existing samples. The fields one can set for existing samples are only those found in the CaseSample table, which boils down to mother, father and status.

I am open to changing if it aids reading the code but when SA and I designed this we deemed it much ado for a very tiny biscuit.

ChrOertlin (approved):

ChrOertlin:
> > shouldn't the existingsample diverge into one that requires a status and one that doesn't? I think we need to enforce status for the samples that require it. would this not make it so that existing samples requiring a status now can pass without it in theory?
> 
> We could definitely consider creating order type specific existing samples, but I am uncertain of whether it is worth it. In part, I do not know if we can perform any validation since the criteria have changed through CG's history; and in part because there are very few fields one can set for existing samples. The fields one can set for existing samples are only those found in the CaseSample table, which boils down to mother, father and status.
> 
> I am open to changing if it aids reading the code but when SA and I designed this we deemed it much ado for a very tiny biscuit.

alright, I understand. It just opens up room for errors in the validation but if that is something that was considered before lets go ahead!

Issue #4044, opened 2025-01-02
Bug(RNA scout upload) Wrong customer used 


Vince-janv:
## Description
Bug report: https://github.com/Clinical-Genomics/bug-reports/issues/27

The upload of RNA-cases to existing DNA-cases uses the wrong customer for some of it's update calls. Instead of using the customer from the target case (DNA) it uses the one from the source case (RNA)

Vince-janv:
Solved by https://github.com/Clinical-Genomics/cg/pull/4046

PR #4046, opened 2025-01-02
Fix(rna-upload)

Vince-janv:
## Description
Solves #4044

### Fixed

- Use correct customer when uploading RNA-omics results to scout


islean (approved):
Nice catch!

PR #4049, opened 2025-01-03
fix(NFAnalysis) Prevent variable being overriden

Vince-janv:
## Description
See issue and bug report for full description. This PR renames the attribute self.config on NFAnalysisAPIs to avoid overriding the one set on the parent class.

### Fixed

- renamed NFAnalysisAPI.config to NFAnalysisAPI.workflow_config_path

### DIFF
--- a/cg/meta/workflow/nf_analysis.py
+++ b/cg/meta/workflow/nf_analysis.py
@@ -63,7 +63,7 @@ class NfAnalysisAPI(AnalysisAPI):
         self.conda_binary: str | None = None
         self.platform: str | None = None
         self.params: str | None = None
-        self.config: str | None = None
+        self.workflow_config_path: str | None = None
         self.resources: str | None = None
         self.tower_binary_path: str | None = None
         self.tower_workflow: str | None = None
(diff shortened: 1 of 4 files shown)


Vince-janv:
Successfully, yet painfully tested on stage...

cg/meta/workflow/nf_analysis.py
```diff
@@ -63,7 +63,7 @@ def __init__(self, config: CGConfig, workflow: Workflow):
         self.conda_binary: str | None = None
         self.platform: str | None = None
         self.params: str | None = None
-        self.config: str | None = None
```
islean, line 66: Very sneaky! Is this only used in the `get_nextflow_config_content`? Just trying to understand the need for this parameter

Vince-janv: Indeed. It's hellish to search for since it is not a defined parameter on the class, but from what i gathered it's only used there

islean (approved):
Looks good as far as I can tell! We should be ready to revert if there are unforeseen consequences though.

PR #4055, opened 2025-01-07
Fix(Dockerfile) Make compatible with new poetry version

Vince-janv:
## Description
Poetry version 2.0.0 was released recently, converting some previous warnings to errors. This causes our dockerfile to fail when installing the dependencies using poetry.

### Fixed
- Copies README file into container
- Split install of requirements our of install of actual cg-repo

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ ENV TRAILBLAZER_SERVICE_ACCOUNT_AUTH_FILE="auth_file"
 
 
 WORKDIR /home/src/app
-COPY pyproject.toml poetry.lock gunicorn.conf.py README.md ./
+COPY pyproject.toml poetry.lock gunicorn.conf.py ./
 
 RUN pip install --no-cache-dir poetry \
     && poetry config virtualenvs.create false \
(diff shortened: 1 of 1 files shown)


henrikstranneheim (approved):
👍

Dockerfile
```diff
@@ -23,14 +23,16 @@ ENV TRAILBLAZER_SERVICE_ACCOUNT_AUTH_FILE="auth_file"
...
+COPY pyproject.toml poetry.lock gunicorn.conf.py README.md ./
 
 RUN pip install --no-cache-dir poetry \
     && poetry config virtualenvs.create false \
-    && poetry install --no-interaction --no-ansi
+    && poetry install --no-interaction --no-ansi --no-root
 
 COPY cg ./cg
 
+RUN poetry install --no-interaction --no-ansi
```
islean, line 34: Why do we run poetry install twice here?

Vince-janv: Indeed. The first one installs all the dependencies but not the actual repo (--no-root). The second one installs the cg. This optimises the docker layering so that a one does not need to reinstall all dependencies every time a file in the cg repo changes

diitaz93 (approved):
👍

PR #4079, opened 2025-01-14
fix(Vulnerability) Update virtualenv

Vince-janv:
## Description

Resolves https://github.com/Clinical-Genomics/cg/security/dependabot/35


diitaz93 (approved):
👍

PR #4098, opened 2025-01-15
add nallo cg workflow config-case

peterpru:
## Description
Add nallo config-case. This generates 3 files needed to start the analysis in nextflow.

For the reviewer: this is the same as for raredisease, except for three things. 
1) For nallo we need bam file paths as input for the sample sheet instead of fastq files. In this PR the bam file paths are collected from housekeeper.
2) For the nallo pipeline we need the sample sheet to have maximum one bam file per line. Instead of having having one sample per line in the samplsheet, we can have multiple. Here the all bam file paths are collected, and for each bam file entry, it will create a sample entry in the sample sheet.
3) no resolve-decompression. Input is bamfiles, so we can skip the entire spring->fastq checking and decompressing.

### Added

- cg workflow nallo config-case

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b add-nallo-config-case
    ```

### How to test

- [x] Do `cg workflow nallo config-case rightalpaca`

### Expected test outcome

- [x] The command should create a rightalpaca_samplesheet.csv, params file and config.yaml
- [x] Files should have content, where params and config are populated from servers, and samplesheet from cg.
image

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


tests/cli/workflow/nf_analysis/test_cli_config_case.py
```diff
@@ -24,7 +24,7 @@
 
 @pytest.mark.parametrize(
     "workflow",
-    NEXTFLOW_WORKFLOWS,
+    NEXTFLOW_WORKFLOWS + [Workflow.NALLO],
```
Vince-janv, line 27: I guess this is temporary, to avoid breaking other parametrized tests?

cg/meta/workflow/nallo.py
```diff
@@ -18,3 +22,72 @@ def __init__(
...
+                sample=case_sample.sample.internal_id,
+                bam_unmapped_read_path=Path(bam_path),
+                family_id=case_sample.case.internal_id,
+                paternal_id=case_sample.get_paternal_sample_id or "0",
+                maternal_id=case_sample.get_maternal_sample_id or "0",
+                sex=self.get_sex_code(case_sample.sample.sex),
+                phenotype=self.get_phenotype_code(case_sample.status),
+            )
+            sample_sheet_entries.extend(sample_sheet_entry.reformat_sample_content)
+        return sample_sheet_entries
```
Vince-janv, line 64: In this context can, can we call it `read_paths` or `read_files`? I feel like unmapped_bam is a detail that can remain in the method that fetches it, but is not relevant in the wider context

cg/meta/workflow/nf_analysis.py
```diff
@@ -252,6 +253,15 @@ def get_paired_read_paths(self, sample=Sample) -> tuple[list[str], list[str]]:
...
         return fastq_forward_read_paths, fastq_reverse_read_paths
 
+    def get_unmapped_bam_read_paths(self, sample=Sample) -> list[str]:
+        """Gather BAM file path for a sample based on the BAM tag."""
+        return [
+            hk_file.full_path
+            for hk_file in self.housekeeper_api.files(
+                bundle=sample.internal_id, tags={AlignmentFileTag.BAM}
+            )
+        ]
```
Vince-janv, line 263: In general i like working with `Path` objects instead of strings in these cases. Is that an easy fix or does it require further changes?

```suggestion
        return [
            Path(hk_file.full_path)
            for hk_file in self.housekeeper_api.files(
                bundle=sample.internal_id, tags={AlignmentFileTag.BAM}
            )
        ]
```

Vince-janv (approved):
Nice 👍

tests/cli/workflow/nf_analysis/test_cli_config_case.py
peterpru: Yes, correct. If I add it to NEXTFLOW_WORKFLOWS, it tests all nf-tests, including e.g. metrics-deliver.

cg/meta/workflow/nallo.py
peterpru: sounds good 👍 changed to read_file, and the method that fetches it to get_bam_read_file_paths.

cg/meta/workflow/nf_analysis.py
peterpru: will test now 👍

peterpru:
Thanks for the review @Vince-janv 

Deployed in stage:
image

Deployed in production:
image

PR #4170, opened 2025-01-30
Fix Pydantic error mapping

islean:
## Description

We have a bug in how we map the Pydantic errors to the validate_order response. The "loc" property of the Pydantic error makes lists incompatible with the current handling.

### Added

-

### Changed

-

### Fixed

- Determining the type of order/case/sample/case_sample error is not by length of loc

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/orders/validation/model_validator/utils.py
```diff
@@ -122,48 +129,54 @@ def get_order_error_details(error_details: list[ErrorDetails]) -> list[ErrorDeta
...
 
 
 def is_case_error(error: ErrorDetails) -> bool:
-    return len(error["loc"]) == 4 and error["loc"][0] == "cases"
+    return "cases" in error["loc"] and "samples" not in error["loc"]
 
 
 def is_case_sample_error(error: ErrorDetails) -> bool:
-    return len(error["loc"]) == 7
+    return "cases" in error["loc"] and "samples" in error["loc"]
```
Vince-janv, line 140: Why do we need to user an in-comparison here compared to just `=`?

cg/services/orders/validation/model_validator/utils.py
```diff
@@ -122,48 +129,54 @@ def get_order_error_details(error_details: list[ErrorDetails]) -> list[ErrorDeta
...
+    return error["loc"][index_for_field_name]
 
 
 def get_order_field_name(error: ErrorDetails) -> str:
     return error["loc"][0]
 
 
 def get_sample_index(error: ErrorDetails) -> int:
-    return error["loc"][1]
+    index_for_index: int = error["loc"].index("samples") + 1
```
Vince-janv, line 171: 🤯

cg/services/orders/validation/model_validator/utils.py
```diff
@@ -122,48 +129,54 @@ def get_order_error_details(error_details: list[ErrorDetails]) -> list[ErrorDeta
...
 
 def get_case_field_name(error: ErrorDetails) -> str:
-    return error["loc"][3]
+    index_for_field_name: int = error["loc"].index("cases") + 3
+    return error["loc"][index_for_field_name]
 
 
 def get_case_sample_field_name(error: ErrorDetails) -> str:
-    return error["loc"][6]
+    index_for_field_name: int = error["loc"].index("samples") + 3
```
Vince-janv, line 162: Under what circumstances would these magic ints change? Ie the 3 here?

Vince-janv (commented):
Still not sure I understand it all 💫

cg/services/orders/validation/model_validator/utils.py
diitaz93: `error["loc"]` is tuple and we can't assure the position where "cases" or "samples"  will be, because some cases have a discriminant placed in the middle, some other don't

cg/services/orders/validation/model_validator/utils.py
```diff
@@ -122,48 +129,54 @@ def get_order_error_details(error_details: list[ErrorDetails]) -> list[ErrorDeta
...
 
 
 def is_case_sample_error(error: ErrorDetails) -> bool:
-    return len(error["loc"]) == 7
+    return "cases" in error["loc"] and "samples" in error["loc"]
 
 
 def is_order_error(error: ErrorDetails) -> bool:
-    return len(error["loc"]) == 1
+    return error["loc"][0] not in ["cases", "samples"]
```
diitaz93, line 144: will this be more stringent? not sure of how the error will look for an order
```suggestion
    return error["loc"][0] == "order"
```

cg/services/orders/validation/model_validator/utils.py
```diff
@@ -122,48 +129,54 @@ def get_order_error_details(error_details: list[ErrorDetails]) -> list[ErrorDeta
...
 
 def get_sample_field_name(error: ErrorDetails) -> str:
-    return error["loc"][2]
+    index_for_field_name: int = error["loc"].index("samples") + 2
+    return error["loc"][index_for_field_name]
 
 
 def get_case_field_name(error: ErrorDetails) -> str:
-    return error["loc"][3]
+    index_for_field_name: int = error["loc"].index("cases") + 3
```
diitaz93, line 157: what if the case has no tag? wouldn't then be 2?

cg/services/orders/validation/model_validator/utils.py
diitaz93, line 171:
```suggestion
    index_for_sample_index: int = error["loc"].index("samples") + 1
```

diitaz93: feel free to disregard

cg/services/orders/validation/model_validator/utils.py
```diff
@@ -122,48 +129,54 @@ def get_order_error_details(error_details: list[ErrorDetails]) -> list[ErrorDeta
...
 
 def get_case_index(error: ErrorDetails) -> int:
-    return error["loc"][1]
+    index_for_index: int = error["loc"].index("cases") + 1
+    return error["loc"][index_for_index]
 
 
 def get_case_sample_index(error: ErrorDetails) -> int:
-    return error["loc"][4]
+    index_for_index: int = error["loc"].index("samples") + 1
```
diitaz93, line 181: maybe the existence of this function is good for clarity but I guess it can now be replaced by `get_sample_index`

cg/services/orders/validation/model_validator/utils.py
islean: I mean, I think we probably can say that `error["loc"][0]` should be "cases" and `error["loc"][3]` should be samples. Is that preferable? As I understand it, we will always have discriminants in there for case orders

cg/services/orders/validation/model_validator/utils.py
islean: Very confusing name 😓 Better suggestions welcome!

cg/services/orders/validation/model_validator/utils.py
islean: For Case orders I do not see why it would not be. All of them support additions of existing samples

cg/services/orders/validation/model_validator/utils.py
islean: If it is an order error the first entry in "loc" will be the field name. So if the order.name is missing, the loc will just be ("name",) or something.

cg/services/orders/validation/model_validator/utils.py
islean: I think the case should always have a tag

cg/services/orders/validation/model_validator/utils.py
islean: Using `get_sample_index` will do the same thing. I do like having this split, but I am also open to removing the get_case_sample_index method if that is preferred

cg/services/orders/validation/model_validator/utils.py
diitaz93: then resolve

cg/services/orders/validation/model_validator/utils.py
diitaz93: keep it if you want

diitaz93 (approved):
you comments brought more clarity 🌟

cg/services/orders/validation/model_validator/utils.py
Vince-janv: Ok, I got a clearer picture from reading the doc-string 👍

cg/services/orders/validation/model_validator/utils.py
Vince-janv: Ok, so it's a tuple with an order which we decide. I was worried it was related to the order of fields received in the request

cg/services/orders/validation/model_validator/utils.py
Vince-janv: I think it makes sense. Just very meta 😉

Vince-janv (approved):
👌

islean:
Deployed to the VMs

Issue #4173, opened 2025-01-31
LIMS strips spaces from container names


Vince-janv:
### Description
LIMS seems to strip all fields of trailing blank spaces. This causes `cg/apps/lims/order.py:79` to crash since the key we're using may contain a trailing blank space but in the dict it has been stripped. Resulting in a KeyError

PR #4174, opened 2025-01-31
Fix(Orders) Add pattern for container_name

Vince-janv:
## Description
Solves [LIMS strips spaces from container names #4173](https://github.com/Clinical-Genomics/cg/issues/4173).

This PR adds validation that the container name follows the same pattern as the sample name.

### Changed

- The container_name field now needs to adhere to the pattern ^[A-Za-z0-9_]*$


Karl-Svard (approved):
I love it!

islean (approved):
💯

Vince-janv:
Deployed in production

Issue #4161, opened 2025-01-28
New endpoints needed for fetching PacBio sequencing data


Karl-Svard:
As a LIMS developer,
I want new endpoints in cg for fetching PacBio data,
So that we can set up post-sequencing QC checks in LIMS.

# Work impact

Answer the following questions:
- Is there currently a workaround for this issue? If so, what is it?
    - Nope.
- How much time would be saved by implementing this feature on a weekly basis?
    - Not a time-saving issue.
- How many users are affected by this issue?
    - All LIMS developers and by proxy prod bioinfo
- Are customers affected by this issue?
    - Nope.

# Acceptance Criteria

- [ ] GET `/pacbio_sequencing_run/{run_name}` returns `list[PacbioSequencingRun]`
- [x] GET `/pacbio_sample_sequencing_metrics?sample=<sample_id>&smrt_cell=<cell_id>` returns `list[PacbioSampleSequencingMetrics]`

# Notes

- Additional information.
- Dependencies.
- Related user stories.

Karl-Svard:
@Vince-janv would love for this feature request to be bumped a bit if possible :)

Vince-janv:
What will the post-sequencing QC checks in LIMS result in? Ie what decisions will be made based on the results?

Karl-Svard:
This will both give us insight into the sequencing performance per run and tell us if the samples have gotten enough data for their given apptag or if they would require further sequencing

Vince-janv:
> This will both give us insight into the sequencing performance per run and tell us if the samples have gotten enough data for their given apptag or if they would require further sequencing

But is this not what we decided not to implement yet, since we don't have the criteria for "OK" sequencing?

Karl-Svard:
The keyword here is "yet". But we already know what yield each apptag is targeting for, and I assume this logic will be needed for the nallo implementation as well which is close to being in place. 

What's your biggest concern for adding it now? Lack of development resources or simply the addition of new features/logic to cg?

Vince-janv:
> The keyword here is "yet". We already know what yield each apptag is targeting for, and I assume this logic will be needed for the nallo implementation as well which is close to being in place.
> 
> What's your biggest concern for adding it now? Lack of development resources or simply the addition of new features/logic to cg?

[...]

PR #4180, opened 2025-02-05
Add pacbio sample sequencing metrics  endpoint

islean:
## Description

Resolves part 1 of #4161 .

### Added

- Endpoint for fetching Pacbio sample sequencing metrics

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/store/crud/read.py
```diff
@@ -1790,3 +1793,21 @@ def get_related_dna_cases_with_samples(self, rna_case: Case) -> list[RNADNAColle
...
+        sequencing_metrics: Query = (
+            self._get_query(table=PacbioSampleSequencingMetrics)
+            .join(Sample)
+            .join(InstrumentRun)
+            .join(RunDevice)
+        )
+        if sample_id:
+            sequencing_metrics = sequencing_metrics.filter(Sample.internal_id == sample_id)
+        if smrt_cell_id:
+            sequencing_metrics = sequencing_metrics.filter(RunDevice.internal_id == smrt_cell_id)
```
islean, line 1812: I will start a discussion here: We are currently enforcing a pattern when filtering queries where we use the `apply_sample_filter` methods for example. This does not follow said logic currently, which is by choice, since I do not find the logic above hard to read at all. Meanwhile, following the pattern comes with a number of issues:

1. It is table specific while we have started to introduce polymorphism. So for example you would have the PacbioSampleSequencingMetrics which inherits from SampleRunMetrics, but would you need to different methods and two different enums with filters - one called apply_sample_run_metrics_filter and one called apply_pacbio_sample_sequencing_metrics_filter? That seems to work against the benefits of the polymorphism.
2. It is not at all obvious to me what we gain from the current pattern. If we just want to group all the filters in the same place, we could have discussion about that. But what is the benefit to having a common entrypoint to all filters?

cg/store/crud/read.py
```diff
@@ -1790,3 +1793,21 @@ def get_related_dna_cases_with_samples(self, rna_case: Case) -> list[RNADNAColle
...
+    ) -> list[PacbioSampleSequencingMetrics]:
+        """
+        Fetches data from PacbioSampleSequencingMetrics filtered on sample_internal_id and/or smrt_cell_id.
+        """
+        sequencing_metrics: Query = (
+            self._get_query(table=PacbioSampleSequencingMetrics)
+            .join(Sample)
+            .join(InstrumentRun)
+            .join(RunDevice)
+        )
```
islean, line 1808: I guess this one should be in base.py as well. But I have a hard time seeing the benefit of those methods as well. To me it adds some obscurity having `self.get_join_pacbio_sample_sequencing_metrics_sample_instrument_run_run_device` here and the names will only get more opaque by more polymorphism.

cg/server/endpoints/sequencing_metrics/pacbio_sequencing_metrics.py
```diff
@@ -0,0 +1,29 @@
...
+from cg.services.sample_run_metrics_service.dtos import PacbioSequencingMetrics
+
+PACBIO_SAMPLE_SEQUENCING_METRICS_BLUEPRINT = Blueprint(
+    "pacbio_sample_sequencing_metrics", __name__, url_prefix="/api/v1"
+)
+PACBIO_SAMPLE_SEQUENCING_METRICS_BLUEPRINT.before_request(before_request)
+
+
+@PACBIO_SAMPLE_SEQUENCING_METRICS_BLUEPRINT.route("/pacbio_sample_sequencing_metrics")
+@is_public
```
diitaz93, line 19: why is this one public and not `/flowcells/<flow_cell_name>/sequencing_metrics`?

islean: Woopsie! For testing purposes, but it definitely should not be. Thank you!

cg/services/sample_run_metrics_service/sample_run_metrics_service.py
diitaz93, line 1: I created a subfolder in `services` for this kind of services called `web_services`, maybe you can move this module there?

cg/server/endpoints/sequencing_metrics/pacbio_sequencing_metrics.py
islean: Made it private again

cg/services/sample_run_metrics_service/sample_run_metrics_service.py
```diff
@@ -1,15 +1,40 @@
...
+        )
+        for db_metric in sequencing_metrics:
+            metric = PacbioSequencingMetrics(
+                hifi_mean_read_length=db_metric.hifi_mean_read_length,
+                hifi_median_read_quality=db_metric.hifi_median_read_quality,
+                hifi_reads=db_metric.hifi_reads,
+                hifi_yield=db_metric.hifi_yield,
+                sample_id=db_metric.sample.internal_id,
+                smrt_cell_id=db_metric.instrument_run.device.internal_id,
+            )
```
diitaz93, line 38: This is super minor, resolve if you disagree: one object of type `PacbioSampleSequencingMetrics` corresponds to many metrics of one sample, not a single metric. Maybe this renaming brings some clarity on that matter
```suggestion
        for sample_metrics in sequencing_metrics:
            metric = PacbioSequencingMetrics(
                hifi_mean_read_length=sample_metrics.hifi_mean_read_length,
                hifi_median_read_quality=sample_metrics.hifi_median_read_quality,
                hifi_reads=sample_metrics.hifi_reads,
                hifi_yield=sample_metrics.hifi_yield,
                sample_id=sample_metrics.sample.internal_id,
                smrt_cell_id=sample_metrics.instrument_run.device.internal_id,
            )
```

tests/services/sample_run_metrics_service/test_sample_metrics_service.py
```diff
@@ -10,7 +11,70 @@ def test_get_metrics_for_flow_cell(
...
     # THEN the metrics should be returned
     assert metrics
+
+
+def test_get_pacbio_metrics(sample_run_metrics_service: SampleRunMetricsService):
+    # GIVEN a SampleRunMetricsService and a database containing Pacbio data
+
+    # WHEN fetching all PacbioSampleSequencingMetrics
+    metrics_request = PacbioSequencingMetricsRequest()
+    metrics = sample_run_metrics_service.get_pacbio_metrics(metrics_request)
```
diitaz93, line 25: add type hint

tests/services/sample_run_metrics_service/test_sample_metrics_service.py
```diff
@@ -10,7 +11,70 @@ def test_get_metrics_for_flow_cell(
...
+
+    # THEN metrics should be returned
+    assert metrics
```
diitaz93, line 28: can we assert something on the content?

tests/services/sample_run_metrics_service/test_sample_metrics_service.py
```diff
@@ -10,7 +11,70 @@ def test_get_metrics_for_flow_cell(
...
+    sample_run_metrics_service: SampleRunMetricsService, pacbio_barcoded_sample_internal_id: str
+):
+    # GIVEN a SampleRunMetricsService and a database containing Pacbio data
+
+    # WHEN fetching a specific PacbioSampleSequencingMetrics
+    metrics_request = PacbioSequencingMetricsRequest(sample_id=pacbio_barcoded_sample_internal_id)
+    metrics = sample_run_metrics_service.get_pacbio_metrics(metrics_request)
+
+    # THEN metrics should be returned
+    assert metrics
```
diitaz93, line 41: I would say add type hint and assert something on the content of the metrics in this test and remove the previous one

tests/services/sample_run_metrics_service/test_sample_metrics_service.py
```diff
@@ -10,7 +11,70 @@ def test_get_metrics_for_flow_cell(
...
+    sample_run_metrics_service: SampleRunMetricsService,
+):
+    # GIVEN a SampleRunMetricsService and a database containing Pacbio data
+
+    # WHEN fetching a specific PacbioSampleSequencingMetrics
+    metrics_request = PacbioSequencingMetricsRequest(sample_id="I do not exist")
+    metrics = sample_run_metrics_service.get_pacbio_metrics(metrics_request)
+
+    # THEN metrics should be returned
+    assert not metrics
```
diitaz93, line 54: it is unclear if this will return `None` or an empty list from this. Maybe we want the test to be strict on that to catch unwanted modifications in the code

tests/services/sample_run_metrics_service/test_sample_metrics_service.py
```diff
@@ -10,7 +11,70 @@ def test_get_metrics_for_flow_cell(
...
+    sample_run_metrics_service: SampleRunMetricsService,
+):
+    # GIVEN a SampleRunMetricsService and a database containing Pacbio data
+
+    # WHEN fetching a specific PacbioSampleSequencingMetrics
+    metrics_request = PacbioSequencingMetricsRequest(smrt_cell_id="internal_id")
+    metrics = sample_run_metrics_service.get_pacbio_metrics(metrics_request)
+
+    # THEN metrics should be returned
+    assert metrics
```
diitaz93, line 67: same as above

diitaz93 (approved):
Looks good, left some comments

cg/services/sample_run_metrics_service/sample_run_metrics_service.py
islean: It is better. I will change and the indented variable as well

cg/store/crud/read.py
islean: But `pacbio_model_join` would be a very confusing name for me. If we have such a method, I would want to understand what is joined to what.

cg/services/sample_run_metrics_service/sample_run_metrics_service.py
islean: I feel a bit hesitant, mainly because it feels like that would imply a restriction which does not exist. Although it is currently only used in the endpoints I added, there is nothing in the service indicating that its purpose is web specific. Can I move it if I add anything to the service which actually makes it more suitable for only web use, and leave it in the current directory if not?

cg/services/sample_run_metrics_service/dtos.py
```diff
@@ -1,9 +1,18 @@
 from pydantic import BaseModel
 
 
-class SequencingMetrics(BaseModel):
+class IlluminaSequencingMetrics(BaseModel):
```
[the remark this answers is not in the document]
islean: Is this not a bit doubling up? Or we are doing some work with simplifying imports to make them cleaner and not repeat themselves, and here we have it in the DTO directory already.

tests/services/sample_run_metrics_service/test_sample_metrics_service.py
islean: I would say so - it tests that supplying no parameters means no filtering, rather than filtering on nothing

cg/services/sample_run_metrics_service/dtos.py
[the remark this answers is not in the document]
diitaz93: I think those are two different things. The file where the dtos live may be called `dto.py` and the class called `DTO`, as a service class lives in a file called `service.py` and is called `Service`.  This suggestions comes from me not having clear which is the database model and which is the DTO, as their name is almost the same. adding DTO at the end will make itmuch clearer I think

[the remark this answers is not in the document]
islean: Makes sense. Another why it might make sense to import packages instead of classes at times though.

tests/services/sample_run_metrics_service/test_sample_metrics_service.py
diitaz93: then check something in the content

cg/store/crud/read.py
islean: I will bring up a discussion on us adding some SQL views which I think could remedy the situation (or maybe it is just moving the discussion). I will resolve for now though

islean:
Deployed to the VMs

PR #4182, opened 2025-02-05
Create one case per sample in fastq orders

### LINKED ISSUE #4181: One Case Per Sample for FASTQ
As a user assigned to the pipeline FASTQ,
I want one case per sample,
So that the order-view in Trailblazer would tell me if a sample failed on amount of reads.

# Acceptance Criteria

- [ ] For each sample, generate one case in any ticket that includes an app-tag related to FASTQ delivery.

...

# Notes

- Additional information.
- Dependencies.
- Related user stories.

# Implementation plan

- [ ] Work item 1
- [ ] Work item 2

diitaz93:
## Description
Closes #4181 
Modify the Fastq storing service to create one case per sample in the cg database, instead of one case containing all samples.

### Changed

- Method `store_order_data_in_status_db` from `StoreFastqOrderService`
- The unit test of the method above

### How to test

- [x] Submit a Fastq order in stage

### Expected test outcome

- [x] Check that one case per sample is created with the correct name

## Review

- [x] Tests executed by SD

## Implementation Plan

- [ ] Deployed to stage:
```shell

```
- [ ] Deployed to production:
```shell

```

### DIFF
--- a/cg/services/orders/storing/implementations/fastq_order_service.py
+++ b/cg/services/orders/storing/implementations/fastq_order_service.py
@@ -45,16 +45,18 @@ class StoreFastqOrderService(StoreOrderService):
         The stored data objects are:
         - Order
         - Samples
-        - One Case containing all samples
-        - For each Sample, a relationship between the sample and the Case
+        - One Case per sample
+        - For each Sample, a relationship between the sample and its Case
         - For each non-tumour WGS Sample, a MAF Case and a relationship between the Sample and the
         MAF Case
         """
         db_order: Order = self._create_db_order(order=order)
-        db_case: Case = self._create_db_case(order=order, db_order=db_order)
         new_samples = []
         with self.status_db.session.no_autoflush:
             for sample in order.samples:
+                db_case: Case = self._create_db_case_for_sample(
+                    sample=sample, order=order, customer=db_order.customer
+                )
                 db_sample: Sample = self._create_db_sample(
                     sample=sample,
                     order_name=order.name,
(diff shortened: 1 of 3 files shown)


cg/services/orders/storing/implementations/fastq_order_service.py
```diff
@@ -80,17 +82,20 @@ def _create_db_order(self, order: FastqOrder) -> Order:
         )
         return self.status_db.add_order(customer=customer, ticket_id=ticket_id)
 
-    def _create_db_case(self, order: FastqOrder, db_order: Order) -> Case:
+    def _create_db_case_for_sample(
+        self, sample: FastqSample, order: FastqOrder, customer: Customer
+    ) -> Case:
         """Return a Case database object."""
-        priority: str = order.samples[0].priority
+        case_name: str = f"{sample.name}-case"
```
islean, line 89: I dislike these naming patterns since the constraints on case names are different from those on samples. Could we have something which guarantees uniqueness - maybe sample.name-ticket_nr or something with a randomly generated string in it?

diitaz93: sure, I will implement

diitaz93: This pattern is used in Microbial fastq, do you think it is important to change it there too? @islean

islean: I would see it as an improvement but I do not know if it a decision I can take on my own

ChrOertlin: What about sample_internal_id-case ?

ChrOertlin (approved):
Looks good! Not a fan of the tiket id in the case-name, considering @islean suggestion about uniqueness and name constraints perhaps sample_internal_id-case is unique and more explicit what this case is for.

cg/services/orders/storing/implementations/fastq_order_service.py
islean: It should work but would block any support of reruns of fastq samples but it would also cause issues if we would decide to add support for reordering fastq samples. Then we would get repeated case names. Hypothetical scenario which might never happen though

diitaz93: but we can discuss it before merging

diitaz93:
## Tests in stage:
One case per sample ✅ 
Screenshot 2025-02-05 at 11 42 23

PR #4184, opened 2025-02-06
fix(dragen) Prevent double booking dragen nodes during demux

Vince-janv:
## Description
During the autumn the SLURM configuration for cg-dragen was updated from stating it has 24 cores to the correct 48 cores. However since we still tell send the SLURM jobs out in cg, requesting 24 cores, SLURM can now allocate two such jobs to the same node.

In theory this means that the second one will just wait until the first one has completed, since they both need access to the "special dragen" board on the machine.

### Changed

- Demultiplexing job now requests 48 cores

### DIFF
--- a/cg/models/slurm/sbatch.py
+++ b/cg/models/slurm/sbatch.py
@@ -23,4 +23,4 @@ class Sbatch(BaseModel):
 class SbatchDragen(Sbatch):
     partition: str = HastaSlurmPartitions.DRAGEN
     nodes: int = 1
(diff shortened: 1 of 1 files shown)


ahdamin (approved):
Looks fine to me. 

cg-dragen has `CPUTot=48`, cg-dragen2 has `CPUTot=48`, and cg-dragen3 has `CPUTot=64`

karlnyr (approved):
Nice fix!

PR #4219, opened 2025-02-18
chore: centralise app config

ChrOertlin:
## Description

Centralise the flask app config values.

### Added

-

### Changed

- Change the class app_config to read in ENV values or set a default if it is not present.

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/server/app.py
+++ b/cg/server/app.py
@@ -60,7 +60,7 @@ def create_app():
 
 def _load_config(app: Flask):
     app.config.update(app_config.model_dump())
-
+    app.secret_key = app_config.secret_key
 
 def _configure_extensions(app: Flask):
     _initialize_logging(app)
--- a/cg/server/app.py
+++ b/cg/server/app.py
@@ -60,7 +60,7 @@ def create_app():
 
 def _load_config(app: Flask):
     app.config.update(app_config.model_dump())
-    
+
 
 def _configure_extensions(app: Flask):
     _initialize_logging(app)
--- a/cg/server/app.py
+++ b/cg/server/app.py
@@ -44,7 +44,7 @@ from cg.store.models import (
     Sample,
     User,
 )
-from app_config import app_config
+from cg.server.app_config import app_config
 
 
 def create_app():
--- a/cg/server/app.py
+++ b/cg/server/app.py
@@ -62,6 +62,7 @@ def _load_config(app: Flask):
     app.config.update(app_config.model_dump())
     app.secret_key = app_config.secret_key
 
+
 def _configure_extensions(app: Flask):
     _initialize_logging(app)
     certs_resp = requests.get("https://www.googleapis.com/oauth2/v1/certs")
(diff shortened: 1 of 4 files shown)


cg/server/ext.py
Vince-janv, line 1: Nice changes 👍

cg/server/app_config.py
```diff
@@ -1,30 +1,33 @@
...
+    google_oauth_client_secret: str = os.getenv("GOOGLE_OAUTH_CLIENT_SECRET", "client_secret")
+    trailblazer_host: str = os.getenv("TRAILBLAZER_HOST", "trailblazer_host")
+    trailblazer_service_account: str = os.getenv("TRAILBLAZER_SERVICE_ACCOUNT", "service_account")
+    trailblazer_service_account_auth_file: str = os.getenv(
+        "TRAILBLAZER_SERVICE_ACCOUNT_AUTH_FILE", "auth_file.json"
+    )
+    freshdesk_url: str = os.getenv("FRESHDESK_URL", "https://company.freshdesk.com")
+    freshdesk_api_key: str = os.getenv("FRESHDESK_API_KEY", "freshdesk_api_key")
+    freshdesk_order_email_id: int = int(os.getenv("FRESHDESK_ORDER_EMAIL_ID", 10))
+    freshdesk_environment: str = os.getenv("FRESHDESK_ENVIRONMENT", "Stage")
```
Vince-janv, line 30: We don't need to explicitly use getenv here, that's what the BaseSettings class does. Was the doc string not clear enough?

ChrOertlin:
# tests
Deployed on cg-services-stage
Could open statusdb view, login and view data
Could submit order via web-ui, so lims values etc are properly read.

Vince-janv (commented):
Nice to stop using the keys to access the values. For the instantiation of AppConfig I think we should either use `os.getenv` or `BaseSettings`. Using both is confusing to me

cg/server/app_config.py
ChrOertlin: In combination with the two different ways of accessing environment files it was not clear to me. I reverted the use ov os.getenv

cg/server/app.py
```diff
@@ -58,8 +59,8 @@ def create_app():
 
 
 def _load_config(app: Flask):
-    app.config.update(app_config.dict())
-    app.secret_key = app.config["cg_secret_key"]
+    app.config.update(app_config.model_dump())
```
islean, line 62: Is the app.config used anymore or could we in theory remove it?

ChrOertlin: app.config is the configuration of the actual FlaskApp object. It requires some values such as the secret key to be set etc.

Vince-janv (approved):
Nice 💯

PR #4221, opened 2025-02-18
Allow all workflows in json parser

islean:
## Description

We have customers wishing to submit orders using a JSON order form for other order types than the one we currently allow. Given that we have validation which we can trust, it should be fine to allow parsing of json orders for these order types as well, relying on the customers getting feedback on missing/invalid data.

It will probably not work perfectly, but it should give them a starting point which does not require them to change the analysis type - which resets the order.

### Added

-

### Changed

- The JSON order form parser is not restricted to a subset of workflows.

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
Test performed locally by just swapping the data_analysis from balsamic to rnafusion for all samples:

image

Vince-janv (approved):
Nice! I'm a bit fuzzy on if we use OrderType or Workflow here, but if we test with Microbial_FASTQ we should be safe 👍

diitaz93 (approved):
Looks good!

islean:
Deployed to the VMs

PR #4224, opened 2025-02-19
remove reference genome tomte

diitaz93:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/orders/validation/order_types/tomte/models/sample.py
```diff
@@ -18,7 +18,7 @@ class TomteSample(Sample):
     phenotype_groups: list[str] | None = None
     phenotype_terms: list[str] | None = None
     post_formalin_fixation_time: int | None = None
-    reference_genome: GenomeVersion
+    reference_genome: Literal[GenomeVersion.HG38] = GenomeVersion.HG38
```
islean, line 21: My understanding is that if we get a request specifying it to be HG19 instead, then we would return an error with this change? Is that what we want or do we just want to override such a value with 38?

diitaz93: There will be no assignment of that variable anymore in the request. The field is removed from the order portal [here](https://github.com/Clinical-Genomics/clinical-genomics-ui/pull/701) and the column will not be in the new order form. However, based on what you say it would be safer to fix it to this value

diitaz93: BUt is there a way in which the assignment doe snot raise an error?

islean: When writing this validation flow, quite a lot of effort went into decoupling it from the order portal (not completely but quite a bit). I would prefer if we do not introduce logic which hinges on us knowing how the order portal works and if cURLing towards a swagger gave us the same behaviour. 

It might be that we want to raise an error if the request specifies something other than hg38 though. Just wanted us to think about this.

islean: We could in theory remove it as a field or mark it as a private attribute, which we only include in the model_dump

diitaz93: We still need to populate the database with this value (that is part of the acceptance criteria), so I think we should keep it somehow. Keeping it hidden is an option too, but I prefer having the fewer hidden arguments as possible

islean: I think I agree. These models do dictate what user input is allowed, and the reference genome does not seem to be part of the customer input for Tomte anymore. 

Ideally we would just have a TomteStoringService where we could set the reference genome and remove it completely from here, but now we still have the CaseStoringService. When should the reference genome be set for the Tomte samples now? So far we have sort of made due with private attributes, but I guess there are drawbacks there as well.

diitaz93: I will remove it

cg/meta/workflow/nf_analysis.py
```diff
@@ -907,11 +906,7 @@ def get_cases_to_store(self) -> list[Case]:
         ]
 
     def get_genome_build(self, case_id: str) -> GenomeVersion:
-        """Return reference genome version for a case.
-        Raises CgError if this information is missing or inconsistent for the samples linked to a case.
-        """
-        case = self.status_db.get_case_by_internal_id(case_id)
-        return get_genome_build(case)
```
ChrOertlin, line 914: this will alter functionality for all other NfAnalysis children.  Not sure if this is intended or not

ChrOertlin (approved):
Looks good. Just be aware that changing the get_genome_version() in the parent to not implemented will cause issues if any of the other children relied on the parent implementation.

cg/meta/workflow/nf_analysis.py
islean: Yeah, what happens when we generate delivery reports for other nf-core pipelines with this change? We seem to be calling this method there.

diitaz93: Where @islean ?

islean: write_delivery_report_file -> get_delivery_report_html -> get_report_data -> get_case_data -> get_case_analysis_data -> get_genome_build. But maybe all others have it overridden already, as you said

islean: But it gets quite hard to troubleshoot with these hybrid classes and the nested calls so it might very well be no problem

islean: You seem to be correct, apart from Jasen not having it implemented yet. But I guess it should have its own implementation either way

diitaz93:
## Tests on stage
Order submitted
Screenshot 2025-02-24 at 10 23 05
Sample saved in the database without reference genome
Screenshot 2025-02-24 at 10 24 56

cg/meta/workflow/utils/genome_build_helpers.py
```diff
@@ -1,33 +0,0 @@
...
-    mapping = {
-        GenomeVersion.GRCh37: GenomeBuild.hg19,
-        GenomeVersion.GRCh38: GenomeBuild.hg38,
-        GenomeVersion.HG19: GenomeBuild.hg19,
-        GenomeVersion.HG38: GenomeBuild.hg38,
-    }
-    # Check if the current instance is in the mapping, raise an error if not
-    if genome not in mapping:
-        raise ValueError(f"Genome build incompatible with scout: {genome.value}")
-    return mapping[genome]
```
Vince-janv, line 33: I don't think you can remove this. It seems to translate our way of writing genome builds to the way scout wants it

Vince-janv (approved):
👍

cg/meta/workflow/utils/genome_build_helpers.py
diitaz93: there are no longer usages for this. This is only called by `cg.meta.upload.scout.uploadscoutapi.upload_rna_genome_build_to_scout` which is only used for Tomte and the returning value for now on will be HG38, so I replaced the function call for this, see cg/meta/upload/scout/uploadscoutapi.py

diitaz93:
## Deployment to hasta
## Deployed to stage:
```shell
fatal: Not a git repository (or any parent up to mount point /home)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
repository is clean
Logging deploy ...
Getting deployer... done.
Getting version of deploy scripts... /home/js.diazboada
done.
Log deploy... done.
cg, version 68.1.0
[js.diazboada@hasta:~] [S_base] $
```
## Deployed to production:
```shell
Log deploy... done.
cg, version 68.1.0
```
## Deployed to CGVS
Screenshot 2025-02-25 at 09 34 42

Issue #4232, opened 2025-02-20
Bug: scout upload failing for Tomte when lacking drop results


fevac:
**Describe the bug**

When running TOMTE,  DROP analyses to compute expression/splicing outiliers are optional depending on the sample type. Drop is only run when the sample type is either blood or fibroblast (for which we have databases available). Currently, the Scout upload implementation to does not allow the drop output to be missing and the upload fails with `FileNotFoundError: No RNA fraser/outrider file was found in housekeeper for`

**To Reproduce**

Try to upload `suitedtoad` in scout

**Expected behavior**
Scout upload should be possible when missing drop files

**Additional context**

This is the part of the code that needs to be updated:  https://github.com/Clinical-Genomics/cg/blob/9de2d6be6e40be13ae17d847eb5ed9dfe5009bcb/cg/meta/upload/scout/uploadscoutapi.py#L399

For both rna_fraser and rna_outrider a warning should be raised instead of an error and then self.scout_api.upload_rna_fraser_outrider should only be done if those files are found

PR #4233, opened 2025-02-20
Fix(Tomte upload) Skip upload if missing files

Vince-janv:
## Description
Optional files are causing the Tomte upload to crash. See #4232.

### Fixed
- If the optional files do not exist, they are not uploaded


cg/apps/scout/scoutapi.py
```diff
@@ -302,25 +302,25 @@ def upload_omics_sample_id(
...
-            outrider_file_path,
         ]
+        if fraser_file_path:
+            upload_command.extend(["--fraser", fraser_file_path])
+
+        if outrider_file_path:
+            upload_command.extend(["--outrider", outrider_file_path])
         try:
             LOG.info(
                 f"Uploading rna fraser file {fraser_file_path} and outrider file {outrider_file_path} to case {case_id} with command {upload_command}"
```
ChrOertlin, line 326:
```suggestion
            LOG.info(
                f"Uploading rna fraser file {fraser_file_path} and outrider file {outrider_file_path} to case {case_id} with command {upload_command}"
```

Would be a bit confusing if outrider_file_path is None

ChrOertlin (approved):
Looks good. Eva mentioned a scout case that can be used to reproduce the bug (i.e. no DROP analysis present), I suppose you can try uploading it with this fix along with another rna fusion case does have a DROP analysis.

cg/meta/upload/scout/uploadscoutapi.py
```diff
@@ -413,13 +404,14 @@ def upload_rna_fraser_outrider_to_scout(
...
-                    cust_id=customer_case.customer.internal_id,
-                )
+                if rna_fraser or rna_outrider:
+                    self.scout_api.upload_rna_fraser_outrider(
+                        fraser_file_path=rna_fraser.full_path,
+                        outrider_file_path=rna_outrider.full_path,
+                        case_id=dna_case_id,
+                        customer_case_name=customer_case.name,
+                        cust_id=customer_case.customer.internal_id,
+                    )
```
diitaz93, line 414: I found the logging a bit confusing, maybe this can help? feel free to disregard
```suggestion
                    )
                else:
                    LOG.warning("No fraser nor outrider files found to upload")
```

diitaz93 (approved):
👍

cg/meta/upload/scout/uploadscoutapi.py
Vince-janv: I don't think this improves the loging situation. Will refrain from adding

Vince-janv:
When tested in stage the upload was skipped since the files were missing ✅

Vince-janv:
Deployed

PR #4247, opened 2025-02-25
Rename Raredisease Case config and Configurator to Nextflow

diitaz93:
## Description
Une one CaseConfig for all nextflow pipelines

### Added

- Method `_get_case_workflow` to get the workflow inside the NetflowConfigurator

### Changed

- Renamed `RarediseaseCaseConfig` to `NextflowCaseConfig`
- Names of fixtures

### Fixed

- Removed currently unused elements from config in Configurator

### How to test

Not testable. Merging to the development branch

## Review

### DIFF
--- a/cg/services/analysis_starter/configurator/implementations/nextflow.py
+++ b/cg/services/analysis_starter/configurator/implementations/nextflow.py
@@ -4,39 +4,25 @@ from pathlib import Path
 from cg.constants import FileExtensions, Priority, Workflow
 from cg.models.cg_config import RarediseaseConfig
 from cg.services.analysis_starter.configurator.abstract_service import Configurator
-from cg.services.analysis_starter.configurator.models.raredisease import RarediseaseCaseConfig
+from cg.services.analysis_starter.configurator.models.nextflow import NextflowCaseConfig
+from cg.store.models import Case
 from cg.store.store import Store
 
 LOG = logging.getLogger(__name__)
 
 
-class RarediseaseConfigurator(Configurator):
+class NextflowConfigurator(Configurator):
     """Configurator for Raredisease analysis."""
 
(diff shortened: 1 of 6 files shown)


cg/services/analysis_starter/configurator/implementations/nextflow.py
```diff
@@ -4,39 +4,25 @@
...
-from cg.services.analysis_starter.configurator.models.raredisease import RarediseaseCaseConfig
+from cg.services.analysis_starter.configurator.models.nextflow import NextflowCaseConfig
+from cg.store.models import Case
 from cg.store.store import Store
 
 LOG = logging.getLogger(__name__)
 
 
-class RarediseaseConfigurator(Configurator):
+class NextflowConfigurator(Configurator):
```
islean, line 14: I do not know if we can get away with a single configurator for all Nextflow pipelines, I just meant that the NextflowCaseConfig should probably be the same for all. Or is the "configure case" action the same for all nextflow pipelines?

ChrOertlin (approved):

cg/services/analysis_starter/configurator/implementations/nextflow.py
ChrOertlin: I was thinking the same, but thought you had a plan for that. If there are more specific NextflowConfigurators for example for tomte etc, then the name should be kept here I suppose

islean: I think there are some things in the `config_case` method which makes multiple configurators reasonable. For example, `get_managed_variants`does not exist in the NfAnalysisAPI but is used in config_case, so it is probably pipeline specific.

diitaz93: That is a good point. It is a bit har to know at this stage but I think we can hold this change until we are sure. Good catch 👍

PR #4277, opened 2025-03-11
(Pipeline Integration) Merge PrepareFastqAPI into the FastqFetcher

islean:
## Description

This PR introduces two new classes, `CaseCompressionData` and `SampleCompressionData` which encapsulate data on a case's/sample's compression status. It moves the functionality of the `PrepareFastqAPI` into the `FastqFetcher` since their purposes overlapped. 

Removing the PrepareFastqAPI completely is out of scope for this PR. It inherits from the MetaAPI and this coupling makes me consider it a separate task. The other parts of the code which use PrepareFastqAPI, such as the DownsamplingAPI should only need the CompressAPI after this PR though.

### Added

- CaseCompressionData
- SampleCompressionData

### Changed

- PrepareFastqAPI not needed in the FastqFetcher

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/compress/compress.py
+++ b/cg/meta/compress/compress.py
@@ -15,7 +15,7 @@ from cg.constants import SequencingFileTag
 from cg.exc import DecompressionCouldNotStartError
 from cg.meta.backup.backup import SpringBackupAPI
 from cg.meta.compress import files
-from cg.models.compression_data import CompressionData
+from cg.models.compression_data import CaseCompressionData, CompressionData, SampleCompressionData
 from cg.store.models import Case, Sample
 
 LOG = logging.getLogger(__name__)
@@ -359,3 +359,22 @@ class CompressAPI:
             if fastq_file.exists():
                 fastq_file.unlink()
                 LOG.debug(f"FASTQ file {fastq_file} removed")
+
+    def get_case_compression_data(self, case: Case) -> CaseCompressionData:
+        """Return an object containing compression data for a case."""
+        sample_compressions: list[SampleCompressionData] = []
+        for sample in case.samples:
+            sample_id: str = sample.internal_id
+            sample_compression_data: SampleCompressionData = self.get_sample_compression_data(
+                sample_id
+            )
+            sample_compressions.append(sample_compression_data)
+        return CaseCompressionData(
+            case_id=case.internal_id, sample_compression_data=sample_compressions
+        )
+
+    def get_sample_compression_data(self, sample_id: str) -> SampleCompressionData:
+        compression_objects: list[CompressionData] = []
+        version: Version = self.hk_api.get_latest_bundle_version(sample_id)
+        compression_objects.extend(files.get_spring_paths(version))
+        return SampleCompressionData(sample_id=sample_id, compression_objects=compression_objects)
--- a/cg/meta/compress/compress.py
+++ b/cg/meta/compress/compress.py
@@ -12,10 +12,11 @@ from cg.apps.crunchy import CrunchyAPI
 from cg.apps.crunchy.files import update_metadata_date
 from cg.apps.housekeeper.hk import HousekeeperAPI
 from cg.constants import SequencingFileTag
+from cg.exc import DecompressionCouldNotStartError
 from cg.meta.backup.backup import SpringBackupAPI
 from cg.meta.compress import files
 from cg.models.compression_data import CompressionData
-from cg.store.models import Sample
+from cg.store.models import Case, Sample
 
 LOG = logging.getLogger(__name__)
 
@@ -101,6 +102,15 @@ class CompressAPI:
         if (not spring_file) or (not spring_file.archive):
             return False
         return bool(spring_file.archive.archived_at)
+
+    def decompress_case(self, case: Case) -> None:
+        """Decompresses all Spring files tied to the case.
+        Raises:
+            DecompressionFailedToStartError if no sample could start decompressing."""
+        if not any(self.decompress_spring(sample.internal_id) for sample in case.samples):
+            raise DecompressionCouldNotStartError(
+                f"No sample could be decompressed for {case.internal_id}"
+            )
 
     def decompress_spring(self, sample_id: str) -> bool:
         """Decompress SPRING archive for a sample.
--- a/cg/meta/compress/compress.py
+++ b/cg/meta/compress/compress.py
@@ -364,9 +364,8 @@ class CompressAPI:
         """Return an object containing compression data for a case."""
         sample_compressions: list[SampleCompressionData] = []
         for sample in case.samples:
-            sample_id: str = sample.internal_id
             sample_compression_data: SampleCompressionData = self.get_sample_compression_data(
-                sample_id
+                sample.internal_id
             )
             sample_compressions.append(sample_compression_data)
         return CaseCompressionData(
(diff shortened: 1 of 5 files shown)


cg/meta/workflow/prepare_fastq.py
```diff
@@ -23,7 +22,6 @@ def __init__(self, store: Store, compress_api: CompressAPI):
         self.store: Store = store
         self.hk_api: HousekeeperAPI = compress_api.hk_api
         self.compress_api: CompressAPI = compress_api
-        self.crunchy_api: CrunchyAPI = compress_api.crunchy_api
```
islean, line 26: This was not used

cg/services/analysis_starter/input_fetcher/implementations/fastq_fetcher.py
```diff
@@ -71,54 +75,41 @@ def _resolve_decompression(self, case_id: str) -> None:
...
-        if self.prepare_fastq_api.is_spring_decompression_needed(case_id):
-            self._decompress_case(case_id)
-            return
-
-        if self.prepare_fastq_api.is_spring_decompression_running(case_id):
+        case_compression_data: CaseCompressionData = self._get_case_compression_data(case_id)
+        if case_compression_data.is_spring_decompression_needed():
+            LOG.info(f"The analysis for {case_id} could not start, decompression is needed")
+            if not case_compression_data.can_at_least_one_sample_be_decompressed():
+                self._set_to_analyze_if_decompressing(case_compression_data)
```
islean, line 82: Should this just raise an error? If we need to decompress but we are unable to - what does that tell us?

cg/services/analysis_starter/input_fetcher/implementations/fastq_fetcher.py
```diff
@@ -142,16 +133,69 @@ def _is_illumina_run_check_applicable(self, case_id) -> bool:
...
-        LOG.info(
-            f"Decompression is running for {case_id}, analysis will be started when decompression is done"
+        case_id: str = case_compression_data.case_id
+        if case_compression_data.is_spring_decompression_running():
+            LOG.info(
+                f"Decompression is running for {case_id}, analysis will be started when decompression is done"
+            )
+            self.status_db.set_case_action(case_internal_id=case_id, action=CaseActions.ANALYZE)
+
+    def _get_case_compression_data(self, case_id: str) -> CaseCompressionData:
```
islean, line 145: I wonder if this method should exist outside of the FastqFetcher. Same thing with _get_sample_compression_data. Then we could get rid of the FastqFetcher dependency in downsampling for example

Vince-janv (commented):
Tried to answer you questions but it felt a bit to big to review in detail. Will give it another pass before you merge

cg/services/analysis_starter/input_fetcher/implementations/fastq_fetcher.py
islean: It is quite tricky for me to discern as well. I think you are correct, although it is hard to see since the checks are performed for all files tied to the case at once. 

But from the first check we know that there is at least one file with no decompression pending and no fastq pair. 

The second check checks if there is any file which can be decompressed. When we get to one of the files with no decompression pending and no fastq pair, the only thing remaining is, as you say, the spring file check. 

I think a missing Spring file should be cause for concern - more than a warning. But do we know what happens with DDN archived spring files? Maybe they go in to this clause?

cg/services/analysis_starter/input_fetcher/implementations/fastq_fetcher.py
islean: I have moved it to the CompressAPI since this comment. The DownsamplingAPI already has access to the compress api so should be good. Althought there is some weird MetaAPI inheritance going on which I think is out of scope for this PR to address

cg/services/analysis_starter/input_fetcher/implementations/fastq_fetcher.py
```diff
@@ -142,16 +138,53 @@ def _is_illumina_run_check_applicable(self, case_id) -> bool:
...
+
+    @staticmethod
+    def _should_skip_sample(case: Case, sample: Sample):
+        """
+        For some workflows, we want to start a partial analysis disregarding the samples with no reads.
+        This method returns true if we should skip the sample.
+        """
+        return case.data_analysis in PIPELINES_USING_PARTIAL_ANALYSES and not sample.has_reads
+
+    def _add_decompressed_fastq_files_to_housekeeper(self, case_id: str) -> None:
```
ChrOertlin, line 158: To me it seems the FastqFetcher has a bit to many responsibilities and I think this has to do with some consideration about when a compression or decompression call should be considered as "resolved". Or in general, when is a certain task fully resolved.

Example for compression:
1. The call is resolved when a compression / decompression job is send to SLURM.
2. The call is resolved when compressed / decompressed files are stored in the file system (i.e. housekeeper).

Right now, compression is resolved once the job is send to SLURM. This is apparent by the need for logic that adds decompressed files to HK in the fastqfetcher. However, this has also as consequence that when you require a compression / decompression elsewhere this part also needs this extra piece of code to add files to housekeeper.

Alternatively in option 2, a compression / decompression job being resolved when processed files are stored in housekeeper would keep such logic centralised. Furthermore, it would reduce the overall need for additional code elsewhere.

I would suggest to add this logic to the compression / decompression class(es). And given our timer based system have it handled as such.

Scenario1:

Caller asks -> can you decompress these files 
Decompresser -> sure thing, found the compressed files and starting a job!

Scenario2:
Caller asks -> can you decompress these files
Decompressor -> hmm these files are not there, but I found their decompressed counterparts I added them to the file storage system.

Perhaps, this is something to consider.

cg/meta/compress/compress.py
```diff
@@ -102,6 +103,15 @@ def _is_spring_archived(self, compression_data: CompressionData) -> bool:
             return False
         return bool(spring_file.archive.archived_at)
 
+    def decompress_case(self, case: Case) -> None:
```
ChrOertlin, line 106: I would consider splitting the decompression and compression logic into two distinct modules. 
They are opposite processes, and classes that require something to be decompressed likely don't care about compression at all. It might help with maintainability in the long run as well.

Vince-janv (approved):
LGTM :+1

cg/services/analysis_starter/input_fetcher/implementations/fastq_fetcher.py
islean: I do agree that the "adding of decompressed files to housekeeper"-functionality feels a bit homeless in our current setup. But I also think it will not be fixed in this PR. I will create an issue where we can track this and potentially act on it in the future.

cg/meta/compress/compress.py
islean: I agree, but touching the compress api was not within the intended scope for this PR. It was only done to facilitate the removal of the prepare fastq api and decouple it from parts of the code where it should not be needed.

cg/services/analysis_starter/input_fetcher/implementations/fastq_fetcher.py
islean: Created https://github.com/Clinical-Genomics/cg/issues/4281.

PR #4279, opened 2025-03-12
Add post processing filter for unfinished run transfers

ChrOertlin:
## Description

Currently the pacbio post processing is performed on runs that are in progress of being transferred. This PR adds functionality to perform a check whether all required files are there before it gets added to the post processing queue.

In terms of pacbio this is to check whether the manifest, and all required files are present WITHOUT manipulating any of the present files (i.e. unzipping the reports folder).

Then a check is added to the "get_unprocessed_runs" that checks wheter a run can start (i.e. has all files present).

### Added

- Extended ```PostProcessingService(ABC)``` with a function ```def can_post_processing_start````
- Extended ```RunValidator(ABC)``` with a function ```def validate_run_files(runData)````
- Added an if statement to ```_get_unprocessed_runs_from_run_names``` that already has the post processing service and now performs the check
```if post_processing_service.can_post_processing_start(name):``` If it TRUE it adds the run to the unprocessed runs list

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### CODE REVIEW

File: cg/services/run_devices/pacbio/run_validator/pacbio_run_validator.py [hunk 1]
```diff
@@ -57,6 +58,12 @@ def ensure_post_processing_can_start(self, run_data: PacBioRunData) -> None:
         self._touch_is_validated(run_data.full_path)
         LOG.debug(f"Run for {run_data.full_path} is validated.")
 
+    def validate_run_files(self, run_data: PacBioRunData):
```
ChrOertlin, line 61: This basically just checks whether all files are present without trying to parse them or anything. First files it checks for is the "transferdone" or manifest files.

Just be aware, is a transfer is never fully completed it will always fail silently with each time-based iteration and adding this behaviour makes it possible that "corrupted" runs will not notify production. The old approach by trying and failing each time would make it noticable. Downside of course is notification fatigue.

cg/services/run_devices/pacbio/post_processing_service.py
```diff
@@ -64,3 +64,14 @@ def is_run_processed(self, run_name: str) -> bool:
...
         return processing_complete_file.exists()
+
+    def can_post_processing_start(self, run_name: str) -> bool:
+        LOG.info(f"Checking if Pacbio post-processing can start for run: {run_name}")
+        try:
+            run_data: PacBioRunData = self.run_data_generator.get_run_data(
+                run_name=run_name, sequencing_dir=self.sequencing_dir
+            )
+            self.run_validator.validate_run_files(run_data)
+        except Exception:
```
islean, line 75: Can we except the more specific `PostProcessingRunFileManagerError` here?

cg/services/run_devices/pacbio/run_validator/pacbio_run_validator.py
```diff
@@ -57,6 +58,12 @@ def ensure_post_processing_can_start(self, run_data: PacBioRunData) -> None:
...
+        """
+        Validate presence of all required run files
```
islean, line 63: Could we document in the docstring what errors this raises? Or make it wrap all errors raised in some RunFileValidationError? To me, it is not clear from the method what kind of validation is performed and the handling one should consider for it.

ChrOertlin: I expanded the docstring!

cg/services/run_devices/pacbio/run_validator/pacbio_run_validator.py
```diff
@@ -57,6 +58,18 @@ def ensure_post_processing_can_start(self, run_data: PacBioRunData) -> None:
         self._touch_is_validated(run_data.full_path)
         LOG.debug(f"Run for {run_data.full_path} is validated.")
 
+    def validate_run_files(self, run_data: PacBioRunData):
```
Vince-janv, line 61: If i understand this correctly, the `get_run_validation_files` checks for the existence of two files and not necessarily all required files. Can we change the docstring and method name to reflect this? We could call the method validate_run_completion

Vince-janv (approved):
Nice, this should solve the issue 🙌

cg/services/run_devices/pacbio/run_validator/pacbio_run_validator.py
ChrOertlin: I changed the docstring to better reflect this, but kept the method name the same.

Issue #3563, opened 2024-08-14
Bug: No delivery option not parsed correctly.


islean:
**Describe the bug**

When uploading an (at least 1508) excel order form with "No delivery" selected as data_delivery, the parsing fails, with an "Unsupported data delivery" error.

**To Reproduce**

Upload a 1508 form with No delivery selected from the dropdown. 

**Expected behavior**

It should be parsed into the no-delivery option.

## Clarification
In the excel orderform the column data delivery is a drop down meny containg the option No delivery since it contains a space it crashes when the excel sheet is uploaded.

The option is only used within the CG org to place control samples

diitaz93:
The parsing of the panel "Inherited cancer" also gets wrongly parsed in that same orderform

<img width="1089" alt="Image" src="https://github.com/user-attachments/assets/688e52ad-dc73-427c-b46c-572f3d03050b" />

PR #4294, opened 2025-03-17
fix( parsing no delivery option)

ChrOertlin:
## Description
closes: #3563
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


ChrOertlin:
File with No Delivery could be uploaded

cg/apps/orderform/excel_orderform_parser.py
```diff
@@ -178,7 +178,10 @@ def get_data_delivery(self) -> str:
         """Determine the order_data delivery type"""
 
         data_delivery: str = self.parse_data_delivery()
-
+        # Orderforms allow for No delivery option, not adjusting it here causes an error when parsing excel order forms.
+        # The .replace() below would otherwise an incompatible DataDelivery option.
```
ChrOertlin, line 182:
```suggestion
```

cg/apps/orderform/excel_orderform_parser.py
```diff
@@ -178,7 +178,8 @@ def get_data_delivery(self) -> str:
         """Determine the order_data delivery type"""
 
         data_delivery: str = self.parse_data_delivery()
-
+        if data_delivery == "no_delivery":
+            return DataDelivery.NO_DELIVERY
```
islean, line 182: I think the problem goes a bit deeper than only data delivery when considering this comment https://github.com/Clinical-Genomics/cg/issues/3563#issuecomment-2631397523. I think any option with a space in it gets read incorrectly.

ChrOertlin: That was considered out of scope

ChrOertlin: Nvm I can't read apparently, I will add it to the list. :)

ChrOertlin: I will have a look at this in a different PR. I changed the Gene list ( Panel ) to Interited cancer but it won't break the tests there... so I will have a closer look where the parsing and changing actually happens.

cg/apps/orderform/excel_orderform_parser.py
```diff
@@ -233,5 +226,24 @@ def parse_orderform(self, excel_path: str) -> None:
 
     @staticmethod
     def _transform_data_delivery(data_delivery: str) -> str:
-        """Transforms the data-delivery parsed in the excel file, to the ones used in cg"""
-        return data_delivery.lower().replace(" + ", "-").replace(" ", "_")
+        """Transforms the data-delivery parsed in the excel file, to the ones used in cg."""
+        try:
+            orderform_to_internal: dict = {
```
ChrOertlin, line 231: I went for this option because it seemed the most robust and clear.

Previous option:
```Python data_delivery.lower().replace(" + ", "-").replace(" ", "_")```

Would create a clash between the allowed options
```fastq qc``` and ```no delivery``` as adjusting the replace in the benefit of one, would make it fail for the other.

Now we have a clear mapping

File: cg/models/orders/validators/excel_sample_validators.py [hunk 4]
```diff
@@ -8,6 +8,8 @@ def parse_panels(panels: str) -> list[str] | None:
     separator = ";" if ";" in panels else None
     if ":" in panels:
         separator = ":"
+    if not separator:
```
ChrOertlin, line 11: The default behaviour of split() is to split on any whitespace if a given separator is none. This breaks in cases when there is only one panel that has a white space in it (i.e. Inherited cancer).

cg/apps/orderform/excel_orderform_parser.py
```diff
@@ -175,20 +175,13 @@ def parse_data_analysis(self) -> str:
...
-        """Determine the order_data delivery type"""
-
-        data_delivery: str = self.parse_data_delivery()
-
-        try:
-            return DataDelivery(data_delivery)
-        except ValueError as error:
-            raise OrderFormError(f"Unsupported Data Delivery: {data_delivery}") from error
+        """Determine the order_data delivery type."""
+        return self.parse_data_delivery()
```
Vince-janv, line 179: I feel like `parse_data_delivery` can just replace `get_data_delivery` since the later does not do anything other than call the first

cg/apps/orderform/excel_orderform_parser.py
Vince-janv: I like this implementation. It clarifies the mess 🧹

cg/models/orders/validators/excel_sample_validators.py
Vince-janv: Wow. Reading this, splitting on space seems unintended. This solution is more explicit 👍

tests/apps/orderform/test_excel_orderform_parser.py
```diff
@@ -168,6 +168,30 @@ def test_parse_mip_orderform(mip_orderform: str, nr_samples_mip_orderform: int):
     assert order_form_parser.project_type == OrderType.MIP_DNA
 
 
+def test_parse_mip_orderform_no_delivery(
```
Vince-janv, line 171: Nice to add a test 👍  The naming implies that we should assert that the delivery is "no-delivery" but this does not seem to be checked.

Vince-janv (approved):
Nice patch! 💯

cg/apps/orderform/excel_orderform_parser.py
```diff
@@ -233,5 +226,24 @@ def parse_orderform(self, excel_path: str) -> None:
...
+                "bam": DataDelivery.BAM,
+                "fastq": DataDelivery.FASTQ,
+                "fastq + analysis": DataDelivery.FASTQ_ANALYSIS,
+                "fastq + analysis + scout": DataDelivery.FASTQ_ANALYSIS_SCOUT,
+                "fastq + Scout": DataDelivery.FASTQ_SCOUT,
+                "fastq qc": DataDelivery.FASTQ_QC,
+                "fastq qc + analysis": DataDelivery.FASTQ_QC_ANALYSIS,
+                "no delivery": DataDelivery.NO_DELIVERY,
+                "scout": DataDelivery.SCOUT,
+                "nipt viewer": DataDelivery.NIPT_VIEWER,  # legacy option; keeping it for backward compatibility
```
islean, line 243: What backwards compatability does this add? The excel order forms are versioned so any file with this option specified would be blocked either way. I think it is clearer with it gone if it does not exist in the current options

islean (approved):
Looks good!

ChrOertlin:
image

cg/apps/orderform/excel_orderform_parser.py
ChrOertlin: Was trying to stay on the safe side here, but I think you are right. The option is not present any longer so it can be removed.

tests/apps/orderform/test_excel_orderform_parser.py
ChrOertlin: added the check

ChrOertlin: I also added an assert that multi-gene panels are parsed properly

PR #4295, opened 2025-03-17
(Pipeline Integration) Add starter factory

islean:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/analysis_starter/constants.py
+++ b/cg/services/analysis_starter/constants.py
@@ -1,0 +1,18 @@
+from cg.constants import Workflow
+
+FASTQ_WORKFLOWS = [
+    Workflow.BALSAMIC,
+    Workflow.BALSAMIC_PON,
+    Workflow.BALSAMIC_UMI,
+    Workflow.FLUFFY,
+    Workflow.MICROSALT,
+    Workflow.MIP_DNA,
+    Workflow.MIP_RNA,
+    Workflow.MUTANT,
+    Workflow.RAREDISEASE,
+    Workflow.RNAFUSION,
+    Workflow.TAXPROFILER,
+    Workflow.TOMTE,
+]
+
+BAM_WORKFLOWS = [Workflow.NALLO]
--- a/cg/services/analysis_starter/factory.py
+++ b/cg/services/analysis_starter/factory.py
@@ -118,7 +118,7 @@ class AnalysisStarterFactory:
             seqera_platform_client: SeqeraPlatformClient = self._get_seqera_platform_client()
             return SeqeraPlatformSubmitter(
                 client=seqera_platform_client,
-                compute_environment_ids=self.cg_config.seqera_platform.compute_environmentsCan,
+                compute_environment_ids=self.cg_config.seqera_platform.compute_environments,
             )
 
     def _get_seqera_platform_client(self) -> SeqeraPlatformClient:
--- a/cg/services/analysis_starter/factory.py
+++ b/cg/services/analysis_starter/factory.py
@@ -1,3 +1,6 @@
+from cg.apps.housekeeper.hk import HousekeeperAPI
+from cg.apps.lims import LimsAPI
+from cg.apps.scout.scoutapi import ScoutAPI
 from cg.constants import Workflow
 from cg.constants.nextflow import NEXTFLOW_WORKFLOWS
 from cg.meta.archive.archive import SpringArchiveAPI
@@ -18,6 +21,9 @@ from cg.services.analysis_starter.configurator.file_creators.params_file.abstrac
 from cg.services.analysis_starter.configurator.file_creators.params_file.raredisease import (
     RarediseaseParamsFileCreator,
 )
+from cg.services.analysis_starter.configurator.file_creators.sample_sheet.abstract import (
+    NextflowSampleSheetCreator,
+)
 from cg.services.analysis_starter.configurator.file_creators.sample_sheet.raredisease import (
     RarediseaseSampleSheetCreator,
 )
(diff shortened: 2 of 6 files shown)


cg/services/analysis_starter/constants.py
```diff
@@ -0,0 +1,18 @@
...
+    Workflow.MIP_DNA,
+    Workflow.MIP_RNA,
+    Workflow.MUTANT,
+    Workflow.RAREDISEASE,
+    Workflow.RNAFUSION,
+    Workflow.TAXPROFILER,
+    Workflow.TOMTE,
+]
+
+BAM_WORKFLOWS = [Workflow.NALLO]
```
islean, line 18: Do we have more? How should these be tracked?

islean: With this design anyone adding a pipeline needs to add it here as well, but it seems quite forgettable

cg/services/analysis_starter/factory.py
```diff
@@ -0,0 +1,125 @@
...
+        gene_panel_creator: GenePanelFileCreator = self._get_gene_panel_file_creator()
+        managed_variants_creator: ManagedVariantsFileCreator = (
+            self._get_managed_variants_file_creator()
+        )
+        return RarediseaseExtension(
+            gene_panel_file_creator=gene_panel_creator,
+            managed_variants_file_creator=managed_variants_creator,
+        )
+
+    def _get_gene_panel_file_creator(self) -> GenePanelFileCreator:
```
islean, line 93: Should all the file creators be instantiated here?

cg/services/analysis_starter/factory.py
```diff
@@ -0,0 +1,125 @@
...
+        return GenePanelFileCreator(scout_api=self.cg_config.scout_api, store=self.store)
+
+    def _get_managed_variants_file_creator(self) -> ManagedVariantsFileCreator:
+        return ManagedVariantsFileCreator(scout_api=self.cg_config.scout_api, store=self.store)
+
+    def _get_input_fetcher(self, case_id: str) -> InputFetcher:
+        workflow: Workflow = self.store.get_case_workflow(case_id)
+        if workflow in FASTQ_WORKFLOWS:
+            spring_archive_api = SpringArchiveAPI(
```
islean, line 102: This should probably be instantiated in the config? Or is this good?

cg/services/analysis_starter/factory.py
```diff
@@ -0,0 +1,125 @@
...
+            )
+        return InputFetcher()
+
+    def _get_submitter(self, case_id: str) -> Submitter:
+        workflow: Workflow = self.store.get_case_workflow(case_id)
+        if workflow in NEXTFLOW_WORKFLOWS:
+            seqera_platform_client: SeqeraPlatformClient = self._get_seqera_platform_client()
+            return SeqeraPlatformSubmitter(
+                client=seqera_platform_client,
+                compute_environment_ids=self.cg_config.seqera_platform.compute_environments,
```
islean, line 121: This reaching through the SeqeraPlatformConfig indicates that our server config is not designed properly.

cg/services/analysis_starter/factory.py
```diff
@@ -0,0 +1,125 @@
...
+            )
+
+    def _get_seqera_platform_client(self) -> SeqeraPlatformClient:
+        return SeqeraPlatformClient(self.cg_config.seqera_platform)
```
islean, line 125: Should this be here?

islean (commented):
Unclarities added

cg/services/analysis_starter/factory.py
```diff
@@ -0,0 +1,125 @@
...
+                account=pipeline_config.account,
+                platform=pipeline_config.platform,
+                resources=pipeline_config.resources,
+                store=self.store,
+                workflow_config_path=pipeline_config.workflow_config_path,
+            )
+            params_file_creator: ParamsFileCreator = RarediseaseParamsFileCreator(
+                lims=self.cg_config.lims_api, store=self.store, params=pipeline_config.params
+            )
+            sample_sheet_creator: RarediseaseSampleSheetCreator = RarediseaseSampleSheetCreator(
```
islean, line 64: Clearly these should be initiated in a dynamic way and not only be Raredisease do we know if the inits are the same? What do we need to pass to fetch the correct one?

cg/services/analysis_starter/factory.py
```diff
@@ -0,0 +1,125 @@
...
+                housekeeper_api=self.cg_config.housekeeper_api,
+                lims=self.cg_config.lims_api,
+                params_file_creator=params_file_creator,
+                pipeline_config=pipeline_config,
+                sample_sheet_creator=sample_sheet_creator,
+                store=self.store,
+                pipeline_extension=self._get_pipeline_extension(workflow),
+            )
+
+    def _get_pipeline_config(self, workflow: Workflow) -> CommonAppConfig:
```
islean, line 80: The fact that all the nextflow configs inherit only from CommonAppConfig leads to our code being riddled with warnings because of unknown attributes.

cg/services/analysis_starter/factory.py
```diff
@@ -0,0 +1,125 @@
...
+        return getattr(self.cg_config, workflow)
+
+    def _get_pipeline_extension(self, workflow: Workflow) -> PipelineExtension:
+        gene_panel_creator: GenePanelFileCreator = self._get_gene_panel_file_creator()
+        managed_variants_creator: ManagedVariantsFileCreator = (
+            self._get_managed_variants_file_creator()
+        )
+        return RarediseaseExtension(
```
islean, line 88: This should be dynamic. But how do we know what to pass? Does it need its own factory?

cg/services/analysis_starter/factory.py
```diff
@@ -0,0 +1,125 @@
...
+                housekeeper_api=self.cg_config.housekeeper_api,
+                data_flow_config=self.cg_config.data_flow,
+            )
+            return FastqFetcher(
+                compress_api=self.cg_config.meta_apis["compress_api"],
+                housekeeper_api=self.cg_config.housekeeper_api,
+                spring_archive_api=spring_archive_api,
+                status_db=self.store,
+            )
+        return InputFetcher()
```
islean, line 113: To be modified with BAM support in the future

cg/services/analysis_starter/factory.py
```diff
@@ -0,0 +1,125 @@
...
+
+class AnalysisStarterFactory:
+    def __init__(self, cg_config: CGConfig):
+        self.cg_config = cg_config
+        self.store: Store = cg_config.status_db
+
+    def get_analysis_starter(self, case_id: str) -> AnalysisStarter:
+        configurator: Configurator = self._get_configurator(case_id)
+        input_fetcher: InputFetcher = self._get_input_fetcher(case_id)
+        submitter: Submitter = self._get_submitter(case_id)
```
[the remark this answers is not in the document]
islean: My thinking initially was that the input fetcher depends on the delivery type, but then I remembered that most delivery types are not only fastq or bam, so at this point, yeah - we should probably pass the workflow.

cg/services/analysis_starter/factory.py
islean: I see our code's primary purpose as facilitating running the pipelines, so I think our baseline should be to design with flexibility in mind. It is not up for me to say whether they will grow, but if a pipeline dev says "we need this sort of file to run our new pipeline", we should have a design which integrates that easily.

cg/services/analysis_starter/factory.py
islean: It is more the fact that the computeEnvIds are not needed by the client at all anymore, but they are set in the config we designed for the client. I guess we should have one config for the submitter and one for the client?

cg/services/analysis_starter/factory.py
```diff
@@ -0,0 +1,125 @@
...
+        if workflow in NEXTFLOW_WORKFLOWS:
+            pipeline_config: CommonAppConfig = self._get_pipeline_config(workflow)
+            config_file_creator = NextflowConfigFileCreator(
+                account=pipeline_config.account,
+                platform=pipeline_config.platform,
+                resources=pipeline_config.resources,
+                store=self.store,
+                workflow_config_path=pipeline_config.workflow_config_path,
+            )
+            params_file_creator: ParamsFileCreator = RarediseaseParamsFileCreator(
```
ChrOertlin, line 61: These likely needs it own function as soon as there are more than the "RareDiseaseParamsFileCreator"

cg/services/analysis_starter/factory.py
ChrOertlin: You can make a function that pulls the correct one from a map for example.

cg/services/analysis_starter/factory.py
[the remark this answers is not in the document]
ChrOertlin: Unless there is more information required than just the workflow it makes sense, if it turns out later you need more information than case_id will allow to fetch more information of course. I'd wait with changing it until it is more final.

cg/services/analysis_starter/factory.py
ChrOertlin: I agree here with Isak, it will make it clear how to extend the current setup without requiring refactoring.

islean: Let's keep the file creator instantiatiations here for now. It might be that we want to keep the file creators separate at some point though.

cg/services/analysis_starter/constants.py
islean: Let's keep it here for now then

cg/services/analysis_starter/factory.py
islean: Yeah, that might be the nicest approach but then we need to fix the init in the parent NextflowSampleSheetCreator class.

cg/services/analysis_starter/factory.py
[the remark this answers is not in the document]
islean: I will keep workflow for now.

Issue #4307, opened 2025-03-25
Bug(cleaning) Sequencing data not completely cleaned


Vince-janv:
### Description
Sequencing data is permanently stored as SPRING files in housekeeper. However it passes through several other directories before it gets there. At the moment only one of the four intermediate directories are cleaned.

The bug was introduced here: https://github.com/Clinical-Genomics/cg/commit/e65ae7363b07b63cb8d851eac69f21d7b98b49d5#r154283894.

Instead of deleting the path given, it reconstructs the run-directory path.

PR #3349, opened 2024-06-18
[DEV] Feature branch to re-route flow to illumina device tables.

ChrOertlin:
# Description

Feature branch to re-route the post-processing database flow from the `Flowcell` -> `SampleLaneSequencingMetrics` model paradigm to `IlluminaFlowCell` -> `IlluminaSequencingRun` -> `IlluminaSampleSequencingMetrics`. 
The choice to use a feature branch is to collect all changes and make sure they are tested. Since we need to alter core logic that requires a database migration we cannot do these step by step.

PR in servers: https://github.com/Clinical-Genomics/servers/pull/1394
Data migration script: https://github.com/Clinical-Genomics/add-new-tech/issues/7

## Deploy
```shell
bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b develop-illumina-devices -a
```

## CLI command changes
Commands that need to be updated in servers

- [x] `cg clean flow-cells` -> `cg clean illumina-runs` (requires servers update, PR: <placeholder>)
- [x] `cg backup encrypt-flow-cells` -> `cg backup encrypt-illumina-runs` (requires servers update, PR: <placeholder>)
- [x] `cg backup flow-cells` -> `cg backup illumina-runs` (requires servers update, PR: <placeholder>)
- [x] `cg backup fetch-flow-cell`-> `cg backup fetch-illumina-run` (no update needed)
- [x] `cg demultiplex finish flow-cell` -> `cg demultiplex finish illumina-run` (no update needed in servers
- [x] `cg get flow-cell` -> `cg get sequencing-run` (Does not require a PR in servers)
- [x] `cg set flowcell` -> `cg set sequencing-run` (Does not require a PR in servers)
- [x] `cg store flow-cell`-> `cg store illumina-run` (Does not require a PR in servers)
- [x] `cg store demultiplexed-flow-cell` -> `cg store demultiplexed-run` (Does not require a PR in servers)
- [x] `cg decompress flow-cell` -> `cg decompress illumina-run` (Does not require a PR in servers)
https://github.com/Clinical-Genomics/servers/pull/1394
## Tests
CO:
- [x] NIPT 
  - [x] `cg -l DEBUG upload nipt case healthycollie --dry-run`
  - [x] Requirements: A NIPT case present on stage - Healthy collie
- [x] Clean illumina runs
  - [x] `cg clean illumina-runs --dry-run`
[...]

PR #4308, opened 2025-03-25
Patch(cleaning) Delete correct sequencing directories

Vince-janv:
## Description
See issue #4307

This PR changes the logic back to what it was before. See this comment: https://github.com/Clinical-Genomics/cg/pull/3349/files#r2011987641

### Fixed

- The deletion now occurs on the actual directory provided

### DIFF
--- a/cg/services/illumina/cleaning/clean_runs_service.py
+++ b/cg/services/illumina/cleaning/clean_runs_service.py
@@ -62,14 +62,12 @@ class IlluminaCleanRunsService:
             self.set_sample_sheet_path_from_housekeeper()
             if self.can_run_directory_be_deleted():
                 if self.dry_run:
-                    LOG.debug(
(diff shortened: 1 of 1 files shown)


cg/services/illumina/cleaning/clean_runs_service.py
```diff
@@ -69,9 +66,7 @@ def delete_run_directory(self) -> None:
                         f"Dry run: Would have removed: {self.sequencing_run_dir_data.get_sequencing_runs_dir()}"
```
islean, line 66: This should match right?

Vince-janv: True, changing there as well 👍

islean (approved):
Nice! Because we loop through the sequencing runs dir and the demultiplexed runs dir in the command right? But the change introduced made us always go to the sequencing runs dir for cleaning?

Vince-janv:
Deployed in prod

PR #4312, opened 2025-03-26
Patch2(cleaning) Delete correct sequencing directories

Vince-janv:
## Description
See #4307 for issue descruption. The previous PR (#4308) fixed the directories that were supposed to be deleted, but the checks performed before deletion still used the wrong path. This PR corrects that and no other usage of the old method exists in the cleaning service

### Changed

- Date check is now done on the correct directory


islean (approved):
💪

Vince-janv:
Deployed in production

PR #4331, opened 2025-04-08
Add rule for resetting capture kit

islean:
## Description

When the bait set/capture kit of a sample is not required but set either way, we should reset it since any errors regarding it will not be fixable in the order portal. The field will not be rendered since it is not needed. This PR adds a rule which sets the capture kit to None if it is provided but not needed, and which returns a CaseSampleError for the field "warnings", which is to be used for customer feedback.

### Added

- reset_optional_capture_kits rule

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/orders/validation/errors/case_sample_errors.py
```diff
@@ -142,6 +142,11 @@ class SexSubjectIdError(CaseSampleError):
     message: str = "Another sample with the same subject id has a different sex"
 
 
+class CaptureKitResetError(CaseSampleError):
+    field: str = "warnings"
+    message: str = "Capture Kit was reset since it will not be used with this application."
```
diitaz93, line 147: The rest of the messages don't end with point
```suggestion
    message: str = "Capture Kit was reset since it will not be used with this application"
```

diitaz93 (commented):
I would like to see a demo of how it works 😜 I am not sure how it looks when the warning is rendered

islean:
This is how the response looks:

image

tests/services/orders/validation_service/workflows/balsamic/test_rules.py
Vince-janv, line 1: Good tests 👍

Vince-janv (approved):
Painful but necessary 😬

islean:
Tested in the FE and feedback is given to the customer 

image

PR #4333, opened 2025-04-09
Choose correct scout instance

islean:
## Description

With Nallo and other pipelines using genome build hg38, we will have two separate Scout instances running. This PR intends to adapt cg to use the correct instance for the correct workflow. In practice, this means uploading Nallo cases to the new instances. When Raredisease is to be implemented, we need some more intricate solution, since the workflow alone will not be enough to distinguish.

Dependent on a servers PR which adds the binary path and config path of the new scout instance. 

### Added

-

### Changed

- Two instances of the Scout api will now be instantiated in the CGConfig.
- The workflow is used to distinguish which scout instance should be used.

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b choose-correct-scout-instance -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/apps/scout/scoutapi.py
+++ b/cg/apps/scout/scoutapi.py
@@ -20,9 +20,9 @@ LOG = logging.getLogger(__name__)
 class ScoutAPI:
     """Interface to Scout."""
 
-    def __init__(self, config, slurm_upload_service: SlurmUploadService):
-        binary_path = config["scout"]["binary_path"]
-        config_path = config["scout"]["config_path"]
+    def __init__(self, scout_config, slurm_upload_service: SlurmUploadService):
+        binary_path = scout_config.binary_path
+        config_path = scout_config.config_path
         self.process = Process(binary=binary_path, config=config_path)
         self.slurm_upload_service = slurm_upload_service
         self.scout_base_command = f"{binary_path} --config {config_path}"
--- a/cg/cli/clean.py
+++ b/cg/cli/clean.py
@@ -122,7 +122,7 @@ def scout_finished_cases(
     context: click.Context, days_old: int, skip_confirmation: bool = False, dry_run: bool = False
 ) -> None:
     """Clean up of solved and archived Scout cases."""
-    scout_api: ScoutAPI = context.obj.scout_api
+    scout_api: ScoutAPI = context.obj.scout_api_37
     bundles: list[str] = []
     for status in [ScoutTag.ARCHIVED, ScoutTag.SOLVED]:
         cases: list[ScoutExportCase] = scout_api.get_cases(status=status, reruns=False)
--- a/cg/cli/upload/mutacc.py
+++ b/cg/cli/upload/mutacc.py
@@ -7,7 +7,7 @@ import rich_click as click
 from cg.apps.mutacc_auto import MutaccAutoAPI
 from cg.apps.scout.scout_export import ScoutExportCase
 from cg.apps.scout.scoutapi import ScoutAPI
-from cg.constants import Workflow
+from cg.cli.upload.utils import get_scout_api
 from cg.constants.cli_options import DRY_RUN
 from cg.meta.upload.mutacc import UploadToMutaccAPI
 from cg.models.cg_config import CGConfig
@@ -29,10 +29,7 @@ def process_solved(
 
     LOG.info("----------------- PROCESS-SOLVED ----------------")
 
-    workflow = context.status_db.get_case_by_internal_id(case_id).data_analysis
-    scout_api: ScoutAPI = (
-        context.scout_api_38 if workflow == Workflow.NALLO else context.scout_api_37
-    )
+    scout_api: ScoutAPI = get_scout_api(cg_config=context, case_id=case_id)
     mutacc_auto_api: MutaccAutoAPI = context.mutacc_auto_api
     mutacc_upload_api = UploadToMutaccAPI(scout_api=scout_api, mutacc_auto_api=mutacc_auto_api)
 
--- a/cg/cli/upload/mutacc.py
+++ b/cg/cli/upload/mutacc.py
@@ -7,6 +7,7 @@ import rich_click as click
 from cg.apps.mutacc_auto import MutaccAutoAPI
 from cg.apps.scout.scout_export import ScoutExportCase
 from cg.apps.scout.scoutapi import ScoutAPI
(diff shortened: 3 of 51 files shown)


tests/conftest.py
```diff
@@ -2213,10 +2213,14 @@ def context_config(
             },
             "tower_workflow": "taxprofiler",
         },
-        "scout": {
```
islean, line 2216: Is this something we want to change? It feels "correct" but could be an instance of "don't fix what's not broken".

cg/cli/clean.py
```diff
@@ -122,7 +122,7 @@ def scout_finished_cases(
     context: click.Context, days_old: int, skip_confirmation: bool = False, dry_run: bool = False
 ) -> None:
     """Clean up of solved and archived Scout cases."""
-    scout_api: ScoutAPI = context.obj.scout_api
+    scout_api: ScoutAPI = context.obj.scout_api_37
```
islean, line 125: Should this loop through both instances? Maybe a question for DN and VJ...

cg/apps/scout/scoutapi.py
```diff
@@ -20,9 +20,9 @@
 class ScoutAPI:
     """Interface to Scout."""
 
-    def __init__(self, config, slurm_upload_service: SlurmUploadService):
-        binary_path = config["scout"]["binary_path"]
-        config_path = config["scout"]["config_path"]
+    def __init__(self, scout_config, slurm_upload_service: SlurmUploadService):
```
islean, line 23: You are not allowed to type hint this because of circular imports. Basically the ScoutConfig data model is by convention defined in the CGConfig module. But the CGConfig module imports the ScoutAPI since it is one of the "MetaAPIs". This seems like a strong indication that our setup with CGConfig and the meta apis should be thought through. But is it within scope for this PR?

diitaz93: not in scope but definetly a red flag

cg/cli/clean.py
diitaz93: That could be a way. Also we could have two services with different frequencies, as I imagine there will be much fewer hg38 cases at the beginning. But out of scope for now, this will be solved for Raredisease

tests/cli/upload/conftest.py
```diff
@@ -13,11 +13,7 @@
 from cg.apps.tb import TrailblazerAPI
 from cg.constants import DELIVERY_REPORT_FILE_NAME
 from cg.constants.constants import FileFormat
-from cg.constants.housekeeper_tags import (
-    HK_DELIVERY_REPORT_TAG,
-    GensAnalysisTag,
-    HkMipAnalysisTag,
-)
+from cg.constants.housekeeper_tags import HK_DELIVERY_REPORT_TAG, GensAnalysisTag, HkMipAnalysisTag
```
diitaz93, line 16: not again the linter 😮‍💨

tests/conftest.py
diitaz93: This fixture should reflect the state of the config file in servers, so I would say lets keep this change

diitaz93 (approved):
LGTM 👍 I assume this comes with a PR in servers to reflect these changes

tests/conftest.py
islean: I was more thinking if we want to rename the old instance in servers, which is currently called "scout" to "scout_37"? Or should we leave the old instance as "scout" in servers?

cg/cli/upload/mutacc.py
```diff
@@ -28,7 +29,10 @@ def process_solved(
 
     LOG.info("----------------- PROCESS-SOLVED ----------------")
 
-    scout_api: ScoutAPI = context.scout_api
+    workflow = context.status_db.get_case_by_internal_id(case_id).data_analysis
+    scout_api: ScoutAPI = (
+        context.scout_api_38 if workflow == Workflow.NALLO else context.scout_api_37
+    )
```
islean, line 35: Maybe a utils method

islean: It is now a utils method

cg/models/cg_config.py
```diff
@@ -449,8 +454,10 @@ class CGConfig(BaseModel):
     pigz: CommonAppConfig | None = None
     run_names_services_: RunNamesServices | None = None
     sample_sheet_api_: IlluminaSampleSheetService | None = None
-    scout: CommonAppConfig = None
-    scout_api_: ScoutAPI = None
+    scout: ScoutConfig = None
+    scout_38: ScoutConfig = None
```
diitaz93, line 458: why do we need to create a new class `ScoutConfig` if we could use the `CommonAppConfig`? Seems nice but implies some code duplication

cg/cli/upload/utils.py
```diff
@@ -20,3 +22,8 @@ def suggest_cases_to_upload(status_db: Store, workflow: Workflow | None = None)
     ]
     for case_obj in records:
         click.echo(case_obj)
+
+
+def get_scout_api(cg_config: CGConfig, case_id: str) -> ScoutAPI:
+    workflow = cg_config.status_db.get_case_by_internal_id(case_id).data_analysis
```
diitaz93, line 28:
```suggestion
    workflow: Workflow = cg_config.status_db.get_case_by_internal_id(case_id).data_analysis
```

diitaz93 (approved):
Looks great ⭐

cg/models/cg_config.py
islean: Maybe not necessary. I can revert it. I just dislike the CommonAppConfig and prefer code duplication over the unused field. But seems as if the CommonAppConfig actually should not lead to problems, if you avoid referencing the container_mount_volume which is not used here

tests/conftest.py
diitaz93: lets leave it for now and maybe rename it later

islean:
Deployed to Hasta:

```
repository is clean
Logging deploy ...
Getting deployer... done.
done.
Log deploy... done.
cg, version 69.3.4
```

Issue #4251, opened 2025-02-27
Allow the user to specify the amount of cases to start with "start-available" for balsamic and MIP-DNA


karlnyr:
As a production bioinformatics member,
I want to have the option to specify the number of cases that we start automatically,
So that we can circumvent possible overlapping starts of cases that will make the analysis fail.

# Work impact

Answer the following questions:
- Is there currently a workaround for this issue? If so, what is it?
    - No, there is a feature in CG currently to limit the number of cases to get when we use the start available. However, we have no way of passing the limit through CLI, nor is the limit hardcoded.
- How much time would be saved by implementing this feature on a weekly basis?
    - Hours spent currently due to the state of hasta, and the amount of restarts we have to do
- How many users are affected by this issue?
    - Production bioinformatics members
- Are customers affected by this issue?
    - Yes, analyses get further delays due to the number of restarts we have to do.

# Acceptance Criteria

- [ ] The user can specify a threshold for the number of cases to start automatically through the CLI
- [ ] This should at least be available for balsamic and MIP, but could be available for all workflows as well

# Notes

- Additional information.
- Dependencies.
- Related user stories.

Vince-janv:
## Technical refinement
- Add a -limit flag to MIP and Balsamic start-available.
- Pass this limit down the call chain to cg.store.crud.read.ReadHandler.get_cases_to_analyze

PR #4381, opened 2025-05-06
Add limit flag for mip and balsamic start-available

beatrizsavinhas:
## Description

Closes https://github.com/Clinical-Genomics/cg/issues/4251#event-17514707005.

### Added

- limit flag for number of cases to start with the start-available command for mip and balsamic.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Deploy this branch on production.

### DIFF
--- a/cg/cli/workflow/balsamic/base.py
+++ b/cg/cli/workflow/balsamic/base.py
@@ -20,7 +20,7 @@ from cg.cli.workflow.balsamic.options import (
 from cg.cli.workflow.commands import ARGUMENT_CASE_ID, link, resolve_compression
 from cg.cli.workflow.utils import validate_force_store_option
 from cg.constants import EXIT_FAIL, EXIT_SUCCESS
-from cg.constants.cli_options import DRY_RUN, FORCE, COMMENT
+from cg.constants.cli_options import COMMENT, DRY_RUN, FORCE
 from cg.exc import AnalysisNotReadyError, CgError
 from cg.meta.workflow.analysis import AnalysisAPI
 from cg.meta.workflow.balsamic import BalsamicAnalysisAPI
@@ -238,14 +238,18 @@ def start(
 
 @balsamic.command("start-available")
 @DRY_RUN
+@click.option(
+    "--limit",
+    help="Maximum number of analyses to start",
+)
 @click.pass_context
-def start_available(context: click.Context, dry_run: bool = False):
+def start_available(context: click.Context, dry_run: bool = False, limit: int = None):
     """Start full workflow for all cases ready for analysis"""
 
     analysis_api: AnalysisAPI = context.obj.meta_apis["analysis_api"]
 
     exit_code: int = EXIT_SUCCESS
-    for case in analysis_api.get_cases_ready_for_analysis():
+    for case in analysis_api.get_cases_ready_for_analysis(limit=limit):
         try:
             context.invoke(start, case_id=case.internal_id, dry_run=dry_run)
         except AnalysisNotReadyError as error:
--- a/cg/meta/workflow/analysis.py
+++ b/cg/meta/workflow/analysis.py
@@ -123,12 +123,12 @@ class AnalysisAPI(MetaAPI):
             case_is_set_to_analyze or case_has_not_been_analyzed or case_latest_analysis_failed
         )
 
-    def get_cases_ready_for_analysis(self):
+    def get_cases_ready_for_analysis(self, limit: int = None):
         """
         Return cases that are ready for analysis. The case is ready if it passes the logic in the
         get_cases_to_analyze method, and it has passed the pre-analysis quality check.
         """
-        cases_to_analyse: list[Case] = self.get_cases_to_analyze()
+        cases_to_analyse: list[Case] = self.get_cases_to_analyze(limit=limit)
         cases_passing_quality_check: list[Case] = [
             case for case in cases_to_analyse if SequencingQCService.case_pass_sequencing_qc(case)
         ]
@@ -375,8 +375,8 @@ class AnalysisAPI(MetaAPI):
         )
         return analyses_to_clean
 
-    def get_cases_to_analyze(self) -> list[Case]:
-        return self.status_db.get_cases_to_analyze(workflow=self.workflow)
+    def get_cases_to_analyze(self, limit: int = None) -> list[Case]:
+        return self.status_db.get_cases_to_analyze(workflow=self.workflow, limit=limit)
 
     def get_cases_to_store(self) -> list[Case]:
         """Return cases where analysis finished successfully,
--- a/cg/meta/workflow/balsamic.py
+++ b/cg/meta/workflow/balsamic.py
@@ -93,9 +93,9 @@ class BalsamicAnalysisAPI(AnalysisAPI):
         """Returns a path where the Balsamic case for the case_id should be located"""
         return Path(self.root_dir, case_id)
 
-    def get_cases_ready_for_analysis(self) -> list[Case]:
+    def get_cases_ready_for_analysis(self, limit: int = None) -> list[Case]:
         """Returns a list of cases that are ready for analysis."""
-        cases_to_analyse: list[Case] = self.get_cases_to_analyze()
+        cases_to_analyse: list[Case] = self.get_cases_to_analyze(limit=limit)
         cases_ready_for_analysis: list[Case] = [
             case for case in cases_to_analyse if self.is_case_ready_for_analysis(case)
         ]
--- a/cg/meta/workflow/mip.py
+++ b/cg/meta/workflow/mip.py
@@ -249,9 +249,9 @@ class MipAnalysisAPI(AnalysisAPI):
                 return True
         return False
 
-    def get_cases_ready_for_analysis(self) -> list[Case]:
+    def get_cases_ready_for_analysis(self, limit: int = None) -> list[Case]:
         """Return cases to analyze."""
-        cases_to_analyse: list[Case] = self.get_cases_to_analyze()
+        cases_to_analyse: list[Case] = self.get_cases_to_analyze(limit=limit)
         cases_ready_for_analysis: list[Case] = [
             case for case in cases_to_analyse if self.is_case_ready_for_analysis(case)
         ]
--- a/cg/cli/workflow/mip/base.py
+++ b/cg/cli/workflow/mip/base.py
@@ -199,14 +199,18 @@ def start(
 
 @click.command("start-available")
 @DRY_RUN
+@click.option(
+    "--limit",
+    help="Maximum number of analyses to start",
+)
 @click.pass_context
-def start_available(context: click.Context, dry_run: bool = False):
+def start_available(context: click.Context, dry_run: bool = False, limit: int = None):
     """Start full analysis workflow for all cases ready for analysis."""
 
     analysis_api: MipAnalysisAPI = context.obj.meta_apis["analysis_api"]
 
     exit_code: int = EXIT_SUCCESS
-    for case in analysis_api.get_cases_ready_for_analysis():
+    for case in analysis_api.get_cases_ready_for_analysis(limit=limit):
         try:
             context.invoke(start, case_id=case.internal_id, dry_run=dry_run)
         except AnalysisNotReadyError as error:


tests/store_helpers.py
beatrizsavinhas, line 1: See comment above - https://github.com/Clinical-Genomics/cg/pull/4381/files#r2077750784.

tests/cli/workflow/mip/test_cli_mip_dna_start.py
```diff
@@ -1,24 +1,49 @@
...
 def test_dry(cli_runner, mip_dna_context):
     """Test mip dna start with --dry option"""
 
-    # GIVEN a mip_dna_context
+    # GIVEN a mip_dna_context with several cases that are ready to analyse
 
-    # WHEN using dry running
-    result = cli_runner.invoke(start_available, ["--dry-run"], obj=mip_dna_context)
+    # GIVEN that the cases do not need decompression
+    with mock.patch.object(MipDNAAnalysisAPI, "resolve_decompression", return_value=None):
```
islean, line 18: This is odd to me - was this part of the flow touched?

tests/cli/workflow/mip/test_cli_mip_dna_start.py
```diff
@@ -1,24 +1,49 @@
...
     assert result.exit_code == EXIT_SUCCESS
 
 
+def test_start_available_with_limit(cli_runner, mip_dna_context, caplog):
+    caplog.set_level(logging.INFO)
+
+    # GIVEN a mip_dna_context with several cases that are ready to analyse
+
+    # GIVEN that the cases do not need decompression
+    with mock.patch.object(MipDNAAnalysisAPI, "resolve_decompression", return_value=None):
```
islean, line 33: Same question here - given that resolving decompression should not have been touched, why do we need to mock it here?

tests/cli/workflow/mip/test_cli_mip_dna_start.py
```diff
@@ -1,24 +1,49 @@
...
-    # GIVEN a mip_dna_context
+    # GIVEN a mip_dna_context with several cases that are ready to analyse
 
-    # WHEN using dry running
-    result = cli_runner.invoke(start_available, ["--dry-run"], obj=mip_dna_context)
+    # GIVEN that the cases do not need decompression
+    with mock.patch.object(MipDNAAnalysisAPI, "resolve_decompression", return_value=None):
+
+        # WHEN using dry running
+        result = cli_runner.invoke(start_available, ["--dry-run"], obj=mip_dna_context)
```
islean, line 21: Could we add to this test something that makes sure that the behaviour is different in this test compared to the next?

cg/meta/workflow/analysis.py
```diff
@@ -123,7 +123,7 @@ def is_case_ready_for_analysis(self, case: Case) -> bool:
             case_is_set_to_analyze or case_has_not_been_analyzed or case_latest_analysis_failed
         )
 
-    def get_cases_ready_for_analysis(self):
+    def get_cases_ready_for_analysis(self, limit: int = None):
```
islean, line 126: Is this used by MIP or Balsamic? I thought they had their own AnalysisAPIs which are changed in the upcoming files?

cg/meta/workflow/mip.py
```diff
@@ -249,12 +249,14 @@ def get_skip_evaluation_flag(self, case_id: str, skip_evaluation: bool) -> bool:
...
-    def get_cases_ready_for_analysis(self) -> list[Case]:
+    def get_cases_ready_for_analysis(self, limit: int = None) -> list[Case]:
         """Return cases to analyze."""
         cases_to_analyse: list[Case] = self.get_cases_to_analyze()
         cases_ready_for_analysis: list[Case] = [
             case for case in cases_to_analyse if self.is_case_ready_for_analysis(case)
         ]
+        if limit:
+            return cases_ready_for_analysis[:limit]
         return cases_ready_for_analysis[:MAX_CASES_TO_START_IN_50_MINUTES]
```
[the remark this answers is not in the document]
beatrizsavinhas: Ah! Nice idea with passing 'MAX_CASES_TO_START_IN_50_MINUTES' as default!!! Will change that.

cg/meta/workflow/analysis.py
beatrizsavinhas: You are correct! This is a remnant of my initial attempt at implementing this which had different logic. Will remove this!

tests/cli/workflow/mip/test_cli_mip_dna_start.py
beatrizsavinhas: Absolutely! I changed the test minimally to make sense (instead of the void test we had before) but I agree it would be better to add one more check!

tests/cli/workflow/mip/test_cli_mip_dna_start.py
beatrizsavinhas: This part of the flow was not touched, however, by making it so that 'mip_dna_context' actually has cases ready to be analysed, 'resolve_decompression' is now called for each case and would raise an error because files/bundles are not found for those cases. So I have to mock that to continue with the test!

tests/cli/workflow/mip/test_cli_mip_dna_start.py
beatrizsavinhas: See my reply to your previous comment!

cg/meta/workflow/balsamic.py
```diff
@@ -93,13 +93,15 @@ def get_case_path(self, case_id: str) -> Path:
         """Returns a path where the Balsamic case for the case_id should be located"""
         return Path(self.root_dir, case_id)
 
-    def get_cases_ready_for_analysis(self) -> list[Case]:
+    def get_cases_ready_for_analysis(
+        self, limit: int = MAX_CASES_TO_START_IN_50_MINUTES
```
diitaz93, line 97: The way this method is called (in Balsamic and MIP-DNA), the default value will never be used: 
1. If you provide a limit through the CLI, that number will be used as limit. 
2. If you don't provide anything, limit will be `None` and that value will be propagated here, overriding the default and not filtering the list. 

If you want the default to be used when no limit is provided you can:
1. Add a default to the flag
2. do a check in this function `limit = MAX_CASES_TO_START_IN_50_MINUTES if limit is None else limit` or something like that

tests/cli/workflow/balsamic/test_compound_commands.py
```diff
@@ -140,6 +132,33 @@ def test_start_available(
     assert f"Starting analysis for {case_id_not_enough_reads}" not in caplog.text
 
 
+def test_start_available_with_limit(
```
diitaz93, line 135: I think it is important to test the behaviour of `BalsamicAnalysisAPI.get_cases_ready_for_analysis` too, with and without limits. Same for MIP-DNA. The tests for the CLI command are complicated: There is no verification that "balsamic_case_tgs_single" is a case ready to analyse, so the fact that it is not in the logs does not guarantee that it was filtered out. Excluding this case from the context fixture will not affect the behaviour of the test.

diitaz93 (commented):
I left some comments on how the default value of the `get_cases_ready_for_analysis` might not be used. It would be important to get some tests on these intermediate functions

cg/cli/workflow/balsamic/base.py
```diff
@@ -238,14 +238,19 @@ def start(
 
 @balsamic.command("start-available")
 @DRY_RUN
+@click.option(
```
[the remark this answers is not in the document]
beatrizsavinhas: Will add the decorator! Was unsure if it would be worth doing since there's only two uses, but since you also suggest I agree it's nice to add!

I thought about the short flag version too, but this flag is not meant to be used often so I think it is better to avoid confusion.

tests/cli/workflow/balsamic/test_compound_commands.py
beatrizsavinhas: I agree there was a flaw to this test because there is no check that there are actually several cases available to analyse!
Would this be an appropriate way to make sure of that?
I have changed it also in the mip tests. Should I change the original balsamic start-available test to reflect this logic?

cg/meta/workflow/balsamic.py
```diff
@@ -93,13 +93,13 @@ def get_case_path(self, case_id: str) -> Path:
...
 
-    def get_cases_ready_for_analysis(self) -> list[Case]:
+    def get_cases_ready_for_analysis(self, limit: int | None = None) -> list[Case]:
         """Returns a list of cases that are ready for analysis."""
         cases_to_analyse: list[Case] = self.get_cases_to_analyze()
         cases_ready_for_analysis: list[Case] = [
             case for case in cases_to_analyse if self.is_case_ready_for_analysis(case)
         ]
-        return cases_ready_for_analysis[:MAX_CASES_TO_START_IN_50_MINUTES]
+        return cases_ready_for_analysis[: limit if limit else MAX_CASES_TO_START_IN_50_MINUTES]
```
beatrizsavinhas, line 102: Added condition to ensure the default limit is used.

cg/meta/workflow/balsamic.py
beatrizsavinhas: You are right! I changed the logic to address Isak's comment and missed that.

Would this be a good way to do this? https://github.com/Clinical-Genomics/cg/pull/4381/files#r2079489594

cg/meta/workflow/balsamic.py
```diff
@@ -93,13 +93,13 @@ def get_case_path(self, case_id: str) -> Path:
         """Returns a path where the Balsamic case for the case_id should be located"""
         return Path(self.root_dir, case_id)
 
-    def get_cases_ready_for_analysis(self) -> list[Case]:
+    def get_cases_ready_for_analysis(self, limit: int | None = None) -> list[Case]:
```
Vince-janv, line 96: Why not just have MAX_CASES_TO_START_IN_50_MINUTES be the default and skip the if/else in the return?
```suggestion
    def get_cases_ready_for_analysis(self, limit: int = MAX_CASES_TO_START_IN_50_MINUTES) -> list[Case]:
```

beatrizsavinhas: See earlier comment from @diitaz93! 😉 
https://github.com/Clinical-Genomics/cg/pull/4381#discussion_r2079130053

tests/cli/workflow/balsamic/test_compound_commands.py
```diff
@@ -140,6 +133,40 @@ def test_start_available(
...
+    case: Case = balsamic_context.status_db.get_case_by_internal_id(
+        internal_id="balsamic_case_tgs_single"
+    )
+    for sample in case.samples:
+        sample.reads = sample.expected_reads_for_sample
+    balsamic_context.status_db.session.commit()
+
+    # GIVEN that there are now 2 cases that are ready to analyse
+    analysis_api = BalsamicAnalysisAPI(config=balsamic_context)
+    assert len(analysis_api.get_cases_ready_for_analysis()) == 2
```
diitaz93, line 155: 👍

tests/cli/workflow/balsamic/test_compound_commands.py
```diff
@@ -140,6 +133,40 @@ def test_start_available(
...
+    mocker.patch.object(BalsamicAnalysisAPI, "resolve_decompression", return_value=None)
+
+    # WHEN running command with limit=1
+    result = cli_runner.invoke(start_available, ["--dry-run", "--limit", 1], obj=balsamic_context)
+
+    # THEN command exits with a successful exit code
+    assert result.exit_code == EXIT_SUCCESS
+
+    # THEN only 1 case is picked up to start
+    assert caplog.text.count("Starting analysis for") == 1
```
diitaz93, line 167: this looks much better 👍

beatrizsavinhas: Thank youuuuu! 😁

tests/cli/workflow/balsamic/test_compound_commands.py
diitaz93: These tests that involve a `cli_runner.invoke` are very hard because there is little you can test apart from the exit status, as relying on testing logs is quite brittle. We want to get rid of these in the future, so I would not invest too much time on them, and instead try to test all the intermediate functions. So I think it is not necessary to change any other start-available test. Having said that, this test looks great, it tests what we need despite relying a bit on logs. I think it is still important to test the `get_cases_ready_for_analysis` with and without limit.

cg/meta/workflow/balsamic.py
Vince-janv: Sorry! Disregard my comment 👍

diitaz93 (approved):
Looks great 💯 I would still add two more tests for the get_cases_ready_for_analysis of MIP and Balsamic

tests/cli/workflow/balsamic/test_compound_commands.py
beatrizsavinhas: Added the additional tests now! Thank you for all the feedback! 🙂

tests/cli/workflow/balsamic/test_balsamic_analysis_api.py
```diff
@@ -0,0 +1,22 @@
+from cg.meta.workflow.balsamic import BalsamicAnalysisAPI
+
+
+def test_get_cases_ready_for_analysis(balsamic_analysis_api: BalsamicAnalysisAPI):
+
+    # GIVEN a balsamic_context with 1 case that is ready for analysis
+
+    # WHEN running the command
+    cases_to_analyse = balsamic_analysis_api.get_cases_ready_for_analysis()
```
diitaz93, line 9:
```suggestion
    cases_to_analyse list[Case] = balsamic_analysis_api.get_cases_ready_for_analysis()
```

tests/cli/workflow/balsamic/test_balsamic_analysis_api.py
```diff
@@ -0,0 +1,22 @@
...
+
+    # THEN only 1 case is retrieved
+    assert len(cases_to_analyse) == 1
+
+
+def test_get_cases_ready_for_analysis_with_limit(balsamic_analysis_api: BalsamicAnalysisAPI):
+    # GIVEN a balsamic_context with 1 case that is ready for analysis
+
+    # WHEN running the command with limit=0
+    cases_to_analyse = balsamic_analysis_api.get_cases_ready_for_analysis(limit=0)
```
diitaz93, line 19:
```suggestion
    cases_to_analyse: list [Case] = balsamic_analysis_api.get_cases_ready_for_analysis(limit=0)
```

tests/cli/workflow/balsamic/test_balsamic_analysis_api.py
beatrizsavinhas: Added!

PR #4383, opened 2025-05-06
(Pipeline integration) Add subprocess submitter

### LINKED ISSUE #4215: Create the SubProcessSubmitter
Create the SubProcessSubmitter.

It should have the methods `submit`, which takes a `CaseConfig` as input. 
This method should:

1. Fetch a template based on the workflow in the CaseConfig. 
2. Fill out the template with its content 
3. Start the command as a subprocess

islean:
## Description

Closes #4215. Adds a SubProcessSubmitter intended to be used for all Slurm pipelines

### Added

- SubProcessSubmitter

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/analysis_starter/configurator/models/microsalt.py
+++ b/cg/services/analysis_starter/configurator/models/microsalt.py
@@ -1,7 +1,6 @@
-from pathlib import Path
-
 from pydantic import BaseModel
 
+from cg.constants import Workflow
 from cg.services.analysis_starter.configurator.abstract_model import CaseConfig
 
 
@@ -24,4 +23,9 @@ class MicrosaltConfigContent(BaseModel):
 
 
 class MicrosaltCaseConfig(CaseConfig):
-    config_file_path: Path
+    binary: str
+    conda_binary: str
+    config_file: str
+    environment: str
+    fastq_directory: str
+    workflow: Workflow = Workflow.MICROSALT
--- a/tests/services/analysis_starter/test_subprocess_submitter.py
+++ b/tests/services/analysis_starter/test_subprocess_submitter.py
@@ -1,0 +1,36 @@
+import subprocess
+from unittest import mock
+
+from cg.services.analysis_starter.configurator.models.microsalt import MicrosaltCaseConfig
+from cg.services.analysis_starter.submitters.subprocess.submitter import SubprocessSubmitter
+
+
+def test_subprocess_submitter():
+    # GIVEN a SubProcessSubmitter
+    subprocess_submitter = SubprocessSubmitter()
+
+    # WHEN submitting a case using a valid case config
+    microsalt_case_config = MicrosaltCaseConfig(
+        binary="/path/to/microsalt/binary",
+        case_id="microsalt_case",
+        conda_binary="/path/to/conda/binary",
+        config_file="/path/to/microsalt_case/config.json",
+        environment="S_microsalt",
+        fastq_directory="/path/to/microsalt_case/fastq",
+    )
+    with mock.patch.object(subprocess, "run") as mock_run:
+        subprocess_submitter.submit(microsalt_case_config)
+
+        # THEN the analysis should have been started
+        expected_args = (
+            "/path/to/conda/binary run --name S_microsalt "
+            "/path/to/microsalt/binary analyse /path/to/microsalt_case/config.json "
+            "--input /path/to/microsalt_case/fastq"
+        )
+        mock_run.assert_called_once_with(
+            args=expected_args,
+            shell=True,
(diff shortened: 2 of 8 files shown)


cg/services/analysis_starter/configurator/models/microsalt.py
```diff
@@ -24,4 +23,9 @@ class MicrosaltConfigContent(BaseModel):
 
 
 class MicrosaltCaseConfig(CaseConfig):
```
islean, line 25: I was a bit unsure of where to put the pipeline parameters (binary, conda_binary etc). Should they be here?

diitaz93: Isn't the conda binary the same for several pipelines? Basically all nextflow? I think that should be a constant somewhere. The Microsalt binary could be defined as a frozen attribute with predefined value

diitaz93 (approved):
Looks good! I like the submitter 🙌  I think we could discuss where to place the conda binary, maybe is not needed as an attribute. It could be defined directly in the command, or in a constant somewhere else and loaded to the command.

tests/services/analysis_starter/test_subprocess_submitter.py
```diff
@@ -0,0 +1,36 @@
...
+    )
+    with mock.patch.object(subprocess, "run") as mock_run:
+        subprocess_submitter.submit(microsalt_case_config)
+
+        # THEN the analysis should have been started
+        expected_args = (
+            "/path/to/conda/binary run --name S_microsalt "
+            "/path/to/microsalt/binary analyse /path/to/microsalt_case/config.json "
+            "--input /path/to/microsalt_case/fastq"
+        )
```
mogmi, line 29: You could consider using f-strings inserting the values from the`microsalt_case_config` here to tie the expected outcome closer to the input, but the way it is might be more readable so it's not clear cut for me.

mogmi (approved):

cg/services/analysis_starter/configurator/models/microsalt.py
islean: @mogmi Exactly, I think it depends on whether it should contain "everything" needed for the case to start, or if it should be only the configurations specific to that case. I think the first option makes sense, whence this is the place to put the pipeline specific parameters as well.

@diitaz93 You are correct, but I do not think a constant is possible. The paths depend on stage/production and are fetched from the cg.yaml. A bit uncertain of what you mean with a frozen attribute for the microsalt binary but is that not what it is in the cg.yaml?

cg/services/analysis_starter/configurator/models/microsalt.py
```diff
@@ -24,4 +23,9 @@ class MicrosaltConfigContent(BaseModel):
...
-    config_file_path: Path
+    binary: str
+    conda_binary: str
```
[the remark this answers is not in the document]
islean: I think the MicrosaltConfig in this case has a faulty type hint. Or I have a very hard time seeing when this would not be set, looking at our config file in servers?

tests/services/analysis_starter/test_subprocess_submitter.py
islean: Yeah, that is probably an improvement!

cg/services/analysis_starter/configurator/models/microsalt.py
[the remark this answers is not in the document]
mogmi: I see that it was added [here](https://github.com/Clinical-Genomics/cg/commit/3bd0c3e7183cd5901257ade3fa95e5146370111a) where it is set to `None` if 
`config["crunchy"]["conda_binary"]` is falsy - my guess is to set it to None if the string is empty maybe?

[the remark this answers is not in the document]
islean: That commit I think only refactored from using "Optional" to the pipe operator. So this might be even older

cg/services/analysis_starter/configurator/models/microsalt.py
islean: Resolved in Huddle

tests/services/analysis_starter/test_subprocess_submitter.py
islean: Done

Issue #3765, opened 2024-09-24
Pass order prio or express status to scout on load


dnil:
As a clinical geneticist at KGG `cust002`, I would like cases I order with `priority` or `express` status to be immediately more visible than cases ordered on routine priority in Scout. See https://github.com/Clinical-Genomics/scout/issues/4892. One obvious option would be CG to pass the `status: prioritized` to Scout when constructing a case load config for a track `rare` case.

This would be one more level safeguarding against issues like https://github.com/Clinical-Genomics/Deviations/issues/698.

karlnyr:
@dnil Could we not use the same priority levels in Scout as we do at cg? 

Those would be priority: research, standard, priority, express

dnil:
We could export that as well if you prefer it, and translate that into a case status of prioritized on load instead. A-ok solution, whatever is easier. But order priority and case status are quite different things, so we couldn't eg just set the scout status to research or normal.

diitaz93:
Assigning @ChrOertlin to schedule a meeting with @dnil to discuss the implementation and the urgency

Vince-janv:
@dnil has this been requested by scout users? I don't see us addressing it in the near future otherwise

dnil:
Thank you for your confidence sir. 😛 

It has indeed been requested by Scout users. And not only this particular KGG clinical geneticist, also others.

It's ok to have a meeting if you must, but if so perhaps one where we just fix it together? To the casual onlooker it in cg involves printing one string to a file if a particular string is set in the db. And in scout involves reading one string from said file and setting a corresponding value in its database.

Vince-janv:
Great. I had some worries regarding only doing this for the `rare` track. If we can apply the translation mentioned to all scout uploads I'm sure this should not be that big of an issue

dnil:
Hi, we got a ping on this one again. I guess we will be busy with other things this week, but it's getting to be around that time. 😊

dnil:
Ping @Clinical-Genomics/sysdev - I think this probably has been left with an outdated status, as feedback is present earlier in the thread. Or do you need anything else?

Vince-janv:
Thanks for the ping @dnil. I agree that status and priority are not really the same and equating them might be problematic in the future. How difficult would it be to add priority to scout cases (and i guess in the load config)? Translating the statusDB ones to the ones you choose to have in scout should be no problem

dnil:
Hi there! 

[...]

PR #4392, opened 2025-05-09
Include prioritized status in ScoutLoadConfig

beatrizsavinhas:
## Description

Closes https://github.com/Clinical-Genomics/cg/issues/3765

### Added

- `status: prioritized` to the scout load config if the priority of the case is `priority` or `express`

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b 3765-pass-case-prio-to-scout-load-config -a
    ```

### How to test

- [x] Run `cg upload create-scout-load-config <case>` on a prioritized case
- [x] Run `cg upload create-scout-load-config <case>` on a prioritized case with this branch
- [x] Run `cg upload create-scout-load-config <case>` on a non prioritized case with this branch

### Expected test outcome

- [x] Check that status is not present in the file generated with master or the file generated for a non prioritized case with this branch.
- [x] Check that status is present in the file generated with this branch for a prioritized case.

## Review

- [X] Tests executed by BSV

## Implementation Plan

- [ ] Deploy this branch in Production


cg/meta/upload/scout/scout_config_builder.py
```diff
@@ -41,6 +41,7 @@ def add_common_info_to_load_config(
         load_config.family = analysis.case.internal_id
         load_config.family_name = analysis.case.name
         load_config.owner = analysis.case.customer.internal_id
+        load_config.status = "prioritized" if analysis.case.is_prioritized else None
```
beatrizsavinhas, line 44: Should this logic be in a separate funtion?

cg/store/models.py
```diff
@@ -572,6 +572,14 @@ def is_uploaded(self) -> bool:
         """Returns True if the latest connected analysis has been uploaded."""
         return self.analyses and self.analyses[0].uploaded_at
 
+    @property
```
beatrizsavinhas, line 575: Is it worth adding a Case property or should we do this in another way?

Vince-janv: I don't think the database models should care about scout statuses. Hence I think the logic should lie somewhere in the scout-uploading logic

cg/meta/upload/scout/scout_config_builder.py
Vince-janv: I would add the a constant called SCOUT_PRIORITIZED_STATUS in `cg/constants/subject.py`

cg/store/models.py
islean: I think for me it is confusing, since I would assume that it just checks if the priority of the case is "priority". This seems Scout specific, so I would prefer not having it on the Case. I would probably just do a

is_prioritized = bool(case.priority in [Priority.priority, Priority.express]

when used instead

cg/meta/upload/scout/scout_config_builder.py
Vince-janv: Finally a situation where we can use IntEnum as intended 🥳 If this looks messy to you can move it to a method
```suggestion
        load_config.status = SCOUT_PRIORITIZED_STATUS if analysis.case.priority >= Priority.priority else None
```

cg/store/models.py
islean: That is true!

Vince-janv (approved):
Nice! Good job testing as well. As a final test, can we just upload one case to scout stage to make sure the extra param works?

cg/meta/upload/scout/scout_config_builder.py
beatrizsavinhas: Ahhh! Great suggestions! Will change!

beatrizsavinhas:
# Tests (part 2)

## Scout upload in stage

Screenshot 2025-05-12 at 14 01 09

PR #4422, opened 2025-05-27
Log internal ids of selected cases

mogmi:
### Added

- Log every case that has been selected for analysis for testing purposes


diitaz93 (approved):
👍

PR #4424, opened 2025-05-28
Fix(Nallo) Link Nallo to tower clean command

Vince-janv:
## Description
Adds Nallo as a valid workflow for the `cg clean tower-past-run-dirs`command.

### Fixed

- Nallo is now a valid option for `cg clean tower-past-run-dirs


islean (approved):
💯

PR #4430, opened 2025-06-05
Feat(Vulture) Add Vulture pre-commit to cleanup dead code

Vince-janv:
## Description
Vulture is a popular tool for removing unused code in python. It was suggested in a meeting that it could be called automatically to help prune the codebase when refactoring modules.

### paths
The paths to the code you want to check. By selecting `cg` we exclude any usages in tests. The whitelist should also be added.

### False positives (The whitelist)
Some stuff is reported as dead code but is actually being used. The bulk of it is unused attributes in classes which felt like a hassel to go through, but also stuff like pydantic fields which are parsed, validated and then dumped are flagged as unused.

### Minimum Confidence
vulture reports three different kinds of confidence 60%, 80% and 100%, which you can set in the config. All our dead code was reported with 60%, hence that's what I set.

### ignore_decorators
Used to ignore functions/methods with given decorators. Used for ignoring functions that are never explicitly called, like CLI functions, endpoints, pydantic validators ETC.

### ignore_names
Used to ignore Classes/methods/functions matching a certain name pattern

## Cleaning procedure
Since the dead code is sorted by size I went from the bottom up and removed stuff that I deemed correctly marked as unused. pycharm and vulture had a 100% match rate, ie, pycharm never found usages that vulture missed.

### Added

- vulture to the pre-commit hooks
- vulture_whitelist.txt for ignoring false positives

### Changed

- Removed almost all methods flagged by vulture

## Tested:
- [x] Sample sheet generation
- [x] Flow cell post-processing
- [x] Starting pipeline (mip-dna)
- [x] Storing pipeline (mip-dna)
- [x] Uploading pipeline (scout, caesar, gens, genotype)
- [x] Placing an order


pyproject.toml
```diff
@@ -95,4 +95,11 @@ ensure_newline_before_comments = "true"
...
 target-version = "py311"
 exclude = ["alembic"]
-ignore = ["E501"]
\ No newline at end of file
+ignore = ["E501"]
+
+[tool.vulture]
+min_confidence = 60
+ignore_decorators = ["@click.*","*route*","@field_validator*","@model_validator*","@app.*","@validator*","@action*","@field_serializer"]
+ignore_names= ["on_model_change","*View(BaseView)"]
```
islean, line 103: Would it be nice to include `*(BaseModel)` here?

islean: Or will that just allow for debris? Thinking of how we can get rid of listing all the BaseModel attributes in the whitelist but maybe best is just to wait for the functionality. It is in Vulture's TODO.

tests/apps/demultiplex/test_override_cycles_validator.py
```diff
@@ -5,6 +5,20 @@
 from cg.exc import OverrideCyclesError
 
 
+def set_run_cycles(
```
islean, line 8: Why do we add this here?

islean (approved):
I think it looks good! My main gripe is with the whitelist which seems hard to maintain but it looks like Vulture is aiming to support excluding types of classes (e.g. all that inherit from BaseModel) in the future: https://github.com/jendrikseipp/vulture/issues/362.

cg/store/models.py
```diff
@@ -964,38 +883,6 @@ def __str__(self) -> str:
         return self.name
 
 
-class SampleLaneSequencingMetrics(Base):
```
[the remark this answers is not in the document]
islean: Is it?

PR #4431, opened 2025-06-05
fix delivery report creation with completed_at date

diitaz93:
## Description
Fix a current bug in production in which the delivery report generation tries to get a HK bundle version with a wrong datetime

### Fixed

- The version would be fetched through `completed_at` instead of `started_at` of the Analysis object
- The `completed_at` date of teh Analysis object will be the timestamp of the bundle generated by HermesAPI

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/meta/workflow/analysis.py
+++ b/cg/meta/workflow/analysis.py
@@ -315,7 +315,7 @@ class AnalysisAPI(MetaAPI):
             LOG.info("Dry-run: StatusDB changes will not be commited")
             return
         self.status_db.update_analysis_completed_at(
-            analysis_id=analysis.id, completed_at=datetime.now()
+            analysis_id=analysis.id, completed_at=self.get_bundle_created_date(case_id)
         )
         self.status_db.update_analysis_comment(analysis_id=analysis.id, comment=comment)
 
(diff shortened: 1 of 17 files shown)


cg/meta/workflow/analysis.py
```diff
@@ -315,7 +315,7 @@ def update_analysis_as_completed_statusdb(
             LOG.info("Dry-run: StatusDB changes will not be commited")
             return
         self.status_db.update_analysis_completed_at(
-            analysis_id=analysis.id, completed_at=datetime.now()
+            analysis_id=analysis.id, completed_at=self.get_bundle_created_date(case_id)
```
diitaz93, line 318: This is the core of the patch, we will sync the bundle date with the analysis completed_at date

islean, line 318: This seems off to me - is this what we want?

islean: Yeah, but using the completed at in a similar way for me would be to still use the datetime.now(). But that is mainly because I do not know if there is one analysis object per housekeeper bundle or not

mogmi: I fully agree about the timestamps, my suggestion is that we don't keep it like this for long and implement a better link pretty soon so we don't get too many analyses in the database with a misleading completed_at.

mogmi (approved):
I think it looks like a good way to get around the immediate issue 👍

PR #4435, opened 2025-06-09
patch wrong analysis set as complete in cases with multiple analyses

diitaz93:
## Description
The AnalysisAPI method `update_analysis_as_completed_statusdb` used in the `store` methods picked up the analysis with the latest `completed_at` date to be stored. Still, by this point all of the analyses of the case had `completed_at = None`, making the function pick up the wrong analysis for updating. This PR changes the sorting to `started_at` instead.

### Added

- New CRUD function `get_latest_started_analysis_for_case` with tests

### Changed

- AnalysisAPI method `update_analysis_as_completed_statusdb` so that it uses the new function

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/store/crud/read.py
+++ b/cg/store/crud/read.py
@@ -15,7 +15,14 @@ from cg.constants.constants import (
     SampleType,
 )
 from cg.constants.sequencing import DNA_PREP_CATEGORIES, SeqLibraryPrepCategory
-from cg.exc import CaseNotFoundError, CgDataError, CgError, OrderNotFoundError, SampleNotFoundError
+from cg.exc import (
+    AnalysisDoesNotExistError,
+    CaseNotFoundError,
+    CgDataError,
+    CgError,
+    OrderNotFoundError,
+    SampleNotFoundError,
+)
 from cg.models.orders.constants import OrderType
 from cg.models.orders.sample_base import SexEnum
 from cg.server.dto.samples.requests import CollaboratorSamplesRequest
@@ -186,6 +193,19 @@ class ReadHandler(BaseHandler):
             case_entry_id=case_entry_id,
             completed_at_date=completed_at_date,
         ).first()
+
+    def get_latest_started_analysis_for_case(self, case_id: str) -> Analysis:
+        """Return the latest started analysis for a case.
+        Raises:
+            AnalysisDoesNotExistError if no analysis is found.
+        """
+        case: Case = self.get_case_by_internal_id(case_id)
+        analyses: list[Analysis] = case.analyses
+        if not analyses:
+            raise AnalysisDoesNotExistError(f"No analysis found for case {case_id}")
+        analyses.sort(key=lambda x: x.started_at, reverse=True)
+        analysis: Analysis | None = analyses[0]
+        return analysis
 
     def get_analysis_by_entry_id(self, entry_id: int) -> Analysis | None:
         """Return an analysis."""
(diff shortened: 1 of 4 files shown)


cg/store/crud/read.py
```diff
@@ -187,6 +194,19 @@ def get_analysis_by_case_entry_id_and_completed_at(
...
+    def get_latest_started_analysis_for_case(self, case_id: str) -> Analysis:
+        """Return the latest started analysis for a case.
+        Raises:
+            AnalysisDoesNotExistError if no analysis is found.
+        """
+        case: Case = self.get_case_by_internal_id(case_id)
+        analyses: list[Analysis] = case.analyses
+        if not analyses:
+            raise AnalysisDoesNotExistError(f"No analysis found for case {case_id}")
+        analyses.sort(key=lambda x: x.started_at, reverse=True)
```
islean, line 206: Would it make sense to have this sorting order on case.analyses already?

diitaz93: Not sure which other function uses that, so as least invasive as possible with the changes

mogmi, line 206:
```suggestion
        started_analyses: list[Analysis] = [a for a in analyses if a.started_at is not None]
        started_analyses.sort(key=lambda a: cast(datetime, a.started_at), reverse=True)
        analysis: Analysis | None = started_analyses[0]
```
Since it's possible for started_at to not be set I think we need to handle it, otherwise this method would throw an exception when trying to sort the list. I propose that we first filter them out and then I add a `cast` to convince the type checker that we at this point are sure that there will be no null-values.

cg/store/crud/read.py
```diff
@@ -187,6 +194,19 @@ def get_analysis_by_case_entry_id_and_completed_at(
             completed_at_date=completed_at_date,
         ).first()
 
+    def get_latest_started_analysis_for_case(self, case_id: str) -> Analysis:
+        """Return the latest started analysis for a case.
+        Raises:
+            AnalysisDoesNotExistError if no analysis is found.
+        """
+        case: Case = self.get_case_by_internal_id(case_id)
```
mogmi, line 202:
```suggestion
        case: Case | None = self.get_case_by_internal_id(case_id)
```

Until we get that version that always returns an item or raises :)

cg/store/crud/read.py
diitaz93: set it as out of scope

tests/store/crud/read/test_read_analysis.py
```diff
@@ -512,3 +514,34 @@ def test_get_cases_for_analysis_multiple_analyses_and_one_analysis_is_not_older_
...
+    """Test returning the latest started analysis for a case."""
+    # GIVEN a store with multiple analyses for a case
+    test_case: Case = helpers.add_case(store=base_store, name="test_case")
+    test_analysis_1: Analysis = helpers.add_analysis(
+        store=base_store, case=test_case, started_at=datetime(2023, 1, 1, 0, 0, 0)
+    )
+    test_analysis_2: Analysis = helpers.add_analysis(
+        store=base_store, case=test_case, started_at=datetime(2023, 1, 2, 0, 0, 0)
+    )
+
```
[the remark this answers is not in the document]
diitaz93: out of scope

PR #4466, opened 2025-06-24
Fix(statusDB Apptags) Add is_archived filter

Vince-janv:
### Added

- Ability to filter on is_archived in the ApplicationView


islean (approved):
💯

PR #4639, opened 2025-10-03
feat(Clean seq-runs) Give option to provide day threshold

Vince-janv:
## Description
Saving the sequencing files for 21 days is superfluous. We want to lower this, but since it is hard-coded this is difficult.

This PR makes it so that one has to provide the --day-threshold option to the cg clean illumina-runs command.

### Added

- Option --day-threshold to cg clean illumina-runs

### Changed

- cg clean illumina-runs no longer always uses 21 days as its threshold


islean (approved):
Looks good to me!

PR #4645, opened 2025-10-06
Fix(Cleaning) Remove analysis based on completion instead of start

Vince-janv:
## Description
This PR changes the cleaning logic to start counting from the completion timestamp rather than the starting timestamp. This is more intuitive since analysis completion times vary significantly.

Additionally removes the default argument for

### Changed
- Changed analysis cleaning logic to be base on completion timestamp
- Removed default value for XXXX


islean (approved):
Looks good to me!

PR #4659, opened 2025-10-20
Modify test cases to include male and unknown samples

mogmi:
### Changed
Previously we only had the sex `SexOptions.FEMALE` this PR modifies to tests and their expectations to also include `SexOptions.MALE` and `SexOptions.UNKNOWN` and their expected results.


diitaz93 (approved):
💯

PR #4665, opened 2025-10-23
Add date filter on requests to LIMS

RasmusBurge-CG:
## Description
The PR aims to reduce the requests sent to Lims, see https://github.com/Clinical-Genomics/bug-reports/issues/43.

### Added

- Filtering based on ordered_at date.
- An option to the `cg transfer lims` command that lets the user specify an age cutoff. If a sample was ordered before this cutoff, it will be filtered out and no request for that sample will be sent to LIMS.

### How to prepare for test

- [x] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b reduce-requests-sent-to-lims
    ```

### How to test
Test the `--max-order-age` flag by removing received_at, prepared_at, last_sequenced_at, and delivered_at for a sample in statusdb. The order for the sample is placed more than one year ago, but is newer than two. 1<Order_at<2. 
- [x] Set the `--max-order-age` to 1 and expect that the values should NOT have been set in StatusDB for the sample with missing values.
- [x] Set the `--max-order-age` to 2 and expect that the values should have been set in StatusDB for the sample with missing values.

### Expected test outcome

- [x] When the `--max-order-age` is set to 1, then the values for the sample with missing dates should not have been updated in StatusDB. 
- [x] When the `--max-order-age` is set to 2, then the values for the sample with missing dates should have been updated in StatusDB with the correct values.

### Older order
Commands:
```
cg transfer lims --status received --max-order-age 1
cg transfer lims --status prepared --max-order-age 1
cg transfer lims --status delivered --max-order-age 1
```

Relevant output:
```
INFO Remove samples ordered before 2024-11-21 11:34:29.939547
```

Result in statusdb, time limit 1:
internal_id | ordered_at | delivered_at | received_at | prepared_at
-- | -- | -- | -- | --
SAMPLE | 2024-01-23 09:55:22 |   |   |  

### Newer order
Commands:
```
cg transfer lims --status received --max-order-age 2
cg transfer lims --status prepared --max-order-age 2
cg transfer lims --status delivered --max-order-age 2
```

Relevant output:
```
INFO Remove samples ordered before 2023-11-22 11:36:59.998541
```

Result in statusdb, time limit 1:
internal_id | ordered_at | received_at | prepared_at | delivered_at
-- | -- | -- | -- | --
SAMPLE | 2024-01-23 9:55:22 | 2024-01-23 0:00:00 |  2024-01-25 00:00:00  | 2024-02-02 0:00:00

## Review

- [x] Tests executed by @RasmusBurge-CG 
- [x] "Merge and deploy" approved by @islean 

## Implementation Plan

- [ ] Deploy this branch on Hasta
- [ ] Inform on slack

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 78.4.2
+current_version = 78.4.3
 commit = True
 tag = True
 tag_name = v{new_version}
--- a/cg/__init__.py
+++ b/cg/__init__.py
@@ -1,2 +1,2 @@
 __title__ = "cg"
-__version__ = "78.4.2"
+__version__ = "78.4.3"
--- a/cg/server/admin.py
+++ b/cg/server/admin.py
@@ -16,6 +16,7 @@ from cg.server.ext import applications_service, db, sample_service
 from cg.server.utils import MultiCheckboxField
 from cg.store.models import Application
 from cg.utils.flask.enum import SelectEnumField
+from cg.server.app_config import app_config
 
 
 class BaseView(ModelView):
@@ -195,6 +196,44 @@ def view_customer_link(unused1, unused2, model, unused3):
         )
 
     return markup
+
+
+def view_ticket_link(unused1, unused2, model, attribute_name):
+    """Column formatter used to add hyperlink to ticket ID, if applicable."""
+    del unused1, unused2
+
+    # Ticket attribute called differently from models, infer attribute name from args
+    ticket_attr = getattr(model, attribute_name)
+    ticket_str = str(ticket_attr).strip()
+
+    # Freshdesk ticket IDs have >=7 digits
+    if len(ticket_str) >= 7:
+        ticket_link = f"{app_config.freshdesk_url}/a/tickets/{ticket_str}"
+        ticket_markup = Markup(f"<a href='{ticket_link}'>{ticket_str}</a>")
+        return ticket_markup
+    else:
+        return ticket_str
+
+
+def view_tickets_links(unused1, unused2, model, unused3):
+    """Column formatter used to add hyperlinks to comma-separated ticket IDs, if applicable. Assumes attribute name 'tickets'."""
+    del unused1, unused2, unused3
+
+    tickets_list = str(model.tickets).strip().split(sep=",")
+
+    tickets_markups = []
+    for ticket in tickets_list:
+        ticket_str = str(ticket).strip()
+
+        # Freshdesk ticket IDs have >=7 digits
+        if len(ticket_str) >= 7:
+            ticket_link = f"{app_config.freshdesk_url}/a/tickets/{ticket_str}"
+            ticket_markup = Markup(f"<a href='{ticket_link}'>{ticket_str}</a>")
+            tickets_markups.append(ticket_markup)
+        else:
+            tickets_markups.append(ticket_str)
+
+    return Markup(", ".join(tickets_markups))
 
 
 class ApplicationView(BaseView):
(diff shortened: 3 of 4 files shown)


cg/cli/transfer.py
```diff
@@ -31,11 +37,20 @@ def transfer_group(context: CGConfig):
 )
 @click.option("--sample-id", help="Lims Submitted Sample id. use together with status.")
 @click.pass_obj
-def check_samples_in_lims(context: CGConfig, status: str, include: str, sample_id: str):
+def check_samples_in_lims(
+    context: CGConfig, order_age_cutoff: int, status: str, include: str, sample_id: str
+):
     """Check if samples have been updated in LIMS."""
+    if order_age_cutoff:
```
islean: There are some subtleties here. If I provide order_age_cutoff=0, this will evaluate to False, meaning that order_date_cutoff gets set to None. Is that what we want?

islean: I also think it would be nice to move away as much logic from the CLI layer as possible. I would just pass the years/the default on to the lims api and handle the conversion to days there.

cg/meta/transfer/lims.py
```diff
@@ -84,6 +94,18 @@ def transfer_samples(
...
+    def _remove_older_orders(
+        self, samples: list[Sample], order_date_cutoff: datetime
+    ) -> list[Sample] | None:
+        no_samples_before_filtering: int = len(samples)
+        for sample_obj in samples:
+            if self._is_sample_ordered_before_order_date_cutoff(
+                sample_obj=sample_obj, order_date_cutoff=order_date_cutoff
+            ):
+                samples.remove(sample_obj)
+        LOG.info(f"Filter out {no_samples_before_filtering-len(samples)} samples")
```
islean:
```suggestion
        LOG.debug(f"Filter out {no_samples_before_filtering-len(samples)} samples")
```
I do not think this is something we might want to know on a day-to-day basis, rather than as validation once it goes live and if something seems off.

islean (commented):
Adding some comments! I think it will filter out the correct samples, but there are some changes that are not too intrusive which I think can make the changes more robust 💪

cg/cli/transfer.py
diitaz93: I agree to move the conversion from years to days to the TransferAPI and use the following to avoid problems with 0 years:
```python
if order_age_cutoff is not None:
```

cg/meta/transfer/lims.py
```diff
@@ -52,15 +53,24 @@ def __init__(self, status: Store, lims: LimsAPI):
...
             samples: list[Sample] = self.status.get_samples_by_internal_id(internal_id=sample_id)
         else:
-            samples: list[Sample] = self._get_samples_to_include(include, status_type)
+            samples: list[Sample] = self._get_samples_to_include(
+                include=include, status_type=status_type
+            )
 
+        if order_date_cutoff:
+            LOG.info(f"Remove samples ordered before {str(order_date_cutoff)}")
+            self._remove_older_orders(samples=samples, order_date_cutoff=order_date_cutoff)
```
diitaz93: If you want to perform the removal based on the existence of the cutoff, it would be nice to finish the execution of the function there, so that we don't do extra logic that does nothing afterwards
```suggestion
        if order_date_cutoff:
            LOG.info(f"Remove samples ordered before {str(order_date_cutoff)}")
            self._remove_older_orders(samples=samples, order_date_cutoff=order_date_cutoff)
            return
```

PR #4669, opened 2025-10-27
Add nallo sample sheet creator

diitaz93:
## Description
Creates a NalloSampleSheetCreator that does not inherit from the NextflowSampleSheet. The motivation for this is that the parent has too much functionality specific to fastq handling, while Nallo does not need that. 

### Added

- NalloSampleSheetCreator class
- Test for the `create` method
- Fixture with expected content

### Changed

- Added NalloSampleSheetCreator to the ConfiguratorFactory and changed the if-else statements for a `match`.

### DIFF
--- a/cg/services/analysis_starter/configurator/file_creators/nextflow/sample_sheet/creator.py
+++ b/cg/services/analysis_starter/configurator/file_creators/nextflow/sample_sheet/creator.py
@@ -16,7 +16,7 @@ from cg.store.store import Store
 LOG = logging.getLogger(__name__)
 
 
-class NextflowSampleSheetCreator(ABC):
+class NextflowFastqSampleSheetCreator(ABC):
 
     def __init__(self, housekeeper_api: HousekeeperAPI, store: Store):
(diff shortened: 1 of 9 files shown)


cg/services/analysis_starter/configurator/file_creators/nextflow/sample_sheet/nallo.py
```diff
@@ -0,0 +1,65 @@
...
+                PlinkSex[case_sample.sample.sex.upper()].value,
+                str(PlinkPhenotypeStatus[case_sample.status.upper()]),
+            ]
+            sample_sheet_entries.append(sample_sheet_entry)
+        return sample_sheet_entries
+
+    def _get_bam_read_file_paths(self, sample: Sample) -> list[Path]:
+        """Gather BAM file path for a sample based on the BAM tag."""
+        return [
+            Path(hk_file.full_path)
```
islean: Seems like we never want this as a path - should we just return the string?

PR #4690, opened 2025-11-05
Activate the configured conda environment in start command

mogmi:
Activate the configured conda environment when starting the analysis


diitaz93 (approved):
Great catch

PR #4693, opened 2025-11-05
Balsamic dev commands integration test

mogmi:
## Added
- Parameterized the two existing balsamic start-available tests to also run the refactored command and expect the same results
- Create (empty) files where the BalsamicAPI will look for them when creating flags for the config case command to make the tests more complete. 

## Changed
- Updated existing balsamic integration tests to include changes made for Balsamic v.18
- Moved non-fixture functions from contest to a new utils module specific to integration test code
- Alphabetically sorted flags sent to the config case command in BalsamicAPI to make it easier to compare them to the refactored counterparts which are also sorted.

### DIFF
--- a/tests/integration/utils.py
+++ b/tests/integration/utils.py
@@ -161,19 +161,19 @@ def expect_to_add_pending_analysis_to_trailblazer(
 
 def expect_lims_sample_request(lims_server: HTTPServer, sample: Sample, bed_name: str):
     lims_server.expect_request(f"/lims/api/v2/samples/{sample.internal_id}").respond_with_data(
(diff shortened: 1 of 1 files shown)


tests/integration/test_balsamic.py
```diff
@@ -192,55 +214,96 @@ def mock_run(*args, **kwargs):
...
+        case "dev-start-available":
+            run_analysis_call = run_analysis_subprocess_mock.run.mock_calls[0]
+
+            assert run_analysis_call == call(
+                args=expected_run_analysis_command,
+                check=False,
+                shell=True,
+                stderr=ANY,
+                stdout=ANY,
+            )
```
diitaz93: I also don't see the difference here apart from the first/second mock call.

PR #4696, opened 2025-11-06
Update integration tests for balsamic18

mogmi:
### Changed

- Updated expected commands to match those made to Balsamic v18


diitaz93 (approved):
💯

Issue #4758, opened 2025-12-05
Bug: Cases with multiple samples not decompressed


Vince-janv:
## Description
A case with multiple samples needing decompression will only have one of its samples decompressed.

## Troubleshooting:
The method `cg.meta.compress.compress.CompressAPI.decompress_case` uses an generator expression in an `any()` expression. Which short-circuits on the first true (in this case decompression starting for the first sample). The subsequent samples are therefore not decompressed.

**Additional context**
Original bug report: https://github.com/Clinical-Genomics/bug-reports/issues/59

PR #4760, opened 2025-12-05
Fix(Spring decompression) Avoid short-cicuit

Vince-janv:
## Description
#4758 describes a bug with not all of the samples in a case being decompressed due to the behaviour if the any function.

This PR removes the use of any to ensure that looping through the samples completes.

### Fixed
- decompress_case now correctly iterates through all samples

### DIFF
--- a/tests/meta/compress/test_decompress_spring_meta.py
+++ b/tests/meta/compress/test_decompress_spring_meta.py
@@ -44,7 +44,7 @@ def test_decompress_case(compress_api: CompressAPI):
     # GIVEN a decompress_spring method that is mocked
     compress_api.decompress_spring = Mock()
     # WHEN decompressing the case
-    compress_api.decompress_case(case.as_type)
+    compress_api.decompress_case(case)
 
     # THEN the method to decompress spring is called for each sample in the case
     assert compress_api.decompress_spring.call_count == 3


tests/meta/compress/test_decompress_spring_meta.py
```diff
@@ -27,3 +32,22 @@ def test_decompress_spring(populated_decompress_spring_api, compression_files, s
...
+    # GIVEN a case with three samples
+    sample_1 = create_typed_mock(Sample)
+    sample_2 = create_typed_mock(Sample)
+    sample_3 = create_typed_mock(Sample)
+    case: TypedMock[Case] = create_typed_mock(
+        Case, samples=[sample_1.as_type, sample_2.as_type, sample_3.as_type]
+    )
+
+    # GIVEN a decompress_spring method that is mocked
+    compress_api.decompress_spring = Mock()
```
islean: I am not super keen on mocking methods in the module we are testing. It is quite a messy module, though. Maybe it is worth doing it in this case? It probably does not make sense to mock that there is no housekeeper bundle for example. 

Leave as it is unless there is an easy solution.

Vince-janv: Like you said, the module is a mess. I don't think creating a complete set-up is worth the time here.

PR #4785, opened 2025-12-16
Feat(taxprofiler) Add extra files to housekeepr

Vince-janv:
## Description
Adds four files to be stored in the housekeeper bundles of taxprofiler analyses

### Added
- storage of four more taxprofiler files


Vince-janv:
Tested together with https://github.com/Clinical-Genomics/hermes/pull/154.

Running `cg workflow taxprofiler <case_id>` saves the correct files in housekeeper

cg/resources/taxprofiler_bundle_filenames.yaml
```diff
@@ -8,6 +8,11 @@
   path: PATHTOCASE/kraken2/k2_pluspf/SAMPLENAME_pe_k2_pluspf.kraken2.kraken2.report.txt
   step: kraken2
   tag: kraken2_report
+- format: txt
+  id: SAMPLEID
+  path: PATHTOCASE/kraken2/k2_pluspf/SAMPLENAME_pe_k2_pluspf.kraken2.kraken2.classifiedreads.txt
```
islean: Just making sure since the issue mentions SAMPLE_ID that SAMPLENAME is correct here?

Vince-janv: Yes. This seems to be a quirk of taxprofiler. The file is actually named using the `sample_name`

Issue #4798, opened 2025-12-22
Bug: cg upload process-solved crashes without case_id


Vince-janv:
**Describe the bug**
The command `cg upload process-solved` crashes when one does not provide the optional argument case_id

**Expected behavior**
Not providing case_id should lead to the method relying on date logic instead.

**Additional context**

PR #4800, opened 2025-12-29
Fix(Mutacc) Provide scout instance

Vince-janv:
## Description
The command `cg upload process-solved` has been crashing since april, see #4798. This is because the command interacts with scout and we now have two scout instances. We use the case_id to determine which instance to use, but case_id is an optional parameter.

This PR adds an option to the CLI command to specify which instance to use.

### Added

- Option to the `cg upload process-solved` command specifying which scout instance to use.

### How to prepare for test
```shell
bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b 4798-fix-process-solved-command
```


Vince-janv:
Writing a test for this module was horrendous. I'm content with manual test on stage to see that the option correctly selects the correct scout instance

PR #4884, opened 2026-02-10
Fix(mutacc-auto) Add --conda flag

Vince-janv:
## Description
We are using conda for mutacc-auto but we're not passing the conda flag.

### Added

- cg now calls the mutacc-auto extract command with the --conda flag


diitaz93 (approved):
👍

Vince-janv:
Deployed to stage and prod

PR #5037, opened 2026-04-17
Case created at non nullable

mogmi:
### Added
- Alembic migration to make case.created_at non-nullable

### Changed

- Case.created_at type annotation from `datetime | None` -> `datetime`


mogmi:
Successfully migrated up and down in stage.

islean (approved):
💯

PR #5056, opened 2026-04-24
Integration test for generating a scout config

mogmi:
### Added

- Integration test that creates a scout config file for a Nallo case and makes sure it can be parsed


mogmi:
Already merged by another PR

PR #5059, opened 2026-04-27
Refactor most "cg store" commands to be called "cg prepare"

mogmi:
## Description

Proposed changes for the commands currently named `cg store` to be renamed to avoid confusion with the `cg workflow <workflow> store`. The purpose of the command is (mostly) to make decompressed Fastq files available for analysis by updating Housekeeper. My suggestion for the new name is `prepare` because:

- Not all files are FASTQs ( demultiplexed-run includes all bundle files)
- The command never starts decompression — so "decompress" would be misleading
- "link" already means filesystem linking in other cg commands, not Housekeeper registration
- "ingest" implies raw data, but the files are already compressed/demultiplexed in the system
- "stage" conflicts with the staging environment terminology used by the team
- "prepare" might be to general but at least covers the shared intent: making files ready for analysis, regardless of the specific mechanism

But the name needs to be discussed - this PR is just to illustrate what changes are needed, less than I anticipated. 

`cg store qc-metrics` is left untouched since it has a very different purpose from the others

### DIFF
--- a/cg/cli/base.py
+++ b/cg/cli/base.py
@@ -21,6 +21,7 @@ from cg.cli.downsample import downsample
 from cg.cli.generate.base import generate as generate_cmd
 from cg.cli.get import get
 from cg.cli.post_process.post_process import post_process_group as post_processing
+from cg.cli.prepare.base import prepare
 from cg.cli.sequencing_qc.sequencing_qc import sequencing_qc
 from cg.cli.set.base import set_cmd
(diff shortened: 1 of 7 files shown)


Vince-janv:
If you disregard `cg store qc-metrics`, `cg store` is only used for storing FASTQ files in housekeeper after decompression. Maybe `cg register-decompressed` could work?

Either way, I think we should remove the `qc-metrics` command since it was a POC that was never fully implemented

PR #5064, opened 2026-04-29
Lazy load test scenarios for analysis starter

mogmi:
## Description

Currently the analysis starter tests using scenarios are some of the slowest running tests in cg. 

This is a proposal to change the scenario fixtures to return a function that will instantiate everything that is necessary for a particular workflow instead of the current eager loading which sets up all workflows for each version of the parameterized tests even though only one workflow is needed at a time. 


diitaz93 (approved):
Looks great ⭐

PR #5111, opened 2026-05-27
Refactor close_order method

kedhammar:
### Added

-

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/deliver_service.py
+++ b/cg/services/deliver_service.py
@@ -31,7 +31,7 @@ class DeliverService:
             except:
                 self.mark_as_delivered_service.unmark_analyses(analyses)
             finally:
-                if self.is_order_closable(order):
+                if self._is_order_closable(order):
                     order.is_open = False
                     if self.freshdesk_client.is_order_open(order):
                         # Discuss if the freshdesk is_order_open should be inside close_ticket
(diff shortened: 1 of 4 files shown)


cg/services/deliver_service.py
```diff
@@ -31,15 +31,15 @@ def deliver_all_cases(self):
...
+                if self.is_order_closable(order):
                     order.is_open = False
                     if self.freshdesk_client.is_order_open(order):
                         # Discuss if the freshdesk is_order_open should be inside close_ticket
                         self.freshdesk_client.close_ticket(order=order)
                     # TODO method to rollback
                 # TODO commit
 
-    def _is_order_closable(self, order: Order) -> bool:
+    def is_order_closable(self, order: Order) -> bool:
```
kedhammar: This method is tested in isolation, so I though we might want it to be public?

kedhammar:
Failing test is due to missing attribute, to be addressed in base branch

tests/services/test_deliver_service.py
```diff
@@ -33,7 +33,6 @@ def test_deliver_case(mocker: MockerFixture):
     # GIVEN a deliver service
     deliver_service = DeliverService(status_db=status_db, trailblazer_api=trailblazer_api)
     mark_analyses_spy = mocker.spy(deliver_service.mark_as_delivered_service, "mark_analyses")
-    close_order_spy = mocker.spy(deliver_service.mark_as_delivered_service, "close_order")
```
kedhammar, line 36: We previously asserted the call to a method that doesn't necessarily perform any actions. I've removed it from the tests that currently do not have the setup to pass through the logic of `is_order_closable`without Mock truthiness affecting the outcome.

Modifying these tests is left as en exercise for the reader 👍

cg/services/deliver_service.py
islean: I do not think that should guide if it is public or not - that should be determined by whether it is intended as a standalone method and needs to be exposed. I think it should be private. As a consequence, testing the functionality would be baked into the testing of the public methods.

There are a lot of good resources on what should be public vs private, but a good starting place could be to think about what would be part of the interface of the class and what is part of the implementation

PR #5115, opened 2026-05-28
DEV - Automatic delivery

diitaz93:
## Description

First iteration of Phase 3 of Automate Production Project.
Minimal delivery implies:
- marking samples as delivered in StatusDB
- marking orders as closed in StatusDB 
- marking analyses as delivered in Trailblazer 

Ideally, to be merged to `master` before the completion of the project -> Hidden CLI commands

To be merged with this branch in Trailblazer https://github.com/Clinical-Genomics/trailblazer/pull/533

### Added

- CLI commands to perform a minimal delivery of a case, an order and all available orders

### Changed

-

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b POC-deliver-all-exceptions -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/deliver_service.py
```diff
@@ -19,25 +19,60 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
                 self.status_db.rollback()
+                LOG.error(
+                    f"Failed to send delivery message for order {order.id}. "
+                    "Rolling back delivery status in Trailblazer and StatusDB."
+                )
+            except Exception:
+                LOG.critical(
+                    f"Unexpected error while delivering order {order.id}. "
+                    "Check delivery status and ticket manually."
+                )
```
kedhammar: Maybe outfactor this block, i.e. the per-order iteration, into a private method to reduce complexity?

cg/services/deliver_service.py
```diff
@@ -19,25 +19,60 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
-                    order.is_open = False
-                    try:
-                        self.freshdesk_client.close_ticket_if_open(order=order)
-                    except HTTPException:
-                        LOG.error(f"Failed to close ticket for order {order.id}.")
-                        self.status_db.rollback()
-            # TODO commit
+                self.status_db.commit_to_store()  # commit after each order to not lose all progress if there is an error in one of the orders
+
+                try:
```
kedhammar: I think `finally` will run regardless of whether the try is excepted or not, whereas `else` will only run if the try is not excepted. Do we want to use `else` here, since I don't expect we will want to try closing orders and tickets for which the mark-as-delivered or freshdesk message failed?

cg/services/deliver_service.py
diitaz93: Yes definetly

cg/services/deliver_service.py
```diff
@@ -19,25 +19,60 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
+                LOG.error(
+                    f"Failed to deliver analyses in Trailblazer for order {order.id}. "
+                    "Aborting delivery message for this order."
+                )
             except HTTPException:
                 self.mark_as_delivered_service.unmark_analyses(analyses)
-                # TODO: Add logs
                 self.status_db.rollback()
+                LOG.error(
+                    f"Failed to send delivery message for order {order.id}. "
```
diitaz93: Realised I'm mixing order IDs and ticket IDs in the logs, we should decide which one is more useful, maybe ticket?

diitaz93: or both?

cg/services/deliver_service.py
```diff
@@ -19,25 +19,60 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
+                try:
+                    if self._is_order_closable(order):
+                        LOG.info(
+                            f"Closing order {order.id} in StatusDB "
+                            f"and ticket {order.ticket_id} in Freshdesk."
+                        )
+                        order.is_open = False
+                    self._freshdesk_close_ticket_if_open(order=order)
+                except TrailblazerAPIHTTPError:
+                    LOG.error(f"Failed to close {order.id}. Aborting closing order in Freshdesk.")
```
kedhammar:
```suggestion
                    LOG.error(f"Failed to check whether order {order.id} should be closed.")
```

cg/services/deliver_service.py
```diff
@@ -19,25 +19,60 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
+                self.status_db.commit_to_store()  # commit after each order to not lose all progress if there is an error in one of the orders
+
+                try:
+                    if self._is_order_closable(order):
+                        LOG.info(
+                            f"Closing order {order.id} in StatusDB "
+                            f"and ticket {order.ticket_id} in Freshdesk."
+                        )
+                        order.is_open = False
+                    self._freshdesk_close_ticket_if_open(order=order)
```
kedhammar: Indent this? It's dependent on whether the order is closable, no?

islean:
```suggestion
                        self._freshdesk_close_ticket_if_open(order=order)
```

cg/services/deliver_service.py
```diff
@@ -19,25 +19,60 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
+                        "Rolling back closing order in StatusDB."
+                    )
+                    self.status_db.rollback()
+                except Exception:
+                    LOG.critical(
+                        f"Unexpected error while closing ticket {order.ticket_id}. "
+                        "Check order and ticket status manually."
+                    )
+                finally:
+                    self.status_db.commit_to_store()
```
diitaz93: I would also update the method `deliver_order` with the same logic

cg/services/deliver_service.py
```diff
@@ -19,25 +19,60 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
+                        order.is_open = False
+                    self._freshdesk_close_ticket_if_open(order=order)
+                except TrailblazerAPIHTTPError:
+                    LOG.error(f"Failed to close {order.id}. Aborting closing order in Freshdesk.")
+                except HTTPException:
+                    LOG.error(
+                        f"Failed to close ticket {order.ticket_id} in Freshdesk. "
+                        "Rolling back closing order in StatusDB."
+                    )
+                    self.status_db.rollback()
```
kedhammar: I'm unsure whether this rollback is necessary. If this scenario ever plays out I think the ticket will be manually closed by prodbioinfo but the order will be forgotten and left open.

cg/services/deliver_service.py
diitaz93: I think the log is wrong, I didn't mean to include freshdesk here

cg/services/deliver_service.py
```diff
@@ -19,25 +19,60 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
+                except Exception:
+                    LOG.critical(
+                        f"Unexpected error while closing ticket {order.ticket_id}. "
+                        "Check order and ticket status manually."
+                    )
```
islean: I would probably suggest that we do not except these suggestions. I can not think of what could possible cause this apart from a server going down mid call or something, and that I am fine with us not excepting and the service breaking for.

islean: Same with the general except Exception above

cg/services/deliver_service.py
kedhammar: Added TODO

cg/services/deliver_service.py
kedhammar: Added TODO

cg/services/deliver_service.py
```diff
@@ -19,25 +19,13 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
             status_db=status_db, trailblazer_api=trailblazer_api
         )
 
-    def deliver_all_cases(self):
+    def deliver_all_cases(self) -> None:
```
diitaz93: rename to `deliver_all_available`? This is not about cases anymore

cg/services/deliver_service.py
```diff
@@ -70,6 +59,63 @@ def deliver_order(self, ticket_id: int, signature: str):
...
+            self.mark_as_delivered_service.mark_analyses(analyses=analyses)
+            self._freshdesk_send_delivery_message(order=order, analyses=analyses)
+        except TrailblazerAPIHTTPError:
+            LOG.error(
+                f"Failed to deliver analyses in Trailblazer for order {order.id}. "
+                f"Aborting delivery message for ticket {order.ticket_id}."
+            )
+            success = False
+        except HTTPException:
+            self.mark_as_delivered_service.unmark_analyses(analyses)
```
RasmusBurge-CG:
```suggestion
            self.mark_as_delivered_service.set_analyses_as_undelivered(analyses)
```
I'm not sure my suggestion is the best name but I find the `unmark_analyses` a bit cryptic.

cg/services/deliver_service.py
```diff
@@ -70,6 +59,63 @@ def deliver_order(self, ticket_id: int, signature: str):
         else:
             LOG.warning("No analysis in the order ready to deliver")
 
+    def _deliver_order(self, order: Order, analyses: list[Analysis]) -> None:
```
islean: We could do centralized exception handling in here without using booleans in this method. I have a screenshot below of how it could look with comments about the (imo minor) caveats

Image

islean: Where we would also remove try/excepts in the lower methods

cg/services/deliver_service.py
```diff
@@ -0,0 +1,128 @@
...
+    def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
+        self.status_db = status_db
+        self.trailblazer_api = trailblazer_api
+        self.mark_as_delivered_service = MarkAsDeliveredService(
+            status_db=status_db, trailblazer_api=trailblazer_api
+        )
+
+    def deliver_all_available(self) -> None:
+        order_dict: dict[Order, list[Analysis]] = self._get_order_analyses_dictionary()
+        if len(order_dict) == 0:
```
diitaz93, line 30: Add error logging for every type of error and (possibly) one at the beginning mentioning the number of orders to deliver

cg/apps/tb/api.py
diitaz93:
```suggestion
    def get_analyses_to_deliver_for_order(self, order_id: int) -> list[TrailblazerAnalysis]:
```

cg/services/deliver_service.py
```diff
@@ -0,0 +1,128 @@
...
+                self.mark_as_delivered_service.unmark_analyses(analyses)
+            except FreshdeskClosingTicketError:
+                # Samples marked TB analysis marked, delivery message sent in FD, potentially order closed in StatusDB
+                order.is_open = True
+                self.status_db.commit_to_store()
+            else:
+                self.status_db.commit_to_store()
+
+    def deliver_case(self, case_id: str, signature: str):
+        analyses_to_deliver: list[Analysis] = self._get_undelivered_analyses_for_case(case_id)
```
diitaz93, line 60: implement freshdesk for case. Could we just call `_deliver_order`?

cg/services/deliver_service.py
```diff
@@ -0,0 +1,128 @@
...
+
+    def deliver_order(self, ticket_id: int, signature: str):
+        order: Order = self.status_db.get_order_by_ticket_id_strict(ticket_id)
+
+        undelivered_trailblazer_analyses = self.trailblazer_api.get_analyses_to_deliver(order.id)
+        uploaded_analyses_to_deliver = self.status_db.get_uploaded_analyses(
+            [analysis.id for analysis in undelivered_trailblazer_analyses]
+        )
+        if uploaded_analyses_to_deliver:
+            self._deliver_order(
```
diitaz93, line 83: as we don't implement rollbacks here, we could potentially leave the order "half baked". Is this acceptable informing the user? or should we iplement rollbacks here too?

cg/apps/tb/api.py
```diff
@@ -245,7 +255,32 @@ def get_analyses_to_deliver(self, order_id: int) -> list[TrailblazerAnalysis]:
         validated_response = AnalysesResponse.model_validate(raw_response)
         return validated_response.analyses
 
+    def get_delivered_analyses_for_order(self, order_id: int) -> list[TrailblazerAnalysis]:
+        url = f"{self.host}/analyses?order_id={order_id}&status[]=completed&delivered=true"
```
islean, line 260: Would we ever want this to deliver delivered analyses which are not completed?

cg/services/deliver_service.py
islean: Probably! But we might want to rename it in that case

cg/services/deliver_service.py
islean: I think we should add rollbacks

cg/services/deliver_service.py
```diff
@@ -0,0 +1,128 @@
...
+    def _get_undelivered_analyses_for_case(self, case_id: str) -> list[Analysis]:
+        case: Case = self.status_db.get_case_by_internal_id_strict(case_id)
+        uploaded_analyses: list[Analysis] = [
+            analysis for analysis in case.analyses if analysis.uploaded_at
+        ]
+        undelivered_trailblazer_analyses: list[TrailblazerAnalysis] = (
+            self.trailblazer_api.get_analyses_to_deliver_for_case(case_id)
+        )
+        undelivered_trailblazer_ids = [analysis.id for analysis in undelivered_trailblazer_analyses]
+        analyses_to_deliver: list[Analysis] = []
```
islean:
```suggestion
        undelivered_trailblazer_analyses: list[TrailblazerAnalysis] = (
            self.trailblazer_api.get_analyses_to_deliver_for_case(case_id)
        )
        undelivered_trailblazer_ids = [analysis.id for analysis in undelivered_trailblazer_analyses]
        return [
            analysis for analysis in case.analyses if analysis.uploaded_at and analysis.trailblazer_id in undelivered_trailblazer_ids
        ]

```

cg/store/crud/read.py
```diff
@@ -1509,6 +1509,17 @@ def get_order_by_ticket_id(self, ticket_id: int) -> Order | None:
...
+    def get_order_by_ticket_id_strict(self, ticket_id: int) -> Order:
+        """
+        Returns the entry in Order matching the given ticket id.
+        Raises:
+            OrderNotFoundError: If no order is found with the given ticket id.
+        """
+        orders: Query = self._get_query(table=Order).filter_by(ticket_id=ticket_id)
+        if not (order := orders.first()):
+            raise OrderNotFoundError(f"Order with ticket ID {ticket_id} not found.")
+        return order
```
islean: Slight preference for changing to "if order := orders.first(): return, else raise" but fine if we leave as is.

cg/services/deliver_service.py
```diff
@@ -0,0 +1,128 @@
...
+        for analysis in uploaded_analyses:
+            if analysis.trailblazer_id in undelivered_trailblazer_ids:
+                analyses_to_deliver.append(analysis)
```
diitaz93:
```suggestion
        undelivered_trailblazer_analyses: list[TrailblazerAnalysis] = (
            self.trailblazer_api.get_analyses_to_deliver_for_case(case_id)
        )
        undelivered_trailblazer_ids = [analysis.id for analysis in undelivered_trailblazer_analyses]
        analyses_to_deliver: list[Analysis] = self.statusdb.get_uploaded_analyses(undelivered_trailblazer_ids)
```

cg/services/deliver_service.py
```diff
@@ -0,0 +1,128 @@
...
+            LOG.warning("No analysis in the order ready to deliver")
+
+    def _deliver_order(
+        self, order: Order, analyses: list[Analysis], signature: str | None = None
+    ) -> None:
+        LOG.info(f"Delivering {len(analyses)} analyses of ticket {order.ticket_id}.")
+        self.mark_as_delivered_service.mark_analyses(analyses=analyses, signature=signature)
+        self.mark_as_delivered_service.close_order_in_status_db_if_closable(order)
+        self._freshdesk_send_delivery_message(order=order, analyses=analyses)
+        self._freshdesk_close_ticket_if_open(order=order)
```
islean, line 96: This needs to only be called if the order is closable as checked in `_is_order_closable` in the MarkAsDeliveredService

cg/services/mark_as_delivered_service.py
```diff
@@ -16,15 +18,58 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
+            )
+        except TrailblazerAPIHTTPError as error:
+            raise TrailblazerAnalysisDeliveryError from error
+        return response
+
+    def unmark_analyses(self, analyses: list[Analysis]):
+        """
+        Remove Trailblazer delivery entries for the provided analyses.
+        This operation is idempotent per analysis: if an analysis has no delivery entry,
+        no change is made and no error is raised.
```
islean, line 45: I would argue that we probably should not mention how TB works here and that CG should not care too much about implementation in another repo.

cg/services/mark_as_delivered_service.py
```diff
@@ -16,15 +18,58 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
+        trailblazer_ids: list[int] = [analysis.trailblazer_id for analysis in analyses]
+        self.trailblazer_api.set_analyses_delivery_status(
+            trailblazer_ids=trailblazer_ids, is_delivered=False, signature=None, auth_token=None
+        )
+
+    def _is_order_closable(self, order: Order) -> bool:
+        """
+        Return True only if
+        - there is a delivered TB analysis for each case and
+        - each sample in the order has a "delivered_at" set
```
islean, line 62: I guess this means that if a sample is ordered in two different orders, say order_1 and order_2, with case_1 in order_1 delivering the sample and case_2 in order_2 not delivering it, then order_2 can not be closed until order_1 has been closed. Probably not an issue though?

cg/services/mark_as_delivered_service.py
```diff
@@ -16,15 +18,58 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
+        Remove Trailblazer delivery entries for the provided analyses.
+        This operation is idempotent per analysis: if an analysis has no delivery entry,
+        no change is made and no error is raised.
+        """
+        trailblazer_ids: list[int] = [analysis.trailblazer_id for analysis in analyses]
+        self.trailblazer_api.set_analyses_delivery_status(
+            trailblazer_ids=trailblazer_ids, is_delivered=False, signature=None, auth_token=None
+        )
+
+    def _is_order_closable(self, order: Order) -> bool:
```
islean, line 58: Given that this check is applicable for Freshdesk interactions I think it makes sense to either add Freshdesk in this service or to move this method outside of this class.

cg/services/mark_as_delivered_service.py
```diff
@@ -16,15 +18,58 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
-            trailblazer_ids=trailblazer_ids, auth_token=auth_token
+        try:
+            response: Response = self.trailblazer_api.set_analyses_delivery_status(
+                auth_token=auth_token,
+                is_delivered=True,
+                signature=signature,
+                trailblazer_ids=trailblazer_ids,
+            )
+        except TrailblazerAPIHTTPError as error:
+            raise TrailblazerAnalysisDeliveryError from error
```
islean: Slight preference for adding fixing this in the TrailblazerAPI instead but also okay to just leave as is.

mogmi: Yeah I agree, it is a bit weird that this service raises a Trailblazer error

tests/services/test_deliver_service.py
```diff
@@ -0,0 +1,500 @@
...
+    )
+
+    # WHEN delivering the order
+    deliver_service.deliver_order(ticket_id=123, signature="CG")
+
+    # THEN the uploaded analyses should have been marked as delivered
+    mark_analyses_call.assert_called_once_with(analyses=[analysis_1, analysis_2], signature="CG")
+
+    # THEN the order should have been closed
+    assert not order.is_open
```
islean, line 474: This test would still be green if we rollbacked everything, since the Trailblazer rollback is a separate call. Would it make sense to add assertions that we have not made the TB rollback call?

cg/services/deliver_service.py
```diff
@@ -0,0 +1,128 @@
...
+        for order, analyses in order_dict.items():
+            try:
+                self._deliver_order(order=order, analyses=analyses)
+            except TrailblazerAnalysisDeliveryError:
+                # Samples marked but nothing in TB nor FD
+                self.status_db.rollback()
+            except (TrailblazerFailedToGetAnalysesError, FreshdeskDeliveryMessageError):
+                # Samples marked TB analysis marked, no FD
+                self.status_db.rollback()
+                self.mark_as_delivered_service.unmark_analyses(analyses)
```
mogmi, line 45: If `unmark_analyses` raises here it is not handled and it will stop the loop so that the rest of the orders in `order_dict` are not delivered - to me that feels like a high price to pay for something that is an attempt to repair the state of one order. I propose that `unmark_analyses` should not propagate any errors raised within it, just except and log them.

cg/services/deliver_service.py
mogmi: A decision for the next iteration where we add freshdesk I would say

cg/apps/tb/api.py
diitaz93: I think those fall under special cases that require manual delivery

mogmi: We can add it if the need arises?

diitaz93: if any

cg/apps/tb/api.py
```diff
@@ -245,7 +255,32 @@ def get_analyses_to_deliver(self, order_id: int) -> list[TrailblazerAnalysis]:
...
+        response = requests.get(url=url, headers=self.auth_header)
+        if not response.ok:
+            raise TrailblazerFailedToGetAnalysesError(response.reason)
+        validated_response = AnalysesResponse.model_validate(response.json())
```
diitaz93, line 264: how likely is it that the pydantic validation fails? Then we don't catch that and the service breaks for all orders

cg/apps/tb/api.py
```diff
@@ -245,7 +255,32 @@ def get_analyses_to_deliver(self, order_id: int) -> list[TrailblazerAnalysis]:
...
         """Raises an AnalysisNotCompletedError if the latest analysis for the case has not completed."""
         if not self.is_latest_analysis_completed(case_id) and not force:
             raise AnalysisNotCompletedError(f"The latest analysis for {case_id} has not completed.")
+
+    def get_analyses_to_deliver_for_case(self, case_id: str) -> list[TrailblazerAnalysis]:
+        endpoint = f"analyses?case_id={case_id}&status[]={AnalysisStatus.COMPLETED}&delivered=false"
+        raw_response = self.query_trailblazer(
+            command=endpoint, request_body={}, method=APIMethods.GET
+        )
+        validated_response = AnalysesResponse.model_validate(raw_response)
```
diitaz93, line 277: same question as above, are pydantic errors a likely scenario we should consider?

cg/apps/tb/api.py
```diff
@@ -245,7 +255,32 @@ def get_analyses_to_deliver(self, order_id: int) -> list[TrailblazerAnalysis]:
...
+        return validated_response.analyses
+
+    def get_all_analyses_to_deliver(self) -> list[TrailblazerAnalysis]:
+        endpoint = f"analyses?status[]={AnalysisStatus.COMPLETED}&delivered=false"
+
+        raw_response = self.query_trailblazer(
+            command=endpoint, request_body={}, method=APIMethods.GET
+        )
+        validated_response = AnalysesResponse.model_validate(raw_response)
```
diitaz93, line 286: same p(y/e)dantic question as above

cg/apps/tb/api.py
islean: sounds good

cg/apps/tb/api.py
islean: It has not failed in all the years we have used the analyses endpoint AFAIK

cg/services/deliver_service.py
diitaz93: We can discuss, I see this in this iteration too

cg/apps/tb/api.py
islean: I do not think it is an error that should be excepted either, it would mean that we have misunderstood how the response from our own service is constructed

tests/apps/tb/test_api.py
```diff
@@ -238,4 +251,254 @@ def test_mark_analyses_as_delivered_fails_with_http_error(
...
+                    "started_at": "2025-05-21",
+                    "completed_at": None,
+                    "out_dir": "/path/to/out_dir",
+                    "config_path": "/path/to/config",
+                }
+            ]
+        }
+    )
+
+    # GIVEN a Successful HTTP called
```
mogmi: I would like to move this GIVEN to line 447,5

cg/services/deliver_service.py
mogmi: Agreed

cg/services/mark_as_delivered_service.py
```diff
@@ -16,15 +18,58 @@ def __init__(self, status_db: Store, trailblazer_api: TrailblazerAPI) -> None:
...
+        return response
+
+    def unmark_analyses(self, analyses: list[Analysis]):
+        """
+        Remove Trailblazer delivery entries for the provided analyses.
+        This operation is idempotent per analysis: if an analysis has no delivery entry,
+        no change is made and no error is raised.
+        """
+        trailblazer_ids: list[int] = [analysis.trailblazer_id for analysis in analyses]
```
mogmi:
```suggestion
        trailblazer_ids: list[int] = [analysis.trailblazer_id for analysis in analyses if analysis.trailblazer_id]
```

Since trailblazer_id is [int | None]

cg/services/deliver_service.py
mogmi: Actually since we don't catch any errors here, on second thought I think nothing would be committed to the database? Only the trailblazer change would remain, but maybe that's what you meant?

cg/services/deliver_service.py
diitaz93: Yes, definitely the ticket can't be closed without doing some checks. Moving back the `is_order_closable` could be a solution, but I think there is an ongoing discussion about the responsibilities of all these classes, and I think it was moved back because of that. This is probably what we need to fix first

cg/services/deliver_service.py
diitaz93: if you see my suggestion below, we could outsource the filtering to a query in the CRUD, instead of a for loop

cg/services/mark_as_delivered_service.py
diitaz93: I added this precisely to know that we should not care about error handling from tb, as it was a question that came up during development

cg/services/mark_as_delivered_service.py
diitaz93: the sample will be delivered by the case that finishes first

cg/store/crud/read.py
diitaz93: I like that too

tests/apps/tb/test_api.py
```diff
@@ -238,4 +251,254 @@ def test_mark_analyses_as_delivered_fails_with_http_error(
...
+
+def test_get_analyses_to_deliver_for_case(
+    valid_google_credentials: IDTokenCredentials,
+    valid_trailblazer_config: dict,
+    mocker: MockerFixture,
+):
+    # GIVEN a TrailblazerAPI
+    tb_api = TrailblazerAPI(valid_trailblazer_config)
+
+    # GIVEN that trailblazer returns an analysis
```
diitaz93:
```suggestion
    # GIVEN that Trailblazer returns an analysis
```

tests/apps/tb/test_api.py
```diff
@@ -238,4 +251,254 @@ def test_mark_analyses_as_delivered_fails_with_http_error(
...
+            config_path="/path/",  # type: ignore pydantic
+            status="pending",
+            priority="high",
+            is_visible=True,
+            type="other",
+            workflow=Workflow.RAREDISEASE,
+        ),
+    ]
+
+    # THEN the trailblazer request should have been made with the expected parameters
```
diitaz93:
```suggestion
    # THEN the Trailblazer request should have been made with the expected parameters
```

tests/apps/tb/test_api.py
```diff
@@ -238,4 +251,254 @@ def test_mark_analyses_as_delivered_fails_with_http_error(
...
+        verify=True,
+    )
+
+
+def test_get_analyses_to_deliver_for_case_no_analysis(
+    valid_google_credentials: IDTokenCredentials,
+    valid_trailblazer_config: dict,
+    mocker: MockerFixture,
+):
+    # GIVEN a trailblazer api
```
diitaz93:
```suggestion
    # GIVEN a TrailblazerAPI
```

tests/apps/tb/test_api.py
```diff
@@ -238,4 +251,254 @@ def test_mark_analyses_as_delivered_fails_with_http_error(
...
+    # THEN an empty list should be returned
+    assert analyses == []
+
+
+def test_get_analyses_to_deliver_for_case_improper_response(
+    valid_google_credentials: IDTokenCredentials,
+    valid_trailblazer_config: dict,
+    mocker: MockerFixture,
+):
+    # GIVEN a trailblazer api
```
diitaz93:
```suggestion
    # GIVEN a TrailblazerAPI
```

tests/apps/tb/test_api.py
```diff
@@ -238,4 +251,254 @@ def test_mark_analyses_as_delivered_fails_with_http_error(
...
+def test_get_all_completed_undelivered_analyses_success(
+    response_with_two_analyses: str,
+    valid_google_credentials: IDTokenCredentials,
+    valid_trailblazer_config: dict,
+    mocker: MockerFixture,
+):
+    # GIVEN a TrailblazerAPI
+    tb_api = TrailblazerAPI(valid_trailblazer_config)
+
+    # GIVEN that trailblazer returns two analyses, one of which is an RSYNC analysis
```
diitaz93:
```suggestion
    # GIVEN that Trailblazer returns two analyses, one of which is an RSYNC analysis
```

tests/apps/tb/test_api.py
```diff
@@ -238,4 +251,254 @@ def test_mark_analyses_as_delivered_fails_with_http_error(
...
+
+def test_get_all_completed_undelivered_analyses_improper_response(
+    valid_google_credentials: IDTokenCredentials,
+    valid_trailblazer_config: dict,
+    mocker: MockerFixture,
+):
+    # GIVEN a TrailblazerAPI
+    tb_api = TrailblazerAPI(valid_trailblazer_config)
+
+    # GIVEN an erroneous http response
```
diitaz93:
```suggestion
    # GIVEN an erroneous HTTP response
```

cg/services/deliver_service.py
mogmi: Decision: we will improve this later

cg/services/mark_as_delivered_service.py
diitaz93: somehow addressed

cg/services/mark_as_delivered_service.py
islean: We will bring it up during the demo

cg/services/deliver_service.py
islean: We will discuss this in the next iteration

mogmi (approved):
Sonar be damned 👍 !

PR #5119, opened 2026-06-03
Allow missing mutant (Original lab address) and (region code)

moedarrah:
## Description

Make `original_lab_address` and `region_code` optional for Mutant/SARS-CoV-2 sample validation.

These fields do not exist in the order portal and are not visible or editable by users. They are derived from `original_lab` and `region` when possible, so missing values should not block order validation or return validation errors to the frontend.

This change is required because the frontend includes errors for fields that are hidden in the UI. Showing validation errors for `original_lab_address` or `region_code` would confuse users, since they cannot fix those fields in the order portal.

## Added

None

## Changed

- Allow `original_lab_address` to be `None`
- Allow `region_code` to be `None`
- Keep the existing auto-fill behavior from `original_lab` and `region`

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

Issue: https://github.com/Clinical-Genomics/cg/issues/5120

### DIFF
--- a/cg/services/orders/validation/order_types/mutant/models/sample.py
+++ b/cg/services/orders/validation/order_types/mutant/models/sample.py
@@ -26,7 +26,7 @@ class MutantSample(Sample):
     _lab_code: str = PrivateAttr(default="SE100 Karolinska")
     organism: str
     organism_other: str | None = None
-    original_lab: OriginalLab | None = None
+    original_lab: OriginalLab
     original_lab_address: str | None = None
     pre_processing_method: PreProcessingMethod
     primer: Primer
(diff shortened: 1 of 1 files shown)


cg/services/orders/validation/order_types/mutant/models/sample.py
```diff
@@ -27,14 +27,14 @@ class MutantSample(Sample):
     organism: str
     organism_other: str | None = None
     original_lab: OriginalLab
-    original_lab_address: str
+    original_lab_address: str | None = None
```
islean, line 30: While I agree that it is not something the customers can fix, I would argue that we would still want the validation to fail for these cases, upon which the customers will have to open a ticket and let us sort this out. These are still fields which are required for us to deliver the analysis and fulfill our criterias to GISAID and FOHM.

PR #5136, opened 2026-06-11
Toggle on DB changes in upload_handler and update TB

mogmi:
## Description

To be deployed together with https://github.com/Clinical-Genomics/servers/pull/1913 

### Added

- upload_handler.completed sets analysis as uploaded in Trailblazer
- Use backoff delay when sending a not acknowledget to nats

### Changed

- Toggle on DB changes in upload_handler

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/events/upload_handler.py
+++ b/cg/services/events/upload_handler.py
@@ -9,17 +9,17 @@ LOG = logging.getLogger(__name__)
 ANALYSIS_UPLOADED_SUBJECT = "analysis.upload_completed"
 
 
-def completed(store: Store, trailblazer_api: TrailblazerAPI):
+def completed(status_db: Store, trailblazer_api: TrailblazerAPI):
     def handler(message: dict):
         analysis_id = message["cg.analysis_id"]
         uploaded_at = message["uploaded_at"]
         try:
             uploaded_at_datetime = datetime.strptime(uploaded_at, "%Y-%m-%dT%H:%M:%SZ")
-            store.update_analysis_uploaded_at(
+            status_db.update_analysis_uploaded_at(
                 analysis_id=analysis_id,
                 uploaded_at=uploaded_at_datetime,
             )
-            if analysis := store.get_analysis_by_entry_id(analysis_id):
+            if analysis := status_db.get_analysis_by_entry_id(analysis_id):
                 trailblazer_api.set_analysis_uploaded(
                     analysis.case.internal_id, uploaded_at=uploaded_at_datetime
                 )
--- a/cg/services/events/upload_handler.py
+++ b/cg/services/events/upload_handler.py
@@ -1,6 +1,7 @@
 import logging
 from datetime import datetime
 
+from cg.apps.tb.api import TrailblazerAPI
 from cg.store.store import Store
 
 LOG = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


cg/services/events/upload_handler.py
```diff
@@ -1,29 +1,32 @@
...
+from cg.apps.tb.api import TrailblazerAPI
 from cg.store.store import Store
 
 LOG = logging.getLogger(__name__)
 
 ANALYSIS_UPLOADED_SUBJECT = "analysis.upload_completed"
 
 
-def completed(store: Store):
+def completed(store: Store, trailblazer_api: TrailblazerAPI):
```
islean:
```suggestion
def completed(status_db: Store, trailblazer_api: TrailblazerAPI):
```
if we generally feel that StatusDB is more descriptive

cg/services/events/upload_handler.py
```diff
@@ -1,29 +1,32 @@
...
             )
+            if analysis := store.get_analysis_by_entry_id(analysis_id):
+                trailblazer_api.set_analysis_uploaded(
+                    analysis.case.internal_id, uploaded_at=uploaded_at_datetime
+                )
         except Exception as error:
             LOG.error(f"Failed to update analysis uploaded_at for analysis with id {analysis_id}")
             LOG.error(f"Failed to handle message {message}.")
             LOG.exception(error)
+            raise error
```
islean, line 30: What is the reasoning for raising/not raising the error here? What prompted the need for a change?

islean (approved):
Should work! Just mainly confused about the change in error handling

cg/services/events/upload_handler.py
```diff
@@ -1,29 +1,32 @@
...
         try:
-            # store.update_analysis_uploaded_at(
-            #     analysis_id=analysis_id,
-            #     uploaded_at=datetime.strptime(uploaded_at, "%Y-%m-%dT%H:%M:%SZ"),
-            # )
-            LOG.info("DB changes toggled off. Would process event by running:")
-            LOG.info(
-                f"store.update_analysis_uploaded_at(analysis_id={analysis_id}, uploaded_at={datetime.strptime(uploaded_at, '%Y-%m-%dT%H:%M:%SZ')})"
+            uploaded_at_datetime = datetime.strptime(uploaded_at, "%Y-%m-%dT%H:%M:%SZ")
+            store.update_analysis_uploaded_at(
```
diitaz93: Not a blocker and risking being too picky here, this method commits already. I'm wondering if Trailblazer fails (I know, maybe unlikely), we will have the changes in StatusDB already persisted, but the logs will mention that the upload failed. Is it worth implementing a rollback here?

tests/services/events/test_upload_handler.py
```diff
@@ -35,7 +49,7 @@ def test_completed_missing_analysis():
     store.update_analysis_uploaded_at = Mock(side_effect=Exception)
 
     # WHEN a completed message is received
-    completed_handler = completed(store)
-    completed_handler(message)
-
-    # THEN it should not raise an exception
+    # THEN it let's the exception propagate
```
diitaz93:
```suggestion
    # THEN it lets the exception propagate
```

diitaz93 (approved):
Wondering about the rollback but it looks good 👍

cg/services/events/upload_handler.py
islean: I wonder what would be easiest to fix - if statusdb updates but TB does not, we can always just join the tables and update TB with the values set in StatusDB. On the other hand, if nothing is updated, I guess we would need to go through the logs to see what was updated afterwards?

mogmi: Since this is related to the change in error handling I will describe my rational for both here. Now that we are not only setting a value in the local database but also in an external service (Trailblazer) I wrote it this way so that there is a chance it will self heal:

Say that we cannot reach Trailblazer at the moment because it is down or due to network issues, then the value will be set in Status DB and then the error from attempting to set it in Trailblazer will propagate to the EventListener class, there a msg.nak (not acknowledged) will be sent with a delay in seconds, telling NATS that we were not able to fully handle the message and to send it again after the delay has passed. 

The first time this happens the delay will be 30s so we try to set it in Trailblazer again in 30s (setting it again in Status DB is fine since it is idempotent). If that works we have self healed and everything is fine, if it did not work we will try again but in 60s and so on (to not hammer the external service and possibly make the situation worse). 

Regarding the rollback my thinking was that Status DB is the source of truth for this value so it would make sense to have it at least there even if we continuously fail to set it in TB.

cg/services/events/upload_handler.py
mogmi: Agreed 👍

cg/services/events/upload_handler.py
mogmi: I put my explanation [here](https://github.com/Clinical-Genomics/cg/pull/5136#discussion_r3403663963) - if the link works 😅

islean: Reasonable 👌

PR #5142, opened 2026-06-15
Skip Freshdesk interactions when data delivery is NO_DELIVERY

### LINKED ISSUE #5132: Make no_delivery cases not do anything in freshdesk during delivery

diitaz93:
### Added

- Static method to determine if the order is `NO_DELIVERY`
- Check if order is not `NO_DELIVERY` when calling freshdesk to send delivery message and closing ticket

### Changed

- Refactoring to keep symetry in exposed methods

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/services/deliver_service.py
+++ b/cg/services/deliver_service.py
@@ -4,6 +4,7 @@ from cg.apps.tb.api import TrailblazerAPI
 from cg.apps.tb.models import TrailblazerAnalysis
 from cg.clients.freshdesk.constants import Status
 from cg.clients.freshdesk.freshdesk_client import FreshdeskClient
+from cg.constants import DataDelivery
 from cg.exc import (
     FreshdeskDeliveryMessageError,
     FreshdeskGetTicketError,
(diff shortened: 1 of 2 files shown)


cg/services/deliver_service.py
```diff
@@ -126,6 +132,10 @@ def _get_order_analyses_dictionary(self) -> dict[Order, list[Analysis]]:
                 order_analyses[analysis.order] = [analysis]
         return order_analyses
 
+    @staticmethod
+    def _is_order_no_delivery(order: Order) -> bool:
+        return order.cases[0].data_delivery == DataDelivery.NO_DELIVERY
```
islean, line 137: Very minor and as soon as there is one order per case this will work, but we do introduce the small risk that the order we are interested in is not the original order of the case we look at, in which case the delivery will mismatch. Probably not worth fixing though?

islean (approved):
Should work although we might have instances where people rerun a whole case in another order which might cause some mismatches!

cg/services/deliver_service.py
diitaz93: I think this is a problem currently present in the upload. If a new order with `data_delivery=NO_DELIVERY ` reaches the upload, it will use the data delivery of the original case. I guess to really fix this we need to restrict the reordering of cases

PR #5146, opened 2026-06-17
Rollback database changes in upload handler

mogmi:
### Fixed

- Rollback any pending database operations when an error was raised so that the database session is cleared up for the next event.


islean (approved):
💯

PR #5185, opened 2026-07-07
Refine FASTQ compression to Spring

### LINKED ISSUE #5181: Spring compression is too resource intensive
## Description
The compression of FASTQ to the Spring format has proved problematic with the current configuration. It consumes a large amount of the storage bandwidth and we also allocate way more Memory and to a smaller degree CPU than needed.

We need to revise how cg invokes Crunchy and in turn Spring to make it more efficient

Vince-janv:
## Description
This PR implements several changes to optimise FASTQ compression to Spring. The most prominent of these is a rework of the memory and time allocation to Slurm. This PR implements linear scaling of these based on the reads contained in the file (fetched from statusDB). Additionally this PR removes the `--check-integrity` flag passed to crunchy, since a failed compression would never pass silently.

### Changed
- Spring now uses local disk on the compute nodes as a tmp dir
- Spring jobs are now submitted to its own partition
- Cores/threads used reduced from 12 to 8
- Memory and time is now set based on reads in file with fallbacks if the file does not follow the current naming pattern
- Memory and time is significantly reduced
- No longer passes the `--check-integrity` flag to crunchy

### Notes
Tests and selected values: https://docs.google.com/spreadsheets/d/1Irl_vS_ktwWfD3VBLR2U3gAsWgnoEzY-HF7HIjrMENc/edit?gid=0#gid=0


cg/apps/crunchy/sbatch.py
```diff
@@ -31,15 +31,15 @@
 
 FASTQ_TO_SPRING_COMMANDS = """
 mkdir -p {tmp_dir}
-{conda_run} crunchy -t 12 --tmp-dir {tmp_dir} compress fastq -f {fastq_first} -s {fastq_second} \
--o {spring_path} --check-integrity --metadata-file
+{conda_run} crunchy -t 8 --tmp-dir {tmp_dir} compress fastq -f {fastq_first} -s {fastq_second} \
```
diitaz93: Could we pass the number of tasks as a variable?

cg/meta/compress/compress.py
```diff
@@ -29,13 +44,42 @@ def __init__(
...
+        Scales both from the run's read count when it can be found in
+        illumina_sample_sequencing_metrics; otherwise falls back to CrunchyAPI's configured
+        fallback_memory/fallback_minutes.
+        """
+        reads: int | None = files.get_reads_for_run(compression_obj, self.status_db)
+        if not reads:
+            return self.crunchy_api.fallback_memory, self.crunchy_api.fallback_minutes
+        memory: int = scale_resource_by_reads(reads, memory_slope, memory_intercept, min_gb, max_gb)
+        minutes: int = scale_resource_by_reads(
+            reads, time_slope, time_intercept, min_minutes, max_minutes
```
diitaz93: suggestion: use named arguments

cg/constants/compression.py
```diff
@@ -6,10 +6,24 @@
...
+COMPRESSION_MINUTES_PER_READ: float = 4.6394e-07
+COMPRESSION_TIME_INTERCEPT: float = 18.3675
+MINUTES_FLOOR: int = 1
+MINUTES_CEIL: int = 1440
+
+# Read-based memory/time scaling for decompression - independent, separately tunable from compression
+DECOMPRESSION_MEMORY_PER_READ: float = 0
+DECOMPRESSION_MEMORY_INTERCEPT: float = 8
+DECOMPRESSION_MINUTES_PER_READ: float = 1.0918e-07
+DECOMPRESSION_TIME_INTERCEPT: float = 5.51
```
diitaz93, line 25: I understand you did some regression to get these values, but it would be nice to know in the future where they came from. I can imagine a system sentinel feeling quite lost when trying to solve a bug and looking at this file

tests/meta/compress/test_compress_files.py
```diff
@@ -40,3 +44,50 @@ def test_get_fastq_files_no_files(compress_api, sample_hk_bundle_no_files, sampl
...
+
+
+def test_get_reads_for_run_name_does_not_match(base_store: Store):
+    """Test get_reads_for_run returns None when the run name can't be parsed."""
+    # GIVEN a compression run whose name doesn't follow the expected naming convention
+    compression_obj = CompressionData(Path("not_a_valid_name"))
+
+    # WHEN looking up the reads for the run
+    # THEN None is returned
+    assert files.get_reads_for_run(compression_obj, base_store) is None
```
diitaz93: For convention, it is also easier to read I think
```suggestion
    # WHEN looking up the reads for the run
    reads = files.get_reads_for_run(compression_obj, base_store)
    
    # THEN None is returned
    assert reads is None
```

tests/meta/compress/test_compress_files.py
```diff
@@ -40,3 +44,50 @@ def test_get_fastq_files_no_files(compress_api, sample_hk_bundle_no_files, sampl
...
+
+
+def test_get_reads_for_run_no_matching_metrics(base_store: Store):
+    """Test get_reads_for_run returns None when no metrics row matches."""
+    # GIVEN a compression run that parses fine but has no matching sequencing metrics row
+    compression_obj = CompressionData(Path("23M7GHLT4_UNKNOWNSAMPLE_S1_L001"))
+
+    # WHEN looking up the reads for the run
+    # THEN None is returned
+    assert files.get_reads_for_run(compression_obj, base_store) is None
```
diitaz93: If you want you can do the same here

cg/apps/crunchy/sbatch.py
```diff
@@ -31,15 +31,15 @@
...
+-o {spring_path} --metadata-file
 rm {pending_path}
 rm -r {tmp_dir}
 """
 
 SPRING_TO_FASTQ_COMMANDS = """
 mkdir -p {tmp_dir}
-{conda_run} crunchy -t 12 --tmp-dir {tmp_dir} decompress spring {spring_path} -f {fastq_first} -s \
+{conda_run} crunchy -t 8 --tmp-dir {tmp_dir} decompress spring {spring_path} -f {fastq_first} -s \
```
diitaz93: same here, I believe we have access to the number of tasks from the slurm config where we call this

diitaz93 (approved):
Looks great ⭐

cg/apps/crunchy/files.py
```diff
@@ -31,14 +32,41 @@ def get_spring_to_fastq_sbatch_path(log_dir: Path, run_name: str = None) -> Path
...
-    """Create a temporary directory and return the path to it"""
+def get_tmp_dir(base: str) -> str:
+    """Return a node-local temp dir path, unique per SLURM job.
 
-    with tempfile.TemporaryDirectory(prefix=prefix, suffix=suffix, dir=base) as dir_name:
-        tmp_dir_path = dir_name
+    The path is expanded by the shell at job runtime (mkdir -p happens inside the
+    generated sbatch script), so `base` never needs to exist on the submission host.
+    """
+    return f"{base.rstrip('/')}/${{SLURM_JOB_ID}}_spring"
```
andrelaszlo, line 41: issue (resource-leak): I saw this changed from using a `TemporaryDirectory` so I was looking at how the cleanup works (since they are normally removed automatically). The problem, I think, is that our jobs are sometimes cancelled and then it probably doesn't matter if we're using a `TemporaryDirectory` - it probably fails to be cleaned up anyway. It seems like we already have some stale tempdirs:

```
[15:01] [hiseq.clinical@compute-0-28:/state/partition1] [base]  $ scontrol show job 14682756 | grep JobState
   JobState=TIMEOUT Reason=TimeLimit Dependency=(null)
[15:01] [hiseq.clinical@compute-0-28:/state/partition1] [base]  $ du -hs 14682756_spring/
1.4G    14682756_spring/
[15:05] [hiseq.clinical@compute-0-28:/state/partition1] [base]  $ ls -lah 14682756_spring/
total 44K
drwxrwx---   3 hiseq.clinical users 4.0K Jul 14 08:39 .
drwxrwxrwt. 14 root           root  4.0K Jul 14 14:50 ..
drwxr-xr-x   2 hiseq.clinical users  36K Jul 14 08:50 tmp.fa37JncCHr
```

It seems like cancelled jobs will first get a SIGTERM, then a SIGKILL. One approach would be to trap the SIGTERM and clean up the tempdir before exiting.

Another approach would be a cron job that cleans up stale tempdirs, or at least some monitoring for free space on `/state/partition1` on each node?

cg/apps/crunchy/sbatch.py
Vince-janv: Good idea! Addressed in ebe661ba69da27d85e551e508fa4806c758690c1

cg/meta/compress/compress.py
Vince-janv: Addressed in ebe661ba69da27d85e551e508fa4806c758690c1

tests/meta/compress/test_compress_files.py
Vince-janv: Addressed in ebe661ba69da27d85e551e508fa4806c758690c1

tests/meta/compress/test_compress_files.py
Vince-janv: Addressed in ebe661ba69da27d85e551e508fa4806c758690c1

cg/apps/crunchy/sbatch.py
Vince-janv: Addressed in ebe661ba69da27d85e551e508fa4806c758690c1

cg/constants/compression.py
Vince-janv: I linked a spreadsheet in the PR description with my tests and some plot. Not sure how to best pass it along to a future maintainer, but at least if they git blame the file they will hopefully find it

cg/apps/crunchy/crunchy.py
```diff
@@ -55,7 +59,9 @@ def create_pending_file(pending_path: Path, dry_run: bool) -> None:
         pending_path.touch(exist_ok=False)
 
     # These are the compression/decompression methods
-    def fastq_to_spring(self, compression_obj: CompressionData, sample_id: str = "") -> int:
+    def fastq_to_spring(
+        self, compression_obj: CompressionData, memory: int, minutes: int, sample_id: str = ""
```
andrelaszlo, line 63: thought: I have been trying out [`NewType`](https://docs.python.org/3/library/typing.html#newtype) a bit lately when juggling units like this (for `memory` and `minutes`, for example). It might be overkill here though!

andrelaszlo (commented):
Not done yet, just posting two comments for now!

cg/apps/crunchy/files.py
Vince-janv: Indeed. I added clean-up on SIGTERM in 0f3a9a290c2a2e31c8f6869d98f91c9c46a9e77a. I think it's a good first step and we'll see if it suffices

cg/apps/crunchy/crunchy.py
Vince-janv: Very cool! I deem it to be overkill though since this PR is essentially just putting a band-aid on a really old module

cg/apps/crunchy/files.py
andrelaszlo: Awesome 👍

andrelaszlo (approved):
Looks great to me!

PR #5224, opened 2026-07-29
Fix(Spring) Increase threads

### LINKED ISSUE #5223: Spring compression only running on one thread
## Description
Since a couple of weeks Spring compression and decompression has only been running on one thread instead of 8. This has resulted in a lot of timed out jobs

Vince-janv:
## Description
This PR makes crunchy use the slurm_cpus_per_task param instead of the slurm_number_tasks.

### Fixed

- Threads used in Spring now linked to cpus per tasks instead of num_tasks


diitaz93 (approved):
👍

cg/apps/crunchy/crunchy.py
```diff
@@ -92,7 +92,7 @@ def fastq_to_spring(
             job_name="_".join([sample_id, compression_obj.run_name, "fastq_to_spring"]),
             log_dir=log_dir.as_posix(),
             memory=memory,
-            number_tasks=self.slurm_number_tasks,
```
islean, line 95: The parameter name is a bit misleading now but I guess with some background it makes sense?

islean (approved):
👍

cg/apps/crunchy/crunchy.py
Vince-janv: Good catch! I meant to change a param further down. Fixed in ab4101bf57c4c5f087d2a31b8297c2c2b1a2112a

PR #5227, opened 2026-07-30
Make listener receive config from AppConfig instead of from CGConfig

### LINKED ISSUE #5214: Move NATS config to AppConfig class
## Description
There are two configs for cg one being the CGConfig class (for all CLI commands) and one being the AppConfig for the web-API. Since the EventListener is deployed in the same way and on the same machine as the API I think its config should be in the AppConfig class to avoid confusion.

## Acceptance criteria
- EventListener is instantiated with parameters from AppConfig instead of CGConfig

diitaz93:
## Description
PR in servers: https://github.com/Clinical-Genomics/servers/pull/1978

Make the event listener command a separate command from `cg`, independent of any type of configs, reading only from environment variables.

### Changed

- The `listen` command is removed from the cg group and defined independently
- The Nats config in CGConfig no longer has the listener variables and the structure is flat (the publisher config is moved one level up)
- The EventListener class no longer receives the NatsConfig as parameter but all the individual paths and variables that it needs

### How to test

- [x] Let GitHub build the Docker container for this branch
- [x] See further instructions in PR in servers https://github.com/Clinical-Genomics/servers/pull/1978

## Review

- [x] Tests executed by SD & LL
- [x] "Merge and deploy" approved by LL

## Implementation Plan

Deploy this branch on:

- [x] CGVS
  - [x] Production
  - [x] Stage
- [x] Hasta
  - [x] Production
  - [x] Stage


cg/server/app_config.py
```diff
@@ -44,5 +44,14 @@ class AppConfig(BaseSettings):
     freshdesk_order_email_id: int = 10
     freshdesk_environment: str = "Stage"
 
+    # NATS configurations
+    nats_server: str = "nats_server"
```
diitaz93: do we still need this here? @mogmi

tests/cli/test_listen.py
```diff
@@ -57,7 +28,7 @@ def cli_runner() -> CliRunner:
 
 @pytest.fixture(autouse=True)
 def mock_store_and_db(mocker: MockerFixture):
-    mocker.patch.object(events_cli, "Store")
+    mocker.patch.object(listen_cli, "Store")
```
diitaz93: is this patch necessary? it seems like it is patched again inside the test

cg/cli/listen.py
```diff
@@ -0,0 +1,69 @@
...
+from cg.services.events import upload_handler
+from cg.services.events.event_listener import EventListener
+from cg.services.events.upload_handler import ANALYSIS_UPLOADED_SUBJECT
+from cg.store.database import initialize_database
+from cg.store.store import Store
+
+LOG = logging.getLogger(__name__)
+
+
+@click.command("listen")
```
diitaz93:
```suggestion
@click.command("listen", hidden=True)
```

cg/server/app_config.py
mogmi: Nope!

tests/cli/test_listen.py
mogmi: Yes, looks like this also was not cleaned up

mogmi: Good catch!

cg/cli/listen.py
```diff
@@ -0,0 +1,69 @@
...
+
+from cg.apps.tb import TrailblazerAPI
+from cg.cli.utils import LOG_LEVELS
+from cg.services.events import upload_handler
+from cg.services.events.event_listener import EventListener
+from cg.services.events.upload_handler import ANALYSIS_UPLOADED_SUBJECT
+from cg.store.database import initialize_database
+from cg.store.store import Store
+
+LOG = logging.getLogger(__name__)
```
mogmi, line 19: We are not using LOG in this command at the moment, but probably nice to keep around

Dockerfile-cli
mogmi, line 1: We should probably rename this file since it is no longer a container for starting the main cli but just the listen cli

mogmi (approved):

Dockerfile-cli
mogmi: If we make that change we also need to update `build_publish_cli_image.yml` since it is referenced there.

Using the same logic the tag name should probably not be cg-cli anymore, but maybe we should save that for a future PR - it would then need to also change in the systemd service in servers.

mogmi (approved):
Looks great ✨

cg/cli/listen.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+@click.command("listen", hidden=True)
```
islean, line 22: What are the benefits of splitting the code base into two different CLIs? Is it because then we can make sure that no one runs cg listen on Hasta where it will not work? 

To me it is a bit strange to have it separate when it is still using code within the cg repository. What happens if we have a listener that needs trailblazer's code? That will need to be written in the trailblazer repository at which point we will have a name clash?

If we want a `listen` command, it feels reasonable to me that it would be in a new repository which calls endpoints exposed by the different interfaces?

diitaz93: To decouple it from the CGConfig. I think another repo using endpoints may be out of scope for the fix we want. The objective of this PR is just to avoid the need to deploy servers to CGVS every time we make a change in the CGConfig

islean: That makes sense. I have some concerns about scalability when other listeners in other repos come into play, but maybe we will be saved by them being in different environments. Either way, name choices should be easier at that point so feel free to resolve for now.

cg/cli/listen.py
diitaz93: added logging

PR #5236, opened 2026-08-07
Integration test deliver

diitaz93:
## Description
Add integration test for teh cli command `cg deliver all-available` 

### Added

- Functions to mock interactions with Trailblazer and Freshdesk using HTTPServer from pytest_httpserver

### Changed

- Split utils module into utils, trailblazer utils and freshdesk utils.

### DIFF
--- a/tests/integration/config/cg-test.yaml
+++ b/tests/integration/config/cg-test.yaml
@@ -10,7 +10,7 @@ email_base_settings:
   smtp_server: "localhost"
 freshdesk:
   api_key: "some_api_key"
-  base_url: "https://example.freshdesk.com"
+  base_url: "http://localhost:8888/freshdesk"
 
 nats:
   server: nats://localhost/nats
--- a/tests/integration/deliver/test_deliver.py
+++ b/tests/integration/deliver/test_deliver.py
@@ -41,10 +41,10 @@ def test_deliver_all_available(
     analysis_2 = helpers.add_analysis(
         store=status_db, case=case_2, trailblazer_id=102, uploaded_at=datetime.now()
     )
-    cust2 = case_2.customer
     analysis_2.order = helpers.add_order(
         store=status_db, ticket_id=67890, customer_id=case_2.customer.id
     )
+    status_db.commit_to_store()
 
     # GIVEN that Trailblazer has analyses ready for delivery
     expect_to_get_all_analyses_to_deliver_from_trailblazer(
--- a/tests/integration/deliver/test_deliver.py
+++ b/tests/integration/deliver/test_deliver.py
@@ -7,11 +7,12 @@ from pytest_httpserver import HTTPServer
 
 from cg.cli.base import base
 from cg.constants.constants import Workflow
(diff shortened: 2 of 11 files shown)


tests/integration/trailblazer_utils.py
```diff
@@ -0,0 +1,186 @@
...
+        % {
+            b"email": environ_email().encode(),
+            b"type": str(analysis_type).encode(),
+            b"case_id": case.internal_id.encode(),
+            b"ticket_id": str(ticket_id).encode(),
+            b"tower_workflow_id": _quoted_string_or_null(tower_workflow_id),
+            b"out_dir": str(out_dir).encode(),
+            b"config_path": _quoted_string_or_null(config_path),
+            b"workflow": str(workflow).upper().encode(),
+            b"workflow_manager": str(workflow_manager).encode(),
```
islean, line 54: Is it possible to do .format for byte strings? If so, it might be nicer to read.

tests/integration/deliver/test_deliver.py
```diff
@@ -0,0 +1,144 @@
...
+from tests.integration.freshdesk_utils import (
+    expect_freshdesk_get_ticket,
+    expect_freshdesk_reply_to_ticket,
+    expect_freshdesk_update_ticket,
+)
+from tests.integration.utils import IntegrationTestPaths
+from tests.store_helpers import StoreHelpers
+
+
+@pytest.mark.xdist_group(name="integration")
```
islean, line 22: Probably out of scope for this PR but should we collect all the magic numbers that we have in the integration tests in a README or something? Because if we reuse the get_analyses utils for fetching stuff from TB you need to match all the ids which would be nice to have somewhere.

tests/integration/config/cg-test.yaml
```diff
@@ -10,7 +10,7 @@ email_base_settings:
   smtp_server: "localhost"
 freshdesk:
   api_key: "some_api_key"
-  base_url: "https://example.freshdesk.com"
+  base_url: "http://localhost:8888/freshdesk"
```
islean, line 13: Was there a reason for changing this?

islean (approved):
Looks great!

tests/integration/trailblazer_utils.py
diitaz93: It is complicated, as I would have to use .encode. I would mark this as out of scope, as I just moved this method; it was written by LL 😅

tests/integration/config/cg-test.yaml
diitaz93: Because I started using the mocks with HTTPServer, before it was just a placeholder

tests/integration/deliver/test_deliver.py
diitaz93: hmmm I would initially say that these functions are not that reusable. They are more like fixtures. But they could be modified so that they receive the response as a parameter, and those could be fixtures. But yes, I would consider that out of scope

PR #5240, opened 2026-08-12
Improve query for sequencing qc service

### LINKED ISSUE #5244: Reduce the spamming in the logs for cg-sequencing-qc service
## Description
As explained in the parent issue, `cg-sequencing-qc` logs are spammy. 

A deeper look into the logs show that a big proportion of the cases selected for QC are excluded with one of the following warnings:

```shell
WARNING Not all samples for case ___ have been post-processed.
```
or
```shell
Sample ___ has too few reads
```

The first issue, cases not being post-processed, should not be that frequent (maybe not even happen at all). This is caused by samples having a value for `last_sequenced_at` but no entry in the SampleRunMetrics table. This should not happen in production, but as there are many samples migrated from cg_stats that existed before the SampleRunMetrics table ever existed, the database has many of these samples.

The second issue, sample not having enough reads, is the main reason for the QC to fail. However, cases that fail QC due to this warning can be quite old. A deeper look shows that many downsampled samples are being caught by this query.

## Proposed solution

Improve the query for fetching cases for sequencing QC so that cases with downsampled samples and samples without sequencing metrics are excluded. This is a start that will reduce from 800 cases to only 155 per run.

We would still need to figure out (clean) the samples of the remaining cases

diitaz93:
## Description
Changes the query that fetches the cases to start sequencing QC.

### Added

- Unit test for the modified CRUD method

### Changed

- Makes the CRUD function use SQLAlchemy directly in the function without calling any filter function
- Remove the unused filter function

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b fix-sequencing-qc-logs -a
    ```

### How to test

- [x] Run `cg sequencing-qc` in master and count how many cases were selected
- [x] Run `cg sequencing-qc` with this branch 

### Expected test outcome

- [x] Check that the command finishes successfully and the number of selected cases is reduced

### master
```shell
$ cg sequencing-qc 2>&1 | tee seq-qc-logs-master.txt
$ grep -c "Sequencing QC status for case" seq-qc-logs-master.txt
799
```

### branch
```shell
$ cg sequencing-qc 2>&1 | tee seq-qc-logs-dev.txt
$ grep -c "Sequencing QC status for case" seq-qc-logs-dev.txt
239
```

## Review

- [x] Tests executed by SD
- [x] "Merge and deploy" approved by

## Implementation Plan

- [x] Deployed to stage:
```shell
Getting deployer... done.
Log deploy... done.
cg, version 88.2.2
[js.diazboada@hasta:~] [S_base] $ up
```
- [x] Deployed to production:
```shell
Log deploy... done.
cg, version 88.2.2
[js.diazboada@hasta:~] [P_base] $
```


cg/store/crud/read.py
```diff
@@ -1618,21 +1619,36 @@ def get_illumina_flow_cell_by_internal_id(self, internal_id: str) -> IlluminaFlo
...
+            .filter(
+                Case.aggregated_sequencing_qc.in_(
+                    [SequencingQCStatus.PENDING, SequencingQCStatus.FAILED]
+                )
+            )
+            .filter(Sample.downsampled_to.is_(None))
+            .filter(
+                or_(
+                    Application.is_external,
+                    and_(Sample.last_sequenced_at.isnot(None), Sample._sample_run_metrics.any()),
```
diitaz93, line 1664: there are surprisingly many cases with samples that have sequencing metrics but no last_sequenced_at

cg/store/crud/read.py
```diff
@@ -1618,21 +1619,36 @@ def get_illumina_flow_cell_by_internal_id(self, internal_id: str) -> IlluminaFlo
...
+                .join(CaseSample.sample)
+                .join(ApplicationVersion)
+                .join(Application)
+            )
+            .filter(
+                Case.aggregated_sequencing_qc.in_(
+                    [SequencingQCStatus.PENDING, SequencingQCStatus.FAILED]
+                )
+            )
+            .filter(Sample.downsampled_to.is_(None))
```
diitaz93, line 1659: Maybe it would be better to set the sequencing qc of cases with downsampled samples to pass, but for now this will exclude them

cg/services/sequencing_qc_service/quality_checks/utils.py
```diff
@@ -177,6 +177,7 @@ def is_first_sample_yield_based_and_processed(case: Case) -> bool:
 
 def is_first_sample_reads_based_and_processed(case: Case) -> bool:
     sample: Sample = case.samples[0]
+    # Then we can assume all entries have sample run metrics?
```
diitaz93: In that case, we could remove the if statement

islean: I'm not sure I follow?

cg/store/crud/read.py
```diff
@@ -1618,21 +1619,36 @@ def get_illumina_flow_cell_by_internal_id(self, internal_id: str) -> IlluminaFlo
         return flow_cell
 
     def get_cases_for_sequencing_qc(self) -> list[Case]:
-        """Return all cases that are ready for sequencing QC."""
+        """
+        Return all cases that have pending or failed sequencing qc with production sequencing data.
+        Excludes from the query:
+        - Cases that passed sequencing QC
+        - Cases with downsampled samples
+        - Cases with internal samples that have not been post-processed yet
```
islean: Could we not get cases with downsampled samples and internal samples that have not been post-processed yet? If we have a case with three samples, where
sample_1 is downsampled
sample_2 is internal and not post-processed
sample 3 is a normal and processed sample,
I would claim the case would be returned.

Also, I find the use of "internal" samples a bit confusing, it feels like it is most commonly used for cust000 control samples and what not, but here I guess it just says that we sequence the samples, which is almost always the case?

islean: Also, we seem to keep any case with an external sample in it. Could be worth mentioning?

cg/store/crud/read.py
islean, line 1659: How does this interplay with the from_sample field? Do all samples with from_sample set also have downsampled_to set?

islean (commented):
I think I fail to see any changes done to the logging? Or are the down sampled samples the main culprits of the logging?

cg/store/crud/read.py
diitaz93: yes they should

islean: There seem to be around 70 samples that have from_sample not null and downsampled_to is null. Maybe small enough to ignore?

diitaz93: Maybe some manual work was done on those 70. I think it is ok to ignore. The most important is that the number of reads is the original one

cg/store/crud/read.py
```diff
@@ -1618,21 +1619,36 @@ def get_illumina_flow_cell_by_internal_id(self, internal_id: str) -> IlluminaFlo
         return flow_cell
 
     def get_cases_for_sequencing_qc(self) -> list[Case]:
-        """Return all cases that are ready for sequencing QC."""
+        """
+        Return all cases that have pending or failed sequencing qc with production sequencing data.
+        Excludes from the query:
+        - Cases that passed sequencing QC
+        - Cases with downsampled samples
```
diitaz93:
```suggestion
        - Cases with only downsampled samples
```

cg/store/crud/read.py
diitaz93: improved the docstring now

tests/store/crud/read/test_read_case.py
```diff
@@ -61,3 +65,170 @@ def test_get_case_by_internal_id_strict_fails(store_with_cases_and_customers: St
...
+    lane: int | None = None,
+) -> Case:
+    """Create a single case/sample setup for sequencing qc filtering tests."""
+    case: Case = helpers.add_case(
+        store=store,
+        internal_id=case_id,
+        name=case_id,
+        aggregated_sequencing_qc=aggregated_sequencing_qc,
+    )
+    sample = helpers.add_sample(
```
islean:
```suggestion
    sample: Sample = helpers.add_sample(
```

islean (approved):
Looks great! 🔥

PR #5266, opened 2026-08-17
Add datetime to demultiplex log file

### LINKED ISSUE #5113: Retain demultiplexing logs for troubleshooting
As a production bioinformatics member,
I want demultiplexing logs not to be overwritten when demultiplexing starts anew,
So that I can troubleshoot what happened even if we try to restaret demultiplexing anew.

# Work impact

Answer the following questions:
- Is there currently a workaround for this issue? If so, what is it?
    - No, currently the logs are not being sent out properly either, see https://github.com/Clinical-Genomics/bug-reports/issues/76
- How much time would be saved by implementing this feature on a weekly basis?
    - Not that much, it does not happen that often, but when it does, it could save us hours.
- How many users are affected by this issue?
    - Prodbioinfo members/system sentinels
- Are customers affected by this issue?
    - Yes, it may cause great delays because we have to wait and see if the error occurs again to see if it was a one off error, or if there is something wrong with the data.

# Acceptance Criteria

- [ ] Demultiplexing logs are not overwritten in the sequencing runs directory between attempts

# Notes

- Additional information.
  - Currently, since the demultiplexing logs are generically named `<flow_cell_internal_id>_demultiplex.std[err|out]`, they are overwritten when we restart demultiplexing. We also automatically restart the demultiplexing, because in the error trap of the demux script, we are removing the `demuxstarted.txt` and thus enabling demux to start again. This is something we would like to retain. Ideally, we should have some more "refined" situations when we believe that a restart will actually achieve something, and when we know that there is corrupted data and restart will not achieve anything. 
- Dependencies.
- Related user stories.

islean:
## Description

Closes #5113. Adds the current time to the name of the demultiplexing log files, which means it should not get overridden when flow cells are redemultiplexed, instead a separate file should be created.

### Added

-

### Changed

- Name of the demultiplexing log now contains a timestamp for the start of the job.

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b demultiplex-logging-overwrite -a
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


islean:
Tested on stage by starting the demultiplexing for a sequencing run, cancelling it, removing the demuxstarted.txt file and restarting the demultiplexing for the same run:

image

The log files exist in parallell.

cg/apps/demultiplex/demultiplex_api.py
```diff
@@ -109,12 +110,15 @@ def demultiplex_sbatch_path(sequencing_run: IlluminaRunDirectoryData) -> Path:
     @staticmethod
     def get_run_name(sequencing_run: IlluminaRunDirectoryData) -> str:
         """Create the run name for the sbatch job."""
-        return f"{sequencing_run.id}_demultiplex"
+        return f"{sequencing_run.id}_demultiplex_{datetime.now().strftime('%Y_%m_%d_%H%M%S')}"
```
mogmi: A potential problem with this is that if this method was called again only a second later we will get a different run name - I see that this does not happen with the current implementation but it could be an easy mistake for future changes. 

Maybe just generate the name once per invocation and store the generated name in memory and return that if it already exists?

diitaz93 (approved):
Looks good 👍

mogmi (approved):
Nice! 🎉

islean:
Deployed to Hasta:

```
repository is clean
Logging deploy ...
Getting deployer... done.
Log deploy... done.
cg, version 88.3.1
[isak.ohlsson@hasta:~] [P_base] $ 
```

PR #5270, opened 2026-08-18
More lightweight test for fluffy

mogmi:
### Changed

- Use local mocked CGConfig in Fluffy test since nothing was used from it and it reduces number of fixtures read for the test from 85 to 2


diitaz93 (approved):
👍

PR #5285, opened 2026-08-25
Use a fresh database for flask tests

mogmi:
## Description

Currently the flask tests reuse whatever database was setup before they were run. This causes issues when we want to run tests out of order or run a subset of the tests. This PR sets up a fresh database for each flask test just as we do with other tests and makes sure the current session on the store is the one that was just created.  


islean (approved):
💯

PR #5293, opened 2026-08-31
sample-transferred-handler

islean:
## Description
Add support for storing external samples via events

### Added

- `external_sample_transferred_handler` function, that stores external samples received from an event and updates the external_sample table
- New housekeeper function `add_new_bundle_and_version` similar to `create_new_bundle_and_version` but without commiting, in an effort to separate the comimtting from the session adding in Housekeeper
- New function in housekeeper `finalize_file_transactions` that copies files to the housekeeper file structure and commits
- Support for publishing events via nats-py without the need to calling the NATS CLI
- CRUD method to get an external sample strictly
- CRUD method to update an external sample
- Tests

### Changed

- Prefix of event names from `cg` to `status_db`

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    bash /home/proj/production/servers/resources/hasta.scilifelab.se/update-tool-stage.sh -e S_cg -t cg -b sample-transferred-handler -a
    ```

### How to test

- [x] Make sure that there is a sample in the external data directory in hasta
- [x] Make sure that the sample exists in StatusDB
- [x] Make sure that the sample has an entry in the ExternalSample table
- [x] Make sure there is no bundle in housekeeper for the given sample
- [x] Run
```shell
$ cg -l DEBUG receive-event external_sample.transfer_completed --event-payload '{"statusdb.sample_internal_id": "<sample_internal_id>", "cluster_location": "/home/proj/stage/external-data/cust000/sample-name>", "transfer_completed_at": "2026-09-03T09:00:00"}'"
```

### Expected test outcome

- [x] Run `housekeeper get bundle <sample_internal_id>` and see that the sample files appear in the bundle
- [x] Check the bundle directory and verify that the files are copied there
- [x] Check that the `transferred_at` has been updated in the ExternalSample table for the given sample
- [ ] In `hiseq.clinical`, check that the message was published to the stream with:
```shell
nats stream view cg-stage
```

## Review

- [x] Tests executed by SD
- [x] "Merge and deploy" approved by SD

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


cg/services/events/event_dispatching.py
```diff
@@ -17,6 +19,7 @@ def __call__(self, config: CGConfig, event_payload: dict) -> None: ...
 EVENT_HANDLERS: dict[str, EventHandler] = {
     "external.customer_uploaded_sample": external_sample_uploaded_handler.handle,
     "external.samples_ordered": external_samples_ordered_handler.handle,
+    EXTERNAL_SAMPLE_TRANSFERRED_SUBJECT: external_sample_transferred_handler.handle,
```
kedhammar, line 22: I feel like either all or none of these should be global variables

cg/services/transfer_to_cluster_service.py
```diff
@@ -54,7 +54,7 @@ def _get_sbatch_command(cg_config: CGConfig, sample: Sample) -> str:
     destination_path.mkdir(parents=True, exist_ok=True)
```
kedhammar, line 54: In the external_sample_transferred_handler, we throw an error if we find the sample dir that was synced to hasta has no sequencing files. Would it make more sense to only transfer dirs that contain the expected files? I feel like the logic and/or error could be further upstream.

Let's say a customer creates a sample dir and the transfer event triggers before the sequencing files are added, then what will happen and how messy will it be to rectify?

cg/services/events/event_handlers/external_sample_transferred_handler.py
```diff
@@ -0,0 +1,80 @@
...
+    and copy them to the corresponding Housekeeper bundle directory. Also update the entry for the
+    sample in the ExternalSample table with the datetime of the transfer.
+    """
+    event = ExternalSampleTransferredEvent.model_validate(event_payload)
+
+    if not (event.cluster_location.glob("*.bam") or event.cluster_location.glob("*fastq.gz")):
+        raise CgError(f"No sequencing files found in directory {event.cluster_location}")
+
+    sample: Sample = config.status_db.get_sample_by_internal_id_strict(event.sample_internal_id)
+    config.status_db.update_external_sample(
```
kedhammar, line 36: This method is hardcoded to only update "transferred_at", so I feel like the name is maybe too generic

cg/services/events/event_dispatching.py
kedhammar: I think all of them, event subjects and payload attributes would be nice to keep as symbols instead of strings if possible.

cg/services/transfer_to_cluster_service.py
kedhammar: Consider in relation to the copy-complete file logic which has yet to be implemented?

cg/services/events/event_publisher.py
```diff
@@ -14,3 +22,33 @@ def publish_command(nats_config, subject: str, data: dict) -> str:
...
     )
     return command
+
+
+def publish(nats_config, subject: str, event_payload: dict) -> None:
+    """Publish an event to NATS JetStream from synchronous code."""
+    asyncio.run(_publish_async(nats_config=nats_config, subject=subject, data=event_payload))
+
+
+async def _publish_async(nats_config, subject: str, data: dict) -> None:
```
kedhammar, line 35: Private methods on top for modules or what was the convention?

cg/services/events/event_publisher.py
```diff
@@ -1,4 +1,12 @@
...
+import ssl
+from pathlib import Path
+from ssl import Purpose, SSLContext, TLSVersion
+
+import nats
+from nats.aio.client import Client
+from nats.js import JetStreamContext
 
 
 def publish_command(nats_config, subject: str, data: dict) -> str:
```
kedhammar, line 15: We've talked about renaming this, maybe OOS for this PR

cg/services/events/event_dispatching.py
diitaz93: Documented the need here https://github.com/Clinical-Genomics/external-data-ingestion/issues/21

cg/services/events/event_publisher.py
diitaz93: Documented need in https://github.com/Clinical-Genomics/external-data-ingestion/issues/22

diitaz93 (approved):
👍

PR #5295, opened 2026-08-31
Refactor order submission to use external order submitter service

diitaz93:
## Description

Moves the order submitted instantiation to the ext.py

### How to test

- [ ] Deploy to CGVS stage
- [ ] Submit and order

### Expected test outcome

- [ ] Check that the order submission is succesful

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


diitaz93:
## Test in stage
Screenshot 2026-09-11 at 16 05 43

cg/server/ext.py
```diff
@@ -123,3 +124,10 @@ def init_app(self, app):
     system_email_id=app_config.freshdesk_order_email_id,
     env=app_config.freshdesk_environment,
 )
+order_submitter = OrderSubmitter(
+    ticket_handler=ticket_handler,
+    storing_registry=storing_service_registry,
+    validation_service=order_validation_service,
+    status_db=db,
+    nats_config=nats_config,
+)
```
mogmi, line 133: I probably missed this discussion so feel free to disregard, but are we not moving away from the pattern where we instantiate everything in one place no matter what will actually be used in that particular context? 

If we move it here we will have one singleton order_submitter that will live as long as the application is running - that could be fine but then you need to be very careful with making any changes to that instance since that change will persist when another request is received. The way we had it before we would get a fresh new instance every time and there is no risk of affecting other requests.

diitaz93: That makes sense, I think we didn't consider that. That is enough for closing this PR

PR #5307, opened 2026-09-10
Set internal_id for existing samples

### LINKED ISSUE #5137: Allow existing samples to be specified in the JSON-orderforms
## Description
Our customers using JSON forms for placing orders would like to existing samples to be supported. The file is generated by their Lab management system which limits the implementation quite a bit.

## Acceptance criteria
- Users can add a the key "existing_sample_name" which corresponds to a string value containing the sample.name of the existing sample.
- This key is optional. Not providing it should result in the same behaviour as master
- If provided, the existing sample should be fetched from the database (filtering on customer and the provided sample name) and the relevant information should be fetched from the database instead from the other JSON keys.
- Once the data is returned to the FE the existing sample should behave exactly as if it had been searched for and selected in the FE dropdown
- If multiple samples owned by the customer match the given sample name an error should be raised an presented to the user

## Notes
- Do they want to edit parents and status

islean:
## Description

Closes #5137.

### Added

-

### Changed

- Filling out existing_sample_name for a sample in the order form triggers us to set the internal_id for said sample, triggering the frontend and validation to treat it as an existing sample.

### Fixed

-

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...

### DIFF
--- a/cg/apps/orderform/orderform_parser.py
+++ b/cg/apps/orderform/orderform_parser.py
@@ -10,7 +10,8 @@ from cg.exc import OrderFormError
 from cg.models.orders.constants import OrderType
 from cg.models.orders.orderform_schema import OrderCase, Orderform, OrderPool
 from cg.models.orders.sample_base import OrderSample
-from cg.store.models import Customer
+from cg.store.models import Customer, Sample
+from cg.store.store import Store
 
 LOG = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


islean:
Validation tested successfully:

image

Uploading tested successfully:

Screenshot 2026-09-23 at 16 13 00

In the orderform, one sample was specified. It had a subject_id matching 8 samples belonging to the same customer and with an application orderable with raredisease. All samples were added within the same case (with the same name as the original sample had set) and that case failed validation due to the repeated subject_ids (as well as the missing relationships).

cg/apps/orderform/orderform_parser.py
```diff
@@ -159,6 +163,42 @@ def generate_orderform(self) -> Orderform:
             pools=self.get_pools(),
         )
 
+    def _fill_out_existing_samples(self, status_db: Store):
+        existing_samples: list[OrderSample] = []
+        for sample in [sample for sample in self.samples if sample.existing_sample]:
+            if not sample.subject_id:
```
diitaz93: I think this is more clear, but up to you
```suggestion
        for sample in self.samples:
            if not sample.existing_sample:
                continue
            if not sample.subject_id:
```

diitaz93 (approved):
Looks great 🚀

cg/apps/orderform/orderform_parser.py
islean: I agree, will implement 🚀

PR #5323, opened 2026-09-18
Allow raw-data as an option when creating application limitations

mogmi:
## Description

Due to a mistake in a previous migration the allowed values for the `workflow` column in the `application_limitations` table was outdated. This PR migrates the database so that the values match the current ones. 

Luckily there are no rows in the database with the options that will be removed so no migration of data is necessary. 

### Fixed

- Correct values for the `workflow` column of `application_limitations`


alembic/versions/2026_09_18_ac66c84f0449_fix_application_limitations_workflow_.py
```diff
@@ -0,0 +1,73 @@
...
+down_revision = "8a0dd32c2550"
+branch_labels = None
+depends_on = None
+
+
+# Outdated Enum accidentally reinstated by add_nallo_to_limitations_table migration
+outdated_workflow_list = (
+    "balsamic",
+    "balsamic-pon",
+    "balsamic-qc",
```
islean, line 23: This seems to get removed - do we have any limitations with balsamic-qc as workflow currently? If so, we might need to either drop them or change their workflows somehow

islean (approved):
Looks good to me!

alembic/versions/2026_09_18_ac66c84f0449_fix_application_limitations_workflow_.py
mogmi: The thing is balsamic-qc was previously removed by a migration here: https://github.com/Clinical-Genomics/cg/blob/8368022e4331e8804fd860c8ec39abc87052fd4e/alembic/versions/2025_02_10_a9568bb3a8b7_remove_balsamic_qc.py

But then it was (probably mistakenly because of copying an outdated list) added back here: https://github.com/Clinical-Genomics/cg/blob/8368022e4331e8804fd860c8ec39abc87052fd4e/alembic/versions/2025_03_06_3a0250e5526d_add_nallo_to_limitations_table.py

So I believe it should be safe to remove it again

islean: Yeah, and we do not allow for balsamic-qc orders anymore. I double checked the database and we have no limitations for that workflow either so I think we are good.

mogmi:
Deployed to stage:
```bash
INFO  [alembic.runtime.migration] Running upgrade 8a0dd32c2550 -> ac66c84f0449, Fix application limitations workflow enum
repository is clean
Logging deploy ...
Getting deployer... done.
Log deploy... done.
cg, version 88.10.1
[15:39] [hiseq.clinical@hasta:~] [S_base]  $
```

Deployed to prod:

```bash
INFO  [alembic.runtime.migration] Running upgrade 8a0dd32c2550 -> ac66c84f0449, Fix application limitations workflow enum
repository is clean
Logging deploy ...
Getting deployer... done.
Log deploy... done.
cg, version 88.10.1
[15:49] [hiseq.clinical@hasta:~] [P_base]  $
```

PR #5329, opened 2026-09-22
Admin view - Directly type in the search field

RasmusBurge-CG:
### Added

- When typing, it will automatically start typing in the search field.

### How to prepare for test

- [ ] Use stage: `us`
    ```shell
    ```

### How to test

- [ ] Do ...

### Expected test outcome

- [ ] Check that ...

## Review

- [ ] Tests executed by

## Implementation Plan

- [ ] Document in ...
- [ ] Inform to ...


diitaz93 (approved):
👍

cg/server/templates/admin/master.html
```diff
@@ -32,4 +32,23 @@
             box-shadow: inset 0 -2px 0 #ddd;
         }
     </style>
+    <script>
+    document.addEventListener('keydown', function (e) {
+      /* Gets search field if there is one. */
+      var search = document.querySelector('input[name="search"]');
```
moedarrah: A small suggestion, use const for search and el, since neither is reassigned.

cg/server/templates/admin/master.html
```diff
@@ -32,4 +32,23 @@
...
+      if (!search) return;
+
+      /* Ensure to not block writing in any field that is not the search field. */
+      var el = document.activeElement;
+      if (el === search || el.tagName === 'INPUT' ||
+          el.tagName === 'TEXTAREA' || el.isContentEditable) return;
```
moedarrah: Could we also exclude BUTTON here? When a button has keyboard focus, pressing Space should activate it, but this handler moves focus to the search instead.

moedarrah (approved):
Looks great! I just have two minor comments.  💯

cg/server/templates/admin/master.html
RasmusBurge-CG: Ah yes! Good point!

cg/server/templates/admin/master.html
RasmusBurge-CG: I thought it was handled by this:

```
      /* Ensure to not block shortcuts and only add single letters to search field. */
      if (e.metaKey || e.ctrlKey || e.altKey) return;
      if (e.key.length !== 1) return;
```

space and enter I would like to keep for the typing. Let me know if I misunderstood :)

moedarrah: Space is a single character, so it passes that check. Excluding `BUTTON` lets Space activate a focused button while still working normally in search. I can show you the difference in the UI

